PR #87, opened 2024-01-10
feat(js): opanai chat completion streaming support

### LINKED ISSUE #39: [JS][OpenAI]Instrumentation support for streaming

mikeldking:
resolves #39 

Adds the ability to get LLM spans from streaming responses. Does this by splitting the stream via `.tee()` and iterating over the chunks.

### DIFF
--- a/js/packages/openinference-instrumentation-openai/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-openai/src/instrumentation.ts
@@ -13,6 +13,7 @@ import {
   SpanKind,
   Attributes,
   SpanStatusCode,
+  Span,
 } from "@opentelemetry/api";
 import { VERSION } from "./version";
 import {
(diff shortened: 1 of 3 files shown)


js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -119,23 +120,27 @@ export class OpenAIInstrumentation extends InstrumentationBase<typeof openai> {
...
                 ...getLLMOutputMessagesAttributes(result),
                 ...getUsageAttributes(result),
               });
+              span.setStatus({ code: SpanStatusCode.OK });
+              span.end();
+            } else {
+              // This is a streaming response
+              // handle the chunks and add them to the span
+              // First split the stream via tee
+              const [leftStream, rightStream] = result.tee();
```
RogerHYang, line 139: Does `tee` create a two-speed scenario? For example:
1. If user cancels the stream, do we go ahead and read the whole thing, incurring additional token costs?
   - What if the stream never ends?
3. If network error happens mid-stream, how does that error bubble up to the user?

mikeldking: I'm not fully sure on either - both valid concerns. Right now I'm trying to avoid an extremely convoluted instrumentation. Let me add some tests and get back to you.

RogerHYang: Sounds good, thanks!

RogerHYang (approved):

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -338,3 +333,27 @@ function getEmbeddingEmbeddingsAttributes(
...
+ * Consumes the stream chunks and adds them to the span
+ */
+async function consumeChatCompletionStreamChunks(
+  stream: Stream<ChatCompletionChunk>,
+  span: Span,
+) {
+  let streamResponse = "";
+  for await (const chunk of stream) {
+    if (chunk.choices.length > 0 && chunk.choices[0].delta.content) {
+      streamResponse += chunk.choices[0].delta.content;
```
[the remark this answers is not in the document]
RogerHYang: p.s. in which case the output value mime type would have to be JSON (instead of TEXT)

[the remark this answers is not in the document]
mikeldking: Good call out, I've completely not added tool call support so here's the ticket https://github.com/Arize-ai/openinference/issues/90

there's also this auto tool calling that I probably need to consider (https://github.com/openai/openai-node?tab=readme-ov-file#automated-function-calls), though it looks to be in beta

PR #127, opened 2024-01-22
docs(python): llama-index example

mikeldking:


python/examples/llama-index/backend/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Use an official Python runtime as a parent image
+FROM python:3.11-slim as builder
```
axiomofjoy, line 2: This Dockerfile only has a single stage, which is the production stage. There is no need for the `builder` alias.

python/examples/llama-index/backend/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends wget \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# Install poetry
+RUN pip install poetry
```
axiomofjoy, line 12: Optional: The final production image does not actually need poetry, which is used to install dependencies. If you want to optimize the production image, you can use poetry to install the dependencies in a `deps` or `builder` stage, then copy the dependencies over to the production image. (Not required, just a cool thing you can do)

python/examples/llama-index/frontend/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+npm run dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
+
+You can start editing the page by modifying `app/page.tsx`. The page auto-updates as you edit the file.
+
+This project uses [`next/font`](https://nextjs.org/docs/basic-features/font-optimization) to automatically optimize and load Inter, a custom Google Font.
```
axiomofjoy, line 21: Can you add instructions on using Docker compose? See the README for the llama-index-express app for an example.

python/examples/llama-index/run.sh
```diff
@@ -0,0 +1,17 @@
...
+# Run the frontend app from the /frontend in the background
+cd frontend && npm run dev &
+
+pid=
+trap 'echo SIGINT; [[ $pid ]] && kill $pid; exit' SIGINT
+trap 'echo SIGTERM; [[ $pid ]] && kill $pid; exit' SIGTERM
+echo Starting script
+sleep 10000 & pid=$!
+wait
+pid=
```
axiomofjoy, line 17: Do we need this file with the current Docker configuration? This looks like it would be used to run the frontend and backend in the same container, but we're not doing that.

axiomofjoy (approved):

python/examples/llama-index/run.sh
mikeldking: Yeah please delete. Not needed

anticorrelator: This is the auto-generated initialization script, I'm thinking maybe we should keep it in case a user wants to run this locally?

mikeldking: @anticorrelator its not I added it 😅

python/examples/llama-index/frontend/.env.example
```diff
@@ -0,0 +1,3 @@
+# Rename this file to `.env.local` to use environment variables locally with `next dev`
+# https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables
+MY_HOST="example.com"
```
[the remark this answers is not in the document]
anticorrelator: .env files are ignored, just the examples are allowed as per the instruction to rename them to `.env.local`

python/examples/llama-index/frontend/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# COPY . /app/
+
+# # Install production dependencies
+# COPY package*.json ./
+# RUN npm install --only=production
+
+# # The command to run when the container starts
+# CMD ["npm", "start"]
+
+
```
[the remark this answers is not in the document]
anticorrelator: oops 🙃

python/examples/llama-index/run.sh
anticorrelator: oh! ok 😅

python/examples/llama-index/backend/Dockerfile
anticorrelator: This is an awesome suggestion, but I'm going to move on for now, thanks!

PR #138, opened 2024-01-25
feat: langchain instrumentor

RogerHYang:
resolves: https://github.com/Arize-ai/phoenix/issues/1960

based on tracer from [Phoenix](https://github.com/Arize-ai/phoenix/blob/21f54bd9fe5d62a472a23629f0011f81854900eb/src/phoenix/trace/langchain/tracer.py#L309)

## TODO
1. figure out why some times the context api token fails to detach
    - this happens with async
2. figure out why some times openai instrumentor span has its grandparent as parent
    - this happens with async
    - likely related to no. 1
    - see FIXME in the unit test


python/instrumentation/openinference-instrumentation-langchain/tests/openinference/instrumentation/langchain/test_instrumentor.py
```diff
@@ -0,0 +1,353 @@
...
+        assert (oai_span.events[0].attributes or {}).get(
+            OTELSpanAttributes.EXCEPTION_TYPE
+        ) == "BadRequestError"
+    assert oai_attributes == {}
+
+    # The remaining span is from the openai instrumentor.
+    openai_span = spans_by_name.popitem()[1]
+    assert openai_span.parent is not None
+    if is_async:
+        # FIXME: it's unclear why the context fails to propagate.
```
axiomofjoy, line 209: Maybe this is a TODO and a follow-up ticket rather than a FIXME

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def _start_trace(self, run: Run) -> None:
+        super()._start_trace(run)
+        span = self._tracer.start_span(run.name)
+        token = context_api.attach(trace_api.set_span_in_context(span))
+        self._runs[run.id] = _Run(span=span, token=token)
+
+    def _end_trace(self, run: Run) -> None:
+        if event_data := self._runs.pop(run.id, None):
+            # FIXME: find out why sometimes token fails to detach, e.g. when it's async
```
axiomofjoy, line 53: Convert to issue and todo?

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+def _input_messages(run_inputs: Mapping[str, Any]) -> Iterator[Tuple[str, List[Message]]]:
+    """Yields chat messages if present."""
+    if not hasattr(run_inputs, "get"):
+        return
+    # There may be more than one set of messages. We'll use just the first set.
+    if not (multiple_messages := run_inputs.get("messages")):
+        return
+    assert isinstance(
```
axiomofjoy, line 201: I see a vulnerability here in that our parsing logic raises assertion errors when the data is of an unexpected format, which will crash `end_trace`. It looks like that will prevent a span from being created. More importantly, is the user's application insulated from that crash, meaning even if an assertion error is raised here, will the user's app at least continue to function with no traces?

axiomofjoy (approved):

axiomofjoy:
Nice job

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
RogerHYang: added try statements

python/instrumentation/openinference-instrumentation-langchain/tests/openinference/instrumentation/langchain/test_instrumentor.py
RogerHYang: seems like a FIXME because it's not inline with expectation, where TODO is more like extra features or something

PR #144, opened 2024-01-26
chore: ignore vendored dependencies

axiomofjoy:
- git ignore vendored dependencies
- add pytest to dspy

Makes it easier to create local virtual environments since we have a bunch of them now and it's hard to keep track of where they are.

Inside the particular project, create environment with `conda create --prefix .venv python=3.9` and activate with `conda activate ./.venv`.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-dspy/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-dspy/pyproject.toml
@@ -37,6 +37,7 @@ instruments = [
 test = [
   "dspy-ai==2.1.0",
   "opentelemetry-sdk",
+  "pytest",
   "requests-mock",
 ]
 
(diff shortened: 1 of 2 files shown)


python/instrumentation/openinference-instrumentation-dspy/pyproject.toml
```diff
@@ -37,6 +37,7 @@ instruments = [
 test = [
   "dspy-ai==2.1.0",
   "opentelemetry-sdk",
+  "pytest",
```
RogerHYang, line 40: why do we need this?

axiomofjoy: I didn't see this was already defined in `dev-requirements.txt`, will remove.

axiomofjoy:
Closing in favor of https://github.com/Arize-ai/openinference/pull/158

PR #155, opened 2024-01-27
fix: enforce explicit parent context at span creation for llama-index

RogerHYang:
Because it's not predictable when the event callbacks are made, relying on automatic context setting can result in incorrect parent-child relationship. This PR enforces the same parent-child relationship provided by Llama-Index by explicitly specifying the parent context at span creation.

## Before 

Sibling spans under `agent_step` become child spans of the first span (`llm`)

Screenshot 2024-01-26 at 5 04 54 PM

## After

Correct siblings under `agent_step`

Screenshot 2024-01-26 at 5 04 26 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
+++ b/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
@@ -239,6 +239,10 @@ class OpenInferenceTraceCallbackHandler(BaseCallbackHandler):
         if parent_id != BASE_TRACE_EVENT:
             if parent_event_data := self._event_data.get(parent_id):
                 context = parent_event_data.context
+        # Instead of relying on automatic context lookup, we set the context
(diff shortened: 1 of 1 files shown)


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -234,14 +235,24 @@ def on_event_start(
...
-        span: trace_api.Span = self._tracer.start_span(name=event_type.value, start_time=start_time)
+        context = None
+        if parent_id != BASE_TRACE_EVENT:
+            if parent_event_data := self._event_data.get(parent_id):
+                context = parent_event_data.context
+        span: trace_api.Span = self._tracer.start_span(
+            name=event_type.value,
+            start_time=start_time,
+            context=context,
+        )
```
mikeldking, line 246: Nit: might want to put a little bit of docstrings here - it's a bit non-self evident.

mikeldking (approved):

PR #158, opened 2024-01-30
feat: dspy instrumentation enhancements

axiomofjoy:
- instruments `Retriever.forward`
- instruments implicit `Module.forward` via `Module.__call__`
- instruments `ColBERTv2.__call__`
- switches `requests-mock` for `responses`, which has wider adoption


python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -93,13 +134,17 @@ def __call__(
...
             span_name,
-            attributes={
-                SpanAttributes.OPENINFERENCE_SPAN_KIND: OpenInferenceSpanKindValues.LLM.value,
-                SpanAttributes.LLM_MODEL_NAME: instance.kwargs.get("model"),
-                SpanAttributes.LLM_INVOCATION_PARAMETERS: json.dumps(invocation_parameters),
-                SpanAttributes.INPUT_VALUE: str(prompt),
-                SpanAttributes.INPUT_MIME_TYPE: OpenInferenceMimeTypeValues.TEXT.value,
-            },
+            attributes=dict(
+                _flatten(
```
mikeldking, line 138: Is the outer dict necessary?

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -110,10 +155,14 @@ def __call__(
             # TODO: parse usage. Need to decide if this
             # instrumentation should be used in conjunction with model instrumentation
             span.set_attributes(
-                {
-                    SpanAttributes.OUTPUT_VALUE: json.dumps(response),
-                    SpanAttributes.OUTPUT_MIME_TYPE: OpenInferenceMimeTypeValues.JSON.value,
-                }
+                dict(
+                    _flatten(
```
mikeldking, line 159: not sure I follow why flatten is need in these cases

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -131,15 +180,48 @@ def __call__(
...
-            attributes={
-                SpanAttributes.OPENINFERENCE_SPAN_KIND: OpenInferenceSpanKindValues.CHAIN.value,
-                SpanAttributes.INPUT_VALUE: json.dumps(kwargs),
-                SpanAttributes.INPUT_MIME_TYPE: OpenInferenceMimeTypeValues.JSON.value,
-            },
+            attributes=dict(
+                _flatten(
+                    {
+                        OPENINFERENCE_SPAN_KIND: CHAIN.value,
+                        INPUT_VALUE: _get_input_value(
```
mikeldking, line 216: nit: _get_input_value isn't actually generic is it? It's specific to predict right?

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -167,3 +253,269 @@ def _prediction_to_output_dict(self, prediction: Any, signature: Any) -> Dict[st
...
+            ),
+        ) as span:
+            try:
+                prediction = wrapped(*args, **kwargs)
+            except Exception as exception:
+                span.set_status(trace_api.Status(trace_api.StatusCode.ERROR, str(exception)))
+                span.record_exception(exception)
+                raise
+            span.set_attributes(
+                dict(
```
mikeldking, line 349: same as above, can't flatten just return a dict?

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
```diff
@@ -52,41 +69,262 @@ class BasicQA(dspy.Signature):  # type: ignore
...
-                    "content": "Washington DC",
-                },
-                "logprobs": None,
-                "finish_reason": "stop",
-            }
-        ],
-        "usage": {"prompt_tokens": 39, "completion_tokens": 396, "total_tokens": 435},
-        "system_fingerprint": None,
-    }
+    responses.add(
```
mikeldking, line 72: nice

mikeldking (approved):

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
axiomofjoy: I can change, but his function is copy-pasted from our other instrumentation libraries and there might be some value in consistency.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
axiomofjoy: I think it's pretty generic? The main purpose it's trying to serve is that we can't just serialize `kwargs` directly since the contents of `kwargs` will depend on whether the user passed arguments positionally or not (e.g., `self.generate_answer(question)` vs. `self.generate_answer(question=question)`). That problem exists anytime we're trying to parse inputs passed via arguments.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
axiomofjoy: It's not needed in several cases, only in cases where the attribute value is a list of objects. I put it everywhere for consistency, but I can change.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
axiomofjoy: `_flatten` is copy-pasted from our other instrumentation libraries. I think there's some value in keeping it the same in case we want to factor it out at some point.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
axiomofjoy: It's not needed in several cases, only in cases where the attribute values include a list of objects. I included everywhere for the sake of consistency, but can remove if it's confusing.

PR #161, opened 2024-01-31
fix: erroneous if statement

RogerHYang:


mikeldking (approved):

axiomofjoy (approved):
ID can be an empty list?

RogerHYang:
> ID can be an empty list?

good point. converted to checking falsiness

PR #174, opened 2024-02-07
docs: Add bedrock tracing tutorial notebook

anticorrelator:
Resolves [#2133](https://github.com/Arize-ai/phoenix/issues/2133)


mikeldking (changes requested):
Hey @anticorrelator let's move this notebook to @arize-ai/phoenix - as it will need to be incorporated into the notebook. Also it should work end-to-end (e.g. have a collector running in the form of phoenix)

PR #175, opened 2024-02-07
feat(langchain): add support for chain metadata in spans

mikeldking:
Adds the ability to serialize langchain specific metadata into the traces via a `langchain` vendor prefix.

Screenshot 2024-02-07 at 2 42 52 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
@@ -36,6 +36,8 @@ from opentelemetry import trace as trace_api
 from opentelemetry.semconv.trace import SpanAttributes as OTELSpanAttributes
 from opentelemetry.util.types import AttributeValue
 
+from .lc_span_attributes import LC_METADATA_PREFIX
+
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
@@ -140,6 +142,7 @@ def _update_span(span: trace_api.Span, run: Run) -> None:
                     _function_calls(run.outputs),
                     _tools(run),
                     _retrieval_documents(run),
+                    _metadata(run),
                 )
             )
         )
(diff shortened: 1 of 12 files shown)


python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -471,6 +474,26 @@ def _retrieval_documents(run: Run) -> Iterator[Tuple[str, List[Mapping[str, Any]
...
+    #     return
+    # assert hasattr(outputs, "get"), f"expected Mapping, found {type(outputs)}"
+    # documents = outputs.get("documents")
+    # assert isinstance(documents, Iterable), f"expected Iterable, found {type(documents)}"
+    if not (metadata := run.extra.get("metadata")):
+        return
+    assert isinstance(metadata, Mapping), f"expected Mapping, found {type(metadata)}"
+    # Yield out each metadata key and value
+    for key, value in metadata.items():
+        yield f"{LC_METADATA_PREFIX}.{key}", value
```
mikeldking, line 494: I guess I should do a dump here if it's not primative?

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/lc_span_attributes.py
```diff
@@ -0,0 +1,8 @@
+# LangChain specific span attributes
+
+LC_METADATA_PREFIX = "langchain.metadata"
```
mikeldking, line 3: Alternatively could make this a semcov... Might be somewhat unifying.

RogerHYang (approved):
LGTM. Just be mindful of how the dot partitioning (i.e. dict nesting) works when the attribute key is not a semconv string.
Screenshot 2024-02-07 at 3 35 07 PM

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/lc_span_attributes.py
mikeldking: adding to semcov in #176

PR #192, opened 2024-02-08
fix: stop attaching context for callback based instrumentors

RogerHYang:
resolves https://github.com/Arize-ai/phoenix/issues/2229

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/examples/chain_metadata.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/examples/chain_metadata.py
@@ -10,7 +10,7 @@ from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProces
 endpoint = "http://127.0.0.1:6006/v1/traces"
 tracer_provider = trace_sdk.TracerProvider()
 trace_api.set_tracer_provider(tracer_provider)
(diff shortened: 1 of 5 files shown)


python/instrumentation/openinference-instrumentation-langchain/examples/chain_metadata.py
```diff
@@ -10,7 +10,7 @@
 endpoint = "http://127.0.0.1:6006/v1/traces"
 tracer_provider = trace_sdk.TracerProvider()
 trace_api.set_tracer_provider(tracer_provider)
-tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter(endpoint=endpoint)))
+tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter(endpoint)))
```
mikeldking, line 13: nit: keyword is a bit safer

mikeldking (approved):

python/instrumentation/openinference-instrumentation-langchain/examples/chain_metadata.py
RogerHYang: That is true, but since these are just examples, I wanted to make them as least verbose as possible. On the other hand, the chance that OTEL sdk would make a breaking change here is pretty minimal

PR #216, opened 2024-02-15
fix: avoid openai dependency at import time

RogerHYang:
convert version lookup to a (lazy) function with a cached result

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_utils.py
+++ b/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_utils.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import warnings
+from functools import lru_cache
 from importlib.metadata import version
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_utils.py
```diff
@@ -24,7 +25,10 @@
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
-_OPENAI_VERSION = tuple(map(int, version("openai").split(".")[:3]))
+
+@lru_cache
+def _openai_version() -> Tuple[int, int, int]:
+    return cast(Tuple[int, int, int], tuple(map(int, version("openai").split(".")[:3])))
```
mikeldking, line 31:
```suggestion
@lru_cache
def _get_openai_version() -> Tuple[int, int, int]:
    return cast(Tuple[int, int, int], tuple(map(int, version("openai").split(".")[:3])))
```
a bit more clear that it's a function rather than a constant

mikeldking (approved):

PR #235, opened 2024-02-21
feat: add status description for openai instrumentor

### LINKED ISSUE #233: add status description to the openai instrumentor
Added a description string alongside the span status code when there's an error. The description should be formatted in the same way as the OTEL SDK as seen [here](https://github.com/open-telemetry/opentelemetry-python/blob/2b9dcfc5d853d1c10176937a6bcaade54cda1a31/opentelemetry-api/src/opentelemetry/trace/__init__.py#L588)

RogerHYang:
resolves #233 

Added a description string alongside the span status code when an exception occurs.

The description is formatted in the same way as the OTEL SDK as seen [here](https://github.com/open-telemetry/opentelemetry-python/blob/2b9dcfc5d853d1c10176937a6bcaade54cda1a31/opentelemetry-api/src/opentelemetry/trace/__init__.py#L588), i.e. 
```python
f"{type(exception).__name__}: {exception}"
```

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
+++ b/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
@@ -231,7 +231,7 @@ class _WithOpenAI(ABC):
                 response_accumulator=response_accumulator,
             )
         _finish_tracing(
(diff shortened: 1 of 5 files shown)


mikeldking (approved):
Don't forget to add some color to the PR description

PR #249, opened 2024-03-01
fix: concurrent tracing for llama-index

### LINKED ISSUE #248: [bug] llama-index callback not able to retain multiple concurrent traces
There are a few issues that are intertwined (one by us and two by llama-index).
1. Our callback implementation can't handle multiple concurrent traces.
    - To do this we need to rely on `trace_map` to differentiate concurrent spans from separate traces.
2. The `trace_map` from llama-index can't handle multiple concurrent traces.
    - It seems to be able to retain only one of traces when there are supposed to be multiple.
3. The `trace_map` from llama-index is sometimes empty when `end_trace` is called.
    - This is a problem because we need to rely on `trace_map` to evict spans by the traces they belong to.

No. 2 can be shown by llama-index's debug handler:
```python
llama_index.core.global_handler = LlamaDebugHandler(print_trace_on_end=True)
```
The output from the debug handler shows that all concurrent traces have identical timing. This is due to the fact that the `trace_map` variable in llama-index can only retain a single set of traces. This is a problem for us regarding issue No. 1 above because we have no way to differentiate concurrent spans other than relying on llama-index's `trace_map` input.

Screenshot 2024-03-01 at 11 24 33 AM

Below is code for testing async concurrent tracing.

```python
import asyncio

import chromadb
from llama_index.core import SimpleDirectoryReader, StorageContext, VectorStoreIndex
from llama_index.vector_stores.chroma import ChromaVectorStore
from openinference.instrumentation.llama_index import LlamaIndexInstrumentor
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk import trace as trace_sdk

endpoint = "http://127.0.0.1:6006/v1/traces"
tracer_provider = trace_sdk.TracerProvider()
tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter(endpoint)))

LlamaIndexInstrumentor().instrument()
# llama_index.core.global_handler = LlamaDebugHandler(print_trace_on_end=True)

chroma_client = chromadb.EphemeralClient()
chroma_collection = chroma_client.create_collection("essays")
documents = SimpleDirectoryReader("./data/paul_graham/").load_data()
vector_store = ChromaVectorStore(chroma_collection)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
index = VectorStoreIndex.from_documents(documents, storage_context=storage_context)


async def main():
    tasks = [
        query_engine.aquery("What did the author do growing up?"),
        query_engine.aquery("Who is the author?"),
        query_engine.aquery("Where did the author go to college?"),
        query_engine.aquery("Where is the author's hometown?"),
    ]
    return await asyncio.gather(*tasks)


if __name__ == "__main__":
    query_engine = index.as_query_engine()
    asyncio.run(main())
```

RogerHYang:
resolves #248

- Handle templating events separately so we don't need to construct a trace tree, so spans can be exported as soon as they are finished.


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -314,90 +330,56 @@ def on_event_end(
...
 
-    def start_trace(self, trace_id: Optional[str] = None) -> None:
-        self._event_data.clear()
+    def start_trace(
+        self,
+        trace_id: Optional[str] = None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        """This is intentionally empty."""
```
mikeldking, line 348: maybe explain why it's intentionally empty

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -314,90 +330,56 @@ def on_event_end(
...
+    which case the LLM event will never be popped and will remain in the container forever.
     https://github.com/run-llama/llama_index/blob/dcef41ee67925cccf1ee7bb2dd386bcf0564ba29/llama_index/llms/base.py#L62
     Therefore, to prevent memory leak, this container is limited to a certain capacity, and when it
-    exceeds that capacity, the oldest straggler by insertion order is popped and are forced to
-    finish tracing.
+    reaches that capacity, the oldest item by insertion order will be popped.
     """  # noqa: E501
 
-    def __init__(self, capacity: int) -> None:
+    def __init__(self, capacity: int = 1000, fn: Optional[Callable[[_Value], None]] = None) -> None:
```
mikeldking, line 372: `fn` is a bit vague. Can you re-name it? this is the callback when an item get's evicted right? `on_evict_fn` or something like that would help make this more readable

mikeldking (approved):

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
RogerHYang: done

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
RogerHYang: done

PR #255, opened 2024-03-06
fix: add support for DSPy's Google model class

axiomofjoy:
- adds support for DSPY's `Google` LLM class

### DIFF
--- a/python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
@@ -2,7 +2,18 @@ import json
 from abc import ABC
 from enum import Enum
 from inspect import signature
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -521,6 +536,37 @@ def _flatten(mapping: Mapping[str, Any]) -> Iterator[Tuple[str, AttributeValue]]
...
+
+
+def _parse_google_response(response: "GenerateContentResponse") -> Dict[str, Any]:
+    """
+    Parses a response from the Google generative AI SDK into a dictionary.
+    """
+
+    return {
+        "text": response.text,
+    }
```
mikeldking, line 567: is it worth doing a follow-up to map these things to semantic conventions?

mikeldking (approved):

PR #304, opened 2024-03-20
fix: use start and end times from langchain

RogerHYang:
This is necessary because the actual timing is no longer real when the handler is called in a separate thread (which can take place long after the run is done).


mikeldking:
@RogerHYang add a bit more color that you need to respect the timing from langchain when running in a separate thread.

Also is this testable? Not blocking but would be good to have some coverage at some point.

mikeldking (approved):

PR #347, opened 2024-03-27
feat: move openai example to express

mikeldking:
- [ ] add custom span via middleware or per route
- [x] add streaming endpoint


js/examples/openai/src/index.ts
```diff
@@ -22,6 +23,23 @@ app.post("/completion", async (req, res) => {
   });
 });
 
+// Streaming
+app.post("/completion/stream", async (req, res) => {
```
mikeldking, line 27: You can add multiple endpoints ala langserve. Let's add one for each. /completion /stream

Parker-Stafford (approved):

PR #362, opened 2024-04-05
feat(langchainjs): patch callbacks module add tracer and input / output support

Parker-Stafford:
- Adds langchainjs instrumentation with support for OpenInference span kind, inputs, outpus

### DIFF
--- a/js/.eslintrc.js
+++ b/js/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
     browser: true,
     es2021: true,
   },
+  ignorePatterns: ["dist/"],
   extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
   overrides: [
     {
--- a/js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
+++ b/js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
@@ -1,26 +1,7 @@
-// import "../instrumentation";
-import { LangChainInstrumentation } from "../../src";
-import {
-  InMemorySpanExporter,
-  SimpleSpanProcessor,
-} from "@opentelemetry/sdk-trace-base";
-import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
+import "../instrumentation";
 import "dotenv/config";
-
-const tracerProvider = new NodeTracerProvider();
-tracerProvider.register();
-
-const memoryExporter = new InMemorySpanExporter();
-const provider = new NodeTracerProvider();
-provider.getTracer("default");
-
-const instrumentation = new LangChainInstrumentation();
-instrumentation.setTracerProvider(tracerProvider);
-tracerProvider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
-
-instrumentation.enable();
-
 import { ChatOpenAI } from "@langchain/openai";
+
 const main = async () => {
   const chatModel = new ChatOpenAI({
     openAIApiKey: process.env.OPENAI_API_KEY,
--- a/js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
+++ b/js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
@@ -1,0 +1,33 @@
+// import "../instrumentation";
+import { LangChainInstrumentation } from "../../src";
+import {
+  InMemorySpanExporter,
+  SimpleSpanProcessor,
+} from "@opentelemetry/sdk-trace-base";
+import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
+import "dotenv/config";
+
+const tracerProvider = new NodeTracerProvider();
+tracerProvider.register();
+
+const memoryExporter = new InMemorySpanExporter();
+const provider = new NodeTracerProvider();
+provider.getTracer("default");
+
+const instrumentation = new LangChainInstrumentation();
+instrumentation.setTracerProvider(tracerProvider);
+tracerProvider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
+
+instrumentation.enable();
+
+import { ChatOpenAI } from "@langchain/openai";
+const main = async () => {
+  const chatModel = new ChatOpenAI({
+    openAIApiKey: process.env.OPENAI_API_KEY,
+  });
+
+  const response = await chatModel.invoke("test");
+  return response;
+};
+
+main();
--- a/js/packages/openinference-instrumentation-langchain/jest.config.js
+++ b/js/packages/openinference-instrumentation-langchain/jest.config.js
@@ -3,4 +3,5 @@ module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
   prettierPath: null,
+  setupFiles: ["dotenv/config"],
 };
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -26,7 +26,10 @@
   "author": "oss-devs@arize.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "@arizeai/openinference-semantic-conventions": "workspace:*"
+    "@arizeai/openinference-semantic-conventions": "workspace:*",
+    "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0",
+    "@opentelemetry/instrumentation": "^0.46.0"
   },
   "devDependencies": {
     "@langchain/community": "^0.0.44",
@@ -39,10 +42,7 @@
     "dotenv": "^16.4.5",
     "jest": "^29.7.0",
     "langchain": "^0.1.30",
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.23.0",
     "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
-    "@opentelemetry/instrumentation": "^0.46.0",
     "@opentelemetry/resources": "^1.19.0",
     "@opentelemetry/semantic-conventions": "^1.19.0"
   }
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -28,12 +28,14 @@
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "workspace:*",
     "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0",
     "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
     "@opentelemetry/instrumentation": "^0.46.0",
     "@opentelemetry/resources": "^1.19.0",
     "@opentelemetry/semantic-conventions": "^1.19.0"
   },
   "devDependencies": {
+    "@langchain/community": "^0.0.44",
     "@langchain/core": "^0.1.52",
     "@langchain/openai": "^0.0.25",
     "@opentelemetry/sdk-trace-base": "^1.19.0",
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -26,12 +26,14 @@
   "author": "oss-devs@arize.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "@arizeai/openinference-semantic-conventions": "workspace:*"
+    "@arizeai/openinference-semantic-conventions": "workspace:*",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0"
   },
   "devDependencies": {
-    "@langchain/community": "^0.0.44",
-    "@langchain/core": "^0.1.52",
     "@langchain/openai": "^0.0.25",
+    "@langchain/core": "^0.1.52",
     "@opentelemetry/sdk-trace-base": "^1.19.0",
     "@opentelemetry/sdk-trace-node": "^1.19.0",
     "@types/jest": "^29.5.11",
@@ -41,9 +43,6 @@
     "langchain": "^0.1.30",
     "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
     "@opentelemetry/resources": "^1.19.0",
-    "@opentelemetry/semantic-conventions": "^1.19.0",
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.23.0",
-    "@opentelemetry/instrumentation": "^0.46.0"
+    "@opentelemetry/semantic-conventions": "^1.19.0"
   }
 }
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -7,21 +7,41 @@
   "module": "dist/esm/index.js",
   "esnext": "dist/esnext/index.js",
   "types": "dist/src/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Arize-ai/openinference.git"
+  },
   "scripts": {
     "prebuild": "rimraf dist & pnpm run version:update",
     "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
     "version:update": "../../scripts/version-update.js",
-    "type:check": "tsc --noEmit"
+    "type:check": "tsc --noEmit",
+    "test": "jest --setupFiles dotenv/config"
   },
   "keywords": [],
+  "files": [
+    "dist",
+    "src"
+  ],
   "author": "oss-devs@arize.com",
   "license": "Apache-2.0",
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "workspace:*",
     "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/instrumentation": "^0.46.0"
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@opentelemetry/resources": "^1.19.0",
+    "@opentelemetry/semantic-conventions": "^1.19.0"
   },
   "devDependencies": {
+    "@langchain/core": "^0.1.52",
+    "@langchain/openai": "^0.0.25",
+    "@opentelemetry/sdk-trace-base": "^1.19.0",
+    "@opentelemetry/sdk-trace-node": "^1.19.0",
+    "@types/jest": "^29.5.11",
+    "@types/node": "^20.12.4",
+    "dotenv": "^16.4.5",
+    "jest": "^29.7.0",
     "langchain": "^0.1.30"
   }
 }
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -26,10 +26,7 @@
   "author": "oss-devs@arize.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "@arizeai/openinference-semantic-conventions": "workspace:*",
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.23.0",
-    "@opentelemetry/instrumentation": "^0.46.0"
+    "@arizeai/openinference-semantic-conventions": "workspace:*"
   },
   "devDependencies": {
     "@langchain/community": "^0.0.44",
@@ -44,6 +41,9 @@
     "langchain": "^0.1.30",
     "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
     "@opentelemetry/resources": "^1.19.0",
-    "@opentelemetry/semantic-conventions": "^1.19.0"
+    "@opentelemetry/semantic-conventions": "^1.19.0",
+    "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0",
+    "@opentelemetry/instrumentation": "^0.46.0"
   }
 }
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -26,13 +26,7 @@
   "author": "oss-devs@arize.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "@arizeai/openinference-semantic-conventions": "workspace:*",
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.23.0",
-    "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
-    "@opentelemetry/instrumentation": "^0.46.0",
-    "@opentelemetry/resources": "^1.19.0",
-    "@opentelemetry/semantic-conventions": "^1.19.0"
+    "@arizeai/openinference-semantic-conventions": "workspace:*"
   },
   "devDependencies": {
     "@langchain/community": "^0.0.44",
@@ -44,6 +38,12 @@
     "@types/node": "^20.12.4",
     "dotenv": "^16.4.5",
     "jest": "^29.7.0",
-    "langchain": "^0.1.30"
+    "langchain": "^0.1.30",
+    "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@opentelemetry/resources": "^1.19.0",
+    "@opentelemetry/semantic-conventions": "^1.19.0"
   }
 }
--- a/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
@@ -1,0 +1,108 @@
+import * as CallbackManagerModule from "@langchain/core/callbacks/manager";
+import {
+  InstrumentationBase,
+  InstrumentationConfig,
+  InstrumentationModuleDefinition,
+  InstrumentationNodeModuleDefinition,
+  isWrapped,
+} from "@opentelemetry/instrumentation";
+import { VERSION } from "./version";
+import { diag } from "@opentelemetry/api";
+import { LangChainTracer } from "./tracer";
+
+const MODULE_NAME = "@langchain/core";
+
+export class LangChainInstrumentation extends InstrumentationBase<
+  typeof CallbackManagerModule
+> {
+  constructor(config?: InstrumentationConfig) {
+    super(
+      "@arizeai/openinference-instrumentation-langchain",
+      VERSION,
+      Object.assign({}, config),
+    );
+  }
+
+  protected init(): InstrumentationModuleDefinition<
+    typeof CallbackManagerModule
+  > {
+    const module = new InstrumentationNodeModuleDefinition(
+      "@langchain/core/callbacks/manager",
+      ["^0.1.0"],
+      this.patch.bind(this),
+      this.unpatch.bind(this),
+    );
+
+    return module;
+  }
+
+  private patch(
+    module: typeof CallbackManagerModule & {
+      openInferencePatched?: boolean;
+    },
+    moduleVersion?: string,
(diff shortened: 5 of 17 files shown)


js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
```diff
@@ -0,0 +1,14 @@
+import "../instrumentation";
```
Parker-Stafford, line 1: will remove examples before final merge

mikeldking: Eh you can leave it. though the folder structure seems a dad strange. Why not just a flat file?

js/packages/openinference-instrumentation-langchain/jest.config.js
```diff
@@ -3,4 +3,5 @@ module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
   prettierPath: null,
+  setupFiles: ["dotenv/config"],
```
mikeldking, line 6: This needs to be removed overall since it won't be a unit test

js/packages/openinference-instrumentation-langchain/package.json
```diff
@@ -7,21 +7,43 @@
...
   "license": "Apache-2.0",
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "workspace:*",
     "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/instrumentation": "^0.46.0"
+    "@opentelemetry/core": "^1.23.0",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@opentelemetry/resources": "^1.19.0",
+    "@opentelemetry/semantic-conventions": "^1.19.0"
```
mikeldking, line 35: these need to be moved to devDependancies. If these are dependancies, it will cause separate copies to be installed for this module.

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  protected init(): InstrumentationModuleDefinition<
+    typeof CallbackManagerModule
+  > {
+    const module = new InstrumentationNodeModuleDefinition<
+      typeof CallbackManagerModule
+    >(
+      "@langchain/core/dist/callbacks/manager.cjs",
+      ["^0.0.0", "^0.1.0"],
```
mikeldking, line 33: why the list? doesn't the former sufice?

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+    if (module?.openInferencePatched) {
+      return module;
+    }
+
+    this._wrap(module.CallbackManager, "configure", (original) => {
+      return (...args: Parameters<typeof original>) => {
+        const inheritableHandlers = args[0];
+        const newInheritableHandlers =
+          this._addTracerToHandlers(inheritableHandlers);
```
mikeldking, line 60:
```suggestion
          this.addTracerToHandlers(inheritableHandlers);
```

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+      }
+      handlers.addHandler(new LangChainTracer(this.tracer), true);
+      return handlers;
+    }
+    const newHandlers = handlers ?? [];
+    if (!newHandlers.some((handler) => handler instanceof LangChainTracer)) {
+      newHandlers.push(new LangChainTracer(this.tracer));
+    }
+    return newHandlers;
+  }
```
mikeldking, line 111: doesn't need to be a method, it's called once.

js/.eslintrc.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     browser: true,
     es2021: true,
   },
+  ignorePatterns: ["dist/"],
```
Parker-Stafford, line 6: eslint ignore any type was not getting added to .d files during build so declared types with any and eslint ignores were not being ignored after being compiled, don't think we need to lint dist anyways

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+  private _addTracerToHandlers(handlers?: CallbackManagerModule.Callbacks) {
+    if (handlers == null) {
+      return [new LangChainTracer(this.tracer)];
+    }
+    if (handlers instanceof CallbackManagerModule.CallbackManager) {
+      if (
+        handlers.inheritableHandlers.some(
+          (handler) => handler instanceof LangChainTracer,
+        )
+      ) {
```
mikeldking, line 100: does this actually hit? also you can write this so that it's more readable
```suggestion
      tracer_already__registered = handlers.inheritableHandlers.some(
          (handler) => handler instanceof LangChainTracer,
        );
      if (
        tracer_already__registered
      ) {
```

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+        return handlers;
+      }
+      handlers.addHandler(new LangChainTracer(this.tracer), true);
+      return handlers;
+    }
+    const newHandlers = handlers ?? [];
+    if (!newHandlers.some((handler) => handler instanceof LangChainTracer)) {
```
mikeldking, line 107: this is duplicate logic as above. You can re-order the logic and only do this once. E.g. move this up

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    this._wrap(module.CallbackManager, "configure", (original) => {
+      return (...args: Parameters<typeof original>) => {
+        const inheritableHandlers = args[0];
+        const newInheritableHandlers =
+          this._addTracerToHandlers(inheritableHandlers);
+        args[0] = newInheritableHandlers;
+        const manager = original.apply(this, args);
+        return manager;
```
mikeldking, line 63:
```suggestion
        return original.apply(this, args);
```

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,107 @@
...
+  private tracer: Tracer;
+  private runs: Record<string, RunWithSpan | undefined> = {};
+  constructor(tracer: Tracer) {
+    super();
+    this.tracer = tracer;
+  }
+  name: string = "OpenInferenceLangChainTracer";
+  protected persistRun(_run: Run): Promise<void> {
+    return Promise.resolve();
+  }
```
mikeldking, line 35: this needed?

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  protected async _startTrace(run: Run) {
+    await super._startTrace(run);
+    if (isTracingSuppressed(context.active())) {
```
mikeldking, line 39: nice

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    runWithSpan.span.end();
+  }
+
+  private getParentSpanContext(run: Run) {
+    if (run.parent_run_id == null) {
+      return;
+    }
+    const maybeParent = this.runs[run.parent_run_id];
```
mikeldking, line 100: this map is infinitely growing no?

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,107 @@
...
+    await super._startTrace(run);
+    if (isTracingSuppressed(context.active())) {
+      return;
+    }
+
+    let activeContext = context.active();
+    const parentCtx = this.getParentSpanContext(run);
+    if (parentCtx != null) {
+      activeContext = trace.setSpanContext(context.active(), parentCtx);
+    }
```
mikeldking, line 47: this bit is a tad confusing to me and requires a bit of explanation I think.

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,107 @@
...
+} from "./utils";
+
+type RunWithSpan = {
+  run: Run;
+  span: Span;
+};
+
+export class LangChainTracer extends BaseTracer {
+  private tracer: Tracer;
+  private runs: Record<string, RunWithSpan | undefined> = {};
```
mikeldking, line 27: This is not getting GC'd

js/packages/openinference-instrumentation-langchain/src/typeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A type-guard function for checking if a value is an object with string keys
+ */
+export function isObject(value: unknown): value is Record<string, unknown> {
+  return (
+    typeof value === "object" &&
+    value != null &&
+    !Array.isArray(value) &&
+    Object.keys(value).every((key) => typeof key === "string")
```
mikeldking, line 19: The function name doesn't match the implementation. Best to name it accordingly `isObjectWithStringKeys`

p.s. why is this check needed for string keys?

js/packages/openinference-instrumentation-langchain/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  [SemanticConventions.MESSAGE_FUNCTION_CALL_ARGUMENTS_JSON]?: string;
+  [SemanticConventions.MESSAGE_TOOL_CALLS]?: LLMMessageToolCall[];
+};
+
+export type LLMMessageFunctionCall = Pick<
+  LLMMessage,
+  "message.function_call_name" | "message.function_call_arguments_json"
+>;
+
+export type LLMMessageToolCalls = Pick<LLMMessage, "message.tool_calls">;
```
mikeldking, line 20: curious why the Pick approach was taken here. Wouldn't it be better to do a Union type of sub-types.

js/packages/openinference-instrumentation-langchain/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+export type LLMMessagesAttributes =
+  | {
+      [SemanticConventions.LLM_INPUT_MESSAGES]: LLMMessage[];
+    }
+  | {
+      [SemanticConventions.LLM_OUTPUT_MESSAGES]: LLMMessage[];
+    };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type GenericFunction = (...args: any[]) => any;
```
mikeldking, line 31: can you skip the any and just make it unknown?

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+export function withSafety<T extends GenericFunction>(fn: T): SafeFunction<T> {
+  return (...args) => {
+    try {
+      return fn(...args);
+    } catch (error) {
+      diag.error(`Failed to get attributes for span: ${error}`);
+      return null;
+    }
+  };
+}
```
mikeldking, line 33: nice!

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Flattens nested attributes into a single level object
+ */
+function flattenAttributes(
+  attributes: Record<string, unknown>,
+  baseKey: string = "",
```
mikeldking, line 40: can you add jsdocs for these? It will help the next dev.

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+  for (const key in attributes) {
+    const newKey = baseKey ? `${baseKey}.${key}` : key;
+    const value = attributes[key];
+
+    if (value == null) {
+      continue;
+    }
+
+    if (isObject(value)) {
+      Object.assign(result, flattenAttributes(value, newKey));
```
mikeldking, line 52: just a note that this is modifying the result here. It looks safe but might want a unit test to make sure you are not modifying anything.

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+  return "UNKNOWN";
+}
+
+function formatIO({
+  io,
+  type,
+}: {
+  io: Run["inputs"] | Run["outputs"];
+  type: "input" | "output";
```
mikeldking, line 87: I'd avoid the word type - it's a reserved field. ioType might work.

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+      normalizedRunType as keyof typeof OpenInferenceSpanKind
+    ];
+  }
+  return "UNKNOWN";
+}
+
+function formatIO({
+  io,
+  type,
+}: {
```
mikeldking, line 85: add JSDoc

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  return {
+    [valueAttribute]: JSON.stringify(io),
+    [mimeTypeAttribute]: MimeType.JSON,
+  };
+}
+
+function getRoleFromMessageData(
+  messageData: Record<string, unknown>,
+): string | undefined {
```
mikeldking, line 124:
```suggestion
): ?string {
```

mikeldking: any reason we wouldn't provide a fallback?

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+    isObject(messageData.kwargs) &&
+    isString(messageData.kwargs.role)
+  ) {
+    return messageData.kwargs.role;
+  }
+}
+
+function getContentFromMessageData(
+  messageKwargs: Record<string, unknown>,
+): string | undefined {
```
mikeldking, line 156: any reason you are using undefined over null?

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+      [SemanticConventions.TOOL_CALL_FUNCTION_NAME]: toolCallName,
+      [SemanticConventions.TOOL_CALL_FUNCTION_ARGUMENTS_JSON]: toolCallArgs,
+    };
+  });
+  return {
+    [SemanticConventions.MESSAGE_TOOL_CALLS]: formattedToolCalls,
+  };
+}
+
+function parseMessage(messageData: Record<string, unknown>): LLMMessage {
```
mikeldking, line 207: JSDoc

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+  return {
+    ...message,
+    ...getFunctionCallDataFromAdditionalKwargs(additionalKwargs),
+    ...getToolCallDataFromAdditionalKwargs(additionalKwargs),
+  };
+}
+
+function formatInputMessages(
+  input: Run["inputs"],
+): LLMMessagesAttributes | null {
```
mikeldking, line 233: inconsistent return type - I think null is better

js/packages/openinference-instrumentation-langchain/src/utils.ts
Parker-Stafford: Yeah so my thought was with the undefined and this applies to the comment below as well was that if there is no role or content I don't want it to show up in the attributes as opposed to showing up as null, I just wanted the key to not exist given the logic on line 211 and 217. I can refactor though to do null checks on the role / content and only add the key if not null.

In terms of fallback, I was assuming we only wanted to put things in attributes that actually existed a fallback seems like it wouldn't make sense in most contexts like other models where the roles are inline

js/packages/openinference-instrumentation-langchain/src/utils.ts
Parker-Stafford: ahh good point, will spread just to be safe

js/packages/openinference-instrumentation-langchain/src/types.ts
Parker-Stafford: yeah it was complaining since unknown is a bit more strict than any, will give another pass and see

js/packages/openinference-instrumentation-langchain/src/types.ts
Parker-Stafford: I think it was just order of operations, i wrote the top type first then extracted from it, might be more clear to do a union of the bjects though so can do that

js/packages/openinference-instrumentation-langchain/src/typeUtils.ts
Parker-Stafford: This was because lc has things typed as Record<string, any> for many of the properties on a run, so just knowing it was an object wasn't enough to make the types happy, I'll give it another pass though to see if i can make this less strict and still get away with it

js/packages/openinference-instrumentation-langchain/src/tracer.ts
Parker-Stafford: the runs? it should when the class gets gc'd which i think is what we want, I'll have to look at the logic a bit more to make sure one of these doesn't get created then infinitely grows but it should be the case that this only gets created when a callback manager gets created and then gets gc'd when the handlers are not used anymore

js/packages/openinference-instrumentation-langchain/src/tracer.ts
Parker-Stafford: will add some comments

js/packages/openinference-instrumentation-langchain/src/tracer.ts
Parker-Stafford: yeah in order to extend BaseTracer, but i should call teh super in here so will do that

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
Parker-Stafford: I believe it can when a callback manager gets copied which is apart of their functionality

js/packages/openinference-instrumentation-langchain/src/tracer.ts
mikeldking: No but you need to evict stuff from this map, otherwise it's hanging on to the mapping for the lifespan of the tracer. This will grow in memory.

js/packages/openinference-instrumentation-langchain/src/utils.ts
mikeldking: makes sense. no fallback it is.

js/packages/openinference-instrumentation-langchain/src/tracer.ts
Parker-Stafford: Ahh i gotcha, yeah i can evict them when the span ends since a parent should never end before it's child correct?

js/packages/openinference-instrumentation-langchain/jest.config.js
Parker-Stafford: Was going to remove before the release but not sure i follow the "since it won't be a unit test"

js/packages/openinference-instrumentation-langchain/package.json
Parker-Stafford: Oh gotcha, i see some of these as direct deps for openai instrumentation, how do we determine when they should be direct vs. dev deps?

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
Parker-Stafford: this is actually a different check, the handlers are either

- null - first check
- a callback maanager with a set of handlers on it - second check
- an array of callback handlers - this check

Each of the check narrows it down to this one where now we know we have an array of handlers and either want to add ours or it already exists.

I don't need the nullish coalescing here though since at this point we know it's an array.

js/packages/openinference-instrumentation-langchain/src/tracer.ts
Parker-Stafford: actually BaseTracer is an abstract class, which cannot be instantiated but only extended / inherited from. the persist Run method is an abstract method which means it must be defined by whatever class inherits from the BaseTracer class since the BaseTracer class doesn't have it's own implementation of it.

https://www.tutorialsteacher.com/typescript/abstract-class so basically have to put something here but it's unused so just matched the type

js/packages/openinference-instrumentation-langchain/src/typeUtils.ts
Parker-Stafford: updated!

js/packages/openinference-instrumentation-langchain/src/types.ts
Parker-Stafford: Going to leave as any requires casting everything and losing type inference for params and return types if changed to unknown

js/packages/openinference-instrumentation-langchain/src/utils.ts
Parker-Stafford: I think this is what we want, result is created at the top of this funciton and returned at the bottom, we want to be modifying it directly as we go long

js/packages/openinference-instrumentation-langchain/src/utils.ts
Parker-Stafford: updated to null

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
Parker-Stafford: I think we need both because of this left-most non zero element in the version, since the major version is 0

> Allows changes that do not modify the left-most non-zero element in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

https://github.com/isaacs/node-semver?tab=readme-ov-file#caret-ranges-123-025-004

js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
Parker-Stafford: oh the structure was from before when i was troubleshooting patching can flatten

js/packages/openinference-instrumentation-langchain/src/tracer.ts
mikeldking: might consider https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

js/packages/openinference-instrumentation-langchain/examples/chat/chat.ts
mikeldking: Okay, might just consider flattening.

js/packages/openinference-instrumentation-langchain/jest.config.js
mikeldking: it's not a unit test if it requires a .env file. I'd probably just set your keys via your shell right now so you don't need this but NBD

js/packages/openinference-instrumentation-langchain/package.json
```diff
@@ -7,21 +7,43 @@
...
+    "@types/node": "^20.12.4",
+    "dotenv": "^16.4.5",
+    "jest": "^29.7.0",
+    "langchain": "^0.1.30",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
+    "@opentelemetry/resources": "^1.19.0",
+    "@opentelemetry/semantic-conventions": "^1.19.0",
+    "@opentelemetry/api": "^1.7.0",
+    "@opentelemetry/core": "^1.23.0",
+    "@opentelemetry/instrumentation": "^0.46.0"
```
mikeldking, line 47: these are peer deps no? I might have gotten this wrong in the other module.

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
mikeldking: hmmm, okay so we would have to update when they releasee 0.2 I guess?

js/packages/openinference-instrumentation-langchain/src/tracer.ts
mikeldking: K, sg

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -0,0 +1,112 @@
...
+      ...safelyFormatIO({ io: run.outputs, ioType: "output" }),
+      ...safelyFormatInputMessages(run.inputs),
+      ...safelyFormatOutputMessages(run.outputs),
+    });
+    if (attributes != null) {
+      span.setAttributes(attributes);
+    }
+
+    runWithSpan.span.end();
+    delete this.runs[run.id];
```
mikeldking, line 98: might want to add a test for this. Make sure it doesn't grow

js/packages/openinference-instrumentation-langchain/src/types.ts
mikeldking: k

js/packages/openinference-instrumentation-langchain/src/utils.ts
mikeldking: k

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    };
+  });
+  return {
+    [SemanticConventions.MESSAGE_TOOL_CALLS]: formattedToolCalls,
+  };
+}
+
+/**
+ * Parses a langchain message into OpenInference attributes.
+ * @param messageData The langchain message data to parse
```
mikeldking, line 239: minor nit. Best to try to follow true JSDoc if possible

```
/**
 * Represents a book.
 * @constructor
 * @param {string} title - The title of the book.
 * @param {string} author - The author of the book.
 */

File: js/packages/openinference-instrumentation-langchain/tsconfig.json [hunk 31]
```diff
@@ -2,9 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "."
+    "rootDir": ".",
+    "types": ["node", "jest"]
   },
   "files": [],
-  "include": ["src/**/*.ts", "test/**/*.ts"],
+  "include": ["src/**/*.ts"],
```
mikeldking, line 9: no type checking in the tests?

mikeldking (approved):

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
Parker-Stafford: yes, i believe so

js/packages/openinference-instrumentation-langchain/src/utils.ts
Parker-Stafford: ahh yeah i removed the types cause they weren't showing up but can add them in, also i see the - in teh spec, but it also gets added automatically on hover in vscode so i was a bit confused, will add

js/packages/openinference-instrumentation-langchain/tsconfig.json
Parker-Stafford: i will add this back just removed cause tests were messy as i was working and it was annoying to fix types before i could run / debug

js/packages/openinference-instrumentation-langchain/jest.config.js
Parker-Stafford: ahh i see what you mean, my keys are in my shell but for some reasonweren't getting picked up without this in iether case will definitely remove before final merge

PR #371, opened 2024-04-12
feat(langchainjs): support top level function and tool calling

Parker-Stafford:
Adds support for
- top level function and tool calling
- prompt templates
- metadata
- token counts

### DIFF
--- a/js/packages/openinference-instrumentation-langchain/src/constants.ts
+++ b/js/packages/openinference-instrumentation-langchain/src/constants.ts
@@ -1,0 +1,23 @@
+import {
+  LLMAttributePostfixes,
+  SemanticAttributePrefixes,
+  ToolAttributePostfixes,
+} from "@arizeai/openinference-semantic-conventions";
+
+const PROMPT_TEMPLATE_PREFIX =
+  `${SemanticAttributePrefixes.llm}.${LLMAttributePostfixes.prompt_template}` as const;
+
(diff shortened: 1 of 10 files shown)


js/packages/openinference-instrumentation-langchain/src/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+  `${PROMPT_TEMPLATE_PREFIX}.template` as const;
+
+export const LLM_FUNCTION_CALL =
+  `${SemanticAttributePrefixes.llm}.function_call` as const;
+
+export const TOOL_NAME =
+  `${SemanticAttributePrefixes.tool}.${ToolAttributePostfixes.name}` as const;
+
+export const TOOL_DESCRIPTION =
+  `${SemanticAttributePrefixes.tool}.${ToolAttributePostfixes.description}` as const;
```
mikeldking, line 23: you can move this to the semantic conventions and re-build and you should be able to import. We can do it after though if that feels safer

mikeldking (approved):

js/packages/openinference-instrumentation-langchain/src/constants.ts
Parker-Stafford: Yup was going to once i had a cleanish slate on the langchainjs pr, just wanted to move fast last night and not switch branches / contexts. Will do before release!

PR #382, opened 2024-04-16
feat(langchainjs): add manual instrumentation method to langchain instrumentation

Parker-Stafford:
Adds a manuallyInstrument method to the LangChainInstrumentation class to allow users to manually instrument Langchain. This is necessary because the `InstrumentationNodeModuleDefinition` will not work with packages that don't have a main in package.json and an index.js see here https://github.com/elastic/require-in-the-middle/blob/main/index.js#L259.

### DIFF
--- a/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
@@ -24,6 +24,7 @@ export class LangChainInstrumentation extends InstrumentationBase<
   }
 
   manuallyInstrument(module: typeof CallbackManagerModule) {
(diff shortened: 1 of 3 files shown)


js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -1,4 +1,4 @@
-import type * as CallbackManagerModule from "@langchain/core/callbacks/manager";
+import * as CallbackManagerModule from "@langchain/core/callbacks/manager";
```
mikeldking, line 1: is the type removed on purpose?

mikeldking (approved):

PR #385, opened 2024-04-16
fix(langchainjs): update fallback span kind to be CHAIN

Parker-Stafford:


js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -103,7 +103,7 @@ function getOpenInferenceSpanKindFromRunType(runType: string) {
       normalizedRunType as keyof typeof OpenInferenceSpanKind
     ];
   }
-  return "UNKNOWN";
+  return OpenInferenceSpanKind.AGENT;
```
mikeldking:
```suggestion
  return OpenInferenceSpanKind.CHAIN;
```

mikeldking (approved):

PR #394, opened 2024-04-22
fix: convert numpy values for protobuf

RogerHYang:
Document scores are sometimes numpy values, which will be discarded by protobuf exporter.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
+++ b/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
@@ -672,9 +672,14 @@ def _flatten(mapping: Mapping[str, Any]) -> Iterator[Tuple[str, AttributeValue]]
         else:
             if isinstance(value, Enum):
                 value = value.value
-            elif isinstance(value, SupportsFloat) and not isinstance(value, (int, float)):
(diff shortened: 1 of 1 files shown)


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -671,6 +672,15 @@ def _flatten(mapping: Mapping[str, Any]) -> Iterator[Tuple[str, AttributeValue]]
...
             if isinstance(value, Enum):
                 value = value.value
+            elif isinstance(value, SupportsFloat) and not isinstance(
+                value,
+                (int, float, Iterable),
+            ):
+                # This for when there are numpy values, which will be rejected by protobuf.
+                # We convert all of them to float, so we don't need a dependency on numpy.
+                # The check on Iterable is to avoid converting numpy arrays to float,
+                # because SupportsFloat is true for numpy arrays.
```
axiomofjoy, line 682:
```suggestion
                # This is for when there are numpy values, which will be rejected by protobuf.
                # We convert all of them to float, so we don't need a dependency on numpy.
                # The check on Iterable is to avoid converting numpy arrays to float,
                # because numpy arrays are instances of SupportsFloat.
```

axiomofjoy (approved):

PR #397, opened 2024-04-23
chore: add symlinks to facilitate editable install

RogerHYang:
```
pip install -e python/openinference-instrumentation/
```

Because those symlinks need to be excluded from mypy, the global mypy config is deprecated in favor of local configs in pyproject.toml.


python/openinference-instrumentation/pyproject.toml
```diff
@@ -40,10 +40,39 @@ Homepage = "https://github.com/Arize-ai/openinference/tree/main/python/openinfer
 path = "src/openinference/instrumentation/version.py"
 
 [tool.hatch.build.targets.sdist]
-include = [
-  "/src",
-  "/tests",
+exclude = [
+  "src/openinference/instrumentation/*/*",
 ]
```
axiomofjoy, line 45: I'm a little confused, doesn't the exclude also need to be in each instrumentation library that depends on `openinference-instrumentation`?

axiomofjoy (approved):

python/openinference-instrumentation/pyproject.toml
axiomofjoy: Resolved offline.

PR #414, opened 2024-05-01
feat: Create `using_attributes` context managers and have openai instrumentation read from it

fjcasti1:


axiomofjoy:
@fjcasti1 Can you break this into separate PRs for the different libraries? The reason being that targeting multiple libraries at once in a single PR will cause a race condition in the release pipeline. It's not hard to resolve manually, but it's simpler to make each PR update only a single package.

fjcasti1:
> @fjcasti1 Can you break this into separate PRs for the different libraries? The reason being that targeting multiple libraries at once in a single PR will cause a race condition in the release pipeline. It's not hard to resolve manually, but it's simpler to make each PR update only a single package.

Of course!

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -1,9 +1,30 @@
...
+    SpanAttributes.USER_ID,
+    SpanAttributes.METADATA,
+    SpanAttributes.TAG_TAGS,
+)
 
-@contextlib.contextmanager
-def suppress_tracing() -> Iterator[None]:
+
+class suppress_tracing(AbstractContextManager):
+    slots = ["_token"]
```
axiomofjoy, line 27: Did you intend to use `__slots__` here? Not sure we need `__slots__` tbh. It is a minor optimization and reduces readability IMO.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,147 @@ def suppress_tracing() -> Iterator[None]:
...
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+    async def __aenter__(self) -> "async_using_attributes":
+        self.attach_context()
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback) -> None:
+        self.detach_context()
         return
```
axiomofjoy: Not sure I see the value in an async context manager in this case because there is no IO, we are just adding to an in-memory data structure in the OTel context. What are your thoughts in terms of why this is needed?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,147 @@ def suppress_tracing() -> Iterator[None]:
...
+            isinstance(v, value_allowed_types)
+            or any(is_list_of(v, t) for t in value_list_allowed_types)
+            for v in d.values()
+            if value_allowed_types or value_list_allowed_types
+        )
+    )
+
+
+def is_list_of(lst: Sequence[object], tp: Type) -> bool:
+    return isinstance(lst, list) and all(isinstance(x, tp) for x in lst)
```
axiomofjoy, line 260: This looks like it wants to be a typeguard.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,147 @@ def suppress_tracing() -> Iterator[None]:
...
+        detach(self._token)
+        return
+
+
+class using_attributes(AbstractContextManager, UsingAttributes):
+    """
+    TBD
+    """
+
+    slots = [
```
axiomofjoy: Same.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,147 @@ def suppress_tracing() -> Iterator[None]:
...
+        "_metadata",
+        "_tags",
+    ]
+
+    def __init__(
+        self,
+        session_id: str = "",
+        user_id: str = "",
+        metadata: Dict[str, str] = {},
+        tags: List[str] = [],
```
axiomofjoy, line 61: These need to be optional types, because the user should have the option to not set them. Might make sense to add these individually in follow-up PRs.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,228 @@ def suppress_tracing() -> Iterator[None]:
...
+    ) -> None:
+        if not isinstance(session_id, str):
+            raise TypeError("session_id must be a string")
+        if not isinstance(user_id, str):
+            raise TypeError("user_id must be a string")
+        if not is_dict_of(metadata, key_allowed_types=(str), value_allowed_types=(str)):
+            raise TypeError("metadata must be a dictionary with string keys and values")
+        if not is_list_of(tags, str):
+            raise TypeError("tags must be a list of strings")
```
axiomofjoy, line 70: Definitely a good idea to do validations on the expected types here. I think as a rule of thumb, we should prefer logging a warning and not adding the attribute instead of raising an exception. The reason being that the expectation of instrumentation is that it will not break your production runtime, even if the user misconfigures it or otherwise misuses it. @anticorrelator Do you agree?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,228 @@ def suppress_tracing() -> Iterator[None]:
...
+        session_id: str = "",
+        user_id: str = "",
+        metadata: Dict[str, str] = {},
+        tags: List[str] = [],
+    ) -> None:
+        if not isinstance(session_id, str):
+            raise TypeError("session_id must be a string")
+        if not isinstance(user_id, str):
+            raise TypeError("user_id must be a string")
+        if not is_dict_of(metadata, key_allowed_types=(str), value_allowed_types=(str)):
```
axiomofjoy, line 67: We probably need more allowed value types, e.g., numbers.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
axiomofjoy: I'd also note it's not safe to use mutable types such as dictionaries and lists as default arguments in Python.

https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/#:~:text=However%2C%20the%20problem%20lies%20with,t%20provide%20an%20item%20list.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,228 @@ def suppress_tracing() -> Iterator[None]:
...
+
+        self._token = attach(ctx)
+        return
+
+    def detach_context(self) -> None:
+        detach(self._token)
+        return
+
+
+class using_session(AbstractContextManager, UsingAttributes):
```
axiomofjoy, line 96: I am not sure we need this level of abstraction using abstract base classes and multiple inheritance to start? A class or function is probably enough. What is this level of abstraction buying us in this case?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +33,228 @@ def suppress_tracing() -> Iterator[None]:
...
+    """
+    TBD
+    """
+
+    def __init__(self, metadata: Dict[str, str]) -> None:
+        super().__init__(metadata=metadata)
+
+    def __enter__(self) -> "using_metadata":
+        self.attach_context()
+        return self
```
axiomofjoy, line 140: Not seeing where this is used?

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
```diff
@@ -12,6 +12,7 @@
     Tuple,
 )
 
+from openinference.instrumentation import get_attributes_from_context
```
axiomofjoy, line 15: Nice 🔥

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_with_span.py
```diff
@@ -67,7 +72,10 @@ def finish_tracing(
                 if value is None:
                     continue
                 try:
-                    self._span.set_attribute(key, value)
+                    if isinstance(value, dict):
+                        self._span.set_attribute(key, json.dumps(value))
+                    else:
+                        self._span.set_attribute(key, value)
```
axiomofjoy, line 78: What case is this meant to handle? A bit of a code smell.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
```diff
@@ -76,7 +79,11 @@ def _start_as_current_span(
             record_exception=False,
             set_status_on_exception=False,
         ) as span:
-            yield _WithSpan(span=span, extra_attributes=dict(extra_attributes))
+            yield _WithSpan(
+                span=span,
+                context_attributes=dict(context_attributes),
```
axiomofjoy, line 84: We should talk about the priority of these attributes.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: It is a typeguard, I forgot that we require more modern python. I was stuck on 3.6 from our SDK

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Removed

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Removed

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Made the input `Optional` and default to `None`. Thanks!

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Removed type validation per @anticorrelator's feedback in our call

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Removed the abstract classes inheritance

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: What is used?

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_with_span.py
fjcasti1: metadata, it's a dictionary when reading from the context and it fails if we don't convert it to JSON

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
fjcasti1: Agreed that a conversation is needed

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
axiomofjoy: Forgot to remove this comment.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
axiomofjoy: Chatted on it, the approach makes sense.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
+        """Raise any exception triggered within the runtime context."""
+        detach(self._token)
+        return None
+
+
+class UsingAttributes:
+    def __init__(
+        self,
+        session_id: str = "",
+        user_id: str = "",
```
axiomofjoy, line 49: These also need to be optional types.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
+            ctx = set_value(SpanAttributes.METADATA, self._metadata, ctx)
+        if self._tags:
+            ctx = set_value(SpanAttributes.TAG_TAGS, self._tags, ctx)
+
+        self._token = attach(ctx)
+        return
+
+    def detach_context(self) -> None:
+        detach(self._token)
+        return
```
axiomofjoy, line 74: Nit: Not sure these `return` statements are doing much for us.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
+        super().__aenter__()
+        return self
+
+
+class using_metadata(UsingAttributes):
+    """
+    TBD
+    """
+
+    def __init__(self, metadata: Dict[str, str]) -> None:
```
axiomofjoy, line 132: The value type needs to be more general, I think.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
-        yield
+
+    def __enter__(self) -> "suppress_tracing":
+        self._token = attach(set_value(_SUPPRESS_INSTRUMENTATION_KEY, True))
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        """Raise any exception triggered within the runtime context."""
+        detach(self._token)
+        return None
```
axiomofjoy, line 42: Not sure this is doing much.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
+        return self
+
+    async def __aenter__(self) -> "using_attributes":
+        await super().__aenter__()
+        return self
+
+
+def get_attributes_from_context() -> Iterator[Tuple[str, AttributeValue]]:
+    for ctx_attr in CONTEXT_ATTRIBUTES:
+        yield ctx_attr, get_value(ctx_attr)
```
axiomofjoy, line 180: We should probably only yield if the value is not `None`. Otherwise, won't we wind up sending a bunch of null valued attributes?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: You're correct, it was remanente from before

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,150 @@ def suppress_tracing() -> Iterator[None]:
...
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        """Raise any exception triggered within the runtime context."""
+        detach(self._token)
+        return None
+
+
+class UsingAttributes:
+    def __init__(
+        self,
+        session_id: str = "",
```
axiomofjoy, line 48:
```suggestion
        self,
        *,
        session_id: str = "",
```

Making it required that these are keyword arguments will give us flexibility to change the order in the future, e.g., by inserting a keyword argument in the middle, without breaking backward compatibility for users.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_with_span.py
fjcasti1: Yeah I can do that instead, good call out

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Removed

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: How about string keys, any type for values?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Good call out. Downstream consumers only append them to the attributes if its not None but its better to do it at creation time as well

fjcasti1: Done

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Defaulting to empty strings achieves the same thing

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Great suggestion

fjcasti1: Great suggestion. I assume you mean only for `using_attributes`. Since `using_user(user_id="xxx")` is too redundant IMO

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +32,222 @@ def suppress_tracing() -> Iterator[None]:
...
+        self._tags = tags
+
+    def attach_context(self) -> None:
+        ctx = get_current()
+        if self._session_id:
+            ctx = set_value(SpanAttributes.SESSION_ID, self._session_id, ctx)
+        if self._user_id:
+            ctx = set_value(SpanAttributes.USER_ID, self._user_id, ctx)
+        if self._metadata:
+            ctx = set_value(SpanAttributes.METADATA, json.dumps(self._metadata), ctx)
```
axiomofjoy, line 65:
```suggestion
            ctx = set_value(SpanAttributes.METADATA, json.dumps(self._metadata, default=str), ctx)
```

axiomofjoy (approved):
lgtm, just need to make sure we don't crash if the metadata is non-json-serializable. add tests as well?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,222 @@ def suppress_tracing() -> Iterator[None]:
...
+            using_session("my-session-id"),
+            using_user("my-user-id"),
+            using_metadata(metadata),
+            using_tags(tags),
+        ):
+            ...
+
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
fjcasti1: DOne

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -12,11 +31,222 @@ def suppress_tracing() -> Iterator[None]:
...
+
+    def __enter__(self) -> "suppress_tracing":
+        self._token = attach(set_value(_SUPPRESS_INSTRUMENTATION_KEY, True))
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        detach(self._token)
+
+
+class UsingAttributes:
```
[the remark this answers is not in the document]
fjcasti1: Done

python/openinference-instrumentation/tests/test_suppress_tracing.py
fjcasti1, line 1: File was renamed and edited, not purely deleted

axiomofjoy:
nice work!

PR #420, opened 2024-05-03
feat: Make Mistral ai instrumentation read context attributes 

fjcasti1:


axiomofjoy (commented):
Rather than changing every test case to use `using_attributes`, can we add one new test case and name it appropriately?

fjcasti1:
> Rather than changing every test case to use `using_attributes`, can we add one new test case and name it appropriately?

I want to test all the scenarios. sync, async, with tools, without, etc.

python/instrumentation/openinference-instrumentation-mistralai/tests/openinference/instrumentation/mistralai/test_instrumentor.py
```diff
@@ -831,9 +933,58 @@ def test_synchronous_streaming_chat_completions_with_tool_call_response_emits_ex
...
+    assert attributes.pop(USER_ID, None) == user_id
+    attr_metadata = attributes.pop(METADATA, None)
+    assert attr_metadata is not None
+    assert isinstance(attr_metadata, str)  # must be json string
+    metadata_dict = json.loads(attr_metadata)
+    assert metadata_dict == metadata
+    attr_tags = attributes.pop(TAG_TAGS, None)
+    assert attr_tags is not None
+    assert len(attr_tags) == len(tags)
+    assert list(attr_tags) == tags
```
axiomofjoy, line 957: it's a best practice i try to follow to avoid abstraction as much as possible in test cases. duplication, especially of something like an assert statement, is very acceptable and even encouraged. the reason being that test cases need to be easy to understand because they need to point you to issues quickly, be stupidly easy to understand, and serve as documentation for other developers. i like the way they put it here: [dry is for code, damp is for tests](https://stackoverflow.com/a/11837973)

fjcasti1: I agree in general but in this case we're testing, on each case, that the attributes match the fixtures. I do not see the benefit of expanding this until we want to make more complex combined tests. However, if it's a must, I can explode the helper function on each test

axiomofjoy:
Definitely agree we should test sync and async, but I think these should be separate test cases. I'm not clear that this feature has any interaction with something like tool calls. They are orthogonal features, aren't they? I'd also mention that by adding `using_attributes` to every test case, we are no longer testing that mistral instrumentation works without the context manager, which is probably the most common situation.

Thinking of it another way, when a test fails, it ideally points you directly to the root cause of the failure. if every test case uses `using_attributes` and `using_attributes` breaks for some reason, I would have many red test cases. if i have a test case called `test_using_attributes_sync` or something like that and only that one test case fails, i immediately know what is wrong and the name of the test accurately reflects the nature of the problem.

axiomofjoy:
Thinking about this some more, I'm not even sure we necessarily need to explicitly test `using_attributes` in this module. Is the plan to have two tests, one async and one sync, for each of the five context managers (`using_attributes`, `using_metadata`, `using_tags`, `using_session_id`, and `using_user_id`) across each of the six instrumentation libraries? That's 60 tests and will grow when we add more libraries, and they are sort of integration tests between `openinference-instrumentation` and `openinference-instrumentation-mistralai` that result in coupling, rather than true unit tests. Imagine if we have to make a breaking change to the signature of the context managers. We'd have to change tests across many libraries. These kinds of tests are valuable to test the integration between `openinference-instrumentation` and other libraries, but we probably can't maintain that many of them.

What if instead we just add one test to `openinference-instrumentation-mistralai` testing that the instrumentation is able to pull session, user, metadata, and tags from OTel context? If we also have tests in `openinference-instrumentation` verifying that each context manager places the appropriate attributes into OTel context, I think that would be enough to be confident that things are working correctly. An integration test between `using_attributes` and one of the client libraries would be helpful, too.

axiomofjoy (approved):
approving based on the agreed-upon changes we discussed @fjcasti1 nice job 👍

PR #431, opened 2024-05-04
feat: Allow context managers as decorators

fjcasti1:


python/openinference-instrumentation/tests/test_context_managers.py
```diff
@@ -23,56 +23,121 @@ def test_suppress_tracing() -> None:
...
+        assert get_value(SpanAttributes.USER_ID) == user_id
+
+    f()
+    assert get_value(SpanAttributes.USER_ID) is None
+
+
+def test_using_metadata_decorator(metadata: Dict[str, Any]) -> None:
+    @using_metadata(metadata)
+    def f() -> None:
+        assert get_value(SpanAttributes.METADATA) == json.dumps(metadata, default=str)
```
axiomofjoy, line 90:
```suggestion
        assert get_value(SpanAttributes.METADATA) == json.dumps(metadata)
```

Nit: I think you can do without the `default` in these tests just because the metadata fixture is JSON serializable.

axiomofjoy (approved):

python/openinference-instrumentation/tests/test_context_managers.py
fjcasti1: Done

PR #437, opened 2024-05-06
feat: Add context manager for prompt template, version, and variables

fjcasti1:


python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -166,6 +184,61 @@ def __init__(self, tags: List[str]) -> None:
...
+class using_prompt_template_version(_UsingAttributesContextManager):
+    """
+    Context manager to add prompt template version to the current OpenTelemetry Context.
+    OpenInference instrumentations will read this Context and pass the prompt template
+    version as a span attribute, following the OpenInference semantic conventions.
+
+    Examples:
+        with using_prompt_template_version("v1.0"):
+            # Tracing within this block will include the span attribute:
+            # "llm.prompt_template.version" = "v1.0"
```
mikeldking, line 215: Let's combine all of these as they should probably all be set at the same time

mikeldking (changes requested):

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
fjcasti1: Much better, what was I thinking?

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -166,6 +184,44 @@ def __init__(self, tags: List[str]) -> None:
...
+            # "llm.prompt_template.variables" = "{\"city\": \"Johannesburg\",
+            \"date\": \"July 11\"}"
+            ...
+    """
+
+    def __init__(
+        self,
+        *,
+        template: str = "",
+        version: str = "",
```
mikeldking, line 215:
```suggestion
        version: Optional[str] = "",
```

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -166,6 +184,44 @@ def __init__(self, tags: List[str]) -> None:
...
+            # "llm.prompt_template.version" = "v1.0"
+            # "llm.prompt_template.variables" = "{\"city\": \"Johannesburg\",
+            \"date\": \"July 11\"}"
+            ...
+    """
+
+    def __init__(
+        self,
+        *,
+        template: str = "",
```
mikeldking, line 214:
```suggestion
        template: Optional[str] = "",
```

mikeldking (approved):
Nice!

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -192,51 +192,33 @@ class using_prompt_template(_UsingAttributesContextManager):
 
     Examples:
         prompt_template = "Please describe the weather forecast for {city} on {date}"
-        with using_prompt_template(prompt_template):
+        prompt_template_variables = "{"city": "Johannesburg", date:"July 11"}"
```
axiomofjoy, line 196:
```suggestion
        prompt_template_variables = {"city": "Johannesburg", date: "July 11"}
```

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -166,6 +184,44 @@ def __init__(self, tags: List[str]) -> None:
         super().__init__(tags=tags)
 
 
+class using_prompt_template(_UsingAttributesContextManager):
+    """
+    Context manager to add prompt template, with its version and variables a to the
```
axiomofjoy, line 189:
```suggestion
    Context manager to add prompt template, with its version and variables to the
```

axiomofjoy (approved):
nice, i think you're using strings rather than dictionaries in the docstrings for metadata, but other than that, looks good. thanks for the thorough docstrings.

fjcasti1:
> nice, i think you're using strings rather than dictionaries in the docstrings for metadata, but other than that, looks good. thanks for the thorough docstrings.

Good catch!

PR #446, opened 2024-05-09
feat: Instrumentation reads session id from metadata

fjcasti1:
This PR makes it so we capture the OpenInference span attribute `session.id` from the lanchain metadata if `session_id`, `conversation_id`, or `thread_id` is present in it. If multiple are present, the order of preference is 
```
session_id > conversation_id > thread_id
```
In addition, we correct the order of preference of context attributes. This is, if we use our context managers to pass metadata, but a different metadata object is passed directly in the call, the latter will be selected over the former.

fix: Overwrite context attributes if passed directly to langchain calls


python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -569,6 +569,12 @@ def _metadata(run: Run) -> Iterator[Tuple[str, str]]:
     if not run.extra or not (metadata := run.extra.get("metadata")):
         return
     assert isinstance(metadata, Mapping), f"expected Mapping, found {type(metadata)}"
+    if LANGCHAIN_THREAD_ID in metadata.keys():
+        yield SESSION_ID, metadata[LANGCHAIN_THREAD_ID]
+    if LANGCHAIN_CONVERSATION_ID in metadata.keys():
+        yield SESSION_ID, metadata[LANGCHAIN_CONVERSATION_ID]
+    if LANGCHAIN_SESSION_ID in metadata.keys():
+        yield SESSION_ID, metadata[LANGCHAIN_SESSION_ID]
```
axiomofjoy, line 577:
```suggestion
    if (session_id := (metadata.get(LANGCHAIN_THREAD_ID) or metadata.get(LANGCHAIN_CONVERSATION_ID) or metadata.get(LANGCHAIN_SESSION_ID)):
        yield SESSION_ID, session_id
```

Suggestion: Avoid `keys` for inclusion checks since it returns an iterator over the dictionary keys and hence runs in linear time in the size of the dictionary.

python/instrumentation/openinference-instrumentation-langchain/tests/openinference/instrumentation/langchain/test_instrumentor.py
```diff
@@ -260,43 +332,221 @@ def test_chain_metadata(
...
-    metadata_dict = json.loads(attr_metadata)
-    assert metadata_dict == metadata
-    attr_tags = attributes.pop(TAG_TAGS, None)
-    assert attr_tags is not None
-    assert len(attr_tags) == len(tags)
-    assert list(attr_tags) == tags
+    if session_id is not None:
+        assert attributes.pop(SESSION_ID, None) == session_id
+    if user_id is not None:
+        assert attributes.pop(USER_ID, None) == user_id
```
axiomofjoy, line 524: You've got redundancy here in the `pop`.

python/instrumentation/openinference-instrumentation-langchain/tests/openinference/instrumentation/langchain/test_instrumentor.py
```diff
@@ -43,7 +43,9 @@
 @pytest.mark.parametrize("is_async", [False, True])
 @pytest.mark.parametrize("is_stream", [False, True])
 @pytest.mark.parametrize("status_code", [200, 400])
+@pytest.mark.parametrize("use_context_attributes", [False, True])
```
axiomofjoy, line 46: I understand this test case is already doing this, but if adding a parameter to a test case requires adding complex branching logic to the test case, it should not be another parameter in the test case, but should be a separate test case.

axiomofjoy (approved):

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
fjcasti1: Nice, didn't know that

python/instrumentation/openinference-instrumentation-langchain/tests/openinference/instrumentation/langchain/test_instrumentor.py
fjcasti1: Done

PR #458, opened 2024-05-14
fix: add footnote for how to flatten attributes based on lists of objects

RogerHYang:
Add footnote to explain how to set attributes based on lists of objects.

### DIFF
--- a/spec/semantic_conventions.md
+++ b/spec/semantic_conventions.md
@@ -8,7 +8,7 @@ operations used by applications. These conventions are used to populate the `att
 The following attributes are reserved and MUST be supported by all OpenInference Tracing SDKs:
 
 | Attribute                              | Type                        | Example                                                                    | Description                                                      |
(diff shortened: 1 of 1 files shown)


spec/semantic_conventions.md
```diff
@@ -7,55 +7,65 @@ operations used by applications. These conventions are used to populate the `att
...
+
+```python
+messages = [{"message.role": "user", "message.content": "hello"}, {"message.role": "assistant", "message.content": "hi"}]
+
+for i, obj in enumerate(messages):
+    for key, value in obj:
+        span.set_attribute(f"input.messages.{i}.{key}", value)
+```
+
+If the objects are further nested, flattening should continue until the attribute values are either simple values, i.e. `bool, str, bytes, int, float` or simple lists, i.e. `List[bool], List[str], List[bytes], List[int], List[float]`
```
mikeldking: can you add a typescript example too. We have some vendors watching our language compatibility

mikeldking (approved):

PR #460, opened 2024-05-14
ci: ensure type ignores in langchain tests run across all versions of langchain

axiomofjoy:
The issue is that our type ignores are required on the old version of LangChain, but not required on the new version of LangChain. If we include the type ignore, the new version of LangChain complains about an usused type ignore. Found the solution [here](https://github.com/python/mypy/issues/8823#issuecomment-1556288015).


RogerHYang (approved):

axiomofjoy:
This issue resolves only one of several issues causing CI to fail. Merging for now, will resolve the other issues.

PR #461, opened 2024-05-14
ci: fix llama-index test cases

axiomofjoy:
LlamaIndex introduced a regression causing streaming LLM responses that result in a 400 to not register their exception and status code information. We are going to ignore this issue as we are about to migrate our LlamaIndex instrumentation away from the existing callback system to the new system.

This test case removes a few lines causing the tests to fail. It also ensures we are pinning `llama-index-core` for our testing.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_callback.py
+++ b/python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_callback.py
@@ -78,6 +78,11 @@ def test_callback_llm(
     metadata: Dict[str, Any],
     tags: List[str],
 ) -> None:
+    if status_code == 400 and is_stream:
+        pytest.xfail(
+            "There's been a regression where LlamaIndex does not invoke `on_event_end` "
+            "when streaming and receiving a 400."
+        )
     n = 10  # number of concurrent queries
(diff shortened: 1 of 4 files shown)


RogerHYang (approved):

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_callback.py
```diff
@@ -227,15 +218,6 @@ def _check_spans(
         assert synthesize_span.status.status_code == trace_api.StatusCode.OK
         assert not synthesize_span.status.description
         assert synthesize_attributes.pop(OUTPUT_VALUE, None) == answer
-    elif (
-        # FIXME: currently the error is propagated when streaming because we don't rely on
-        # `on_event_end` to set the status code.
-        status_code == 400 and is_stream
-    ):
-        assert synthesize_span.status.status_code == trace_api.StatusCode.ERROR
```
RogerHYang, line 235: maybe comment out and leave a note. instead of deleting

axiomofjoy: addressed in 00dd9eee4f2fd9738953e67720f89cac00efbf7b

PR #462, opened 2024-05-15
feat: support llama-index new instrumentation paradigm under feature flag `use_experimental_instrumentation`

### LINKED ISSUE #288: 🗺️ LlamaIndex instrumentation based tracing
Scope out a strategy where we can have both callbacks and instrumentation simultaneously and we can switch to instrumentation incrementally.

RogerHYang:
resolves #288 

Notes
- New instrumentation requires `llama-index-core>=0.10.37`.
- Setting `use_experimental_instrumentation=True` generates spans only from the new instrumentation system.
- Setting `use_experimental_instrumentation="both"` generates spans from both systems at the same time.

## Missing
1. Streaming events


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/__init__.py
```diff
@@ -30,13 +32,44 @@ def _instrument(self, **kwargs: Any) -> None:
             OpenInferenceTraceCallbackHandler,
         )
 
-        import llama_index.core
+        if _legacy_llama_index():
```
mikeldking, line 35: I think we could just publish with a `use_callback_handler` flag. That way we could get this in more hands early?

python/examples/llama-index/backend/poetry.lock
```diff
@@ -2,87 +2,87 @@
 
 [[package]]
 name = "aiohttp"
-version = "3.9.4"
+version = "3.9.5"
```
mikeldking, line 5: should we do this as follow-up or is needed for testing?

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,507 @@
...
+logger = logging.getLogger(__name__)
+
+
+class _Span(
+    BaseSpan,
+    extra=Extra.allow,
+    keep_untouched=(singledispatchmethod, property),
+):
+    _otel_span: trace_api.Span = PrivateAttr()
+    _span_kind: Optional[str] = PrivateAttr()
```
mikeldking, line 100: should this just fall back to 'chain' for now?

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,507 @@
...
+    @process_event.register
+    def _(self, event: AgentRunStepStartEvent) -> None:
+        if not self._span_kind:
+            self._span_kind = AGENT
+        if input := event.input:
+            self[INPUT_VALUE] = input
+
+    @process_event.register
+    def _(self, event: AgentRunStepEndEvent) -> None:
+        # FIXME: not sure what to do here
```
mikeldking, line 152: can you elaborate?

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,507 @@
...
+        ...
+
+    @process_event.register
+    def _(self, event: AgentToolCallEvent) -> None:
+        tool = event.tool
+        if name := tool.name:
+            self[TOOL_NAME] = name
+        self[TOOL_DESCRIPTION] = tool.description
+        self[TOOL_PARAMETERS] = json.dumps(tool.get_parameters_dict())
```
mikeldking, line 161: are there more params we could capture?

mikeldking (approved):
Let's ship behind a flag and get feedback. We should compile a list of "enhancements"

python/examples/llama-index/backend/poetry.lock
RogerHYang: Yea, we can do this as a follow-up.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
RogerHYang: It is falling back to chain when the span ends (and span_kind is still None).

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
RogerHYang: Yes, but they are on the instance and we would need to figure out how to find and exclude the API_KEYs so we are not grabbing too much.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/__init__.py
RogerHYang: OK. I'll add a flag that _adds_ instrumentation on top of the old paradigm so they're both exported at the same time for ease of comparison.

python/examples/llama-index/backend/app/api/routers/chat.py
```diff
@@ -99,20 +99,15 @@ async def chat(
...
-            async for token in response.async_response_gen():
-                if await request.is_disconnected():
-                    break
-                full_response = full_response + token
-                yield token
-            span.set_attribute(SpanAttributes.OUTPUT_VALUE, full_response)
-            span.end()
-
-        return StreamingResponse(event_generator(), media_type="text/plain")
+    last_message_content, messages = await parse_chat_data(data)
```
axiomofjoy, line 102: Are the spans appearing as normal, or do we need to keep this hack in the example?

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
```diff
@@ -0,0 +1,521 @@
...
+    respx_mock: MockRouter,
+    in_memory_span_exporter: InMemorySpanExporter,
+    nodes: List[Document],
+    chat_completion_mock_stream: Tuple[List[bytes], List[Dict[str, Any]]],
+    session_id: str,
+    user_id: str,
+    metadata: Dict[str, Any],
+    tags: List[str],
+) -> None:
+    n = 10  # number of concurrent queries
```
axiomofjoy, line 88: Is the threadpool needed on this test case?

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
```diff
@@ -0,0 +1,521 @@
...
+        (False, True),
+        (True, False),
+        # FIXME: stream + async is not supported by LlamaIndex as of v0.9.33
+        # (True, True),
+    ],
+)
+@pytest.mark.parametrize(
+    "status_code", [200]
+)  # 400 has been removed because retries have been added and can't be easily modified
+@pytest.mark.parametrize("use_context_attributes", [False, True])
```
axiomofjoy, line 73: I think we're overusing parameterization here. Parameterization is definitely useful for many test cases, but in this case, it introduces complex branching logic that makes the test case harder to understand. I like the way this comment puts it: [DRY is for code, DAMP is for tests](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests/11837973#11837973).

axiomofjoy (commented):
A few non-blocking comments.

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
RogerHYang: This is a bit of trade off because the code that is doing the checking is fairly massive, and if we abstract it out as a helper function, `pytest` doesn't seem to do a very good job of pin-pointing the exact assert statement that is failing.

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
RogerHYang: I don't seem to have a problem with the callstack using the IDE if that's what you're alluding to.

Screenshot 2024-05-16 at 11 06 29 AM

axiomofjoy: I think I'm not clear what purpose the threadpool is serving in this test case. Could we check the same things with a single question running in the main thread?

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
axiomofjoy: Definitely a challenge. It looks like there's a workaround [here](https://stackoverflow.com/questions/41522767/pytest-assert-introspection-in-helper-function). If we have multiple test cases, we also don't necessarily need to check every attribute for every test case, only the ones that are important to the particular test case.

This is non-blocking.

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_handler.py
RogerHYang: Since the instrumentor is a singleton, there's no a priori assurance that the it will handle concurrency. So this is a simulation as confirmation.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,509 @@
...
+            self[f"{EMBEDDING_EMBEDDINGS}.{i}.{EMBEDDING_VECTOR}"] = vector
+
+    @process_event.register
+    def _(self, event: LLMPredictStartEvent) -> None:
+        if not self._span_kind:
+            self._span_kind = LLM
+        self[LLM_PROMPT_TEMPLATE] = event.template.get_template()
+        vars = event.template.template_vars
+        args = {**event.template.kwargs, **(event.template_args if event.template_args else {})}
+        template_args = {var: arg for var in vars if (arg := args.get(var)) is not None}
```
axiomofjoy, line 190:
```suggestion
        variable_names = event.template.template_vars
        args = {**event.template.kwargs, **(event.template_args if event.template_args else {})}
        template_variables = {var_name: var_value for var_name in variable_names if (var_value := args.get(var_name)) is not None}
```

Suggestion: More verbose names would help here.

axiomofjoy (approved):
Nice job, Roger.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,509 @@
...
+    def _(self, event: RetrievalStartEvent) -> None:
+        if not self._span_kind:
+            self._span_kind = RETRIEVER
+        self._process_query_type(event.str_or_query_bundle)
+
+    @process_event.register
+    def _(self, event: RetrievalEndEvent) -> None:
+        for i, node in enumerate(event.nodes):
+            self[f"{RETRIEVAL_DOCUMENTS}.{i}.{DOCUMENT_ID}"] = node.node_id
+            if content := node.get_content():
```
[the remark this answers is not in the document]
RogerHYang: The content [defaults](https://github.com/run-llama/llama_index/blob/767de070b231fb328b6c0640c2e002c9c7af0a83/llama-index-core/llama_index/core/schema.py#L358) to `""`, so it's the same as `None`.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/package.py
```diff
@@ -1,2 +1,2 @@
-_instruments = ("llama-index >= 0.10.0", "llama-index-core >= 0.10.0")
+_instruments = ("llama-index-core >= 0.10.5",)
```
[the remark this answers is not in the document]
RogerHYang: Not really, because everything we import is `from llama_index.core`.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,509 @@
...
+        self._process_query_type(event.query)
+        self[RERANKER_TOP_K] = event.top_n
+        self[RERANKER_MODEL_NAME] = event.model_name
+        for i, node in enumerate(event.nodes):
+            self[f"{RERANKER_INPUT_DOCUMENTS}.{i}.{DOCUMENT_ID}"] = node.node_id
+            if content := node.get_content():
+                self[f"{RERANKER_INPUT_DOCUMENTS}.{i}.{DOCUMENT_CONTENT}"] = content
+            if (score := node.get_score()) is not None:
+                self[f"{RERANKER_INPUT_DOCUMENTS}.{i}.{DOCUMENT_SCORE}"] = score
+            if metadata := node.metadata:
```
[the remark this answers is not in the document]
RogerHYang: probably don't to waste an attribute on `{}`

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/__init__.py
```diff
@@ -35,8 +57,39 @@ def _instrument(self, **kwargs: Any) -> None:
...
+            dispatcher.add_span_handler(self._event_handler.span_handler)
+
     def _uninstrument(self, **kwargs: Any) -> None:
         import llama_index.core
 
         llama_index.core.global_handler = self._original_global_handler
         self._original_global_handler = None
+
+        if self._with_experimental_instrumentation and not _legacy_llama_index():
+            if self._event_handler is None:
```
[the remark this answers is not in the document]
RogerHYang: The parent class [guards](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f8758c6902ed725864ff677f739829aa6bce2078/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L127) against that.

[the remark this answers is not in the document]
RogerHYang: this check here is actually unnecessary

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -0,0 +1,509 @@
...
+    @process_event.register
+    def _(self, event: LLMPredictStartEvent) -> None:
+        if not self._span_kind:
+            self._span_kind = LLM
+        self[LLM_PROMPT_TEMPLATE] = event.template.get_template()
+        vars = event.template.template_vars
+        args = {**event.template.kwargs, **(event.template_args if event.template_args else {})}
+        template_args = {var: arg for var in vars if (arg := args.get(var)) is not None}
+        if template_args:
+            self[LLM_PROMPT_TEMPLATE_VARIABLES] = json.dumps(template_args)
```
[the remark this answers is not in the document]
RogerHYang: good idea. thx!

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
RogerHYang: updated

PR #491, opened 2024-05-22
fix(js): fallback patching flag for openAI for immutable modules

mikeldking:
In some runtimes the module exported is made immutable by the bundler (webpack) or the runtime (deno). This adds a fallback to a closure variable. The old flag is preserved just so that we can still check if a module is patched directly

### DIFF
--- a/js/packages/openinference-instrumentation-openai/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-openai/src/instrumentation.ts
@@ -41,6 +41,19 @@ import { isTracingSuppressed } from "@opentelemetry/core";
 const MODULE_NAME = "openai";
 
 /**
+ * Flag to check if the openai module has been patched
+ * Note: This is a fallback in case the module is made immutable (e.x. Deno, webpack, etc.)
+ */
+let _isOpenInferencePatched = false;
+
+/**
+ * function to check if instrumentation is enabled / disabled
+ */
+export function isPatched() {
+  return _isOpenInferencePatched;
+}
+
+/**
  * Resolves the execution context for the current span
  * If tracing is suppressed, the span is dropped and the current context is returned
  * @param span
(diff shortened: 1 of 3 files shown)


js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -318,16 +331,35 @@ export class OpenAIInstrumentation extends InstrumentationBase<typeof openai> {
...
+    moduleExports: typeof openai & { openInferencePatched?: boolean },
+    moduleVersion?: string,
+  ) {
     diag.debug(`Removing patch for ${MODULE_NAME}@${moduleVersion}`);
     this._unwrap(moduleExports.OpenAI.Chat.Completions.prototype, "create");
+    this._unwrap(moduleExports.OpenAI.Completions.prototype, "create");
     this._unwrap(moduleExports.OpenAI.Embeddings.prototype, "create");
+
+    try {
+      _isOpenInferencePatched = false;
```
axiomofjoy, line 357: Same as above.

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -318,16 +331,35 @@ export class OpenAIInstrumentation extends InstrumentationBase<typeof openai> {
       },
     );
 
-    module.openInferencePatched = true;
+    try {
+      _isOpenInferencePatched = true;
```
axiomofjoy, line 335: This line does not need to be part of the `try` clause.

axiomofjoy (approved):

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
mikeldking: fair

PR #514, opened 2024-06-03
feat: get messages on chat model start

RogerHYang:
This [copies the behavior](https://github.com/langchain-ai/langchain/blob/c01467b1f4f9beae8f1edb105b17aa4f36bf6573/libs/core/langchain_core/tracers/langchain.py#L115) from langchain tracer.

## Before
Screenshot 2024-06-03 at 12 29 16 PM

## After
Screenshot 2024-06-03 at 12 29 00 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
@@ -24,6 +24,8 @@ from typing import (
 from uuid import UUID
 
 import wrapt  # type: ignore
+from langchain_core.load import dumpd
+from langchain_core.messages import BaseMessage
 from langchain_core.tracers.base import BaseTracer
 from langchain_core.tracers.schemas import Run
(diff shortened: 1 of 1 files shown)


mikeldking (approved):

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -157,6 +159,38 @@ def on_tool_error(self, error: BaseException, *args: Any, run_id: UUID, **kwargs
...
+        parent_run_id: Optional[UUID] = None,
+        metadata: Optional[Dict[str, Any]] = None,
+        name: Optional[str] = None,
+        **kwargs: Any,
+    ) -> Run:
+        """Start a trace for an LLM run."""
+        start_time = datetime.now(timezone.utc)
+        if metadata:
+            kwargs.update({"metadata": metadata})
+        chat_model_run = Run(
```
axiomofjoy, line 178: This seems like it might be duplicated from a parent class method?

axiomofjoy (approved):

PR #529, opened 2024-06-11
chore: update llama-index docker example

RogerHYang:
Screenshot 2024-06-10 at 9 31 53 PM


mikeldking (approved):

RogerHYang:
> Noticing the double embedding spans. Does it capture embeddings on both?

The semantic conventions are generated by the firing of the `event`, which is only captured by one of the spans, i.e. the spans closest to the event.

PR #540, opened 2024-06-24
feat: Add guardrails span kind

nate-mar:
Add guardrails span kind to list of span kinds

### DIFF
--- a/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
+++ b/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
@@ -341,7 +341,7 @@ export enum OpenInferenceSpanKind {
   RERANKER = "reranker",
   EMBEDDING = "embedding",
   AGENT = "agent",
-  GUARDRAILS = "guardrails",
+  GUARDRAIL = "guardrail",
 }
 
 export enum MimeType {
--- a/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
+++ b/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
@@ -341,6 +341,7 @@ export enum OpenInferenceSpanKind {
   RERANKER = "reranker",
   EMBEDDING = "embedding",
   AGENT = "agent",
+  GUARDRAILS = "guardrails",
 }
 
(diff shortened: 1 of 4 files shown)


nate-mar:

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -341,6 +341,7 @@ export enum OpenInferenceSpanKind {
   RERANKER = "reranker",
   EMBEDDING = "embedding",
   AGENT = "agent",
+  GUARDRAILS = "guardrails",
```
mikeldking, line 344:
```suggestion
  GUARDRAIL = "guardrail",
```
Can we make it singular? Also what would be the typical contents of a guardrail span?

mikeldking (changes requested):
Let's add spec changes and also python changes

nate-mar:
@mikeldking will do -- can you point me to where in the spec file I should add this span kind? I can't seem to locate a place in the spec where these span kinds are comprehensively listed; I only see a list of example span kinds under the `openinference.span_kind` attribute.

mikeldking:
> @mikeldking will do -- can you point me to where in the spec file I should add this span kind? I can't seem to locate a place in the spec where these span kinds are comprehensively listed; I only see a list of example span kinds under the `openinference.span_kind` attribute.

Yes, as this is technically an attribute value it's not bound by any convention. But at least describing the value here would be ideal: https://github.com/Arize-ai/openinference/blob/main/spec/traces.md

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -341,6 +341,7 @@ export enum OpenInferenceSpanKind {
   RERANKER = "reranker",
   EMBEDDING = "embedding",
   AGENT = "agent",
+  GUARDRAIL = "guardrail",
```
mikeldking, line 344: These changes require something called a changeset to be deployed: https://github.com/Arize-ai/openinference/blob/main/js/DEVELOPMENT.md#changesets

mikeldking (approved):
Exciting!

PR #578, opened 2024-07-09
feat: helper function to `get_current_span()` for langchain

RogerHYang:
part of https://github.com/Arize-ai/phoenix/issues/3775

1. `langchain-core>=0.1.5` is needed because `var_child_runnable_config` was added by this [commit](https://github.com/langchain-ai/langchain/commit/4e4b11961432203f4b53a63cc428ea631d7cc970#diff-a7db1a221ffd895fa02a23efd3dcecbad235fe1abe1f2859246bf2dbd1430aa6R99) on Dec 29, 2023.
2. Note that OTEL has a similar [facility](https://opentelemetry.io/docs/languages/python/instrumentation/#get-the-current-span) (see below), but we can't use that for LangChain, because we rely on LangChain's own context management for parent-child-span associations.

```python

current_span = trace.get_current_span()
```

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -1,14 +1,17 @@
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Collection
+from typing import TYPE_CHECKING, Any, Callable, Collection, Optional
+from uuid import UUID
 
 from openinference.instrumentation.langchain.package import _instruments
 from openinference.instrumentation.langchain.version import __version__
 from opentelemetry import trace as trace_api
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
+from opentelemetry.trace import Span
 from wrapt import wrap_function_wrapper  # type: ignore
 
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -68,3 +77,30 @@ def __call__(
...
+
+
+def __getattr__(name: str) -> Any:
+    if name == get_current_span.__name__:
+        if tuple(map(int, version("langchain-core").split(".")[:3])) < (0, 1, 47):
+            raise ImportError("get_current_span requires langchain-core>=0.1.47")
+        return get_current_span
+    if name == LangChainInstrumentor.__name__:
+        return LangChainInstrumentor
+    raise AttributeError(name)
```
axiomofjoy, line 106: Can we move this near the top of the module?

axiomofjoy (approved):

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
RogerHYang: I think this is somewhat ancillary, so it shouldn't be first thing a reader sees when the file is opened.

PR #579, opened 2024-07-09
feat: helper functions to get string span id and trace id

RogerHYang:


python/openinference-instrumentation/src/openinference/instrumentation/helpers.py
```diff
@@ -0,0 +1,9 @@
+from opentelemetry.trace import Span
+
+
+def str_span_id(span: Span) -> str:
+    return span.get_span_context().span_id.to_bytes(8, "big").hex()
+
+
+def str_trace_id(span: Span) -> str:
+    return span.get_span_context().span_id.to_bytes(16, "big").hex()
```
axiomofjoy, line 9:
```suggestion
    return span.get_span_context().trace_id.to_bytes(16, "big").hex()
```

trace id?

python/openinference-instrumentation/src/openinference/instrumentation/helpers.py
```diff
@@ -0,0 +1,9 @@
+from opentelemetry.trace import Span
+
+
+def str_span_id(span: Span) -> str:
+    return span.get_span_context().span_id.to_bytes(8, "big").hex()
+
+
+def str_trace_id(span: Span) -> str:
```
axiomofjoy, line 8:
```suggestion
def get_span_id_as_string(span: Span) -> str:
    return span.get_span_context().span_id.to_bytes(8, "big").hex()


def get_trace_id_as_string(span: Span) -> str:
```

Suggestion

axiomofjoy (approved):

PR #585, opened 2024-07-15
feat(bedrock): add converse api support for bedrock

### LINKED ISSUE #531: [feature request] Traces with Bedrock Converse API (+Claude 3)
When running evals with Claude v3, the OpenInference library doesn't support [Anthropic's message api](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html) (it assumes "prompt" is in the `invoke_model` call, which is [only true for Claude 2](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-text-completion.html)

* Ability to have traces with Claude 3

* Currently reverted to using Claude 2 for now.

**Additional context**

cjunkin:
resolves #531

Adds support for AWS Bedrock's [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) to support tracing for Claude 3 and use of converse with Claude 2.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-bedrock/README.md
+++ b/python/instrumentation/openinference-instrumentation-bedrock/README.md
@@ -6,6 +6,26 @@ This package implements OpenInference tracing for `invoke_model` and `converse` 
 
 [![pypi](https://badge.fury.io/py/openinference-instrumentation-bedrock.svg)](https://pypi.org/project/openinference-instrumentation-bedrock/)
 
+## Supported Models
+
+Find the list of Bedrock-supported models and their IDs [here](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html#model-ids-arns).
+
+> **NOTE**: Future testing is planned for additional models.
+
+| Model                               | Supported Methods    |
+| ----------------------------------- | -------------------- |
+| `Anthropic Claude 2.0`              | converse, invoke     |
+| `Anthropic Claude 2.1`              | converse, invoke     |
+| `Anthropic Claude 3 Sonnet 1.0`     | converse             |
+| `Anthropic Claude 3.5 Sonnet`       | converse             |
+| `Anthropic Claude 3 Haiku`          | converse             |
+| `Meta Llama 3 8b Instruct`          | converse             |
+| `Meta Llama 3 70b Instruct`         | converse             |
+| `Mistral AI Mistral 7B Instruct`    | converse             |
+| `Mistral AI Mixtral 8X7B Instruct`  | converse             |
+| `Mistral AI Mistral Large`          | converse             |
+| `Mistral AI Mistral Small`          | converse             |
+
 ## Installation
 
 ```shell
--- a/python/instrumentation/openinference-instrumentation-bedrock/README.md
+++ b/python/instrumentation/openinference-instrumentation-bedrock/README.md
@@ -2,7 +2,7 @@
 
 Python autoinstrumentation library for AWS Bedrock calls made using `boto3`.
 
-This package implements OpenInference tracing for `invoke_model` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
+This package implements OpenInference tracing for `invoke_model` and `converse` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
 
 [![pypi](https://badge.fury.io/py/openinference-instrumentation-bedrock.svg)](https://pypi.org/project/openinference-instrumentation-bedrock/)
 
@@ -57,6 +57,9 @@ Instrumenting `boto3` is simple:
 BedrockInstrumentor().instrument()
 ```
 
+> **NOTE**:
+> OpenInference supports both [`invoke_model`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html#BedrockRuntime.Client.invoke_model) and [`converse`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html#). For models that use the Messages API, such as Anthropic Claude 3 and Anthropic Claude 3.5, please use the [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) instead.
+
 Now, all calls to `invoke_model` are instrumented and can be viewed in the `phoenix` UI.
 
 ```python
@@ -66,6 +69,40 @@ prompt = b'{"prompt": "Human: Hello there, how are you? Assistant:", "max_tokens
 response = client.invoke_model(modelId="anthropic.claude-v2", body=prompt)
 response_body = json.loads(response.get("body").read())
 print(response_body["completion"])
+```
+
+Now, all calls to `converse` are instrumented and can be viewed in the `phoenix` UI.
+
+```python
+session = boto3.session.Session()
+client = session.client("bedrock-runtime")
+
+message1 = {
+            "role": "user",
+            "content": [{"text": "Create a list of 3 pop songs."}]
+}
+message2 = {
+        "role": "user",
+        "content": [{"text": "Make sure the songs are by artists from the United Kingdom."}]
+}
+messages = []
+
+messages.append(message1)
+response = client.converse(
+    modelId="anthropic.claude-3-5-sonnet-20240620-v1:0",
+    messages=messages
+)
+out = response["output"]["message"]
+messages.append(out)
+print(out.get("content")[-1].get("text"))
+
+messages.append(message2)
+response = client.converse(
+    modelId="anthropic.claude-v2:1",
+    messages=messages
+)
+out = response['output']['message']
+print(out.get("content")[-1].get("text"))
 ```
 
 ## More Info
--- a/python/instrumentation/openinference-instrumentation-bedrock/README.md
+++ b/python/instrumentation/openinference-instrumentation-bedrock/README.md
@@ -6,11 +6,12 @@ This package implements OpenInference tracing for `invoke_model` and `converse` 
 
 [![pypi](https://badge.fury.io/py/openinference-instrumentation-bedrock.svg)](https://pypi.org/project/openinference-instrumentation-bedrock/)
 
+> [!NOTE]\
+> The Converse API was introduced in botocore [v1.34.116](https://github.com/boto/botocore/blob/develop/CHANGELOG.rst). Please use v1.34.116 or above to utilize converse.
+
 ## Supported Models
 
-Find the list of Bedrock-supported models and their IDs [here](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html#model-ids-arns).
-
-> **NOTE**: Future testing is planned for additional models.
+Find the list of Bedrock-supported models and their IDs [here](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html#model-ids-arns). Future testing is planned for additional models.
 
 | Model                               | Supported Methods    |
 | ----------------------------------- | -------------------- |
@@ -34,7 +35,7 @@ pip install openinference-instrumentation-bedrock
 
 ## Quickstart
 
-> **NOTE**:
+> [!IMPORTANT]\
 > OpenInference for AWS Bedrock supports both [`invoke_model`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html#BedrockRuntime.Client.invoke_model) and [`converse`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html#). For models that use the Messages API, such as Anthropic Claude 3 and Anthropic Claude 3.5, use the [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) instead.
 
 In a notebook environment (`jupyter`, `colab`, etc.) install `openinference-instrumentation-bedrock`, `arize-phoenix` and `boto3`.
--- a/python/instrumentation/openinference-instrumentation-bedrock/README.md
+++ b/python/instrumentation/openinference-instrumentation-bedrock/README.md
@@ -14,6 +14,9 @@ pip install openinference-instrumentation-bedrock
 
 ## Quickstart
 
+> **NOTE**:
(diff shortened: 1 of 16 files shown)


cjunkin:

python/instrumentation/openinference-instrumentation-bedrock/README.md
```diff
@@ -14,6 +14,9 @@ pip install openinference-instrumentation-bedrock
 
 ## Quickstart
 
+> **NOTE**:
+> OpenInference for AWS Bedrock supports both [`invoke_model`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html#BedrockRuntime.Client.invoke_model) and [`converse`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html#). For models that use the Messages API, such as Anthropic Claude 3 and Anthropic Claude 3.5, you can use the [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) instead.
```
Parker-Stafford, line 18: this says "you can" is it the case that they must use the converse api for it to work properly? If so would adjust wording

python/instrumentation/openinference-instrumentation-bedrock/README.md
```diff
@@ -68,6 +71,40 @@ response_body = json.loads(response.get("body").read())
...
+}
+message2 = {
+        "role": "user",
+        "content": [{"text": "Make sure the songs are by artists from the United Kingdom."}]
+}
+messages = []
+
+messages.append(message1)
+response = client.converse(
+    modelId="anthropic.claude-3-5-sonnet-20240620-v1:0",
```
Parker-Stafford, line 92: is here a simpler way to call this? Like an alias for these id's or something

python/instrumentation/openinference-instrumentation-bedrock/examples/bedrock_example.py
```diff
@@ -49,3 +48,55 @@
...
     print(response_body["completion"])
+
+def converse_example():
+    message1 = {
+            "role": "user",
+            "content": [{"text": "Create a list of 3 pop songs."}]
+    }
+    message2 = {
+            "role": "user",
+            "content": [{"text": "Make sure the songs are by artists from the United Kingdom."}]
```
Parker-Stafford, line 59: nit: coud maybe consider having the first message be a "system" (don't know if claude has system messages) message or something (common in prompting) and message 2 be a user message and simplify the example a bit while still showing the multiple messages

system_mes
```suggestion
   system_message = { "role": "system", "content": [{"text": "You are an expert at creating music playlists"}]}
   user_message = {"role": "user", "content": [{"text": "can you create a list of 3 pop songs by artists from the United Kingdom"}]}
```

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,75 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+                    SpanAttributes.OPENINFERENCE_SPAN_KIND,
+                    OpenInferenceSpanKindValues.LLM.value,
+                )
+
+                if model_id := kwargs.get("modelId"):
+                    _set_span_attribute(span, SpanAttributes.LLM_MODEL_NAME, model_id)
+
+                if request_messages := kwargs.get("messages"):
+                    # Extract the most recent request message in message history list
+                    current_request_message = request_messages[-1]
```
Parker-Stafford, line 183: why do we only get the last message here?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,75 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+
+                    if request_content := current_request_message.pop("content", ""):
+                        # Currently only supports single text-based data
+                        # Future implementation can be extended to handle different media types
+                        if prompt := request_content[-1].get("text"):
+                            _set_span_attribute(span, SpanAttributes.INPUT_VALUE, prompt)
```
Parker-Stafford, line 189: nit: i'm not a pythonista is there a way to simplify this if nesting? if messages, if content, if prompt etc. nothing wrong with this just a preference - will leave it up to you and the python folks!

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,75 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+
+                    invocation_parameters = safe_json_dumps(current_request_message)
+                    _set_span_attribute(
+                        span, SpanAttributes.LLM_INVOCATION_PARAMETERS, invocation_parameters
+                    )
```
Parker-Stafford, line 194: why do invocation parameters come from current_request_message? these are things like temperature etc. are they stored on the individual messages?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
Parker-Stafford: err is this coming in a follow up pr?

python/instrumentation/openinference-instrumentation-bedrock/tests/test_instrumentor.py
```diff
@@ -271,6 +271,208 @@ def test_invoke_client_with_missing_tokens(
         )
     assert attributes == {}
 
+@pytest.mark.parametrize("use_context_attributes", [False, True])
```
Parker-Stafford, line 274: a test for multiple messages would be good although that seems like it's mayb ecomign later?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,75 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+
+        def dispatcher(method_name):
+            if method_name == "invoke_model":
+                return instrumented_response_invoke
+            elif method_name == "converse":
+                return instrumented_response_converse
+            else:
+                raise ValueError(f"Unsupported method: {method_name}")
+
+        return dispatcher
```
Parker-Stafford, line 235: when does this get called? is this going to cause an issue if a method that isn't invoke or converse gets called on the client? I don't have a great understanding of how this works but would be very carefl about throwing in here since something that is supposed to be patching behind the scenes will now be breaking application logic potentially

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
cjunkin: The converse API supports a back-and-forth conversation with LLMs. Therefore, "messages" in the request call is a list that contains all previous chat history. Since the current span is tracking the most recent prompt, I have it so that we get the last prompt in that list. However, this can be changed so that we include all previous prompt/responses from the chat history in the span

cjunkin: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html#

python/instrumentation/openinference-instrumentation-bedrock/README.md
cjunkin: I don't believe so unfortunately

https://us-east-1.console.aws.amazon.com/bedrock/home?region=us-east-1#/providers?model=anthropic.claude-3-5-sonnet-20240620-v1:0

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
cjunkin: This is a mistake. Will fix

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
cjunkin: _model_invocation_wrapper is only ever called/assigned to inovke_model and converse so no other method for the client should touch this wrapper. I also thought of using a partial function instead of this dispatcher function. Would love to get some feedback on what might be safest.

cjunkin:
Failing tox check, will fix

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
Parker-Stafford: Yeah i think typically we get all the messages sent on every call since that span will be describing what happens when all of those messages are sent not just one of them. Even though it may be implied by other spans that this call was only adding this one message, it's still the case that this call contained all of the messages which affects teh output

cjunkin: Resolved. Chat context is stored as llm.input_messages. Since there is no semantic convention available for system prompts, converse called with system prompts adds a message with the role "system" at the start of llm.input_messages.

cjunkin:
__init__ is becoming quite cumbersome and is likely in need of a refactor to maintain clarity and manageability?

python/instrumentation/openinference-instrumentation-bedrock/README.md
```diff
@@ -2,10 +2,30 @@
 
 Python autoinstrumentation library for AWS Bedrock calls made using `boto3`.
 
-This package implements OpenInference tracing for `invoke_model` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
+This package implements OpenInference tracing for `invoke_model` and `converse` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
 
 [![pypi](https://badge.fury.io/py/openinference-instrumentation-bedrock.svg)](https://pypi.org/project/openinference-instrumentation-bedrock/)
 
+## Supported Models
```
anticorrelator, line 9: this table is great!

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -74,19 +77,23 @@ def _client_wrapper(
 
         if bound_arguments.arguments.get("service_name") == "bedrock-runtime":
             client = cast(InstrumentedClient, client)
+
             client._unwrapped_invoke_model = client.invoke_model
-            client.invoke_model = _model_invocation_wrapper(tracer)(client)
+            client.invoke_model = _model_invocation_wrapper(tracer)(client)("invoke_model")
+
+            client._unwrapped_converse = client.converse
+            client.converse = _model_invocation_wrapper(tracer)(client)("converse")
```
anticorrelator, line 85: as discussed, let's break the wrapper for converse and invoke model apart so we don't have one wrapper doing two completely different things

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,99 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+                if model_id := kwargs.get("modelId"):
+                    _set_span_attribute(span, SpanAttributes.LLM_MODEL_NAME, model_id)
+
+                if inference_config := kwargs.get("inferenceConfig"):
+                    invocation_parameters = safe_json_dumps(inference_config)
+                    _set_span_attribute(
+                        span, SpanAttributes.LLM_INVOCATION_PARAMETERS, invocation_parameters
+                    )
+
+                if system_prompts := kwargs.get("system", []):
```
anticorrelator, line 187: small suggestion: let's process the system messages and output messages in one sequence so we don't have to worry about finicky ordering issues like needing to keep track of the message number offset

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
```diff
@@ -158,7 +164,99 @@ def instrumented_response(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
+                            _set_span_attribute(
+                                span, f"{span_prefix}.message.role", request_msg_role
+                            )
+                            _set_span_attribute(
+                                span, f"{span_prefix}.message.content", request_msg_prompt
+                            )
+
+                response = wrapped_client._unwrapped_converse(*args, **kwargs)
+                if (response_message := response.get("output", {}).get("message")) and (
+                    response_content := response_message.get("content", [None])[0]
```
anticorrelator, line 220: what's the structure of "content", in general it's assuming the model might be multimodal right? Is there a way we can guarantee that we're getting a text field or maybe even the first text field?

Given what we discussed, it might be worthwhile to get all the text fields in the content and concatenate them. We're more likely to get a report suggesting the organization of our span is bad than people noticing that data should be there, but isn't

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py
anticorrelator: ^ agree we don't need a dispatcher, we can define two separate wrapper functions since we explicitly wrap both converse and invoke earlier in the file

python/instrumentation/openinference-instrumentation-bedrock/README.md
```diff
@@ -2,10 +2,30 @@
 
 Python autoinstrumentation library for AWS Bedrock calls made using `boto3`.
 
-This package implements OpenInference tracing for `invoke_model` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
+This package implements OpenInference tracing for `invoke_model` and `converse` calls made using a `boto3` `bedrock-runtime` client. These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize `phoenix`](https://github.com/Arize-ai/phoenix).
```
anticorrelator, line 5: Please add a note to the documentation that shows the minimum boto version required for converse support

python/instrumentation/openinference-instrumentation-bedrock/pyproject.toml
```diff
@@ -35,10 +35,10 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "boto3 >= 11.28.57",
```
anticorrelator, line 38: Was this really 11 before? Wow, my bad

anticorrelator (approved):
This is great! Excited to get this in

PR #591, opened 2024-07-16
fix(openai): missing span when stream response is used as context manager

RogerHYang:
OpenAI responses can be used as a context manager. For example, see [here](https://github.com/langchain-ai/langchain/blob/dc42279eb55fbb8ec5175d24c7b30fe7b502b6d1/libs/partners/openai/langchain_openai/chat_models/base.py#L513) in LangChain. When that happens, the `__enter__` method on the wrapped stream object is called and the stream object is [returned](https://github.com/openai/openai-python/blob/435a5805ccbd5939a68f7f359ab72e937ef86e59/src/openai/_streaming.py#L103-L104), escaping our wrapper. Subsequent iteration of the stream is un-noticed by our span, and the span is missing because it's never exported. This PR overrides the `__enter__` method so the wrapped object does not escape.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_stream.py
+++ b/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_stream.py
@@ -8,6 +8,7 @@ from typing import (
     Protocol,
     Tuple,
     Union,
+    cast,
 )
 
 from openinference.instrumentation.openai._utils import _finish_tracing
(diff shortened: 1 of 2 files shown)


mikeldking:
@RogerHYang I'm not sure I understand the issue from the PR title. Can you add a code snippet of the usage pattern?

RogerHYang:
@mikeldking I added more info in the description

mikeldking:
> @mikeldking I added more info in the description

Very clear, thank you

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_stream.py
```diff
@@ -110,6 +110,24 @@ async def __anext__(self) -> Any:
             self._process_chunk(chunk)
             return chunk
 
+    def __enter__(self) -> Any:
+        obj = self.__wrapped__.__enter__()
+        if obj is self.__wrapped__:
+            return self
+        return obj
```
mikeldking, line 117: I would add the text in the PR here.

mikeldking (approved):

PR #628, opened 2024-07-25
fix: allow `raw` to be dict or object for llama-index v0.10.58

### LINKED ISSUE #627: [bug] accommodate type changes in the latest version of llama index
```
AssertionError: raw must be Mapping, found <class 'openai.types.chat.chat_completion.ChatCompletion'>
```

RogerHYang:
resolves #627


mikeldking (approved):
Thanks for jumping and fixing CI 🙏

PR #636, opened 2024-07-26
nemo guardrails

harrisonchu:


harrisonchu:
Screenshot 2024-08-07 at 4 37 55 PM

detailed visibility into datasets guardrails will have to be instrumented directly since nemo is more of a generic flow execution framework it seems

PR #657, opened 2024-07-31
feat(openinference-core): js support for prompt templates

### LINKED ISSUE #654: [js][openinference-core] inheritable promptTemplate support

Parker-Stafford:
resolves #654

### DIFF
--- a/js/packages/openinference-core/src/trace/promptTemplate.ts
+++ b/js/packages/openinference-core/src/trace/promptTemplate.ts
@@ -1,0 +1,75 @@
+import {
+  PROMPT_TEMPLATE_TEMPLATE,
+  PROMPT_TEMPLATE_VARIABLES,
+  PROMPT_TEMPLATE_VERSION,
+} from "@arizeai/openinference-semantic-conventions";
+import { Attributes, Context, createContextKey } from "@opentelemetry/api";
+
+export const PROMPT_TEMPLATE_TEMPLATE_KEY = createContextKey(
+  `OpenInference SDK Context Key ${PROMPT_TEMPLATE_TEMPLATE}`,
+);
+
+export const PROMPT_TEMPLATE_VARIABLES_KEY = createContextKey(
+  `OpenInference SDK Context Key ${PROMPT_TEMPLATE_VARIABLES}`,
+);
+
+export const PROMPT_TEMPLATE_VERSION_KEY = createContextKey(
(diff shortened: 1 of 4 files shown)


js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -301,6 +301,12 @@ export const PROMPT_TEMPLATE_VARIABLES =
 export const PROMPT_TEMPLATE_TEMPLATE =
   `${PROMPT_TEMPLATE_PREFIX}.template` as const;
 
+/**
+ * A prompt template version
+ */
+export const PROMPT_TEMPLATE_VERSION =
```
Parker-Stafford, line 307: not sure how we want to handle these we might be behind on a few of them so can accumulate here then do a patch and rebase before merging the feture branch

js/packages/openinference-core/src/trace/promptTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+  `OpenInference SDK Context Key ${PROMPT_TEMPLATE_VERSION}`,
+);
+
+export function setPromptTemplateAttributes({
+  context,
+  template,
+  variables,
+  version,
+}: {
+  context: Context;
```
mikeldking, line 26: I like this structure but if we do it we should probably make it consistent across all

js/packages/openinference-core/src/trace/promptTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+  template?: string;
+  variables?: Record<string, unknown>;
```
mikeldking, line 28: Make sure to add docstrings here as people will not read our docs.

js/packages/openinference-core/src/trace/promptTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+);
+
+export function setPromptTemplateAttributes({
+  context,
+  template,
+  variables,
+  version,
+}: {
+  context: Context;
+  template?: string;
```
mikeldking, line 27: maybe we make this required? Otherwise it's kinda weird....

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
mikeldking: can you cherry pick this and PR main? That's the safest option. In theory this should work. We can try it if you want.

PR #760, opened 2024-08-06
feat(haystack): Add support for ChatPromptBuilder and ChatOpenAIGenerator components

21ShisodeParth:


python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -166,25 +206,49 @@ def __call__(
...
-                                SpanAttributes.LLM_TOKEN_COUNT_PROMPT: response["meta"][0]["usage"][
-                                    "prompt_tokens"
-                                ],
-                                SpanAttributes.LLM_TOKEN_COUNT_TOTAL: response["meta"][0]["usage"][
-                                    "total_tokens"
-                                ],
-                                SpanAttributes.LLM_OUTPUT_MESSAGES: response["replies"],
-                            }
+                if "Chat" in component.__class__.__name__:
+                    reply = response["replies"][0]
```
harrisonchu, line 210: we can be even more defensive here e.g.
```python
replies = response.get("replies")
if replies is None or len(replies) == 0:
     pass
reply = replies[0]
usage = reply.meta.get("usage", {})
```

harrisonchu (approved):
approved.

however, let's work together to refactor this if we decide to keep on supporting more components in haystack. there are well known patterns to tame the explosion of complexity here

PR #830, opened 2024-08-09
docs(annotations): add annotations to openai example

Parker-Stafford:
partially resolves: https://github.com/Arize-ai/phoenix/issues/4140

- adds feedback to open ai instrumentation example

May want to move this into phoenix repo

https://github.com/user-attachments/assets/bbdfeb47-eab7-4491-a3b0-2851f39346f7

### DIFF
--- a/js/examples/openai/compose.yml
+++ b/js/examples/openai/compose.yml
@@ -1,6 +1,6 @@
 services:
   phoenix:
-    image: arizephoenix/phoenix:nightly
+    image: arizephoenix/phoenix:version-4.22.0
     ports:
       - "6006:6006"
   backend:
@@ -10,7 +10,7 @@ services:
     environment:
       - OPENAI_API_KEY=${OPENAI_API_KEY}
       - COLLECTOR_ENDPOINT=http://phoenix:6006/v1/traces
-      - PHOENIX_REST_ENDPOINT=http://phoenix:6006/v1
+      - PHOENIX_API_ENDPOINT=http://phoenix:6006
       - PROD_CORS_ORIGIN=http://localhost:3000
     healthcheck:
       test: ["CMD", "wget", "--spider", "http://localhost:8000"]
--- a/js/examples/openai/compose.yml
+++ b/js/examples/openai/compose.yml
@@ -1,6 +1,6 @@
 services:
   phoenix:
-    image: arizephoenix/phoenix:latest
+    image: arizephoenix/phoenix:nightly
     ports:
       - "6006:6006"
   backend:
@@ -10,6 +10,7 @@ services:
     environment:
       - OPENAI_API_KEY=${OPENAI_API_KEY}
       - COLLECTOR_ENDPOINT=http://phoenix:6006/v1/traces
+      - PHOENIX_REST_ENDPOINT=http://phoenix:6006/v1
       - PROD_CORS_ORIGIN=http://localhost:3000
     healthcheck:
       test: ["CMD", "wget", "--spider", "http://localhost:8000"]
--- a/js/examples/openai/frontend/Dockerfile
+++ b/js/examples/openai/frontend/Dockerfile
@@ -16,7 +16,7 @@ COPY --from=deps /app/node_modules ./node_modules
 COPY . .
 # Set the chat api url. This environment variable must be set at build time.
 # See https://github.com/vercel/next.js/discussions/44628#discussioncomment-7040424 for more information.
-ENV NEXT_PUBLIC_CHAT_API http://localhost:8000/api/chat
+ENV NEXT_PUBLIC_API_ENDPOINT http://localhost:8000
 RUN npm run build
 
 # Production image, copy all the files and run next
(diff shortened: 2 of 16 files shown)


js/examples/openai/compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   phoenix:
-    image: arizephoenix/phoenix:latest
+    image: arizephoenix/phoenix:nightly
```
Parker-Stafford, line 3: latest doesn't have annotations right now

js/examples/openai/frontend/Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY --from=deps /app/node_modules ./node_modules
 COPY . .
 # Set the chat api url. This environment variable must be set at build time.
 # See https://github.com/vercel/next.js/discussions/44628#discussioncomment-7040424 for more information.
-ENV NEXT_PUBLIC_CHAT_API http://localhost:8000/api/chat
+ENV NEXT_PUBLIC_API_ENDPOINT http://localhost:8000
```
Parker-Stafford, line 19: must prefix these with next_public in order for them to work on the frontend

js/examples/openai/frontend/app/components/transform.ts
```diff
@@ -6,6 +6,30 @@ export const isValidMessageData = (rawData: JSONValue | undefined) => {
...
 
+const getSpanIdFromAnnotations = (annotations?: JSONValue[]) => {
+  if (!annotations) return;
+  for (const annotation of annotations) {
+    if (
+      annotation != null &&
+      typeof annotation === "object" &&
+      "spanId" in annotation &&
+      typeof annotation.spanId === "string"
+    ) {
```
Parker-Stafford, line 17: next build didn't accept a typeguard here

js/examples/openai/frontend/package.json
```diff
@@ -2,17 +2,18 @@
...
     "lint": "next lint"
   },
   "dependencies": {
-    "ai": "^2.2.27",
+    "ai": "^3.1.20",
     "dotenv": "^16.3.1",
     "next": "^14.1.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "sswr": "^2.1.0",
```
Parker-Stafford, line 16: added to solve peer deps issue with ai

js/examples/openai/compose.yml
mikeldking: We could pin to this for now? https://github.com/Arize-ai/phoenix/releases/tag/arize-phoenix-v4.22.0

mikeldking (approved):
this is so cool!

js/examples/openai/frontend/app/components/transform.ts
```diff
@@ -6,6 +6,30 @@ export const isValidMessageData = (rawData: JSONValue | undefined) => {
   return true;
 };
 
+const getSpanIdFromAnnotations = (annotations?: JSONValue[]) => {
+  if (!annotations) return;
```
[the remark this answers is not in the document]
Parker-Stafford: yeah there is a fair amount of context here but the easiest way on the server to write something to the response that isn't text is via an annotation - so the server writes the spanId as an annotation on the message (theses are vercel ai-sdk annotations) then we parse it out to get it on the message that is passed around to ui components (that's whats happening here)

The crazy logic you commented on below is because annotations are of type JSONValue[] and JSONValue is of type string | number | null | {[x:string]: JSONValue} or something like that -

 i tried putting the span id in the header but it was swallowed an if you just write it to the stream you have to parse it out so it doesn't get rendered as apart of the message so this felt the cleanest

js/examples/openai/frontend/app/components/transform.ts
Parker-Stafford: lol yeah i don't know how they are compiling but i pulled this out into a typeguard function and the build process failed saying spanId is not a property of string...

js/examples/openai/frontend/app/components/ui/chat/chat-item.tsx
```diff
@@ -1,13 +1,125 @@
...
+  ...buttonProps
+}: {
+  onClick: () => void;
+  children: React.ReactNode;
+} & ButtonHTMLAttributes<HTMLButtonElement>) {
+  return (
+    <button
+      className={
+        "ease-in-out duration-300 hover:bg-gray-200 px-2 rounded active:bg-gray-400"
+      }
```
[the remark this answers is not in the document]
Parker-Stafford: yeah i can switch to svg's and fill them or find a way to signify that with the emoji buttons

js/examples/openai/frontend/app/components/ui/chat/chat-item.tsx
```diff
@@ -1,13 +1,125 @@
...
+  );
+};
+
 export default function ChatItem(message: Message) {
+  const [isSubmittingFeedback, setIsSubmittingFeedback] = useState(false);
+  const [notification, setNotification] = useState<{
+    message: string;
+    variant: "error" | "success";
+  } | null>(null);
+  const onFeedbackClick = useCallback(
```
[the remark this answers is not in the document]
Parker-Stafford: yeah, we'd probably get some thigns for free, loading states etc. - was thinking fetch keeps things simple in a different way. Always hard to tell what level someone will be coming into an example like this

PR #844, opened 2024-08-12
fix: implement prompt template attributes for prompt builder and add helper functions

axiomofjoy:
- implement llm attributes for prompt builders and add corresponding unit test
- refactor existing logic into helper functions
- treat `ChatPromptBuilder` as a generic chain


python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -375,49 +211,122 @@ def __call__(
...
+    """
+    yield OPENINFERENCE_SPAN_KIND, span_kind
+
+
+def _get_input_attributes(arguments: Mapping[str, Any]) -> Iterator[Tuple[str, Any]]:
+    """
+    Yields input attributes.
+    """
+    masked_arguments = dict(_mask_embedding_vectors(key, value) for key, value in arguments.items())
+    yield INPUT_VALUE, safe_json_dumps(masked_arguments)
```
RogerHYang, line 318: missing `INPUT_MIME_TYPE`?

python/instrumentation/openinference-instrumentation-haystack/tests/openinference/haystack/test_instrumentor.py
```diff
@@ -614,6 +615,62 @@ def test_openai_generator_llm_span_has_expected_attributes(
...
+    assert len(spans) == 2
+    span = spans[0]
+    attributes = dict(span.attributes or {})
+    assert attributes.get(OPENINFERENCE_SPAN_KIND) == "LLM"
+    assert attributes.get(LLM_PROMPT_TEMPLATE) == "Where is {{ city }}?"
+    assert isinstance(
+        prompt_template_variables_json := attributes.get(LLM_PROMPT_TEMPLATE_VARIABLES), str
+    )
+    assert json.loads(prompt_template_variables_json) == {"city": "Munich"}
+
```
RogerHYang, line 672: When possible, I would recommend that we account for all the attributes via simply `assert atttrbutes.pop(..., None) is not None` and ending up with `assert attributes == {}`

RogerHYang (approved):

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
axiomofjoy: Good catch.

Issue #561, opened 2024-06-28
[vision][use-case] create an image tracing use-case example under openinference


mikeldking:

fjcasti1:
Relevant to this:
* https://docs.llamaindex.ai/en/stable/examples/multi_modal/gpt4v_multi_modal_retrieval/?h=tesla+images
* https://docs.llamaindex.ai/en/stable/examples/multi_modal/mm_agent/

PR #845, opened 2024-08-12
docs: Add multimodal llama index notebook

fjcasti1:
Closes https://github.com/Arize-ai/openinference/issues/561


mikeldking (changes requested):
Hey @fjcasti1 , maybe we keep the flat structure in examples and just have the notebook in the phoenix repo. We have better notebook linting and the example seems pretty specific to phoenix

PR #851, opened 2024-08-12
feat(crewAI): Added trace config, context attributes, suppress tracing for CrewAI

shreyabsridhar:
Resolves: #731 
Resolves: #748 
Resolves: #719 

Notes:
- Trace config is implemented and tested
- Suppress tracing is implemented
- Context attributes is implemented and testing framework is in place but cannot be tested until crewAI tests are implemented

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -3,8 +3,9 @@ from enum import Enum
 from inspect import signature
 from typing import Any, Callable, Iterator, List, Mapping, Optional, Tuple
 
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
```diff
@@ -206,6 +212,7 @@ def __call__(
                 raise
             span.set_status(trace_api.StatusCode.OK)
             span.set_attribute(OUTPUT_VALUE, response)
+            span.set_attributes(dict(get_attributes_from_context()))
```
fjcasti1, line 215: We want to set these attributes first, check other implementations if you need more details but you can simply add it to the `start_as_current_span` call

fjcasti1 (approved):

PR #874, opened 2024-08-13
ci: update tox-uv

axiomofjoy:


RogerHYang (approved):
can you update development.md also?

PR #894, opened 2024-08-15
feat: add re-ranker support in haystack

axiomofjoy:
- Improves logic distinguishing retrievers and rerankers
- Implements reranker spans

Screenshot 2024-08-14 at 8 39 17 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
@@ -98,11 +98,7 @@ class _ComponentWrapper(_WithTracer):
                 span.set_attributes(
                     {
                         **dict(_get_span_kind_attributes(LLM)),
-                        **dict(
-                            _get_llm_prompt_template_attributes_from_prompt_builder(
-                                component, run_bound_args
-                            )
-                        ),
+                        **dict(_get_llm_prompt_template_attributes(component, run_bound_args)),
                     }
                 )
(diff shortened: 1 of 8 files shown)


python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -212,42 +223,77 @@ def _get_component_type(component: Component) -> ComponentType:
...
+
+def _has_ranker_io_types(run_method: Callable[..., Any]) -> bool:
+    """
+    Uses heuristics to infer if a component has a ranker-like `run` method.
+    """
+    if (input_types := _get_run_method_input_types(run_method)) is None or (
+        output_types := _get_run_method_output_types(run_method)
+    ) is None:
+        return False
+    has_documents_parameter = input_types.get("documents") is List[Document]
```
RogerHYang, line 278: how does this work: `is List[Document]`?

If i do this below I get False:
```python
[1,2] is List[int]
```

RogerHYang (approved):

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
axiomofjoy: Checking that the type is the same. So `List[int] == List[int]`.

RogerHYang: oh, then `is` is the customary choice

PR #921, opened 2024-08-16
feat(core): export withSafety utilities remove custom `isAttributeValue` to use @opentelemetry/core's

Parker-Stafford:


js/packages/openinference-core/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./trace";
+export * from "./utils";
```
Parker-Stafford, line 2: exporting utils here to get access to withSafety and safelyJSONStringify and safelyJSONParse - that, it also exports some commonly used type utils - if we want to handle this in a different way can switch

this will make https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-core/src/utils/index.ts#L10 available on this package

mikeldking (approved):
noice

PR #935, opened 2024-08-20
ci(python): fail fast

RogerHYang:
CI is not for development: it should fail as fast as possible.

https://docs.pytest.org/en/8.3.x/reference/reference.html#command-line-flags


python/tox.ini
```diff
@@ -77,10 +77,10 @@ commands_pre =
...
-  test: pytest -n auto {posargs:tests}
-  ci: ruff format . --diff
-  ci: ruff check . --no-fix
-  ci: mypy .
-  ci: pytest -n auto tests
+  test: pytest -n auto --nf {posargs:.}
+  ci: ruff format --diff {posargs:.}
+  ci: ruff check --no-fix {posargs:.}
+  ci: mypy {posargs:.}
+  ci: pytest -n auto -x -ra {posargs:.}
```
mikeldking, line 86: What does postargs do? does it let you add your own args afterwards?

mikeldking (approved):

python/tox.ini
RogerHYang: yes, it adds a bit more flexibility if you want to say type-check or lint a specific sub folder

PR #939, opened 2024-08-20
feat: Tool calling for Anthropic instrumentor

21ShisodeParth:


python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
```diff
@@ -281,6 +281,108 @@ async def test_anthropic_instrumentation_async_messages(
     assert not attributes
 
 
+@pytest.mark.vcr(
```
harrisonchu, line 284: TIL. will use this for my future tests. thanks!

harrisonchu (approved):

PR #941, opened 2024-08-21
feat(guardrails): limit support to versions <0.5.2

harrisonchu:
after version 0.5.2 guardrails has major changes that makes the autoinstrumentor here useless. The package after that version also has native instrumentation that works quite well so we'll limit our tox testing to versions < 0.5.2 and for anyone that uses this instrumentor with a version of guardrails >= 0.5.2 the auto instrumentor will be a no op.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-guardrails/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-guardrails/pyproject.toml
@@ -38,7 +38,7 @@ instruments = [
   "guardrails-ai",
 ]
 test = [
-  "guardrails-ai <= 0.5.2",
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-guardrails/pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "guardrails-ai >= 0.4.5",
+  "guardrails-ai",
```
mikeldking, line 38: should this just be a range?

mikeldking (approved):

PR #942, opened 2024-08-21
feat(js): Added tool logic to openAI instrumentation and modified test accordingly

shreyabsridhar:
Resolves: #909


js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -409,6 +410,24 @@ function getLLMInputMessagesAttributes(
...
 
+/**
+ * Converts each tool definition into a json schema
+ */
+function getLLMToolsJSONSchema(
+  body: ChatCompletionCreateParamsBase,
+): Attributes {
+  if (!body.tools) {
+    // If tools is undefined, return an empty object
+    return {} as Attributes;
```
Parker-Stafford, line 421: is this as Attributes cast necessary here? if not would remove or just return null here, null should be spreadable (i.e., you can still do ...getLLMToolsJSONSchema(body)

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -409,6 +410,24 @@ function getLLMInputMessagesAttributes(
...
+  }
+  return body.tools.reduce((acc, tool, index) => {
+    const toolJsonSchema = JSON.stringify(tool);
```
Parker-Stafford, line 424: nit use safelyJSONStringify from @openinference/core

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -409,6 +410,24 @@ function getLLMInputMessagesAttributes(
...
+    const key = `${SemanticConventions.LLM_TOOLS}.${index}.${SemanticConventions.TOOL_JSON_SCHEMA}`;
+    acc[key] = toolJsonSchema;
+    return acc;
+  }, {} as Attributes);
```
Parker-Stafford, line 428: i think you can put this type on line 423 instead of casting it down here, acc: Attributes - sometimes that seems to work with reduce sometimes it can be finicky, if you can would do that to avoid the cast here

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
shreyabsridhar: Done, removed casting. Didn't replace with null because:
`Type 'null' is not assignable to type 'Attributes'`

mikeldking: Empty attributes is okay here I think. It's a container and having no attributes is the "empty" state

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
mikeldking: agreed, always good to be safe

mikeldking (approved):

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
shreyabsridhar: Done

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
shreyabsridhar: Done

PR #946, opened 2024-08-22
chore: update pnpm version and pin in package.json

Parker-Stafford:
https://github.com/pnpm/pnpm/releases/tag/v9.7.0
https://github.com/pnpm/pnpm/pull/8363

In order for the packageManger field to be picked up global pnpm must be >=9.7.0.

After that when you're in the js directory any pnpm commands will use the version specified in package json

### DIFF
--- a/.github/workflows/typescript-publish.yaml
+++ b/.github/workflows/typescript-publish.yaml
@@ -26,7 +26,7 @@ jobs:
             - name: Setup PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 8
+                  version: 9.8.0 
             - name: setup pnpm config
               run: pnpm config set store-dir $PNPM_CACHE_FOLDER
             - name: Install Dependencies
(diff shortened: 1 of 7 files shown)


.github/workflows/typescript-publish.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
             - name: Setup PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 8
+                  version: 9.8.0
```
mikeldking, line 29: this has to be specified?

mikeldking (approved):

.github/workflows/typescript-publish.yaml
Parker-Stafford: it doesnt have to specify the fullpath but figured we might as well keep it in sync with whats in pacakge.json, which does have to specify the whole path

PR #947, opened 2024-08-22
feat(mistralai)!: support v1

harrisonchu:
migrate to support v1.0.2 of mistral client and above


harrisonchu:
checkpoint. sync stream and non sync works

Screenshot 2024-08-22 at 6 49 21 PM

harrisonchu:
note to self for last commit. streaming needs to have it's response accumulator re-written becaue it no longer looks like the non streaming response

mikeldking:
@anticorrelator can you do a pass on this?

python/instrumentation/openinference-instrumentation-mistralai/src/openinference/instrumentation/mistralai/_stream.py
```diff
@@ -87,29 +84,79 @@ def __next__(self) -> Any:
...
+    def get_attributes(self) -> Iterator[Tuple[str, AttributeValue]]:
+        if self._self_response_accumulator is not None:
+            yield from self._self_response_accumulator.get_attributes()
+
+    def get_extra_attributes(self) -> Iterator[Tuple[str, AttributeValue]]:
+        if self._self_response_accumulator is not None:
+            yield from self._self_response_accumulator.get_extra_attributes()
+
+
+class _AsyncStream:
```
anticorrelator, line 120: does this also need to be an object proxy?

python/instrumentation/openinference-instrumentation-mistralai/tests/openinference/instrumentation/mistralai/test_instrumentor.py
```diff
@@ -818,21 +843,18 @@ def test_synchronous_streaming_chat_completions_emits_expected_span(
...
-        return mistral_sync_client.chat_stream(
-            model="mistral-large-latest",
-            messages=[
-                ChatMessage(
-                    content="Who won the World Cup in 2018? Answer in one word, no punctuation.",
-                    role="user",
-                )
+    def mistral_stream() -> Generator[CompletionEvent, None, None]:
+        mistral_client = Mistral(api_key="redacted")
+        return mistral_client.chat.stream(  # type: ignore
```
anticorrelator, line 848: can we assert that the instrumented and uninstrumented stream both pass a type check a user might expect? (both for sync and async tests)

python/instrumentation/openinference-instrumentation-mistralai/src/openinference/instrumentation/mistralai/_chat_wrapper.py
```diff
@@ -144,15 +146,7 @@ def _parse_args(
         request_data: Dict[str, Any] = {}
```
anticorrelator, line 146: after binding arguments, use `apply_defaults` to ensure that we capture defaulted arguments as well

anticorrelator (approved):
This generally looks good to me! I left a couple comments about ensuring that defaults are properly applied to bound call signatures and ensuring that the returned stream objects pass type checks

python/instrumentation/openinference-instrumentation-mistralai/src/openinference/instrumentation/mistralai/_chat_wrapper.py
anticorrelator: ^ bound defaults, changed tests to check only explicitly set parameters

Issue #785, opened 2024-08-07
🗺️ Vercel AI SDK


mikeldking:
Vercel AI SDK 3.3 introduces OTEL support, which means there should be spans from the inside of the AI SDK's execution. We should possibly introduce a span processor taht maps these to OpenInference semantic conventions.

## Spike
- [x] #835
- [x] #836
- [x] [ai-sdk] map ai semantic conventions to openinference in a span processor or exporter
- [x] #958
- [x] #943
- [x] [ai-sdk] docs in phoenix
- [ ] #1028

mikeldking:
https://vercel.com/blog/vercel-ai-sdk-3-3

Parker-Stafford:
## Dogfooding use cases:

All of these will be using this example: 
https://github.com/Arize-ai/openinference/tree/main/js/examples/next-openai-telemetry-app

Follow the readme there and paste the code snippet for your use case as instructed below

### Embeddings
add the following in chat/route.ts
imports:
```typescript
import { streamText, embed, embedMany } from "ai";
```

Add in POST function
```typescript
  await embed({
    model: openai.embedding("text-embedding-ada-002"),
    value: "hello, you are a great bot",
    experimental_telemetry: {
      isEnabled: true,
      metadata: { example: "value" },
    },
  });

  await embedMany({
    model: openai.embedding("text-embedding-ada-002"),
    values: ["hello, you are a great bot", prompt],
    experimental_telemetry: {
      isEnabled: true,
      metadata: { example: "value" },
    },
  });

```

### Stream / GenerateObject
run:
```shell
pnpm install zod --ignore-workspace
```

add the following in chat/route.ts
Imports:
```typescript
import { streamText, streamObject } from "ai"; // add stream object here
import { z } from "zod";
```
Add in POST function:
```typescript
  const objectStream = await streamObject({
    model: openai("gpt-3.5-turbo"),
    maxTokens: 100,
    messages: [
      { content: "hello, you are a great bot", role: "system" },
      { content: "please create me 3 objects like this", role: "user" },
    ],
    schema: z.object({
      description: z.string().describe("The generated text"),
      number: z.number().describe("A generated number"),
    }),
    experimental_telemetry: {
      isEnabled: true,
      metadata: { example: "value" },
    },
  });
  const objectStreamResponse = await objectStream.toTextStreamResponse().text();
```

### Tools

run:
```shell
pnpm install zod --ignore-workspace
```

add the following in chat/route.ts
Imports:
[...]

PR #949, opened 2024-08-23
fix(vercel): add input / output to tool spans, remove duplicate token counts

Parker-Stafford:
- removes double token counts from `ai.generateText` spans and their children `ai.generateText.doGenerate`, now only the `LLM` span (doGenerate) will have token counts
- adds `input.value` and `output.value` (and respective mime types) to tool spans

see here: https://github.com/Arize-ai/openinference/issues/785#issuecomment-2305233727
and here: https://github.com/Arize-ai/openinference/issues/785#issuecomment-2305278582
for more info

### DIFF
--- a/js/packages/openinference-vercel/src/utils.ts
+++ b/js/packages/openinference-vercel/src/utils.ts
@@ -368,9 +368,8 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
           };
         case AISemanticConventions.TOKEN_COUNT_COMPLETION:
         case AISemanticConventions.TOKEN_COUNT_PROMPT:
-          // Do not capture token counts for chain spans as they are captured by the LLM spans below
-          // Capturing on chains as well results in double token counts
-          if (spanKind === OpenInferenceSpanKind.CHAIN) {
+          // Do not capture token counts for non LLM spans to avoid double token counts
+          if (spanKind !== OpenInferenceSpanKind.LLM) {
             return openInferenceAttributes;
           }
           return {
--- a/js/packages/openinference-vercel/src/utils.ts
+++ b/js/packages/openinference-vercel/src/utils.ts
@@ -368,11 +368,50 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
           };
         case AISemanticConventions.TOKEN_COUNT_COMPLETION:
         case AISemanticConventions.TOKEN_COUNT_PROMPT:
-        case AISemanticConventions.TOOL_CALL_NAME:
-        case AISemanticConventions.TOOL_CALL_ARGS:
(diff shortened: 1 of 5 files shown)


js/packages/openinference-vercel/src/utils.ts
```diff
@@ -368,12 +368,51 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
           };
         case AISemanticConventions.TOKEN_COUNT_COMPLETION:
         case AISemanticConventions.TOKEN_COUNT_PROMPT:
+          // Do not capture token counts for chain spans as they are captured by the LLM spans below
+          // Capturing on chains as well results in double token counts
+          if (spanKind === OpenInferenceSpanKind.CHAIN) {
```
Parker-Stafford, line 373: could flip this conditional and only capture for llm span kinds, currently this is the same either way since there are only llms and chains that have these attributes but might be more future proof if we don't expect any other span kind to care about token counts.

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -368,12 +368,51 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
...
+            ...openInferenceAttributes,
+            [openInferenceKey]: attributes[convention],
+          };
         case AISemanticConventions.TOOL_CALL_NAME:
-        case AISemanticConventions.TOOL_CALL_ARGS:
           return {
             ...openInferenceAttributes,
             [openInferenceKey]: attributes[convention],
           };
+        case AISemanticConventions.TOOL_CALL_ARGS: {
```
Parker-Stafford, line 385: not sure what to put for `input.value` on tool spans so put args?

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -368,12 +368,51 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
...
+          let argsAttributes = {
+            [openInferenceKey]: attributes[convention],
+          };
+          // For tool spans, capture the arguments as input value
```
Parker-Stafford, line 389: don't want to clobber more important io for other span kinds, same as below - not sure if this the right call or not but seems conservative

js/packages/openinference-vercel/src/utils.ts
mikeldking: I think you should flip this.

js/packages/openinference-vercel/src/utils.ts
mikeldking: Yeah they go into the input for now. We don't have any dedicated slot

mikeldking (approved):

PR #952, opened 2024-08-23
feat(core): remove automatic loggin in withSafety function

Parker-Stafford:
- adds the ability to have a custom message to log in functions wrapped with `withSafety`, if not passed in does not log anything.

We use `withSafety` in various packages throughout open inference. In some cases we do so expecting the function to succeed, but don't want to throw in someones application when we hit an unexpected error. In this case, it makes sense to log and swallow to return null. In other cases, specifically with safelyJSONParse we use it to merely see if something is parsable - and the error / message is not really worth logging as it is more of a validation check of some unknown things structure (i.e., we expect it to fail some of the time).

It's debatable whether or not we want to use safelyJSONParse in situations where we are checking the structure of something (vs. parsing something that we expect to be parsable).

In either case the message to log is not one size fits all. For example, seeing this in the console of our examples, when this is expected to happen sometimes can be quite jarring:

`Failed to get attributes for span: SyntaxError: Unexpected token H in JSON at position 0`

In this case we did not fail to get attributes and spans were processed as expected.

### DIFF
--- a/js/packages/openinference-core/src/utils/index.ts
+++ b/js/packages/openinference-core/src/utils/index.ts
@@ -7,12 +7,17 @@ export * from "./typeUtils";
  * @param fn - A function to wrap with a try-catch block.
  * @returns A function that returns null if an error is thrown.
  */
-export function withSafety<T extends GenericFunction>(fn: T): SafeFunction<T> {
+export function withSafety<T extends GenericFunction>(
+  fn: T,
+  fallbackMessage?: string,
+): SafeFunction<T> {
   return (...args) => {
     try {
       return fn(...args);
     } catch (error) {
-      diag.error(`Failed to get attributes for span: ${error}`);
+      if (fallbackMessage) {
+        diag.error(`${fallbackMessage} ${error}`);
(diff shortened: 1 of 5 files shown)


js/packages/openinference-core/src/utils/index.ts
```diff
@@ -7,12 +7,17 @@ export * from "./typeUtils";
  * @param fn - A function to wrap with a try-catch block.
  * @returns A function that returns null if an error is thrown.
  */
-export function withSafety<T extends GenericFunction>(fn: T): SafeFunction<T> {
+export function withSafety<T extends GenericFunction>(
+  fn: T,
+  fallbackMessage?: string,
```
mikeldking, line 12: I would make this an object that has an on error callback. That keeps this generic and scale forward. FallbackMessage as a positional argument just makes this function too specific

mikeldking: I guess it could be an on error calllback

mikeldking (approved):

PR #963, opened 2024-08-26
feat(langchainjs): support langchain `^0.2.0`

### LINKED ISSUE #911: [js] support langchain core 0.2.0
Flowwise and langchain have moved to core 0.2 - which will not get instrumented or might face compatibility issues.

- [ ] bump core support
- [ ] create a compatibility matrix in the README

Parker-Stafford:
resolves #911 

Adds support for langchain `^0.2.0`


js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -69,19 +69,46 @@ export class LangChainInstrumentation extends InstrumentationBase<
     if (module?.openInferencePatched || _isOpenInferencePatched) {
       return module;
     }
-    this.tracer;
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const instrumentation = this;
+
+    if ("_configureSync" in module.CallbackManager) {
```
mikeldking, line 75: this requires a doc-string or at least make this more readable that you are using ducktyping to detect the version. E.g. even re-writing this as 
```suggestion
    const cbManagerIsModern = "_configureSync" in module.CallbackManager;
    if (cbManagerIsModern) {
```

Would make it more clear

js/packages/openinference-instrumentation-langchain/src/instrumentationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function addTracerToHandlers(
+  tracer: Tracer,
+  handlers?: CallbackManagerModuleV1.Callbacks,
+): CallbackManagerModuleV1.Callbacks;
+export function addTracerToHandlers(
+  tracer: Tracer,
+  handlers?: CallbackManagerModuleV2.Callbacks,
+): CallbackManagerModuleV2.Callbacks;
+export function addTracerToHandlers(
```
mikeldking, line 14: this requires a docstring as to how it works

js/packages/openinference-instrumentation-langchain/src/instrumentationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+  const tracerAlreadyRegistered = handlers.inheritableHandlers.some(
+    (handler) => handler instanceof LangChainTracer,
+  );
+  if (tracerAlreadyRegistered) {
+    return handlers;
+  }
+  // There are some slight differences in the CallbackHandler interface between v1 and v2
+  // We support both versions and our tracer is compatible with either as it will extend the class from the installed version
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mikeldking, line 44: duplicate code comment? Seems like this should be in the top level docstring

js/packages/openinference-instrumentation-langchain/src/instrumentationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+      handlers.push(new LangChainTracer(tracer) as any);
+    }
+    return handlers;
+  }
+  const tracerAlreadyRegistered = handlers.inheritableHandlers.some(
+    (handler) => handler instanceof LangChainTracer,
+  );
+  if (tracerAlreadyRegistered) {
+    return handlers;
+  }
```
mikeldking, line 41: I'm not sure as a reader I groc these two different code paths. What's the distinction between inheritable an not inheritable?

js/packages/openinference-instrumentation-langchain/src/tracer.ts
```diff
@@ -42,7 +42,26 @@ export class LangChainTracer extends BaseTracer {
   }
 
   protected async _startTrace(run: Run) {
-    await super._startTrace(run);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore _startTrace only exists on langchain ^0.1.0 BaseTracer and has been replaced in 0.2 by onRunCreate, we support both 0.1 and 0.2
+    if (typeof super._startTrace === "function") {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
mikeldking, line 48: lots of duplicate comments. Can you put it in the jsdoc above the method?

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -16,39 +16,34 @@ import {
...
-  return (...args) => {
-    try {
-      return fn(...args);
-    } catch (error) {
-      diag.error(`Failed to get attributes for span: ${error}`);
-      return null;
-    }
-  };
-}
+const utilErrorCallback = (message: string) => (error: unknown) => {
```
mikeldking, line 37:
```suggestion
const onError = (message: string) => (error: unknown) => {
```

js/packages/openinference-instrumentation-langchain/src/utils.ts
```diff
@@ -16,39 +16,34 @@ import {
...
+import { withSafety } from "@arizeai/openinference-core";
 
 export const RETRIEVAL_DOCUMENTS =
   `${SemanticAttributePrefixes.retrieval}.${RetrievalAttributePostfixes.documents}` as const;
 
 /**
- * Wraps a function with a try-catch block to catch and log any errors.
- * @param fn - A function to wrap with a try-catch block.
- * @returns A function that returns null if an error is thrown.
+ * Callback function to handle errors.
```
mikeldking, line 35: Not the most helpful docstring

js/package.json
```diff
@@ -40,5 +40,10 @@
   "packageManager": "pnpm@9.8.0",
   "eslintIgnore": [
     "examples/**/*"
-  ]
+  ],
+  "pnpm": {
+    "overrides": {
+      "@langchain/openai@^0.1>@langchain/core": "^0.1.63"
+    }
+  }
```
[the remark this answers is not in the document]
Parker-Stafford: overrides have to be set at the root, they can't be set in the subpackages, these are both dev dependencies and are necessary to make the tests pass, so should be fine

js/packages/openinference-instrumentation-langchain/package.json
```diff
@@ -27,13 +27,16 @@
...
+    "@arizeai/openinference-core": "workspace:*",
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/core": "^1.25.1",
     "@opentelemetry/instrumentation": "^0.46.0"
   },
   "devDependencies": {
-    "@langchain/core": "^0.1.63",
-    "@langchain/openai": "^0.0.25",
+    "@langchain/core": "^0.2.30",
+    "@langchain/coreV1": "npm:@langchain/core@^0.1.0",
```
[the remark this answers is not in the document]
Parker-Stafford: ahh yeah can update these

js/packages/openinference-instrumentation-langchain/src/instrumentation.ts
```diff
@@ -1,4 +1,5 @@
-import type * as CallbackManagerModule from "@langchain/core/callbacks/manager";
+import type * as CallbackManagerModuleV2 from "@langchain/core/callbacks/manager";
+import type * as CallbackManagerModuleV1 from "@langchain/coreV1/callbacks/manager";
```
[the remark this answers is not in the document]
Parker-Stafford: will double check, they are imported as types and only dev dependencies so i don't think they would be but will verify

js/packages/openinference-instrumentation-langchain/src/instrumentationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (Array.isArray(handlers)) {
+    const tracerAlreadyRegistered = handlers.some(
+      (handler) => handler instanceof LangChainTracer,
+    );
+    if (!tracerAlreadyRegistered) {
+      // There are some slight differences in the CallbackHandler interface between v1 and v2
+      // We support both versions and our tracer is compatible with either as it will extend the class from the installed version
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      handlers.push(new LangChainTracer(tracer) as any);
```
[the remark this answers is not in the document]
Parker-Stafford: See the comment above (now above the function) - there is a slight difference between the CallbackHandlers in v0.1 and v0.2 our tracer should be compatible with either since it properly extends the BaseTracer (a callback handler) from whatever the installed version is 

langchain basically typescript doesn't know if the pre-existing handlers in this case are v0.1 or v0.2  and thinks the tracer is definitely v0.2 so it can't be added to the Array of handlers - casting as v0.1 creates teh same problem in the opposite direction so just went with any for ease here - tried doing some generic stuff but wasn't able to get it to be happy

js/packages/openinference-instrumentation-langchain/src/instrumentationUtils.ts
Parker-Stafford: I added to the comment for clarity but will put here too, basically the handlers are either undefined, an array of handlers or a CallbackManager class - 
if it's an array we just add it

if its a class, then it has a field called inheritableHandlers where the handlers go, we check there and now (just added) it's handlers to see if there is a langchain tracer - if there isn't we add it.

js/package.json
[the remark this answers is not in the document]
Parker-Stafford:
```
 WARN  The field "pnpm.overrides" was found in /Users/parkerstafford/repos/openinference/js/packages/openinference-instrumentation-langchain/package.json. This will not take effect. You should configure "pnpm.overrides" at the root of the workspace instead.

```

mikeldking (approved):
Very cool that it's type safe and has both versions.

js/packages/openinference-instrumentation-langchain/package.json
```diff
@@ -27,13 +27,19 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "workspace:*",
+    "@arizeai/openinference-core": "workspace:*",
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/core": "^1.25.1",
     "@opentelemetry/instrumentation": "^0.46.0"
   },
+  "peerDependencies": {
```
Parker-Stafford, line 35: @mikeldking i realized we are importing BaseTracer to extend for the tracer - this is used as an actual dep... because of that I think we need this here, does this make sense to you?

js/package.json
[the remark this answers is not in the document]
mikeldking: Got it

js/packages/openinference-instrumentation-langchain/package.json
mikeldking: Yeah - in that case it is a peer dep agree

PR #1010, opened 2024-09-10
feat: id generator with separate source of randomness

### LINKED ISSUE #1009: [bug] should use a different source of randomness, e.g. secrets, so span identifiers are not affected by seed from application
[id_generator](https://github.com/open-telemetry/opentelemetry-python/blob/9697f805d1e5ba53cc55e2502ad1de0cc7c5900a/opentelemetry-sdk/src/opentelemetry/sdk/trace/id_generator.py) is affected by application seeds

Screenshot 2024-09-10 at 7 51 52 AM

RogerHYang:
resolves #1009

### DIFF
--- a/python/openinference-instrumentation/src/openinference/instrumentation/config.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/config.py
@@ -1,6 +1,7 @@
 import os
 from contextlib import contextmanager
 from dataclasses import dataclass, field, fields
+from secrets import randbits
 from typing import (
     Any,
     Callable,
@@ -21,7 +22,16 @@ from opentelemetry.context import (
     detach,
     set_value,
 )
-from opentelemetry.trace import Link, Span, SpanKind, Tracer, use_span
+from opentelemetry.sdk.trace import IdGenerator
+from opentelemetry.trace import (
+    INVALID_SPAN_ID,
+    INVALID_TRACE_ID,
+    Link,
+    Span,
+    SpanKind,
+    Tracer,
+    use_span,
+)
 from opentelemetry.util.types import Attributes, AttributeValue
 
 from openinference.semconv.trace import (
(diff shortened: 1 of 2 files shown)


python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -343,10 +353,32 @@ def end(self, end_time: Optional[int] = None) -> None:
...
 
+class _IdGenerator(IdGenerator):
+    """
+    An IdGenerator that uses a different source of randomness to
+    avoid being affected by seeds set by user application.
+    """
+
+    def generate_span_id(self) -> int:
+        while (span_id := randbits(64)) == INVALID_SPAN_ID:
+            continue
```
mikeldking, line 364: what exactly is an invalid_span_id?

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -343,10 +353,32 @@ def end(self, end_time: Optional[int] = None) -> None:
...
+
 class OITracer(wrapt.ObjectProxy):  # type: ignore[misc]
     def __init__(self, wrapped: Tracer, config: TraceConfig) -> None:
         super().__init__(wrapped)
         self._self_config = config
+        self._self_id_generator = _IdGenerator()
+
+    @property
+    def id_generator(self) -> IdGenerator:
+        return self._self_id_generator
```
mikeldking, line 381: A bit confused - how does writing our own ID generator fix this? wouldn't otel be ripe with this problem? How is the implementation above different than the OTEL one that it side-steps the problem?

I'm guessing we confirmed the instrumentation was creating duplicates?

python/openinference-instrumentation/src/openinference/instrumentation/config.py
RogerHYang: it's all [zeros](https://github.com/open-telemetry/opentelemetry-python/blob/9697f805d1e5ba53cc55e2502ad1de0cc7c5900a/opentelemetry-api/src/opentelemetry/trace/span.py#L578-L579)

looping is based on otel [implementation](https://github.com/open-telemetry/opentelemetry-python/blob/9697f805d1e5ba53cc55e2502ad1de0cc7c5900a/opentelemetry-sdk/src/opentelemetry/sdk/trace/id_generator.py#L52)

python/openinference-instrumentation/src/openinference/instrumentation/config.py
RogerHYang: it's not affected by `seed` as shown in the unit test

RogerHYang: > I'm guessing we confirmed the instrumentation was creating duplicates?

not confirmed (because we don't have access), but this eliminates one of the possibilities

python/openinference-instrumentation/src/openinference/instrumentation/config.py
mikeldking: Got it, thanks for the link

python/openinference-instrumentation/src/openinference/instrumentation/config.py
mikeldking: hmm, I see - it seems sorta unlikely to me that we would have that rampant of a problem due to a seed and it also feels like this inherently would be causing issues for any instrumentation out there.

I'm cool to stamp this PR for some confirmation but it feels like it might not solve the problem.

RogerHYang: > it feels like it might not solve the problem

Agree. This is just to eliminate one avenue of problems. The rationale derives from the potential severity of the problem, not the confirmed existence of the problem.

mikeldking: Got it, sounds good

mikeldking (approved):

PR #1025, opened 2024-09-10
fix: missed ref_doc_id in retrieval

romnovi:
`DOCUMENT_ID` is actually `node.node_id` now.
It's difficult to track documents which are found by retrieval stage.
I suggest adding `ref_doc_id` in the attributes.


mystdeim:

RogerHYang:
@mystdeim After reviewing the llama-index source code, it seems that `ref_doc_id` is a deprecated property and serves as an alias for `node_id`, meaning they should have the same values. Could you please provide more details about the issues you're encountering with `node_id`?

https://github.com/run-llama/llama_index/blob/bee0e9d5ab43b5ddca672a75c4a0cc9a7810a6ff/llama-index-core/llama_index/core/schema.py#L351-L356

RogerHYang:
OK I now understand the issue. Thank you for bringing this to our attention.

Let me discuss it with the team first, and follow up with you afterwards.

mikeldking:
Apologies from us that this never made it in. I think we never fully resolved this. Is it still something of interest?

PR #1043, opened 2024-09-18
feat(mistral): add agent support

harrisonchu:


python/instrumentation/openinference-instrumentation-mistralai/examples/agent_completions.py
```diff
@@ -0,0 +1,25 @@
+from mistralai import Mistral
+from opentelemetry import trace as trace_api
+from opentelemetry.sdk import trace as trace_sdk
+from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor
+
+from openinference.instrumentation.mistralai import MistralAIInstrumentor
+
+tracer_provider = trace_sdk.TracerProvider()
+tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
+trace_api.set_tracer_provider(tracer_provider)
```
anticorrelator, line 10: I think for our examples we should be setting the tracer provider directly in the `instrument` function, I think it's just a bit better of a pattern

anticorrelator (approved):

mikeldking:
@harrisonchu can we merge?

PR #1048, opened 2024-09-19
feat(core): add trace config

### LINKED ISSUE #821: [instrumentation-js] TraceConfig for openinference-core
Mimic the behavior of the trace config for python - e.g. in terms of scope and coherence with the environment variables

Parker-Stafford:
resolves #821 

- adds support for trace config in open inference core
- This allows for configuring various aspects of tracing, right now limited to data masking


js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -73,12 +77,21 @@ function getExecContext(span: Span) {
...
+    traceConfig: TraceConfigOptions;
+  }) {
+    const { instrumentationConfig, traceConfig } = config ?? {};
     super(
       "@arizeai/openinference-instrumentation-openai",
       VERSION,
-      Object.assign({}, config),
+      Object.assign({}, instrumentationConfig),
     );
+    this.openInferenceTracer = new OpenInferenceTracer({
```
Parker-Stafford, line 91: this would need to be done like this for every instrumentation

js/packages/openinference-instrumentation-openai/test/openai.test.ts
```diff
@@ -9,7 +9,11 @@ import { context } from "@opentelemetry/api";
 const tracerProvider = new NodeTracerProvider();
 tracerProvider.register();
 
-const instrumentation = new OpenAIInstrumentation();
+const instrumentation = new OpenAIInstrumentation({
```
Parker-Stafford, line 12: tests are currently failing, as expected

js/packages/openinference-core/package.json
```diff
@@ -34,10 +34,8 @@
   },
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "workspace:*",
-    "@opentelemetry/core": "^1.25.1"
-  },
-  "peerDependencies": {
-    "@opentelemetry/api": ">=1.0.0 <1.9.0"
+    "@opentelemetry/core": "^1.25.1",
+    "@opentelemetry/api": "^1.9.0"
```
Parker-Stafford, line 38: this is a direct dep i think, also we need 1.9 since other packages are using 1.9 and the span interface has changed

js/packages/openinference-core/src/trace/trace_config/OpenInferenceTracer.ts
```diff
@@ -0,0 +1,122 @@
...
+  ): ReturnType<F> | undefined {
+    const formattedArgs = formatStartActiveSpanParams(arg2, arg3, arg4);
+    if (formattedArgs == null) {
+      return;
+    }
+    const { opts, ctx, fn } = formattedArgs;
+    const { attributes } = opts ?? {};
+    return this.tracer.startActiveSpan(
+      name,
+      { ...opts, attributes: undefined },
```
Parker-Stafford, line 92: delegating masking to the OpenInferenceSpan so we don't have to duplicate it here

js/packages/openinference-core/src/trace/trace_config/maskingRules.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Masks (redacts) input text in LLM input messages.
+ * Will mask information stored under the key `llm.input_messages.[i].message.content`.
+ * * @example
+ * ```typescript
+ *  maskOutputTextRule.condition({
+ *      config: {hideInputText: true},
+ *      key: "llm.input_messages.[i].message.content"
+ *  }) // returns true so the rule applies and the value will be redacted
+ */
+const maskInputTextRule: MaskingRule = {
```
Parker-Stafford, line 16: pulled more complicated masking rules out  to write doc comments and wrote more simple ones inline (in the array below)

Parker-Stafford: this rule is based on the same config value (hideINputText) as the rule on line 55, they can be combined into one rule if we think that it's better to have one rule per field, this matches the python implementation, but i think it might be better to merge them

js/packages/openinference-core/src/trace/trace_config/maskingRules.ts
```diff
@@ -0,0 +1,199 @@
...
+    key.includes(SemanticConventions.EMBEDDING_VECTOR),
+  action: () => undefined,
+};
+
+/**
+ * A list of {@link MaskingRule}s that are applied to span attributes to either redact or remove sensitive information.
+ */
+const maskingRules: MaskingRule[] = [
+  {
+    condition: ({ config, key }) =>
```
Parker-Stafford, line 148: simpler rules

js/packages/openinference-core/tsconfig.esm.json
```diff
@@ -3,7 +3,11 @@
   "compilerOptions": {
     "outDir": "dist/esm",
     "rootDir": "src",
-    "tsBuildInfoFile": "dist/esm/tsconfig.esm.tsbuildinfo"
+    "tsBuildInfoFile": "dist/esm/tsconfig.esm.tsbuildinfo",
+    "baseUrl": "./src",
+    "paths": {
+      "@core/*": ["*"]
+    }
```
Parker-Stafford, line 10: todo, remove or fix this, it works in ide and typecheck but fails to build

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -212,17 +225,20 @@ export class OpenAIInstrumentation extends InstrumentationBase<typeof openai> {
...
-          const span = instrumentation.tracer.startSpan(`OpenAI Completions`, {
-            kind: SpanKind.INTERNAL,
-            attributes: {
-              [SemanticConventions.OPENINFERENCE_SPAN_KIND]:
-                OpenInferenceSpanKind.LLM,
-              [SemanticConventions.LLM_MODEL_NAME]: body.model,
-              [SemanticConventions.LLM_INVOCATION_PARAMETERS]:
-                JSON.stringify(invocationParameters),
-              ...getCompletionInputValueAndMimeType(body),
+          const span = instrumentation.openInferenceTracer.startSpan(
```
Parker-Stafford, line 228: just formatting changes from instrumentation.tracer -> instrumentation.openInferenceTracer

js/packages/openinference-core/src/trace/trace_config/OpenInferenceSpan.ts
```diff
@@ -0,0 +1,90 @@
...
+  AttributeValue,
+  Link,
+} from "@opentelemetry/api";
+import { TraceConfig } from "./types";
+import { mask } from "./maskingRules";
+
+/**
+ * A wrapper around the OpenTelemetry {@link Span} interface that masks sensitive information based on the passed in {@link TraceConfig}.
+ */
+export class OpenInferenceSpan implements Span {
```
mikeldking, line 17: I thought in python this is just called MaskedSpan? I think that might be simpler in name.

Also is there no base class to extend?

js/packages/openinference-core/src/trace/trace_config/OpenInferenceTracer.ts
```diff
@@ -0,0 +1,122 @@
...
+    fn = arg4 as F;
+  }
+
+  opts = opts ?? {};
+  ctx = ctx ?? context.active();
+
+  return { opts, ctx, fn };
+}
+
+export class OpenInferenceTracer implements Tracer {
```
mikeldking, line 50: I think in python this is OITracer to not confuse in other packages. Is there also no base class?

js/packages/openinference-core/src/trace/trace_config/OpenInferenceSpan.ts
Parker-Stafford: It is, that name just felt very restrictive to me, but if we only need it to mask things i can change it.

There are several implementations of the Span interface, there isn't a base one to my knowledge. ONe of them might be a good candidate for extending. Not sure if we want to tie our selves to a specific implmentation or just compose the "MaskedSpan" with wahtever span teh tracer creates

js/packages/openinference-core/src/trace/trace_config/OpenInferenceTracer.ts
Parker-Stafford: I can update the name, pretty much same comment here as above in terms of base class, Didn't want to tie to any one implementation and instead just compose with whatever the tracer is that is provided in each instrumentation, felt the least intrusive

js/packages/openinference-core/src/trace/trace_config/OpenInferenceSpan.ts
Parker-Stafford: can take a look into this and tracer to see if there is a good implementation to extend

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  hideInputImages?: boolean;
+  hideInputText?: boolean;
+  hideOutputText?: boolean;
+  hideEmbeddingVectors?: boolean;
+  base64ImageMaxLength?: number;
+};
+
+export type TraceConfig = Readonly<Required<TraceConfigOptions>>;
+
+export type ConfigKey = keyof TraceConfig;
```
mikeldking, line 18:
```suggestion
export type TraceConfigKey = keyof TraceConfig;
```

js/packages/openinference-core/jest.config.js
```diff
@@ -3,4 +3,7 @@ module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
   prettierPath: null,
+  moduleNameMapper: {
+    "^@core/(.*)$": "<rootDir>/src/$1",
+  },
```
mikeldking, line 8: Can we nix this alias for now?

js/packages/openinference-core/src/trace/trace_config/OITracer.ts
```diff
@@ -0,0 +1,116 @@
+import {
```
mikeldking, line 1: minor nit: maybe "tace-config" would be more consistent with the module names?

js/packages/openinference-core/src/trace/trace_config/OITracer.ts
```diff
@@ -0,0 +1,116 @@
...
+  ): ReturnType<F> | undefined {
+    const formattedArgs = formatStartActiveSpanParams(arg2, arg3, arg4);
+    if (formattedArgs == null) {
+      return;
+    }
+    const { opts, ctx, fn } = formattedArgs;
+    const { attributes } = opts ?? {};
+    return this.tracer.startActiveSpan(
+      name,
+      { ...opts, attributes: undefined },
```
mikeldking, line 90: just a reminder that we need to "inherit" the context attributes here as a follow-up

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  hideOutputText?: boolean;
+  hideEmbeddingVectors?: boolean;
+  base64ImageMaxLength?: number;
+};
+
+export type TraceConfig = Readonly<Required<TraceConfigOptions>>;
+
+export type TraceConfigKey = keyof TraceConfig;
+
+export type TraceConfigMetadata =
```
mikeldking, line 20: this type name is kinda confusing - it's like the TraceConfigFlag or something like that

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  | {
+      default: boolean;
+      envKey: string;
+      type: "boolean";
```
mikeldking, line 24: can you add doctrings here of examples?

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  | {
+      default: number;
+      envKey: string;
+      type: "number";
+    };
+
+export type MaskingRuleArgs = {
+  config: TraceConfig;
+  key: string;
+  value?: AttributeValue;
```
mikeldking, line 35: same here - docstrings

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export type MaskingRule = {
+  condition: (args: MaskingRuleArgs) => boolean;
+  action: () => AttributeValue | undefined;
```
mikeldking, line 40: docstrings

js/packages/openinference-core/src/trace/trace_config/types.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export type OpenInferenceActiveSpanCallback = (span: OISpan) => void;
```
mikeldking, line 43: docstring

js/packages/openinference-core/test/trace_config/OITracer.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      );
+      expect(mockSpan.setAttributes).toHaveBeenCalledWith({
+        key1: "value1",
+        "input.value": REDACTED_VALUE,
+      });
+
+      expect(span).toBeInstanceOf(OISpan);
+    });
+  });
+  it("should handle overloads correctly", () => {
```
mikeldking, line 91: did you mean overrides?

mikeldking (approved):

js/packages/openinference-core/jest.config.js
Parker-Stafford: sure, just because the module is pretty small? i just didn't love the `../../` everywhere

js/packages/openinference-core/src/trace/trace_config/OITracer.ts
Parker-Stafford: oh yeah let me look into that, can write some tests and fix it if it's not working

js/packages/openinference-core/src/trace/trace_config/types.ts
Parker-Stafford: good idea

js/packages/openinference-core/test/trace_config/OITracer.test.ts
Parker-Stafford: no i meant the overloads for this function 
(name, fn)
(name, opts, fn)
(name, opts, context, fn)

js/packages/openinference-core/jest.config.js
mikeldking: Oh I see. It's most commont I think to just make the root of a repo be @ I think. https://nextjs.org/docs/app/building-your-application/configuring/absolute-imports-and-module-aliases

Parker-Stafford: oh gotcha, yeah we could do that wasn't sure since these were in te monorepo if root made sense, i removed it for now

PR #1054, opened 2024-10-07
feat: Add `get_chain_root_span` utility for langchain instrumentation

### LINKED ISSUE #1052: [feature request] Add an easy way to get the root span from a child span
I want to implement a feedback sending mechanism from the front-end of a chat bot to a **python** backend.
For LangChain, it is possible to add a link in the chain that fetches the current span_id using `from openinference.instrumentation.langchain import get_current_span`, but this only gives the span_id of that link.
It is much more natural to get the root span id, which is the correct one to annotate with the user-sent feedback, but it's currently not easy to accomplish programmatically. See discussion here: https://github.com/Arize-ai/phoenix/discussions/4800

An easy to use mechanism to get the root span_id of the currently running trace, ideally in a way that is agnostic to the currently instrumented LLM framework solution, but also a specific one for each framework would be nice.

I've considered creating the root span manually and then calling a chain within that with the automatically instrumented langchain, but that is very inelegant.

anticorrelator:
partially resolves #4158
resolves #1052 

Adds a utility to get the most recent chain span that can be found traversing up the trace tree. This will help grab the most appropriate span for sending user feedback.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -73,7 +73,7 @@ class LangChainInstrumentor(BaseInstrumentor):  # type: ignore
         if not run:
             return None
 
-        run_id = run.parent_run_id
+        run_id = run.parent_run_id  # start with the first ancestor
 
         while run_id:
             span = self.get_span(run_id)
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -64,18 +64,25 @@ class LangChainInstrumentor(BaseInstrumentor):  # type: ignore
     def get_span(self, run_id: UUID) -> Optional[Span]:
         return self._tracer.get_span(run_id) if self._tracer else None
 
-    def get_root_chain_spans(self, run_id: UUID) -> Optional[List[Span]]:
-        span = self.get_span(run_id)
-        root_chain_spans = []
-        while span and span.get_span_context().is_valid:
-            span_id = span.get_span_context().span_id
-            tracer = self._tracer
-            assert tracer
-            if span_id in tracer._root_span_ids:
-                root_chain_spans.append(span)
+    def get_ancestors(self, run_id: UUID) -> Optional[List[Span]]:
+        ancestors = []
+        tracer = self._tracer
+        assert tracer
 
-            span = tracer._parent_span_by_span_id.get(span_id)  # get parent span
-        return root_chain_spans if root_chain_spans else None
+        run = tracer.run_map.get(str(run_id))
+        if not run:
+            return None
+
+        run_id = run.parent_run_id
+
+        while run_id:
+            span = self.get_span(run_id)
+            if span:
+                ancestors.append(span)
+
+            run = tracer.run_map.get(str(run_id))
+            run_id = run.parent_run_id
+        return ancestors if ancestors else None
 
 
 class _BaseCallbackManagerInit:
@@ -119,7 +126,7 @@ def get_current_span() -> Optional[Span]:
     return LangChainInstrumentor().get_span(run_id)
 
 
-def get_current_root_chain_spans() -> Optional[List[Span]]:
+def get_ancestor_spans() -> Optional[List[Span]]:
     import langchain_core
 
     run_id: Optional[UUID] = None
@@ -133,4 +140,4 @@ def get_current_root_chain_spans() -> Optional[List[Span]]:
             break
     if not run_id:
         return None
-    return LangChainInstrumentor().get_root_chain_spans(run_id)
+    return LangChainInstrumentor().get_ancestors(run_id)
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -4,7 +4,7 @@ from uuid import UUID
 
 from opentelemetry import trace as trace_api
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
-from opentelemetry.trace import Span
+from opentelemetry.sdk.trace import Span
 from wrapt import wrap_function_wrapper  # type: ignore
 
 from openinference.instrumentation import OITracer, TraceConfig
@@ -110,13 +110,12 @@ def get_current_chain_root_span() -> Optional[Span]:
     from openinference.instrumentation.langchain._tracer import IS_CHAIN_SPAN, _spans_by_span_id
 
     span = get_current_span()
-    while span and span.get_span_context().is_valid:
-        if span.attributes.get(IS_CHAIN_SPAN):
+    while span and span.get_span_context().is_valid:  # type: ignore[no-untyped-call]
+        if span.attributes and span.attributes.get(IS_CHAIN_SPAN):
             return span
         # Get parent span ID
-        parent_span_context = span.parent
-        if not parent_span_context or not parent_span_context.span_id:
+        parent_span = span.parent
+        if not parent_span or not parent_span.span_id:
             break
-        parent_span_id = parent_span_context.span_id
-        span = _spans_by_span_id.get(parent_span_id)
+        span = _spans_by_span_id.get(parent_span.span_id)
     return None
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -108,6 +108,7 @@ def get_current_span() -> Optional[Span]:
 
 def get_current_chain_root_span() -> Optional[Span]:
     from openinference.instrumentation.langchain._tracer import _spans_by_span_id
+
     span = get_current_span()
     while span and span.get_span_context().is_valid:
         if span.attributes.get("is_chain_span"):
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -129,6 +129,13 @@ def get_current_span() -> Optional[Span]:
 
 
 def get_ancestor_spans() -> Optional[List[Span]]:
+    """
+    Retrieve the ancestor spans for the current LangChain run.
+
+    This function traverses the LangChain run tree from the current run's parent up to the root,
+    collecting the spans associated with each ancestor run. The list is ordered from the immediate
+    parent of the current run to the root of the run tree.
+    """
     import langchain_core
 
     run_id: Optional[UUID] = None
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -104,3 +104,12 @@ def get_current_span() -> Optional[Span]:
     if not run_id:
         return None
     return LangChainInstrumentor().get_span(run_id)
+
+
+def get_current_chain_root_span() -> Optional[Span]:
+    from openinference.instrumentation.langchain._tracer import current_chain_root_span_stack
+
+    stack = current_chain_root_span_stack.get()
+    if stack:
+        return stack[-1]
+    return None
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -107,9 +107,6 @@ def get_current_span() -> Optional[Span]:
 
 
 def get_current_chain_root_span() -> Optional[Span]:
-    from openinference.instrumentation.langchain._tracer import current_chain_root_span_stack
+    from openinference.instrumentation.langchain._tracer import current_chain_root_span
 
-    stack = current_chain_root_span_stack.get()
-    if stack:
-        return stack[-1]
-    return None
+    return current_chain_root_span.get()
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -107,6 +107,15 @@ def get_current_span() -> Optional[Span]:
 
 
 def get_current_chain_root_span() -> Optional[Span]:
-    from openinference.instrumentation.langchain._tracer import current_chain_root_span
-
-    return current_chain_root_span.get()
+    from openinference.instrumentation.langchain._tracer import _spans_by_span_id
+    span = get_current_span()
+    while span and span.get_span_context().is_valid:
+        if span.attributes.get("is_chain_span"):
+            return span
+        # Get parent span ID
+        parent_span_context = span.parent
+        if not parent_span_context or not parent_span_context.span_id:
+            break
+        parent_span_id = parent_span_context.span_id
+        span = _spans_by_span_id.get(parent_span_id)
+    return None
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -4,7 +4,8 @@ from uuid import UUID
 
 from opentelemetry import trace as trace_api
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
-from opentelemetry.sdk.trace import Span
+
+from opentelemetry.trace import Span
 from wrapt import wrap_function_wrapper  # type: ignore
 
 from openinference.instrumentation import OITracer, TraceConfig
@@ -66,14 +67,16 @@ class LangChainInstrumentor(BaseInstrumentor):  # type: ignore
 
     def get_root_chain_span(self, run_id: UUID) -> Optional[Span]:
         span = self.get_span(run_id)
-        span_id = span.get_span_context().span_id  # type: ignore[no-untyped-call]
-        while span and span.get_span_context().is_valid:  # type: ignore[no-untyped-call]
-            if span_id in self._tracer._root_span_ids:
+        while span and span.get_span_context().is_valid:
+            span_id = span.get_span_context().span_id
+            tracer = self._tracer
+            assert tracer
+            if span_id in tracer._root_span_ids:
                 return span
-            parent_span = span.parent
-            if not parent_span or not parent_span.span_id:
+
+            span = tracer._parent_span_by_span_id.get(span_id)  # get parent span
+            if not span:
                 break
-            span = self._tracer._spans_by_span_id.get(parent_span.span_id)
         return None
 
 
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
@@ -107,11 +107,11 @@ def get_current_span() -> Optional[Span]:
 
 
 def get_current_chain_root_span() -> Optional[Span]:
-    from openinference.instrumentation.langchain._tracer import _spans_by_span_id
+    from openinference.instrumentation.langchain._tracer import IS_CHAIN_SPAN, _spans_by_span_id
 
     span = get_current_span()
     while span and span.get_span_context().is_valid:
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -140,6 +145,10 @@ def _start_trace(self, run: Run) -> None:
             context=parent_context,
             start_time=start_time_utc_nano,
         )
+
+        if run.run_type.lower() == "chain" and span:
+            span.set_attribute(IS_CHAIN_SPAN, True)
```
RogerHYang, line 150: 1. What are the benefits of modifying the span, versus keeping this information external to the span, e.g. in a tuple or a custom object?
2. What if in the future users want a different `run_type`? How adaptable is this approach: do we need to reserve additional attribute keywords for this purpose?

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -34,6 +34,7 @@
 from opentelemetry import context as context_api
 from opentelemetry import trace as trace_api
 from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY
+from opentelemetry.sdk.trace import Span
```
RogerHYang, line 37: What are the benefits of depending on the SDK version of the `Span` instead of the abstract version of from the Trace API? Which specific implementation detail of the SDK are we relying on?

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -104,6 +106,9 @@ def __delitem__(self, key: K) -> None:
             super().__delitem__(key)
 
 
+_spans_by_span_id: Dict[int, Span] = _DictWithLock[int, Span]()
```
RogerHYang, line 109: Although it probably makes no functional difference, but regarding code organization, what are the benefits of making this a module variable (instead of an instance variable)?

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
```diff
@@ -86,12 +90,68 @@ async def f(_: Any) -> Optional[Span]:
...
+        tasks = [loop.run_in_executor(executor, RunnableLambda(f).invoke, 1) for _ in range(n)]
+        await asyncio.gather(*tasks)
+
+    root_span_after_execution = get_current_chain_root_span()
+    assert root_span_after_execution is None, "Root span should be None after execution"
+
+    assert len(root_spans_during_execution) == n, "Did not capture all root spans during execution"
+
+    spans = in_memory_span_exporter.get_finished_spans()
+    assert len(spans) == n, f"Expected {n} spans, but found {len(spans)}"
```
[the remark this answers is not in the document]
anticorrelator: Yeah, I briefly had one of the tests use a `RunnableSequence` instead but it was really hard for me to figure out how to satisfy mypy simultaneously for both the 3.9 and 3.12 run. I can keep working at it if you think it's worth it

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
anticorrelator: The abstract `Span` object doesn't have the `attributes` or `parent` attributes defined on the object, so our ancestor-walking method will fail type checks without using the fully specified sdk Span

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
anticorrelator: it's primarily to make it easily definable in the `__init__` module without needing to retrieve the active instrumentor

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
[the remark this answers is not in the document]
RogerHYang: If it's just an issue with typing, we could do `type: ignore` since static typing is not a primary concern for testing purposes.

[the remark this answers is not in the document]
anticorrelator: the type: ignore works (I think) for 3.12 checks but for 3.9 checks I get an "unused type:ignore" error, happy to try again but I felt a little bit like it was unnecessary

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
RogerHYang: The otel instrumentor is a singleton, so calling `LangChainInstrument()` will work, as shown [here](https://github.com/Arize-ai/openinference/blob/ff43e9ddc0a5f683f80d09139247ad194d6c29af/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py#L106).

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
RogerHYang: Since the parent relations can be obtain from the (abstract) `get_span_context` method, the only reason to depend on the SDK is because we need to write to and read back values from the span attributes, is that correct?

RogerHYang: Since both of these use cases can be mitigated by having our own bookkeeping, ideally we could depend only on the API and no the SDK, so that we can keep the instrumentor agnostic regarding the `tracer_provider` that's given.

RogerHYang: On the other hand, if you don't think it's worthwhile to abide by just the API, then we should add a validation at instantiation time to confirm the SDK implementation, so there's no ambiguity at the user interface level for the instrumentor.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
anticorrelator: I moved all the bookkeeping onto the tracer

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
anticorrelator: I moved all the bookkeeping to the tracer, so I reverted this to use the abstract API

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,20 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+        while span and span.get_span_context().is_valid:
+            span_id = span.get_span_context().span_id
+            tracer = self._tracer
+            assert tracer
+            if span_id in tracer._root_span_ids:
+                return span
+
+            span = tracer._parent_span_by_span_id.get(span_id)  # get parent span
+            if not span:
+                break
```
RogerHYang, line 78: redundant logic
```suggestion
```

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,20 @@ def _uninstrument(self, **kwargs: Any) -> None:
     def get_span(self, run_id: UUID) -> Optional[Span]:
         return self._tracer.get_span(run_id) if self._tracer else None
 
+    def get_root_chain_span(self, run_id: UUID) -> Optional[Span]:
```
RogerHYang, line 67: I think by making the following signature change, we could make this more general and at the same time simplify our internal logic, given the following compositional equivalence. This simplifies our backend logic because we only need a `Dict[UUID, Optional[UUID]]` to track whose parent is whom, and we can just call on the preexisting `self.run_map`.
```python
get_root_chain_span(run_id) is get_span(get_root_chain_run_id(run_id))
```
We just have to also update `get_span` to take `None` as argument, which is a trivial change.
```suggestion
    def get_root_chain_run_id(self, run_id: UUID) -> Optional[UUID]:
```

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,20 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+        span = self.get_span(run_id)
+        while span and span.get_span_context().is_valid:
+            span_id = span.get_span_context().span_id
+            tracer = self._tracer
+            assert tracer
+            if span_id in tracer._root_span_ids:
+                return span
+
+            span = tracer._parent_span_by_span_id.get(span_id)  # get parent span
```
RogerHYang, line 76: Strictly speaking, we should not be using private methods in this scope.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -104,3 +118,20 @@ def get_current_span() -> Optional[Span]:
...
+    config = langchain_core.runnables.config.var_child_runnable_config.get()
+    if not isinstance(config, dict):
+        return None
+    for v in config.values():
+        if not isinstance(v, langchain_core.callbacks.BaseCallbackManager):
+            continue
+        if run_id := v.parent_run_id:
+            break
+    if not run_id:
+        return None
```
RogerHYang, line 136: If you have time, would you mind DRY-ing this up, since it's the same as line 107-119. Thx!

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
RogerHYang: FYI, [`self.run_map`](https://github.com/langchain-ai/langchain/blob/2a1029c53c9c14b177eb61743dc0501f97dd0738/libs/core/langchain_core/tracers/core.py#L117) seems to be all you need.

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
```diff
@@ -92,6 +95,72 @@ async def f(_: Any) -> Optional[Span]:
...
+
+    root_span_after_execution = get_current_root_chain_span()
+    assert root_span_after_execution is None, "Root span should be None after execution"
+
+    assert (
+        len(root_spans_during_execution) == 2 * n
+    ), "Did not capture all root spans during execution"
+
+    assert (
+        len(set(id(span) for span in root_spans_during_execution)) == 2 * n
```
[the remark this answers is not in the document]
anticorrelator: The RunnableLambdas create their own chain, so they are their own root

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
```diff
@@ -92,6 +95,72 @@ async def f(_: Any) -> Optional[Span]:
...
+        assert root_span is not None, "Root span should not be None during execution (async)"
+        root_spans_during_execution.append(root_span)
+        return x + 1
+
+    sequence: RunnableSerializable[int, int] = RunnableLambda[int, int](f) | RunnableLambda[
+        int, int
+    ](f)
+
+    for _ in range(n):
+        await sequence.ainvoke(1)
```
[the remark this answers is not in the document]
anticorrelator: the callables are entirely blocking anyway, I don't think gather does anything different

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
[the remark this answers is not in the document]
RogerHYang: > The RunnableLambdas create their own chain, so they are their own root

If that's the case then this is not really testing the tree climbing procedure. Is that correct?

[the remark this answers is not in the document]
anticorrelator: It's not, though if like you suggested we yield the the ancestor chains we can test how many there are. I didn't end up doing that since it feels like a confusing interface?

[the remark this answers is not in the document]
RogerHYang: This is not related to yielding the ancestor chain. All i'm pointing out is that we are _not_ testing what we have implemented.

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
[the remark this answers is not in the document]
RogerHYang: Do you want to add a sleep like we have done in an [existing test](https://github.com/Arize-ai/openinference/blob/c3ce79402a05e99640b905274a16a8c0af8c04a2/python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py#L76)? I'm just pointing out that f we aren't testing it concurrently then there is no need to have `n` of them: a single one should suffice for a test that is serial.

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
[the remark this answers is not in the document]
anticorrelator: I know that, I'm suggesting yielding the ancestor chain can be an easy way to do that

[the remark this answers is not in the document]
RogerHYang: > I know that

So are you suggesting that the logic you have implemented so far doesn't need to be tested?

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
RogerHYang: I'm just pointing out that this is a simple copy-paste. I'm not suggesting adding abstractions or anything like that. But it's up to you.

python/instrumentation/openinference-instrumentation-langchain/tests/test_instrumentor.py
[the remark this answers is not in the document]
RogerHYang: I’m not saying it has to be tested, but I’ve noticed the logic isn’t being tested. Since we seem to agree on that observation, I’d like to understand why leaving it untested is considered the better option.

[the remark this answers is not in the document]
anticorrelator: I am specifically offering an alternative strategy to testing whether or not it's the right ancestor directly: if we change this implementation to yield ancestors we can do (I think) a good enough test by counting how many ancestors are yielded each time instead of directly checking the ancestor directly.

Regarding needing to test it: I can go either way on it, I'm not positive this test is needed though I agree it would be nice. I'm using tests as a guide here to scaffold the implementation, not necessarily be exhaustive.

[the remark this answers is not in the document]
RogerHYang: > I'm not positive this test is needed though I agree it would be nice.  I'm using tests as a guide here to scaffold the implementation, not necessarily be exhaustive.

While I understand the test may not seem essential right now, it’s important to consider that this repo is updated only occasionally. Given the infrequent changes and the high likelihood that any future updates may be handled by someone new, having tests that are both thorough and precise can be incredibly beneficial. These tests can serve as reliable documentation and help prevent potential issues when the codebase is revisited later.

> I am specifically offering an alternative strategy to testing 

As an aside, I previously mentioned that the self.run_map from the base instrumentor already stores the family tree of UUIDs. You can leverage this to simplify both the implementation and the testing, as it’s the ultimate source of truth for the family tree.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
RogerHYang: > my preference for first implementations

I’m suggesting that if you don’t anticipate making further updates to this repo in the near future, it’s generally a good practice to replace the copy-pasted sections with a helper function. However, if you think it’s better to leave that for the next maintainer, that’s also fine.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,19 @@ def _uninstrument(self, **kwargs: Any) -> None:
     def get_span(self, run_id: UUID) -> Optional[Span]:
         return self._tracer.get_span(run_id) if self._tracer else None
 
+    def get_root_chain_spans(self, run_id: UUID) -> Optional[List[Span]]:
```
RogerHYang, line 67: Unless you disagree, “root” typically refers to just a single node, so naming the function that way would be misleading if it intends to return a list. Based on my reading, this function returns all ancestors that are chains, and skips those that are not, so if the final root of the tree is not a `chain`, it is actually not returned.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,19 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+        root_chain_spans = []
+        while span and span.get_span_context().is_valid:
+            span_id = span.get_span_context().span_id
+            tracer = self._tracer
+            assert tracer
+            if span_id in tracer._root_span_ids:
+                root_chain_spans.append(span)
+
+            span = tracer._parent_span_by_span_id.get(span_id)  # get parent span
+        return root_chain_spans if root_chain_spans else None
```
RogerHYang, line 78: nit. might as well just return an empty list, since it makes no practical difference but simplifies the types.
```suggestion
        return root_chain_spans
```

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,19 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
         return self._tracer.get_span(run_id) if self._tracer else None
 
+    def get_root_chain_spans(self, run_id: UUID) -> Optional[List[Span]]:
+        span = self.get_span(run_id)
+        root_chain_spans = []
+        while span and span.get_span_context().is_valid:
+            span_id = span.get_span_context().span_id
+            tracer = self._tracer
+            assert tracer
+            if span_id in tracer._root_span_ids:
```
RogerHYang, line 74: As mentioned earlier, in this function you could use the existing `tracer.run_map` from the base `Tracer` to simplify your implementation, rather than accessing private attributes in this scope. Since `tracer.run_map` already provides all the necessary functionality, there’s no need to re-implement everything from scratch: the parts highlighted by arrows in the screenshot below would give you the [`parent_run_id`](https://github.com/langchain-ai/langsmith-sdk/blob/5e21109e1cd70ee19db29941c128223ef360a40b/python/langsmith/schemas.py#L254) and [`run_type`](https://github.com/langchain-ai/langsmith-sdk/blob/5e21109e1cd70ee19db29941c128223ef360a40b/python/langsmith/schemas.py#L225) that you're using for your tree traversal algorithm.

Screenshot 2024-10-16 at 8 11 51 AM

RogerHYang (approved):
Approved to unblock. Concerns are noted in comments

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,26 @@ def _uninstrument(self, **kwargs: Any) -> None:
     def get_span(self, run_id: UUID) -> Optional[Span]:
         return self._tracer.get_span(run_id) if self._tracer else None
 
+    def get_ancestors(self, run_id: UUID) -> Optional[List[Span]]:
```
mikeldking, line 67:
```suggestion
    def get_ancestor_spans(self, run_id: UUID) -> Optional[List[Span]]:
```
just to be explicit about the return type

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/__init__.py
```diff
@@ -64,6 +64,26 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+        run_id = run.parent_run_id  # start with the first ancestor
+
+        while run_id:
+            span = self.get_span(run_id)
+            if span:
+                ancestors.append(span)
+
+            run = tracer.run_map.get(str(run_id))
+            run_id = run.parent_run_id
+        return ancestors if ancestors else None
```
mikeldking, line 85: I think for the simplicity of the typing (and maybe just my FP brain) - I think an empty array might be preferable to represent "none"?

mikeldking (approved):

PR #1065, opened 2024-10-16
fix: relax constraints, allow python3.13

MartinoMensio:
Hi,

I changed the max python version to allow other 3.x versions. The local pytest passes so I assume that there is nothing breaking in python3.13?

Let me know if this sounds reasonable :)

Related to https://github.com/Arize-ai/phoenix/issues/4621

Best,
Martino


MartinoMensio:

RogerHYang:
I think we'll have to do this one at a time, so let us do that separately. We'l put you as coauthor. Thanks again

PR #1085, opened 2024-10-30
feat: add tool call id

RogerHYang:
resolves https://github.com/Arize-ai/phoenix/issues/4908


spec/semantic_conventions.md
```diff
@@ -7,83 +7,87 @@ operations used by applications. These conventions are used to populate the `att
...
+| `tool.name`                            | String                      | `"WeatherAPI"`                                                             | The name of the tool being utilized                                                 |
+| `tool.parameters`                      | JSON string                 | `"{ 'a': 'int' }"`                                                         | The parameters definition for invoking the tool                                     |
+| `tool_call.function.arguments`         | JSON string                 | `"{'city': 'London'}"`                                                     | The arguments for the function being invoked by a tool call                         |
+| `tool_call.function.name`              | String                      | `"get_current_weather"`                                                    | The name of the function being invoked by a tool call                               |
+| `tool_call.id`                         | string                      | `"call_62136355"`                                                          | The id of the by a tool call (useful when thare are more than one at the same time) |
```
axiomofjoy, line 69:
```suggestion
| `tool_call.id`                         | string                      | `"call_62136355"`                                                          | The id of the called tool (useful when thare are more than one at the same time) |
```

spec/semantic_conventions.md
axiomofjoy, line 1: Probably makes sense to add a `tool.id` attribute as well so we can connect the called tool with the schema description.

RogerHYang: that doesn't exist because the id is to connect tool call messages from the AID with tool call results sent back to the AID

spec/semantic_conventions.md
RogerHYang: The `call` is the principal object here, not the `tool`, e.g. the same tool can be called with different arguments.

RogerHYang: added tool.id

spec/semantic_conventions.md
RogerHYang: added `tool.id`

Parker-Stafford (approved):

PR #1093, opened 2024-10-31
fix: remove httpx dependency

RogerHYang:
this fixes Conda

### DIFF
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -50,9 +50,10 @@ commands_pre =
   bedrock-latest: uv pip install -U boto3
   mistralai: uv pip install --reinstall {toxinidir}/instrumentation/openinference-instrumentation-mistralai[test]
   mistralai-latest: uv pip install -U mistralai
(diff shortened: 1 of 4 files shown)


python/tox.ini
```diff
@@ -50,7 +50,10 @@ commands_pre =
   bedrock-latest: uv pip install -U boto3
   mistralai: uv pip install --reinstall {toxinidir}/instrumentation/openinference-instrumentation-mistralai[test]
   mistralai-latest: uv pip install -U mistralai
-  openai: uv pip install --reinstall {toxinidir}/instrumentation/openinference-instrumentation-openai[test]
+  openai: uv pip uninstall -r test-requirements.txt
```
axiomofjoy, line 53: What makes his uninstall necessary?

axiomofjoy (approved):

python/tox.ini
RogerHYang: for running locally, because the env is cached

PR #1124, opened 2024-11-16
fix: capture anthropic model name from response

RogerHYang:
when the model argument is latest, we should record the actual model name from the response

Screenshot 2024-11-16 at 12 14 22 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_wrappers.py
@@ -29,7 +29,7 @@ from openinference.semconv.trace import (
 if TYPE_CHECKING:
     from pydantic import BaseModel
 
-    from anthropic.types import Usage
+    from anthropic.types import Message, Usage
 
 
 class _WithTracer(ABC):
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_wrappers.py
```diff
@@ -326,11 +328,16 @@ def _get_llm_token_counts(usage: "Usage") -> Iterator[Tuple[str, Any]]:
     yield LLM_TOKEN_COUNT_COMPLETION, usage.output_tokens
 
 
-def _get_llm_model(arguments: Mapping[str, Any]) -> Iterator[Tuple[str, Any]]:
+def _get_llm_model_arg(arguments: Mapping[str, Any]) -> Iterator[Tuple[str, Any]]:
     if model_name := arguments.get("model"):
         yield LLM_MODEL_NAME, model_name
 
 
+def _get_llm_model(message: "Message") -> Iterator[Tuple[str, Any]]:
```
Parker-Stafford, line 336: little confused by the stuff above, when does one get used and when does the other get used?

RogerHYang: i have renamed them for better clarity

axiomofjoy (approved):

PR #1147, opened 2024-12-05
feat: openinference tracer

### LINKED ISSUE #1149: [instrumentation] convenience methods for chain spans

axiomofjoy:
resolves #1149


python/openinference-instrumentation/tests/test_config.py
```diff
@@ -128,16 +128,17 @@ def test_settings_from_env_vars_and_code(
...
-    os.environ[OPENINFERENCE_HIDE_OUTPUT_TEXT] = str(hide_output_text)
-    os.environ[OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH] = str(base64_image_max_length)
+    monkeypatch.setenv(OPENINFERENCE_HIDE_INPUTS, str(hide_inputs))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_OUTPUTS, str(hide_outputs))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_INPUT_MESSAGES, str(hide_input_messages))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_OUTPUT_MESSAGES, str(hide_output_messages))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_INPUT_IMAGES, str(hide_input_images))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_INPUT_TEXT, str(hide_input_text))
+    monkeypatch.setenv(OPENINFERENCE_HIDE_OUTPUT_TEXT, str(hide_output_text))
+    monkeypatch.setenv(OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH, str(base64_image_max_length))
```
axiomofjoy, line 141: This was interfering with other tests.

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
```diff
@@ -1,4 +1,14 @@
-from .config import REDACTED_VALUE, OITracer, TraceConfig, suppress_tracing
+from .config import (
+    REDACTED_VALUE,
+    OITracer,
+    TraceConfig,
+    TracerProvider,
+    get_input_attributes,
+    get_output_attributes,
+    get_span_kind_attributes,
```
mikeldking, line 8: isn't this ultimately just one attribute? Feels a tad off but I guess I get it.

mikeldking (approved):

python/openinference-instrumentation/tests/test_manual_instrumentation.py
```diff
@@ -0,0 +1,1659 @@
...
+
+    def test_cannot_set_input_on_non_openinference_span(
+        self,
+        in_memory_span_exporter: InMemorySpanExporter,
+        tracer: OITracer,
+    ) -> None:
+        with tracer.start_as_current_span("non-openinference-span") as span:
+            with pytest.raises(ValueError) as exc_info:
+                span.set_input("input")
+            assert str(exc_info.value) == "Cannot set input attributes on a non-OpenInference span"
```
RogerHYang, line 193: nit. instead of relying on runtime errors, would using `@overload` on `def start_span(` to return a `Protocol` help with type checks?

python/openinference-instrumentation/pyproject.toml
```diff
@@ -66,7 +70,21 @@ testpaths = [
...
+  "src/openinference/instrumentation/guardrails/*",
+  "src/openinference/instrumentation/haystack/*",
+  "src/openinference/instrumentation/instructor/*",
+  "src/openinference/instrumentation/langchain/*",
+  "src/openinference/instrumentation/litellm/*",
+  "src/openinference/instrumentation/llama_index/*",
+  "src/openinference/instrumentation/mistralai/*",
+  "src/openinference/instrumentation/openai/*",
+  "src/openinference/instrumentation/smolagents/*",
+  "src/openinference/instrumentation/vertexai/*",
```
RogerHYang, line 87: Thanks for the fix!

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -334,7 +399,124 @@ def _cast_value(
...
+            if pydantic is not None and isinstance(obj, pydantic.BaseModel):
+                return obj.model_dump()
+            if isinstance(obj, datetime):
+                return obj.isoformat()
+            return super().default(obj)
+        except Exception:
+            return str(obj)
+
+
+def safe_json_dumps_io_value(obj: Any, **kwargs: Any) -> str:
```
RogerHYang, line 490: nit. this is also used to serialize parameters, so it's not always for io values

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -349,20 +531,57 @@ def set_attribute(
...
         span.end(end_time)
 
+    def set_input(
+        self,
+        value: Any,
+        *,
+        mime_type: Optional[OpenInferenceMimeType] = None,
+    ) -> None:
+        if OPENINFERENCE_SPAN_KIND not in self._self_important_attributes:
+            raise ValueError("Cannot set input attributes on a non-OpenInference span")
```
RogerHYang, line 555: nit. maybe it's generally better to avoid errors at runtime?

RogerHYang (approved):
Nice work!

python/openinference-instrumentation/src/openinference/instrumentation/__init__.py
axiomofjoy: This refers to OpenInference span kind, just the single attribute.

axiomofjoy: Let me know if you can think of a better naming pattern. Because these are imported from `openinference.instrumentation`, the previous name of `get_openinference_span_kind` did not seem to capture that a dictionary of key-value pairs was being returned. We also don't have to expose these helper functions as user-facing to start.

python/openinference-instrumentation/tests/test_manual_instrumentation.py
axiomofjoy: It's a good question. The return type on `start_span` and `start_as_current_span` would need to be a union of normal OTel `Span`, `ChainSpan`, `AgentSpan`, and `ToolSpan`. If in the end we wind up with a complex union return type of many different span types, my impression is it doesn't wind up being easier to understand or figure out than a single type with clear error messages.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
axiomofjoy: Yeah, I was wondering about that. I am kind of hoping that this is the sort of error a user hits just once that winds up being helpful. The alternative would be that people would be able to add tools to chain spans for example and then be confused about why it is not showing up in the UI. I think we can relax if it proves to be a problem, but I'd like to start out strict.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -334,7 +399,124 @@ def _cast_value(
...
+        value = str(value)
+    elif normalized_mime_type is OpenInferenceMimeTypeValues.JSON:
+        if not isinstance(value, str):
+            value = safe_json_dumps_io_value(value)
+    else:
+        value, normalized_mime_type = _infer_serialized_io_value_and_mime_type(value)
+    attributes = {
+        OUTPUT_VALUE: value,
+    }
+    if normalized_mime_type is not None:
```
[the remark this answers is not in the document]
axiomofjoy: going to punt on this one since it is an optimization and would require changing quite a few tests, but not opposed to tackling in a follow up

Issue #1077, opened 2024-10-22
[testing][js] Exclude tests from build dist


cephalization:
They contribute to npm package size as well as cause duplicate test runs during local development

axiomofjoy:
Good call

morgan-coded:
Hi, I saw the contributing note that non-trivial work should wait for maintainer confirmation. Is this issue still a good candidate for an outside PR?

The scope I'd keep to is JS package build config only: keep tests typechecked, but make the build emit source-only so `dist/test` does not need to be deleted after the fact. Should that include a changeset, or would you prefer no changeset for build config only?

morgan-coded:
Re-ran the build at main (`9182abb`) to check this still holds: `openinference-core`'s build compiles the test suite into `dist/test` — 36 files, 276 KB — then `postbuild` deletes it with `rimraf dist/test`. The published `openinference-instrumentation-langchain` 4.0.18 tarball shows where the delete-after-build approach leads: it ships 20 compiled `dist/examples/` entries, because that tsconfig compiles `examples/` into dist and its `postbuild` deletes only `dist/test`.

The offer from July stands — build config only, tests stay typechecked, emit goes source-only. If this is better handled in-house or isn't a priority, no response needed; I'll leave it there.

PR #1163, opened 2024-12-09
feat(js): Fix ESM builds

### LINKED ISSUE #1144: [feature request] Proper packaging for JS ESM builds
When looking at the `dist` code on npm for `@arizeai/openinference-vercel` and the various other `@arizeai/*` packages, I noticed the `esm` builds appear incorrect. The builds are using extension-less ESM-style imports and exports, e.g., `export { isOpenInferenceSpan } from "./utils";`, and they do not declare `"type": "module"`. in the `package.json` file. These modules won't work in a true ESM environment, and importing the package from a `.mjs` file, relies on the `cjs` build at `dist/src`. 

Most users are probably importing into a TS environment with a bundler, and the only thing they'd miss out on here is tree shaking, but the fix is relatively simple so you might be interested in updating your boilerplate to do the following:

1. Use extensions in all your local file imports `import { foo } from './util'` -> `import { foo } from './util.js'`
2. Ensure the `dist/esm` folder has a package.json with `{ "type": "module" }`. An npm script like this would do the trick:  
```
"postbuild": "echo '{\"type\": \"module\"}' > ./dist/esm/package.json"
```
3. Set an exports statement in your root `package.json`, e.g.,:
```
  "exports": {
    ".": {
      "import": "./dist/esm/index.js",
      "require": "./dist/src/index.js",
      "default": "./dist/src/index.js"
    }
  },
```

Another nit, `@opentelemetry/sdk-trace-base` seems to be required for `@arizeai/openinference-vercel`, but it is only listed as a `devDependency`. 

I'd be happy to open a PR if you're interested.

cephalization:
- Advertise ESM dist files in each package via package.json `exports` field
- Create stub package.json files with `"type": "module"` in each esm dist folder, to let node know to use esm imports
- Add `tsc-alias` tool to build process to update all relative imports in emitted esm dist files to adhere to ESM import specification requirements (add .js to all import extensions)

I tested newly emitted `dist/` outputs for `openinference-instrumentation-openai` in an ESM only node project and we no longer get `import` related errors as we now correctly import esm code. However, **auto-instrumentation itself does not yet work in ESM mode.**

Resolves #1144

### DIFF
--- a/js/.changeset/smart-donuts-attack.md
+++ b/js/.changeset/smart-donuts-attack.md
@@ -6,7 +6,8 @@
 "@arizeai/openinference-core": major
 ---
 
-Experimental ESM support
+
+ESM support
 
 Packages are now shipped as "Dual Package" meaning that ESM and CJS module resolution
 should be supported for each package.
--- a/js/.changeset/smart-donuts-attack.md
+++ b/js/.changeset/smart-donuts-attack.md
@@ -1,0 +1,15 @@
+---
+"@arizeai/openinference-instrumentation-langchain": major
+"@arizeai/openinference-instrumentation-openai": major
+"@arizeai/openinference-semantic-conventions": major
+"@arizeai/openinference-vercel": major
+"@arizeai/openinference-core": major
+---
+
+Experimental ESM support
+
+Packages are now shipped as "Dual Package" meaning that ESM and CJS module resolution
+should be supported for each package.
+
+Support is described as "experimental" because opentelemetry describes support for autoinstrumenting
+ESM projects as "ongoing". See https://github.com/open-telemetry/opentelemetry-js/blob/61d5a0e291db26c2af638274947081b29db3f0ca/doc/esm-support.md
--- a/js/.changeset/smart-donuts-attack.md
+++ b/js/.changeset/smart-donuts-attack.md
@@ -6,7 +6,6 @@
 "@arizeai/openinference-core": major
 ---
 
-
 ESM support
 
 Packages are now shipped as "Dual Package" meaning that ESM and CJS module resolution
--- a/js/packages/openinference-core/package.json
+++ b/js/packages/openinference-core/package.json
@@ -9,9 +9,16 @@
   "description": "OpenInference Core provides utilities shared by all OpenInference SDK packages.",
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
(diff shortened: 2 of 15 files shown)


js/.changeset/smart-donuts-attack.md
```diff
@@ -0,0 +1,15 @@
+---
+"@arizeai/openinference-instrumentation-langchain": major
+"@arizeai/openinference-instrumentation-openai": major
+"@arizeai/openinference-semantic-conventions": major
+"@arizeai/openinference-vercel": major
+"@arizeai/openinference-core": major
+---
+
+Experimental ESM support
```
mikeldking, line 9:
```suggestion

ESM support
```

js/packages/openinference-core/package.json
```diff
@@ -9,10 +9,17 @@
   "description": "OpenInference Core provides utilities shared by all OpenInference SDK packages.",
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
+    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json && tsc-alias -p tsconfig.esm.json",
+    "postbuild": "echo '{\"type\": \"module\"}' > ./dist/esm/package.json; rimraf dist/test",
```
mikeldking, line 13: no way to ignore the test directory in the build step?

js/packages/openinference-instrumentation-langchain/package.json
```diff
@@ -12,12 +12,19 @@
     "url": "https://github.com/Arize-ai/openinference.git"
   },
   "scripts": {
-    "prebuild": "rimraf dist & pnpm run version:update",
-    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
+    "prebuild": "rimraf dist && pnpm run version:update",
+    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json && tsc-alias -p tsconfig.esm.json",
+    "postbuild": "echo '{\"type\": \"module\"}' > ./dist/esm/package.json && rimraf dist/test",
```
mikeldking, line 17: same here - feels a tad weird that we build the tests. I guess that already was there though 😓

mikeldking (approved):

js/packages/openinference-core/package.json
cephalization: It's because we share the same tsconfig for both tests and src, and then build with TSC using that config.

They don't provide an exclude flag at the cli level either.

I filed an issue for this a month or so back, will tackle it in the future

cephalization: See #1077

PR #1188, opened 2025-01-13
fix: Add support for output messages for sync/async

### LINKED ISSUE #1060: [bug][litellm] output message attributes are not instrumented for litellm
Output messages attributes seem to be missing for LiteLLM. The example below uses DSPy and OpenAI.

Screenshot 2024-10-11 at 10 10 04 AM

To reproduce, run the [DSPy tracing tutorial](https://colab.research.google.com/github/Arize-ai/phoenix/blob/main/tutorials/tracing/dspy_tracing_tutorial.ipynb).

nate-mar:
resolves #1060 

Screenshot 2025-01-13 at 1 11 37 AM


mikeldking:
Possibly also resolves https://github.com/Arize-ai/openinference/issues/1189

nate-mar:
The stream tests I added work in that the stream parameter works with the litellm client and the updated handling. However, the result provided by the litellm client when using a mock_response is not actually representative of how streaming will work in a normal setting unfortunately. 

in actuality, the streamed chunks won't be populated automatically for the response. so I will need to remove the handling I added, as it only works for litellm client mocked responses. I will add a separate PR for streaming and will perhaps drop a note to the lite-lm maintainers to add some clarity around the client when using mock_response in the case of streaming.

mikeldking:
pinging @anticorrelator here to take a look

cc @nate-mar

nate-mar:
Github pushes are down atm so I can't push my latest changes. FYI

python/instrumentation/openinference-instrumentation-litellm/tests/test_instrumentor.py
```diff
@@ -540,7 +548,7 @@ def test_image_generation_url(
 
     mock_response_image_gen = ImageResponse(
         created=1722359754,
-        data=[{"b64_json": None, "revised_prompt": None, "url": "https://dummy-url"}],
+        data=[ImageObject(b64_json=None, revised_prompt=None, url="https://dummy-url")], # type: ignore
```
nate-mar, line 551: ImageObject in the external package does not have typed values in its __init__, unlike ImageResponse

python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
   "openinference-instrumentation>=0.1.17",
   "openinference-semantic-conventions>=0.1.9",
   "wrapt",
+  "setuptools",
```
nate-mar, line 34: fixes build error

python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
```diff
@@ -39,6 +40,7 @@ test = [
   "opentelemetry-sdk",
   "opentelemetry-instrumentation-httpx",
   "tenacity",
+  "tokenizers==0.20.3; python_version == '3.8'"
```
nate-mar, line 43: fixes build error; related to https://github.com/huggingface/tokenizers/issues/1691

nate-mar:
>Possibly also resolves https://github.com/Arize-ai/openinference/issues/1189

Yep, should resolve this too.

anticorrelator:
hi @nate-mar if we want to handle streaming it looks like we need to add another handler for the `CustomStreamWrapper` output type: https://github.com/BerriAI/litellm/blob/00f50bc2018eb45dd300e6d7d5ad8b3ab0ff5410/litellm/litellm_core_utils/streaming_handler.py#L59.

All of their streaming logic is there and we can hook into what they're already doing to instrument the stream (such as compiling the final streamed message and attaching it to our span).

Currently we don't handle this streaming return type and only handle `ModelResponse`: https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py#L155

nate-mar:
Sounds good @anticorrelator ! 

For now, if you wouldn't mind giving me a review on this one so we can get it across the line first, that'd be great. Thanks!

anticorrelator (approved):
lgtm!

PR #1203, opened 2025-01-15
fix: Update openinference-vercel semantic convention mapping for output

luke-moehlenbrock:
When using openinference-vercel with the Vercel AI SDK, spans don't have the output attribute.

This is due to Vercel updating their semantic conventions to use `ai.response.text` for the model response instead of `ai.result.text`.

`ai.response.text` was added to the semantic conventions and `ai.result.text` was marked as deprecated in [this commit](https://github.com/vercel/ai/commit/dad775f494157502eb5c30a49c57b8f206f127dc) and `ai.result.text` was removed entirely in [this commit](url)

With the current mapping, there's no output attribute which results in the Phoenix UI only showing the input for the span:
Screenshot 2025-01-15 at 5 37 46 PM
Screenshot 2025-01-15 at 5 37 55 PM

Updating the prefix from `result` to `response` fixes this:
Screenshot 2025-01-15 at 5 38 11 PM
Screenshot 2025-01-15 at 5 38 19 PM

### DIFF
--- a/js/packages/openinference-vercel/src/AISemanticConventions.ts
+++ b/js/packages/openinference-vercel/src/AISemanticConventions.ts
@@ -13,7 +13,7 @@ const AIPrefixes = {
   telemetry: "telemetry",
   prompt: "prompt",
   toolCall: "toolCall",
-  result: "result",
+  result: "response",
 } as const;
 
 const AIUsagePostfixes = {


luke-moehlenbrock:

mikeldking:
Thanks @luke-moehlenbrock ! Great catch.

js/packages/openinference-vercel/src/AISemanticConventions.ts
```diff
@@ -13,7 +13,7 @@ const AIPrefixes = {
   telemetry: "telemetry",
   prompt: "prompt",
   toolCall: "toolCall",
-  result: "result",
+  result: "response",
```
mikeldking, line 16:
```suggestion
  response: "response",
```
Shouldn't this be the case here? @Parker-Stafford

Parker-Stafford: yeah for consistency lets do that, then you'll want to change these
https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-vercel/src/AISemanticConventions.ts#L53-L60
 to response as well and key off `AIPrefixes.response`

Parker-Stafford: and these
https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-vercel/src/utils.ts#L460-L471

luke-moehlenbrock: Thanks for the suggestions @Parker-Stafford and @mikeldking! I went through and changed result to response throughout the package

mikeldking:
Thanks for the fix!

PR #1244, opened 2025-01-31
feat: Add prompt-related semantic conventions

anticorrelator:
resolves https://github.com/Arize-ai/phoenix/issues/6152

- Adds `prompt`- based semantic conventions
- These are not nested under the `llm` prefix as they aren't necessarily associated with llm calls
  - Consider a case where two prompts are pulled and one is chosen by an agent


cephalization (approved):
I am convinced by the reasoning for top level prompt field. Should we update the README tables as well?

PR #1255, opened 2025-02-04
ci: use paths-filter in py and js CI

### LINKED ISSUE #1214: [ci] add noops python and typescript so that CI checks can be enforced for monorepo setup
we need python / typescript no-op checks so that we can enforce CI

RogerHYang:
resolves #1214 

example screenshot of python ci

### DIFF
--- a/.github/workflows/typescript-CI.yaml
+++ b/.github/workflows/typescript-CI.yaml
@@ -2,7 +2,7 @@ name: Typescript CI
 
 on:
   push:
-    branches: [ main, langchainjs ]
+    branches: [ main ]
   pull_request:
     paths:
       - "js/**"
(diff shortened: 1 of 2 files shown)


.github/workflows/typescript-CI.yaml
```diff
@@ -1,54 +1,73 @@
...
 on:
-    push:
-        branches: [main, langchainjs]
-    pull_request:
-        paths:
-            - "js/**"
-    # Allows you to run this workflow manually from the Actions tab
-    workflow_dispatch:
+  push:
+    branches: [ main, langchainjs ]
```
mikeldking, line 5:
```suggestion
    branches: [ main ]
```

mikeldking (approved):

PR #1277, opened 2025-02-08
feat(smolagents): updates to latest and makes examples use opentelemetry-instrument

codefromthecrypt:
This update smolagents code where recent versions drifted:
* [remove of `ManagedAgent`](https://github.com/huggingface/smolagents/pull/484)
* [removal of `max_tokens` from all models](https://github.com/huggingface/smolagents/pull/475)

This also polishes the examples directory and adds a README there, some of the notable changes below:

* use `opentelemetry-instrument`, so that you can easily copy/paste examples without needing to edit in instrumentation
* remove explicit code that is the same as openai default behavior
* add env.example to document important settings that apply to main or otel code
* added missing tests for `opentelemetry-instrument` entrypoint and `._tracer` is a `OITracer`

I can do make the main README, use opentelemetry-instrument, too if this seems alright.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/README.md
+++ b/python/instrumentation/openinference-instrumentation-smolagents/README.md
@@ -14,42 +14,58 @@ pip install openinference-instrumentation-smolagents
 
 ## Quickstart
 
-This quickstart shows you how to instrument your guardrailed LLM application 
+This quickstart shows you how to instrument your LLM agent application.
 
-Install required packages.
+You've already installed openinference-instrumentation-smolagents. Next is to install packages for smolagents,
+Phoenix and `opentelemetry-instrument`, which exports traces to it.
 
 ```shell
-pip install smolagents arize-phoenix opentelemetry-sdk opentelemetry-exporter-otlp
+pip install smolagents arize-phoenix opentelemetry-sdk opentelemetry-exporter-otlp-proto-grpc opentelemetry-distro
 ```
 
-Start Phoenix in the background as a collector. By default, it listens on `http://localhost:6006`. You can visit the app via a browser at the same address. (Phoenix does not send data over the internet. It only operates locally on your machine.)
+Start Phoenix in the background as a collector, which listens on `http://localhost:6006` and default gRPC port 4317.
+Note that Phoenix does not send data over the internet. It only operates locally on your machine.
 
 ```shell
 python -m phoenix.server.main serve
 ```
 
-Set up `SmolagentsInstrumentor` to trace your crew and send the traces to Phoenix at the endpoint defined below.
+Create an example like this:
 
 ```python
-from opentelemetry import trace
-from opentelemetry.sdk.trace import TracerProvider
-from opentelemetry.sdk.trace.export import BatchSpanProcessor
-
-from openinference.instrumentation.smolagents import SmolagentsInstrumentor
-from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
-from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor
-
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
fyi @mikeldking I tested all of these with phoenix. works fine "on my laptop" ™️

codefromthecrypt:
buffed out a couple other drifts and documented them in the PR desc

nate-mar:
Nice! you're on a roll @codefromthecrypt ! :smiling 

Can we leave at least one example, or perhaps add to the readme that you could instrument without opentelemtry-instrument, in case you need some specific customization. Correct me if I'm wrong, but I think there are certain things that you can't do with the no-code approach (batch vs simple span processor?).

codefromthecrypt:
hey @nate-mar, and thanks for the feedback!

I added a ENV var that exports spans every second. This should help for the higher latency examples, which before this would only flush when the program completes or 5s whichever comes first.

I also added a note to the README about what to copy/paste alternatively.

Next q is should I make corresponding updates to the main smolagents instrumentation README? It currently has manual tracing setup there. Besides fuzz I've left that one alone so far.

nate-mar:
Awesome, thanks! This is great!

Sure, go for it re: updating the main smolagents readme ! I would only just mention the same thing about including a blurb/section about the manual config for additional flexibility if needed. and noted about the metrics!

codefromthecrypt:
Ok I pared down the instructions to least necessary. The top-level example needs no special env variables, due to not using openai etc, so that's a bit simpler. PTAL!

python/instrumentation/openinference-instrumentation-smolagents/README.md
```diff
@@ -14,42 +14,57 @@ pip install openinference-instrumentation-smolagents
 
 ## Quickstart
 
-This quickstart shows you how to instrument your guardrailed LLM application 
+This quickstart shows you how to instrument your LLM agent application.
```
codefromthecrypt, line 17: I noticed this copy/paste elsewhere, so will try to scrub for that if I end up nudging another package.

python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
```diff
@@ -95,6 +95,15 @@ def openai_api_key(monkeypatch: pytest.MonkeyPatch) -> str:
     return api_key
 
 
+class TestInstrumentor:
```
codefromthecrypt, line 98: added this to make sure things don't drift and we lose registration by accident

python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
```diff
@@ -95,6 +96,19 @@ def openai_api_key(monkeypatch: pytest.MonkeyPatch) -> str:
...
+    def test_entrypoint_for_opentelemetry_instrument(self) -> None:
+        (instrumentor_entrypoint,) = entry_points(
+            group="opentelemetry_instrumentor", name="smolagents"
+        )
+        instrumentor = instrumentor_entrypoint.load()()
+        assert isinstance(instrumentor, SmolagentsInstrumentor)
+
+    # Ensure we're using the common OITracer from common openinference-instrumentation pkg
+    def test_oitracer(self, instrument: Any) -> None:
```
codefromthecrypt, line 108: shouldn't need this with autouse

```suggestion
    def test_oitracer(self) -> None:
```

nate-mar (approved):
Amazing! thanks again!

codefromthecrypt:
thanks! great timing as I'm online temporarily today and tomorrow

PR #1278, opened 2025-02-08
feat: add entrypoint for use in opentelemetry-instrument

codefromthecrypt:
openinference can become automatically enabled with [opentelemetry's zero code approach for python](https://opentelemetry.io/docs/zero-code/python/#configuring-the-agent), by adding entrypoints into the pyproject.toml.

Specifically, after this, as long as you have the dependencies of your main code, plus openinference-instrumentation-XXXX, and env exported with standard variables, if you launch with `opentelemetry-instrument`, traces will magically appear.

```bash
opentelemetry-instrument python main.py
```

I've added this configuration for all instrumentation, similar to smolagents done in #1276. I've also added a unit test as that helps ensure refactorings won't break this later.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-anthropic/pyproject.toml
@@ -39,6 +39,9 @@ instruments = [
     "anthropic >= 0.30.0",
 ]
 
+[project.entry-points.opentelemetry_instrumentor]
+anthropic = "openinference.instrumentation.anthropic:AnthropicInstrumentor"
+
 [project.urls]
 Homepage = "https://github.com/Arize-ai/openinference/tree/main/python/instrumentation/openinference-instrumentation-anthropic"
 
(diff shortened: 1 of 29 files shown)


codefromthecrypt:
crewai-latest fail seems unrelated, but I'll look into it

codefromthecrypt:
@nate-mar appreciate if you could fix crew-ai-latest. I think crewai-tools needs to be bumped when that's running, which also ends up with a newer version of openai, that seems to either drift the recordings or hit [this in litellm](https://github.com/BerriAI/litellm/pull/8083). I ran out of time and I'm not going to be available a while from monday (though I will be tomorrow sunday UTC+8 morning). If I happen to be away and this becomes ready, feel free to push commits or rebase this without my permission.

ciao!

nate-mar:
No problem! I’m pretty sure I already know what the issue is with
crewai-latest. I’ll take care of it. Appreciate all the work you’ve done on
this!!

nate-mar (approved):

PR #1312, opened 2025-02-27
ci: fix type check on unit tests for anthropic

RogerHYang:


mikeldking (approved):
Never used?

RogerHYang:
> Never used?

doesn't look like it

PR #1373, opened 2025-03-17
fix: attach span to context

### LINKED ISSUE #1372: [bug] context propagation not working with openai agents
The context is not set on openai agents, making it difficult to compose. As reported by a user. Seems this does work with other OTel instrumentors.

RogerHYang:
resolves #1372

Screenshot 2025-03-17 at 8 30 01 AM


axiomofjoy (approved):
🔥

Issue #1379, opened 2025-03-18
[feature request] `NoOpTracer` compatible with `OITracer`


deepankarm:
Do we have a `NoOpOITracer` compatible with the `OITracer` interface? I can use the `NoOpTracer` from `opentelemetry.trace`, but that and `OpenInferenceSpan` have some additional methods that are not compatible.

While `suppress_tracing` is useful in selected contexts, when phoenix is not available (e.g.- unit tests), `NoOpOITracer` enables us to have a global tracer and leave the rest of the code as-is.

mikeldking:
Slack thread https://arize-ai.slack.com/archives/C04R3GXC8HK/p1742210539779959

PR #1380, opened 2025-03-18
feat: noop tracer

deepankarm:
Implement a draft version of `NoOpOITracer` as discussed in https://github.com/Arize-ai/openinference/issues/1379

This class is just to show an example of what I'm using locally. There are no tests added  and there might be better ways of achieving this (feel free to discard the PR, if so) 

I'm initiating the tracer like this 

```python
from openinference.instrumentation.config import NoOpOITracer, TraceConfig
from opentelemetry.trace import NoOpTracer as OTelNoOpTracer

tracer = NoOpOITracer(wrapped=OTelNoOpTracer(), config=TraceConfig())
```

### DIFF
--- a/python/openinference-instrumentation/src/openinference/instrumentation/config.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/config.py
@@ -892,6 +892,48 @@ class OITracer(wrapt.ObjectProxy):  # type: ignore[misc]
         return lambda f: async_wrapper(f) if asyncio.iscoroutinefunction(f) else sync_wrapper(f)
 
 
+class NoOpOITracer(OITracer):
+    def start_span(
+        self,
+        name: str,
+        context: Optional[Context] = None,
+        kind: SpanKind = SpanKind.INTERNAL,
+        attributes: Attributes = None,
+        links: Optional["Sequence[Link]"] = (),
+        start_time: Optional[int] = None,
+        record_exception: bool = True,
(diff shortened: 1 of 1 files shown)


deepankarm:

deepankarm:
recheck

axiomofjoy:
@deepankarm Definitely understand the need for a `NoOpTracer` that works with our additional methods. With respect to the code snippet above, we don't typically recommend directly instantiating a `tracer` instance, but rather using `TracerProvider` and the `get_tracer` method. `OITracer` and the new `NoOpOITracer` are not considered to be public interfaces, and I would like to keep them private/ undocumented to allow us flexibility to change in the future.

Looking at the code, it seems like OTel typically disables instrumentation via the `OTEL_SDK_DISABLED` env var. If we just made our `TracerProvider` return `NoOpOITracer` when that env var is set, would that suit your needs?

axiomofjoy:
Thanks @deepankarm for this fix! I made a few tweaks and added a test [here](https://github.com/Arize-ai/openinference/pull/1409). I would have pushed to your branch, but your `main` branch on your fork is protected.

axiomofjoy:
Hey @deepankarm, after taking a closer look, I'm noticing that things seem to be working as expected when using `OTEL_SDK_DISABLED`. Can you try using that env var and see if it suits your purpose?

axiomofjoy:
Closing this PR out for now, but feel free to follow up in the thread!

PR #1400, opened 2025-03-21
fix: logging fix for smolagents: Model recursive subclasses

IlyaGusev:
The 1.12.0 smolagents version introduced the ApiModel class, which is a Model subclass. Some of the Models there now inherit ApiModel instead of Model, breaking some logic here.

This PR fixes it by recursively getting all Model subclasses.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
@@ -19,6 +19,15 @@ from openinference.instrumentation.smolagents.version import __version__
 _instruments = ("smolagents >= 1.2.2",)
 
 
+def get_all_subclasses(cls):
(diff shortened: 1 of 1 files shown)


IlyaGusev:

njbrake:
@IlyaGusev  looks like this is related to https://github.com/Arize-ai/openinference/pull/1403:  we noticed the issue as well  😆  and fix it by looking at what exported classes inherit from Model. Since ApiModel inherits from Model we used `issubclass`. Not sure which approach is more appropriate for the openinference patch.

axiomofjoy:
I'll take a look at both approaches. Thanks so much for the quick fixes.

axiomofjoy:
@IlyaGusev Not sure, but there might be an issue with this approach where a grandchild of the `Model` class would be wrapped multiple times?

```
Model -> ApiModel -> Subclass of ApiModel
```

Merged in #1403 and closing this PR. Thanks so much for the contribution and let us know if the issue you are seeing is resolved.

PR #1416, opened 2025-03-23
fix: allow prerelease versions of smolagents

sysradium:
When working with a smolagents library of the `main` branch you would typically get something like:
```
DependencyConflict: requested: "smolagents >= 1.2.2" but found: "smolagents 1.13.0.dev0"
```
I understand that technically you can use flag:
```py
SmolagentsInstrumentor().instrument(skip_dep_check=True)
```
However maybe it would have been easier just to allow prerelease versions by default? 1.13.0.dev0 >= 1.2.2 satisfies the constraint in my opinion :)

ref. https://github.com/huggingface/smolagents/blob/main/pyproject.toml#L7


sysradium:

axiomofjoy (approved):

axiomofjoy:
Thanks @sysradium, and sorry for the slow review. Did not know you could do this, nice find!

Issue #1418, opened 2025-03-24
[bug]: OpenAI Agents Processor Doesn't handle tool output of int


njbrake:
**Describe the bug**
`.venv/lib/python3.12/site-packages/openinference/instrumentation/openai_agents/_processor.py`

If obj.output is an int, an exception follows

**To Reproduce**
Run something where the attribute output is an int

**Expected behavior**
There shouldn't be an exception

**Screenshots**

```
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/agents/tracing/spans.py", line 237, in __exit__
    self.finish(reset_current=reset_current)
  File "venv/lib/python3.12/site-packages/agents/tracing/spans.py", line 222, in finish
    self._processor.on_span_end(self)
  File "venv/lib/python3.12/site-packages/agents/tracing/setup.py", line 65, in on_span_end
    processor.on_span_end(span)
  File "venv/lib/python3.12/site-packages/openinference/instrumentation/openai_agents/_processor.py", line 151, in on_span_end
    for k, v in _get_attributes_from_function_span_data(data):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/openinference/instrumentation/openai_agents/_processor.py", line 269, in _get_attributes_from_function_span_data
    if obj.output[0] == "{" and obj.output[-1] == "}":
       ~~~~~~~~~~^^^
TypeError: 'int' object is not subscriptable
```

**Desktop (please complete the following information):**
$ uv pip freeze | grep openinference
openinference-instrumentation==0.1.24
openinference-instrumentation-langchain==0.1.37
openinference-instrumentation-openai-agents==0.1.2
openinference-semantic-conventions==0.1.15

**Additional context**

PR #1419, opened 2025-03-24
fix: openai agent with int tool return

njbrake:
Fixes https://github.com/Arize-ai/openinference/issues/1418


python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
```diff
@@ -264,9 +264,9 @@ def _get_attributes_from_function_span_data(
     if obj.input:
         yield INPUT_VALUE, obj.input
         yield INPUT_MIME_TYPE, JSON
-    if obj.output:
+    if obj.output is not None:
```
njbrake, line 267: to handle cases where obj.output might be a valid value like 0

RogerHYang:

RogerHYang:
This is released in `openinference-instrumentation-openai-agents==0.1.3`

PR #1425, opened 2025-03-25
fix(beeai): empty data serialization

GALLLASMILAN:
The serialisation process works when the memory object or some other data is missing. 

In the second commit [3ba99bad78b0b95a600e5f70680da08e06f37aa3](https://github.com/Arize-ai/openinference/pull/1425/commits/3ba99bad78b0b95a600e5f70680da08e06f37aa3) I only wrapped the whole `getSerializedObjectSafe` method in try/catch block. Git generated more changes, but there are not more changes. 

in the third commit [4ecc1b3931cb924e6616731a7de6949fe0f31acb](https://github.com/Arize-ai/openinference/pull/1425/commits/4ecc1b3931cb924e6616731a7de6949fe0f31acb) I only wrapped the emitter callbacks in try/catch block as well.

### DIFF
--- a/js/packages/openinference-instrumentation-beeai/src/middleware.ts
+++ b/js/packages/openinference-instrumentation-beeai/src/middleware.ts
@@ -325,7 +325,10 @@ export function createTelemetryMiddleware(
             role: msg.role,
           }));
         } catch (e) {
-          diag.warn("Instrumentation error. Unable to map messages to history", e);
+          diag.warn(
+            "Instrumentation error. Unable to map messages to history",
+            e,
+          );
         }
       },
     );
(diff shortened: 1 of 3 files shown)


js/packages/openinference-instrumentation-beeai/src/middleware.ts
```diff
@@ -309,16 +313,20 @@ export function createTelemetryMiddleware(
...
+          generatedMessage = {
+            role: dataObject.role,
+            text: dataObject.text,
+          };
+          history = memory.messages.map((msg) => ({
+            text: msg.text,
+            role: msg.role,
+          }));
+        } catch (e) {
+          diag.warn("Instrumentation error", e);
```
Parker-Stafford, line 328: Only comment is these messages could be a little bit more specific to the block the try catch is wrapping so that debugging is a bit easier. If the blocks are not cohesive enough a more generic message is fine.

In this case something like:

```suggestion
          diag.warn("Instrumentation error. Unable to map messages to history", e);
```

Parker-Stafford (approved):

PR #1429, opened 2025-03-25
fix: include cache tokens in prompt tokens for anthropic

RogerHYang:


python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/anthropic/_messages.py
```diff
@@ -197,8 +197,12 @@ def _attributes_from_message(
             if TYPE_CHECKING:
                 assert_never(block)
     usage = message.usage
-    if usage.input_tokens:
-        yield LLM_TOKEN_COUNT_PROMPT, usage.input_tokens
+    if prompt_tokens := (
+        usage.input_tokens
+        + (usage.cache_creation_input_tokens or 0)
+        + (usage.cache_read_input_tokens or 0)
+    ):
```
mikeldking, line 204: it might be worth for LLMs and future readers to link out https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching here - describing how we actually describe the prompt tokens

mikeldking (approved):

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/anthropic/_messages.py
RogerHYang: added

Issue #1442, opened 2025-03-28
[bug] OpenAI-Agents SDK: Invalid type dict in attribute 'output.value' value sequence.


njbrake:
**Describe the bug**
Sorry I don't have too much info here because the logs aren't showing me a stack trace. However, when I run an agent, I get this error

`Invalid type dict in attribute 'output.value' value sequence. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or None`.

I'm not sure where this error is being thrown, but I'm hoping that someone more familiar with the repository will know where this is coming from :) 

**To Reproduce**
Run the OpenAI-Agents agent 

**Expected behavior**
I would expect no error.

**Screenshots**

**Desktop (please complete the following information):**
$ uv pip freeze | grep openinference
openinference-instrumentation==0.1.25
openinference-instrumentation-langchain==0.1.37
openinference-instrumentation-llama-index==3.3.3
openinference-instrumentation-openai-agents==0.1.4
openinference-instrumentation-smolagents==0.1.8
openinference-semantic-conventions==0.1.16

**Additional context**
I'm sorry I don't have much detail here but I'm hoping this is enough for a maintainer to spot the issue 😬

njbrake:
Ok I'm not super confident that this is the right place to put the fix, but the linked PR does at least remove the error by turning stuff like tuples and dicts into lists and strings.

njbrake:
Fixed vie #1443

PR #1443, opened 2025-03-28
fix: handle error: invalid type dict in attribute 'output.value' value sequence

njbrake:
#1442


RogerHYang (approved):
Thank you! We really appreciate this!

PR #1456, opened 2025-04-01
docs: create gemini_tracing.py

Jgilhuly:


python/instrumentation/openinference-instrumentation-langchain/examples/gemini_tracing.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from langchain_google_genai import ChatGoogleGenerativeAI
+from phoenix.otel import register
+
+# Set Environment Variables
+os.environ["GEMINI_API_KEY"] = ""
```
mikeldking, line 7: you can just assert this at the top of a file rather than hard coding the assignment

python/instrumentation/openinference-instrumentation-langchain/examples/gemini_tracing.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from langchain_google_genai import ChatGoogleGenerativeAI
+from phoenix.otel import register
```
mikeldking, line 4: if doing this we need this in the requirements.txt

python/instrumentation/openinference-instrumentation-langchain/examples/gemini_tracing.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Set Environment Variables
+os.environ["GEMINI_API_KEY"] = ""
+os.environ["PHOENIX_CLIENT_HEADERS"] = "api_key="
+os.environ["PHOENIX_COLLECTOR_ENDPOINT"] = "https://app.phoenix.arize.com"
```
mikeldking, line 9: I would probably not hard code these. users of OI probably don't necessarily use cloud

mikeldking (changes requested):

PR #1489, opened 2025-04-04
feat(llama-index py): capture reasoning, cache, and audio tokens

### LINKED ISSUE #1483: [Token Count] LlamaIndex

caroger:
resolves #1483


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -619,11 +619,37 @@ def _get_token_counts_from_object(usage: object) -> Iterator[Tuple[str, Any]]:
...
+        if (
+            completion_audio_tokens := getattr(completion_tokens_details, "audio_tokens", None)
+        ) is not None:
+            yield LLM_TOKEN_COUNT_COMPLETION_DETAILS_AUDIO, completion_audio_tokens
     if (total_tokens := getattr(usage, "total_tokens", None)) is not None:
         yield LLM_TOKEN_COUNT_TOTAL, total_tokens
 
+    # Anthropic
+    if (input_tokens := getattr(usage, "input_tokens", None)) is not None:
+        yield LLM_TOKEN_COUNT_PROMPT, input_tokens
```
mikeldking, line 643: I think we need to verify whether you need to sum this with the cache tokens for anthropic like @RogerHYang did - see https://github.com/Arize-ai/openinference/pull/1429

caroger: addressed!

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_token_counts.py
```diff
@@ -2,30 +2,86 @@
...
+        # the token details in recorded response in "TestTokenCounts.test_openai.yaml"
+        # was hard coded/manually altered for test assertion
+        for convention, expected_count in [
+            (LLM_TOKEN_COUNT_PROMPT, 9),
+            (LLM_TOKEN_COUNT_COMPLETION, 10),
+            (LLM_TOKEN_COUNT_TOTAL, 19),
+            (LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ, 1),
+            (LLM_TOKEN_COUNT_PROMPT_DETAILS_AUDIO, 2),
+            (LLM_TOKEN_COUNT_COMPLETION_DETAILS_REASONING, 3),
+            (LLM_TOKEN_COUNT_COMPLETION_DETAILS_AUDIO, 4),
```
mikeldking, line 60: Feels slightly brittle that you are asserting hard-coded numbers that are in a vcr recording. Is it possible to  assert this value based on the usage metrics on the output of llm.chat on line 47 so there are no magic numbers?

mikeldking (approved):
LGTM! passing to @RogerHYang

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_token_counts.py
caroger: addressed

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_callback.py
```diff
@@ -617,12 +617,11 @@ def _get_token_counts_from_object(usage: object) -> Iterator[Tuple[str, Any]]:
...
     """
-    if (prompt_tokens := getattr(usage, "prompt_tokens", None)) is not None:
-        yield LLM_TOKEN_COUNT_PROMPT, prompt_tokens
-    if (completion_tokens := getattr(usage, "completion_tokens", None)) is not None:
-        yield LLM_TOKEN_COUNT_COMPLETION, completion_tokens
-    if (total_tokens := getattr(usage, "total_tokens", None)) is not None:
-        yield LLM_TOKEN_COUNT_TOTAL, total_tokens
+
+    def get_value(obj: object, key: str) -> Any:
+        return getattr(obj, key, None)
```
RogerHYang, line 622: since this is used so often, maybe define a general helper that combines both
```python
def get_value(obj: Union[Mapping[str, Any], Any], key: str) -> Any:
    if isinstance(obj, Mapping):
        return obj.get(key)
    return getattr(obj, key, None)
```

RogerHYang (approved):
make sure the update the lower bounds for deps in project.toml, e.g. openinference-semantic-conventions>=0.1.17

PR #1495, opened 2025-04-05
feat: portkey chat completions

harrisonchu:
Support for simple port key autoinstrumentation for chat completion
Screenshot 2025-04-05 at 5 57 29 PM
Screenshot 2025-04-05 at 5 57 40 PM

Screenshot 2025-04-05 at 5 57 49 PM

```
I am writing a new instrumentation integration with a package called `portkey_ai`.

I want you to bootstrap the minimal scaffolding needed to start this project. Reference everyfile underneath the `openinference-instrumentation-groq` folder to get a sense for what you should scaffold.

do not assume anything about the portkey implementation, just setup the scaffolding 
```


python/instrumentation/openinference-instrumentation-portkey/pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+
+
+[project.optional-dependencies]
+instruments = [
+    "portkey_ai >= 0.1.0",
+]
+test = [
+    "portkey_ai >= 0.1.0",
+    "opentelemetry-sdk",
+    "responses",
```
caroger, line 45: why do specify dependencies both here and in `test-requirements.txt`?

for the non-latest test target, we should be testing against the minimum version that we say we support which should be `"portkey_ai== 0.1.0"`

python/instrumentation/openinference-instrumentation-portkey/pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+module = [
+    "portkey_ai.*",
+    "datasets",
+    "openinference.instrumentation.portkey.*",
+    "wrapt",
+]
+
+[tool.ruff]
+line-length = 100
+target-version = "py38"
```
caroger, line 94: I think we've dropped support for python3.8. should bump target-version to py39

python/instrumentation/openinference-instrumentation-portkey/tests/test_instrumentor.py
```diff
@@ -0,0 +1,54 @@
...
+) -> None:
+    portkey = import_module("portkey_ai")
+    client = portkey.Portkey(
+        api_key="REDACTED",
+        virtual_key="REDACTED",
+    )
+    resp = client.chat.completions.create(
+        messages=[{"role": "user", "content": "What's the weather like?"}], model="gpt-4o-mini"
+    )
+    print(resp)
```
caroger, line 27: remove `print()`

python/tox.ini
```diff
@@ -98,6 +100,10 @@ commands_pre =
   anthropic-latest: uv pip install -U anthropic
   smolagents: uv pip install --reinstall {toxinidir}/instrumentation/openinference-instrumentation-smolagents[test]
   autogen: uv pip install --reinstall {toxinidir}/instrumentation/openinference-instrumentation-autogen[test]
+  portkey: uv pip uninstall -r test-requirements.txt
```
caroger, line 103: update if endup consolidating the test requirements into `project.toml`

mikeldking:
I see two spans with duplicate token counts. This will definitely make cost and token counting difficult. If needed I would make the outer span a chain span with maybe portkey in the name, then have a nested LLM span

python/instrumentation/openinference-instrumentation-portkey/pyproject.toml
caroger: disregard - the latest standard is to put all test dependencies under `test-requirements.txt`

python/tox.ini
caroger: disregard

python/instrumentation/openinference-instrumentation-portkey/tests/test_instrumentor.py
```diff
@@ -0,0 +1,54 @@
...
+    spans = in_memory_span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    span = spans[0]
+    attributes = dict(span.attributes or {})
+
+    expected_attributes = {
+        f"{SpanAttributes.LLM_INPUT_MESSAGES}.0.{MessageAttributes.MESSAGE_ROLE}": "user",
```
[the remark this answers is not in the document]
caroger: oh i was talking more about how you set the expected value in this `table`.

For example - the first attribute (MESSAGE_ROLE) has a hardcoded expected value of "user". Instead of hardcoding it, i believe everything that got recorded in the cassette response is also available and accessible by keying into the response object `resp = client.chat.completions.create....` that you got back from the LLM call (L24 above)

again just a NIT :)

harrisonchu:
> I see two spans with duplicate token counts. This will definitely make cost and token counting difficult. If needed I would make the outer span a chain span with maybe portkey in the name, then have a nested LLM span

i might've gotten that screenshot when i was mid development, i do have tests asserting only 1 span and just tested and only get one span back
Screenshot 2025-04-10 at 7 00 25 PM

python/instrumentation/openinference-instrumentation-portkey/tests/test_instrumentor.py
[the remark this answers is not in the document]
harrisonchu: ahh makes sense!

caroger (approved):

PR #1510, opened 2025-04-12
feat(langchain-js): Instrument langchain tool messages

cephalization:
If a langchain message history contained tool messages, it would blow up instrumented results into a big blob of json. Now they render properly in the UI.

### Before

image

### After

image


mikeldking (approved):
Don't forget to add a changeset

PR #1524, opened 2025-04-17
feat(mcp): mcp python context propagation

anuraaga:
I am working with @codefromthecrypt on MCP instrumentation for context propagation and he suggested I can send a PR here for it. This adds instrumentation for MCP Python SDK as an implementation of the discussion in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/246.

The key point is picking `requst.params._meta` as the carrier for context. If a different field was defined in the protocol in the future, it would be relatively simple to migrate to it as the instrumentation otherwise doesn't change at all.

I have not included any examples / quickstart yet since it seemed a bit awkward to me because the instrumentation doesn't generate telemetry itself - it means either having examples that include other instrumentation, or manually creating spans, which both seemed out of scope for a example. In the future if generating MCP telemetry, this problem would be solved. But let me know if you have any advice on this point.


anuraaga:

python/instrumentation/openinference-instrumentation-mcp/tests/collector.py
```diff
@@ -0,0 +1,42 @@
...
+
+@dataclass
+class Telemetry:
+    traces: list[ResourceSpans] = field(default_factory=list)
+
+    def clear(self) -> None:
+        self.traces.clear()
+
+
+class OTLPServer(HTTPServer):
```
anuraaga, line 17: We need to start up an HTTP collector mostly since the `stdio` transport starts a subprocess. I considered just returning the parent span context in the tool's response, but figured it's worth setting this up now so it can be extended to actually generating telemetry in the future without new test setup

python/instrumentation/openinference-instrumentation-mcp/tests/test_instrumenter.py
```diff
@@ -0,0 +1,89 @@
...
+                StdioServerParameters(
+                    command=sys.executable,
+                    args=[server_script],
+                    env={"MCP_TRANSPORT": "stdio", "OTEL_EXPORTER_OTLP_ENDPOINT": otlp_endpoint},
+                )
+            ) as (reader, writer), ClientSession(reader, writer) as client:
+                await client.initialize()
+                yield client
+        case "sse":
+            proc = await asyncio.create_subprocess_exec(
```
anuraaga, line 35: While the SSE version could start the server in the same process, it seemed simplest to keep it consistent with stdio for these tests

codefromthecrypt:
@anuraaga good point on example.. how about raising a PR on [this](https://github.com/elastic/observability-examples/tree/main/genai-function-calling/openai-agents) pointing the requirements.txt to your branch here? It should solve the broken trace. This example already uses openinference for openai-agents and MCP in a way that this should fix, I think.

codefromthecrypt:
probably after folks buy-in in general, the next step could be to find where arize examples live and make something like I did, except in typescript for the https://github.com/Arize-ai/phoenix/tree/main/js/packages/phoenix-mcp server. that would be pretty cool

python/instrumentation/openinference-instrumentation-mcp/tests/collector.py
codefromthecrypt: agree this is the neatest way out I can think of, and also the MCP general discussion has some hints about OTLP proxying on the agent (MCP-client side), so this doubles as a hint of a potential future.

anuraaga:
> how about raising a PR on [this](https://github.com/elastic/observability-examples/tree/main/genai-function-calling/openai-agents) pointing the requirements.txt to your branch here? It should solve the broken trace.

Good idea. Just one extra requirements dependency fixed the trace

https://github.com/elastic/observability-examples/pull/62

> the next step could be to find where arize examples live and make something like I did, except in typescript for the https://github.com/Arize-ai/phoenix/tree/main/js/packages/phoenix-mcp server

Makes sense, will work on Typescript next.

python/instrumentation/openinference-instrumentation-mcp/tests/collector.py
axiomofjoy: Makes sense.

axiomofjoy:
This is a solid foundation to build on. Small comments and questions above, but it's ready to merge after those are addressed.

High-level question: My understanding is that MCP also allows servers to invoke clients. Will we need to handle context propagation in the opposite direction?

axiomofjoy (commented):
This is a solid foundation to build on. Small comments and questions above, but it's ready to merge after those are addressed.

High-level question: My understanding is that MCP also allows servers to invoke clients. Will we need to handle context propagation in the opposite direction?

python/instrumentation/openinference-instrumentation-mcp/tests/test_instrumenter.py
axiomofjoy: Makes sense

mikeldking (approved):
excited to try this out! Will merge and publish - just need to get the release-please in place.

axiomofjoy (approved):

axiomofjoy:
It's out: https://pypi.org/project/openinference-instrumentation-mcp/

Keep cooking

anuraaga:
Thanks for the help! Especially https://github.com/Arize-ai/openinference/pull/1524/commits/82b397266cf0a0a3a60313d09191acce0fd27647 sorry I had forgot to fix that minimum version, but you found it.

anuraaga:
> High-level question: My understanding is that MCP also allows servers to invoke clients. Will we need to handle context propagation in the opposite direction?

This is a good point - I think the standard tools / resources wrappers on top of the protocol (what MCP SDK calls `FastMCP` some times) does not expose sending requests from server to client, while the underlying protocol layer indeed supports it. I'll look at getting demo code of that case set up to explore it - I'm a bit scared of how the trace may render if the parent/child get reversed in the middle of a request, I don't think I've ever tried something like that before.

axiomofjoy:
> his is a good point - I think the standard tools / resources wrappers on top of the protocol (what MCP SDK calls `FastMCP` some times) does not expose sending requests from server to client, while the underlying protocol layer indeed supports it. I'll look at getting demo code of that case set up to explore it - I'm a bit scared of how the trace may render if the parent/child get reversed in the middle of a request, I don't think I've ever tried something like that before.

Awesome, excited to see what you find

codefromthecrypt:
For the follow-up,I think the best use case in the MCP protocol to showcase reverse propagation is [sampling](https://modelcontextprotocol.io/docs/concepts/sampling) (server asking a client to do an LLM completion). I elaborated that here and hope it helps! https://github.com/Arize-ai/openinference/issues/1543

PR #1548, opened 2025-04-22
fix: allow instrumentation to work with later versions of beeai-framework

codefromthecrypt:
Before, we could only work with the initially supported version. Also, the auto-instrumentation entrypoint had a typo.

I also fixed an imprecise dep in the examples, so now the requirements.txt can run all of them.


codefromthecrypt:
added workaround to https://github.com/i-am-bee/beeai-framework/issues/789

codefromthecrypt:
@matoushavlena FYI I have this working and can confirm it works outside of phoenix, too (as the otel promise says). Here's a kibana screen shot of an example I use.

Screenshot 2025-04-23 at 2 13 21 PM

codefromthecrypt:
Obviated by #1552

PR #1549, opened 2025-04-22
fix: instructor output bug fix

satya-turing:


python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
```diff
@@ -118,7 +158,11 @@ def patched_new_func(*args: Any, **kwargs: Any) -> Any:
                             OPENINFERENCE_SPAN_KIND: OpenInferenceSpanKindValues.TOOL,
                             INPUT_VALUE_MIME_TYPE: "application/json",
                             # TODO(harrison): figure out why i cant use args with _get_input_value
-                            INPUT_VALUE: kwargs,
+                            INPUT_VALUE: self._get_input_value(attributes),
```
mikeldking, line 161: it's not setting the MIME type here. is this rendering correctly in Phoenix?

python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
```diff
@@ -85,6 +89,41 @@ class _PatchWrapper:
...
+    def _get_messages(cls, request_params: Any) -> Dict[str, Any]:
+        llm_messages = {}
+        if messages := request_params.get("messages"):
+            prefix = f"{LLM_INPUT_MESSAGES}"
+            if isinstance(messages, Iterable) and not isinstance(messages, (str, bytes)):
+                for idx, message in enumerate(messages):
+                    llm_messages[f"{prefix}.{idx}.{MESSAGE_CONTENT}"] = message["content"]
+                    llm_messages[f"{prefix}.{idx}.{MESSAGE_ROLE}"] = message["role"]
+            elif isinstance(messages, str):
+                llm_messages[f"{prefix}.0.{MESSAGE_CONTENT}"] = messages
```
mikeldking, line 102: does this even render correctly? If it's not a list then it means technically it's a non-converse type of llm call - i would just set it as the `input.value` here?

satya-turing: You're right—this isn't required. We're already assigning the messages object to the input.value, so this part is redundant. I'll go ahead and remove this piece of code.

python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
satya-turing: Yes, the MIME type is being set in the lines(Line No. 159) above.

mikeldking (approved):

PR #1551, opened 2025-04-23
chore: revise usage of openinference-instrumentation-mcp

codefromthecrypt:
This revises usage of openinference-instrumentation-mcp to ensure the instrumentation is initialized prior to importing the MCP modules.

cc @anuraaga in case I goofed something

### DIFF
--- a/js/packages/openinference-instrumentation-mcp/README.md
+++ b/js/packages/openinference-instrumentation-mcp/README.md
@@ -18,19 +18,16 @@ The client and server must be manually instrumented due to the non-traditional m
 be registered as usual using the `registerInstrumentations` function.
 
 ```typescript
-import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
+// First, configure and start your OpenTelemetry NodeSDK per documentation.
+
+// Then, import and initialize MCPInstrumentation before importing the MCP modules.
 import { MCPInstrumentation } from "@arizeai/openinference-instrumentation-mcp";
-import * as MCPClientModule from "@modelcontextprotocol/sdk/client/index";
-import * as MCPServerModule from "@modelcontextprotocol/sdk/server/index";
-
-const provider = new NodeTracerProvider();
-provider.register();
-
 const mcpInstrumentation = new MCPInstrumentation();
-// MCP must be manually instrumented as it doesn't have a traditional module structure
+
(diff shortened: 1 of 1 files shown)


js/packages/openinference-instrumentation-mcp/README.md
```diff
@@ -18,19 +18,16 @@ The client and server must be manually instrumented due to the non-traditional m
 be registered as usual using the `registerInstrumentations` function.
 
 ```typescript
```
anuraaga, line 20: One note is any change here should be applied to the langchain readme as well which has the same issue.

https://github.com/Arize-ai/openinference/tree/main/js/packages/openinference-instrumentation-langchain

However, I don't know if we should suggest this usage. FWIU, it's only in very recent versions of Node where it became even possible to have both `import` and `require` in the same files and it's still considered bad practice with many linters likely rejecting it and needing ignores on the lines.

There is a debug warning based on order here

https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L153

I think it's better to leave the import style closer to normal and potentially just ack the potential debug warn - either way, we probably need to add an ack on it to convince users why they are disabling the linter. Or another approach to remove that warning entirely could be to "give up" on auto-instrumentation and commit to manual for now since we know it doesn't work. It means if the auto-instrumentation issues get fixed in the future, we'd then restore back to supporting it. In the meantime, we remove the registration of auto-instrumentation here

https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-mcp/src/mcp.ts#L91
https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts#L72

which I think will cause that warning to go away.

codefromthecrypt:
hmm good points @anuraaga, and actually you are right that this warning isn't critical. "when in doubt, leave it out" wins for now. Let's spend edit time for bigger wins, since almost no one will turn on that diag logger anyway.

js/packages/openinference-instrumentation-mcp/README.md
codefromthecrypt: I like the idea of not being auto when we know it needs to be manual. Any improvements on this I suspect the leads here will appreciate!

PR #1552, opened 2025-04-23
feat(python): update beeai version

GALLLASMILAN:
Release-As: 0.3.0


mikeldking:
Hey @GALLLASMILAN ! LGTM but it looks like there was already a PR open to address some of this from @codefromthecrypt - I'm good with either. This one has more submodule dependancies so just wanted to make sure the best of both is merged in.

codefromthecrypt:
@GALLLASMILAN so there are some other cleanups in this which I've tested #1548 

Also, should we add gitignore for uv.lock as I don't think other projects check them in here..

python/instrumentation/openinference-instrumentation-beeai/src/openinference/instrumentation/beeai/__init__.py
```diff
@@ -28,7 +28,7 @@
 
 from .middleware import create_telemetry_middleware
 
-_instruments = ("beeai-framework >= 0.1.10, < 0.1.17",)
+_instruments = ("beeai-framework >= 0.1.10, < 0.1.18",)
```
codefromthecrypt, line 31: a lot of instrumentation (most?) don't put an upper bound as doing so is continual maintenance (basically revlock until someone makes a PR like this and then it gets released. I ran into revlock here and would run into again immediately as 0.1.18 is out now

python/instrumentation/openinference-instrumentation-beeai/pyproject.toml
```diff
@@ -33,10 +33,12 @@ dependencies = [
...
 [project.optional-dependencies]
 instruments = [
-  "beeai-framework==0.1.16",
+  "beeai-framework[duckduckgo,wikipedia]==0.1.17",
+  "pydantic-settings>=2.9.1",
 ]
 test = [
-  "beeai-framework==0.1.16",
+  "beeai-framework[duckduckgo,wikipedia]==0.1.17",
+  "pydantic-settings>=2.9.1",
```
codefromthecrypt, line 41: fyi this is fixed in 0.1.18

python/instrumentation/openinference-instrumentation-beeai/src/openinference/instrumentation/beeai/__init__.py
GALLLASMILAN: Unfortunately, the beeai framework is developing very quickly, and we haven't released the first major version yet. This means that the framework does not yet use standard semantic versioning. The braking changes can be hidden in the patch update. This is the reason why we are testing each new published version.
When we use the framework and the instrumentation, we always use the specific fixed version.

GALLLASMILAN (commented):
@codefromthecrypt @mikeldking due to committing `uv.lock` file, we have better control of what versions we use in the repo. Is there any reason to remove it from the git? 
It's ready for the next review.

python/instrumentation/openinference-instrumentation-beeai/src/openinference/instrumentation/beeai/__init__.py
codefromthecrypt: Fair enough as long as leads here can turn around releases as you make them.

Do you mind looking at the PR i made? There were a couple lints i noticed and some places where versions were not coherent. I'm happy to close mine out as long as yours works with today's version of beeai!

GALLLASMILAN: I've updated the instrumentation using the current version BeaAI. 
I had a look at your PR and I fixed the `entry-points` names due to your change (thanks). 
We should use the version range until we release the first major version of BeeAI.

GALLLASMILAN:
Thanks @codefromthecrypt for the collaboration. @mikeldking it's ready to merge.

mikeldking (approved):

PR #1555, opened 2025-04-24
feat(mcp-js): move instrumentation to transport layer

anuraaga:
This reworks the context propagation instrumentation for MCP to instrument `Transport` instead of `Client`/`Server`. I had intentionally picked the latter at first due to concerns of the implications of propagating back from a server to a client, but with more thought, it was too much concern. Notably, if treating as two nodes talking to each other, it seems just fine to have A -> B -> A since the second A just happens to be the same server, but is not the same RPC method. This is the point notably brought up in https://github.com/Arize-ai/openinference/pull/1524#pullrequestreview-2776853857 where MCP explicitly supports a server calling back to the client. This PR adds this usage to the test cases to demonstrate.

Aside from server calling client, we found a couple of other advantages of instrumenting `Transport`

- Some libraries such as vercel AI can accept the `Transport` from `@modelcontextprotocol/sdk` while using hand-written clients. By instrumenting `Transport`, this will be instrumented without explicitly instrumenting vercel AI
- Some libraries don't use `Client`/`Server` at all to set up an MCP stream. Notably, proxies such as in [mcp-inspector](https://github.com/modelcontextprotocol/inspector/blob/6ab7ac3e1ac45b30a63e9eec8a5ee25b4136b109/server/src/mcpProxy.ts#L14) do this. Instrumenting `Transport` handles this case too.

The downside is that for any new transport, instrumentation needs to be updated - this is probably not a frequent event though. The instrumentation logic is the same so it is mostly a registration step.

/cc @codefromthecrypt


codefromthecrypt (approved):
Thanks so much for this @anuraaga! I had hoped both vercel and inspector would be traceable with openinference by my next talk (cloud native summit early next week in auckland)

Bravo and thanks for the support!

@mikeldking fyi another MCP win

mikeldking (approved):
This is great @anuraaga. Cool to  see the new streamable module too.

PR #1560, opened 2025-04-25
fix: removes spurious print(2) from python beeai and converts rest to loggers

codefromthecrypt:
Currently, if you run beeai with this instrumentation, you'll get '2' printed to the console occasionally.

This removes the spurious 2 and converts the rest of the prints to log errors per style in otel. https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L383

cc @GALLLASMILAN @mikeldking

p.s. most otel instrumentation do not do a lot of defensive catching of errors. once the dust settles, another PR might be wise to remove some of this.


mikeldking (approved):
Thanks @codefromthecrypt. Makes sense to me

PR #1562, opened 2025-04-25
fix: Start span before API call

davidgmonical:
resolves: https://github.com/Arize-ai/openinference/issues/1318

Start the span before we call the original llitellm funcs in order to properly instrument and capture the latency

Before:
Screenshot 2025-04-24 at 11 10 10 PM

After:
Screenshot 2025-04-24 at 11 08 52 PM


mikeldking (approved):
Thanks for this @davidgmonical ! Might get a python expert but it LGTM

mikeldking (approved):

PR #1563, opened 2025-04-25
feat(mcp-python): move instrumentation to transport layer

anuraaga:
This is the same as #1555 for Python. There is some more effort here because of the message passing model used causing context breaks within the SDK between transport and handler. While we could continue to instrument at the higher level than handler, I thought it's still better to keep as similar between the languages as possible and instead instrumented the message passing itself as well for in-process context propagation.

I also went ahead and bumped the supported version because there were significant changes to how client message handling is done

https://github.com/modelcontextprotocol/python-sdk/releases/tag/v1.6.0

It seemed simpler to raise the target since this is new instrumentation, though targeting both admittedly has a much larger impact on the tests than the instrumentation itself.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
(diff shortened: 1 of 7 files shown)


codefromthecrypt (approved):
thanks for keeping the code coherent across languages

PR #1570, opened 2025-04-29
feat(tracing): Add dynamic session_id and metadata assignment support

### LINKED ISSUE #1421: [ENHANCEMENT] Tracing -> Allow dynamic session_id and metadata assignment in execution time
No, just want to enhance tracing extensibility

Currently, the `@using_session` and `@using_metadata` decorators require `session_id` and `metadata` values to be set at definition time, which limits dynamic assignment at runtime. This is problematic when handling dynamically generated data, such as request-based session IDs or metadata.

**Proposal**  
- Modify the `__call__` method in `using_session` and `using_metadata` to allow passing dynamic values using `kwargs` at runtime.  
- Suggested modification in `__call__`:

```python
def __call__(self, func):
    async def wrapper(*args, **kwargs):
        session_id = kwargs.get('session_id', self._session_id)
        if session_id:
            with using_session(session_id=session_id):
                return await func(*args, **kwargs)
        else:
            return await func(*args, **kwargs)
    return wrapper
```

Using context manager inside the target function and modifying its signature to allow additional tracing information

**Additional context**
**Use Case**
Instead of setting values at definition time, allow passing them dynamically at runtime:

```python
@using_session(session_id=None)
@using_metadata(metadata=None)
async def process_data(data):
    print(f"Processing data with session_id: {data}")

# Passing dynamic values at runtime
await process_data("input_data", session_id="dynamic-session-id", metadata={"key": "value"})
```

This would provide greater flexibility when handling dynamic data.

Related Docs: [Phoenix Tracing - Add Metadata](https://docs.arize.com/phoenix/tracing/how-to-tracing/add-metadata/customize-spans)

fjprobos:
## Issue
Resolves #1421

## Description
This PR enhances the tracing functionality by allowing dynamic assignment of `session_id` and `metadata` at runtime through the `@using_session` and `@using_metadata` decorators. This change provides greater flexibility when handling dynamically generated data, such as request-based session IDs or metadata.

## Changes
- Modified `__call__` method in `using_session` and `using_metadata` classes to support dynamic values through kwargs
- Added support for runtime assignment of session_id and metadata
- Updated tests to cover new dynamic assignment functionality

## Testing
- Added new test cases for dynamic session_id and metadata assignment
- Verified existing functionality remains unchanged
- All tests pass with tox

## Example Usage
```python
@using_session(session_id=None)
@using_metadata(metadata=None)
async def process_data(data):
    print(f"Processing data with session_id: {data}")

# Passing dynamic values at runtime
await process_data("input_data", session_id="dynamic-session-id", metadata={"key": "value"})
```

## Related Documentation
- [Phoenix Tracing - Add Metadata](https://docs.arize.com/phoenix/tracing/how-to-tracing/add-metadata/customize-spans)

## Manual check

-> Resulting session_id and metadata were verified in postgres collector db.

```
from phoenix.otel import register
import openai

from openinference.instrumentation import using_session, using_metadata

tracer_provider = register(
    project_name="test-dynamic-decorators", 
    auto_instrument=True, 
    endpoint="http://0.0.0.0:6006/v1/traces"
)

client = openai.OpenAI()
response = client.chat.completions.create(
    model="gpt-4o",
    messages=[{"role": "user", "content": "Write a haiku."}],
)
print(response.choices[0].message.content)


@using_session("default-session-id")
def process_with_session(data, session_id=None):
    client = openai.OpenAI()
    response = client.chat.completions.create(
        model="gpt-4o",
        messages=[{"role": "user", "content": "Write a haiku."}],
    )
    print(response.choices[0].message.content)

@using_metadata({"default": "value", "environment": "test"})
def process_with_metadata(data, metadata=None):
    client = openai.OpenAI()
    response = client.chat.completions.create(
        model="gpt-4o",
        messages=[{"role": "user", "content": "Write a haiku."}],
    )
    print(response.choices[0].message.content)

@using_session("default-session-id")
@using_metadata({"default": "value", "environment": "test"})
def process_with_both(data, session_id=None, metadata=None):
    client = openai.OpenAI()
    response = client.chat.completions.create(
        model="gpt-4o",
        messages=[{"role": "user", "content": "Write a haiku."}],
    )
    print(response.choices[0].message.content)

if __name__ == "__main__":
    print("Example 1: Using using_session with default value")
    result = process_with_session("test data")
    print(result)
    
    print("\nExample 1: Using using_session with dynamic value")
    result = process_with_session("test data", session_id="dynamic-session-123")
    print(result)
    
    print("\nExample 2: Using using_metadata with default value")
    result = process_with_metadata("test data")
    print(result)
    
    print("\nExample 2: Using using_metadata with dynamic value")
    result = process_with_metadata("test data", metadata={"dynamic": "value", "custom": "data"})
    print(result)
    
    print("\nExample 3: Using both decorators with default values")
    result = process_with_both("test data")
    print(result)
    
    print("\nExample 3: Using both decorators with dynamic values")
    result = process_with_both("test data", session_id="dynamic-session-456", metadata={"dynamic": "value", "custom": "data"})
    print(result)
```


fjprobos:

axiomofjoy:
> > Hey @fjprobos, thanks so much for your contribution. I definitely hear your pain on the `using_metadata` and `using_session` decorators. They are primarily intended as context managers, and their usefulness as decorators is much less clear since the values they take are usually dynamic, as you point out.
> > If I understand the approach of the current PR correctly, it looks like it changes `using_session` to change the input signature of the decorated function to accept additional keyword arguments for session ID. This breaks an expectation I have of decorators to be removable without breaking application code at the call site. After discussing with the team, the approach we would like to explore is to add an `openinference.instrumentation.context` decorator that accepts a `process_input` argument that allow the user to set certain attributes in the OTel context so they are attached to all spans within the context.
> > ```python
> > import openinference.instrumentation as oi
> > import random
> > 
> > def process_input(input_messages, user_id):
> >     return oi.get_session_attributes(session_id=f"{user-id}-{random.random()}")
> > 
> > @oi.context(process_input=process_input)
> > def f(input_messages, user_id) -> str:
> >     # code here
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We may also add similar affordances for context variables to the existing `tracer.chain` and `tracer.tool` decorators.
> > We'll address this change as soon as one of our core devs free up. As a temporary workaround, `using_session` does work as a context manager, so the following pattern is possible:
> > ```python
> > def f():
> >     with using_session(<dynamic-session-id>):
> >         # your code
> > ```
> 
> Hi @axiomofjoy ,
> 
> Thank you for your response and for considering my contribution. I understand your concerns about decorator removability and the proposed `context` decorator approach.
> 
> However, I have some concerns about the proposed solution:
> 
> 1. **Dynamic Value Assignment**: The `process_input` approach requires defining how to generate session_id and metadata at function definition time, rather than allowing arbitrary values to be passed at call time. This limits flexibility in scenarios where we need to:
>    
>    * Pass session IDs from external sources (e.g., request headers)
>    * Set metadata based on runtime conditions
>    * Handle multiple different session IDs in the same function call
> 2. **Use Case Limitations**: While the context manager approach works as a temporary solution, it's not ideal for:
>    
>    * Functions that are called from multiple places with different session IDs
>    * Scenarios where we need to pass session IDs through multiple layers of function calls
>    * Cases where we want to maintain the same function signature across different implementations
> 3. **Alternative Suggestion**: Could we explore a solution that:
>    
>    * Maintains the ability to pass arbitrary values at call time
>    * Preserves function signature compatibility
>    * Doesn't break existing code when decorators are removed
>    * Works seamlessly with the existing tracing infrastructure
> 
> After considering the alternatives, I still lean towards the current implementation that modifies the function signature. While it does have the drawback of requiring parameter changes if the decorator is removed, it provides the most straightforward and flexible way to pass dynamic values at call time. The current implementation:
> 
> * Allows for immediate and direct value assignment
> * Works seamlessly with existing function calls
> * Maintains clarity in the code
> * Provides the flexibility needed for real-world use cases
> 
> I'd be happy to discuss alternative approaches that could satisfy both the need for dynamic value assignment and the requirement for decorator removability.
> 
> Best regards, Francisco

Hey @fjprobos, thanks for the thoughtful reply.

If the session ID or metadata is contained in or derived from the input to the function in a consistent and well-defined way, then I think a decorator pattern could make sense. But if the session ID or metadata is completely separate/ decoupled from the input to the function, context managers feel like the most appropriate pattern to me.

```python
with using_session("session-id"):
    output = run_chain("How are you?")
```

```python
with using_metadata({"category": "weather"}):
    output = run_agent("How is the weather")
```

This lets you pass arbitrary values for session ID and metadata that are decoupled from the input to the function. This sounds to me like what you are after, but let me know if I am missing something.

fjprobos:
> Hey @fjprobos, thanks for the thoughtful reply.
> 
> If the session ID or metadata is contained in or derived from the input to the function in a consistent and well-defined way, then I think a decorator pattern could make sense. But if the session ID or metadata is completely separate/ decoupled from the input to the function, context managers feel like the most appropriate pattern to me.
> 
> ```python
> with using_session("session-id"):
>     output = run_chain("How are you?")
> ```
> 
> ```python
> with using_metadata({"category": "weather"}):
>     output = run_agent("How is the weather")
> ```
> 
> This lets you pass arbitrary values for session ID and metadata that are decoupled from the input to the function. This sounds to me like what you are after, but let me know if I am missing something.

Hi @axiomofjoy ,

Thank you for your response. Indeed, we are currently using the context manager approach.

However, I'd like to share some practical challenges we've encountered with this:

1. **Code Verbosity**: When we need to apply session IDs or metadata in multiple places, we end up with nested context managers that can make the code harder to read and maintain. For example:

```python
# Current approach with context managers
with using_session("session-id"):
    with using_metadata({"category": "weather"}):
            output = func("How are you?")
```

2. **Maintenance Overhead**: When we need to add or modify tracing attributes across multiple function calls, we have to update each context manager individually. This becomes particularly challenging in larger codebases where the same function might be called from different places with different tracing requirements.

3. **Error Prone**: It's easy to forget to wrap a function call in the appropriate context managers, especially when working with multiple developers or in complex code paths.

The decorator approach we proposed would allow us to:
- Keep the tracing configuration close to the function definition
- Reduce code duplication
- Make it easier to maintain consistent tracing across the codebase
- Still maintain the flexibility to override values when needed

I understand your concerns about coupling, but in practice, we've found that having the tracing configuration as part of the function definition makes the code more maintainable and less prone to errors. Not sure if there is any better approach. I ran an OpenAI Deep Research about this, but didn't find a nicer solution: It suggested `contextvars` but that could have race condition issues.

Regards,

Francisco

axiomofjoy:
> > Hey @fjprobos, thanks for the thoughtful reply.
> > If the session ID or metadata is contained in or derived from the input to the function in a consistent and well-defined way, then I think a decorator pattern could make sense. But if the session ID or metadata is completely separate/ decoupled from the input to the function, context managers feel like the most appropriate pattern to me.
> > ```python
> > with using_session("session-id"):
> >     output = run_chain("How are you?")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> > with using_metadata({"category": "weather"}):
> >     output = run_agent("How is the weather")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This lets you pass arbitrary values for session ID and metadata that are decoupled from the input to the function. This sounds to me like what you are after, but let me know if I am missing something.
> 
> Hi @axiomofjoy ,
> 
> Thank you for your response. Indeed, we are currently using the context manager approach.
> 
> However, I'd like to share some practical challenges we've encountered with this:
> 
> 1. **Code Verbosity**: When we need to apply session IDs or metadata in multiple places, we end up with nested context managers that can make the code harder to read and maintain. For example:
> 
> ```python
> # Current approach with context managers
> with using_session("session-id"):
>     with using_metadata({"category": "weather"}):
>             output = func("How are you?")
> ```
> 
> 2. **Maintenance Overhead**: When we need to add or modify tracing attributes across multiple function calls, we have to update each context manager individually. This becomes particularly challenging in larger codebases where the same function might be called from different places with different tracing requirements.
> 3. **Error Prone**: It's easy to forget to wrap a function call in the appropriate context managers, especially when working with multiple developers or in complex code paths.
> 
> The decorator approach we proposed would allow us to:
> 
> * Keep the tracing configuration close to the function definition
> * Reduce code duplication
> * Make it easier to maintain consistent tracing across the codebase
> * Still maintain the flexibility to override values when needed
> 
> I understand your concerns about coupling, but in practice, we've found that having the tracing configuration as part of the function definition makes the code more maintainable and less prone to errors. Not sure if there is any better approach. I ran an OpenAI Deep Research about this, but didn't find a nicer solution: It suggested `contextvars` but that could have race condition issues.
> 
> Regards,
> 
> Francisco

Hey @fjprobos, with respect to the issue of deep nesting, I think this can be addressed with something like this.

```python
with using_session("session-id"), using_metadata({"category": "weather"}):
    output = func("How are you?")
```

We also have a `using_attributes` context manager that might help:

```python
with using_attributes(
    session_id="session_id",
    metadata={"key": "value"}
):
    output = func("How are you?")
```

I agree that using a context manager incurs a few extra lines of code. I'm not sure I see how passing a session ID to a context manager instead of directly into a function incurs significant maintenance overhead, though.

I'm not against decorators for context attributes like session ID. I think some of the patterns I mentioned previously could make sense. However, I am hesitant to introduce decorators that mutate an input signature, since this breaks a [principle of transparency](https://stackoverflow.com/a/15347207) that in my understanding is fundamental to the decorator pattern. This means certain guarantees around composability of decorators and the ability to remove decorators without breaking application code no longer hold. If a user wants to create a function that accepts session ID and metadata, they can always define such a function in their application code.

fjprobos:
> > > Hey @fjprobos, thanks for the thoughtful reply.
> > > If the session ID or metadata is contained in or derived from the input to the function in a consistent and well-defined way, then I think a decorator pattern could make sense. But if the session ID or metadata is completely separate/ decoupled from the input to the function, context managers feel like the most appropriate pattern to me.
> > > ```python
> > > with using_session("session-id"):
> > >     output = run_chain("How are you?")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```python
> > > with using_metadata({"category": "weather"}):
> > >     output = run_agent("How is the weather")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This lets you pass arbitrary values for session ID and metadata that are decoupled from the input to the function. This sounds to me like what you are after, but let me know if I am missing something.
> > 
> > 
> > Hi @axiomofjoy ,
> > Thank you for your response. Indeed, we are currently using the context manager approach.
> > However, I'd like to share some practical challenges we've encountered with this:
> > 
> > 1. **Code Verbosity**: When we need to apply session IDs or metadata in multiple places, we end up with nested context managers that can make the code harder to read and maintain. For example:
> > 
> > ```python
> > # Current approach with context managers
> > with using_session("session-id"):
> >     with using_metadata({"category": "weather"}):
> >             output = func("How are you?")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. **Maintenance Overhead**: When we need to add or modify tracing attributes across multiple function calls, we have to update each context manager individually. This becomes particularly challenging in larger codebases where the same function might be called from different places with different tracing requirements.
> > 3. **Error Prone**: It's easy to forget to wrap a function call in the appropriate context managers, especially when working with multiple developers or in complex code paths.
> > 
> > The decorator approach we proposed would allow us to:
> > 
> > * Keep the tracing configuration close to the function definition
> > * Reduce code duplication
> > * Make it easier to maintain consistent tracing across the codebase
> > * Still maintain the flexibility to override values when needed
> > 
> > I understand your concerns about coupling, but in practice, we've found that having the tracing configuration as part of the function definition makes the code more maintainable and less prone to errors. Not sure if there is any better approach. I ran an OpenAI Deep Research about this, but didn't find a nicer solution: It suggested `contextvars` but that could have race condition issues.
> > Regards,
> > Francisco
> 
> Hey @fjprobos, with respect to the issue of deep nesting, I think this can be addressed with something like this.
> 
> ```python
> with using_session("session-id"), using_metadata({"category": "weather"}):
>     output = func("How are you?")
> ```
> 
> We also have a `using_attributes` context manager that might help:
> 
> ```python
> with using_attributes(
>     session_id="session_id",
>     metadata={"key": "value"}
> ):
>     output = func("How are you?")
> ```
> 
> I agree that using a context manager incurs a few extra lines of code. I'm not sure I see how passing a session ID to a context manager instead of directly into a function incurs significant maintenance overhead, though.
> 
> I'm not against decorators for context attributes like session ID. I think some of the patterns I mentioned previously could make sense. However, I am hesitant to introduce decorators that mutate an input signature, since this breaks a [principle of transparency](https://stackoverflow.com/a/15347207) that in my understanding is fundamental to the decorator pattern. This means certain guarantees around composability of decorators and the ability to remove decorators without breaking application code no longer hold. If a user wants to create a function that accepts session ID and metadata, they can always define such a function in their application code.

It's all good @axiomofjoy 

Thank you for your thoughtful feedback. I understand the importance of adhering to design patterns and the value they bring to maintainability and clarity. While I believe there are cases where deviating from these patterns can be beneficial, especially when clearly documented, I respect your expertise and the broader context you have in making these decisions.

I appreciate the opportunity to discuss this and trust your judgment in determining the best path forward for the project. Please feel free to close this PR if you believe it's the right decision.

Thanks again for this good discussion.

axiomofjoy:
> > > > Hey @fjprobos, thanks for the thoughtful reply.
> > > > If the session ID or metadata is contained in or derived from the input to the function in a consistent and well-defined way, then I think a decorator pattern could make sense. But if the session ID or metadata is completely separate/ decoupled from the input to the function, context managers feel like the most appropriate pattern to me.
> > > > ```python
> > > > with using_session("session-id"):
> > > >     output = run_chain("How are you?")
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ```python
> > > > with using_metadata({"category": "weather"}):
> > > >     output = run_agent("How is the weather")
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This lets you pass arbitrary values for session ID and metadata that are decoupled from the input to the function. This sounds to me like what you are after, but let me know if I am missing something.
> > > 
> > > 
> > > Hi @axiomofjoy ,
> > > Thank you for your response. Indeed, we are currently using the context manager approach.
> > > However, I'd like to share some practical challenges we've encountered with this:
> > > 
> > > 1. **Code Verbosity**: When we need to apply session IDs or metadata in multiple places, we end up with nested context managers that can make the code harder to read and maintain. For example:
> > > 
> > > ```python
> > > # Current approach with context managers
> > > with using_session("session-id"):
> > >     with using_metadata({"category": "weather"}):
> > >             output = func("How are you?")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 2. **Maintenance Overhead**: When we need to add or modify tracing attributes across multiple function calls, we have to update each context manager individually. This becomes particularly challenging in larger codebases where the same function might be called from different places with different tracing requirements.
> > > 3. **Error Prone**: It's easy to forget to wrap a function call in the appropriate context managers, especially when working with multiple developers or in complex code paths.
> > > 
> > > The decorator approach we proposed would allow us to:
> > > 
> > > * Keep the tracing configuration close to the function definition
> > > * Reduce code duplication
> > > * Make it easier to maintain consistent tracing across the codebase
> > > * Still maintain the flexibility to override values when needed
> > > 
> > > I understand your concerns about coupling, but in practice, we've found that having the tracing configuration as part of the function definition makes the code more maintainable and less prone to errors. Not sure if there is any better approach. I ran an OpenAI Deep Research about this, but didn't find a nicer solution: It suggested `contextvars` but that could have race condition issues.
> > > Regards,
> > > Francisco
> > 
> > 
> > Hey @fjprobos, with respect to the issue of deep nesting, I think this can be addressed with something like this.
> > ```python
> > with using_session("session-id"), using_metadata({"category": "weather"}):
> >     output = func("How are you?")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We also have a `using_attributes` context manager that might help:
> > ```python
> > with using_attributes(
> >     session_id="session_id",
> >     metadata={"key": "value"}
> > ):
> >     output = func("How are you?")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I agree that using a context manager incurs a few extra lines of code. I'm not sure I see how passing a session ID to a context manager instead of directly into a function incurs significant maintenance overhead, though.
> > I'm not against decorators for context attributes like session ID. I think some of the patterns I mentioned previously could make sense. However, I am hesitant to introduce decorators that mutate an input signature, since this breaks a [principle of transparency](https://stackoverflow.com/a/15347207) that in my understanding is fundamental to the decorator pattern. This means certain guarantees around composability of decorators and the ability to remove decorators without breaking application code no longer hold. If a user wants to create a function that accepts session ID and metadata, they can always define such a function in their application code.
> 
> It's all good @axiomofjoy
> 
> Thank you for your thoughtful feedback. I understand the importance of adhering to design patterns and the value they bring to maintainability and clarity. While I believe there are cases where deviating from these patterns can be beneficial, especially when clearly documented, I respect your expertise and the broader context you have in making these decisions.
> 
> I appreciate the opportunity to discuss this and trust your judgment in determining the best path forward for the project. Please feel free to close this PR if you believe it's the right decision.
> 
> Thanks again for this good discussion.

Thanks for the discussion @fjprobos. Will do.

PR #1576, opened 2025-04-30
feat(beeai): updates so we can run 0.1.19

codefromthecrypt:
this also removes a spurious uv.lock file not used here.

note: this setup is very strict and if you don't use exactly the same version as published you will get no spans. I think this should be revisited as it is time consuming for people who don't remember this.

cc @mikeldking


python/instrumentation/openinference-instrumentation-beeai/pyproject.toml
```diff
@@ -33,10 +33,10 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "beeai-framework[duckduckgo,wikipedia]==0.1.18",
+  "beeai-framework[duckduckgo,wikipedia]==0.1.19",
```
codefromthecrypt, line 36: this is also very unusual in openinference. usually it is a min range, where it is exact here

mikeldking: I agree - though it might be that the framework is a bit unstable. Feel free to just change to a min version. cc GALLLASMILAN

mikeldking (approved):

python/instrumentation/openinference-instrumentation-beeai/pyproject.toml
codefromthecrypt: thanks I set the min version to match what is in instruments `_instruments = ("beeai-framework >= 0.1.10",)` then made tests use min latest (0.1.19).

This will prevent disappearing spans which so far haven't happened with same caveat as all the other <1.x libraries here... there can be shifts and we want to keep tests running and respond if something breaks!

python/instrumentation/openinference-instrumentation-beeai/pyproject.toml
```diff
@@ -33,10 +33,10 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "beeai-framework[duckduckgo,wikipedia]==0.1.18",
+  "beeai-framework >= 0.1.10",
 ]
 test = [
-  "beeai-framework[duckduckgo,wikipedia]==0.1.18",
```
codefromthecrypt, line 39: I removed the qualification here because we don't actually instrument the duckduckgo or wikipedia code at the moment.

mikeldking (commented):
LGTM

PR #1614, opened 2025-05-12
fix: Agno duplicate model invoke

### LINKED ISSUE #1613: [bug] Agno Instrumentation unique model traces
**Describe the bug**
Currently the agno model traces are duplicated in a nested structure.

dirkbrnd:
There is a bug in Agno instrumentation that causes duplicate model traces to be produced.

Screenshot 2025-05-12 at 21 43 20

Fixes #1613

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/tests/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-agno/tests/test_instrumentor.py
@@ -68,11 +68,10 @@ def test_agno_instrumentation(
             name="News Agent",  # For best results, set a name that will be used by the tracer
             model=OpenAIChat(id="gpt-4o-mini"),
             tools=[DuckDuckGoTools()],
-            debug_mode=True,
         )
         agent.run("What's trending on Twitter?")
     spans = in_memory_span_exporter.get_finished_spans()
-    assert len(spans) == 4
+    assert len(spans) == 2
     checked_spans = 0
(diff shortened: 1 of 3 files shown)


python/openinference-instrumentation/src/openinference/instrumentation/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.28"
+__version__ = "0.1.29"
```
mikeldking, line 1: this is automated via release-please via conventional commits. Ill rebase and remove!

python/instrumentation/openinference-instrumentation-agno/tests/test_instrumentor.py
```diff
@@ -110,4 +109,4 @@ def test_agno_instrumentation(
             assert attributes.get("llm.model_name") == "gpt-4o-mini"
             assert attributes.get("llm.provider") == "OpenAI"
             assert span.status.is_ok
-    assert checked_spans == 4
+    assert checked_spans == 2
```
mikeldking, line 112: I think the tests might be flaky due to the duckdockgo tool. Might want to move off of a tool that requires a network call.

mikeldking (approved):
Thanks for this. I think there are a few more things we probably can do - parse the output messages for LLM spans and possibly add the token counts (though composition does work).

Will release this as it will help a lot with the telemetry

PR #1618, opened 2025-05-13
refactor: Add better LLM instrumentation for Agno Instrumentor

dirkbrnd:


python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -213,24 +203,20 @@ async def arun(
...
                     span.set_status(trace_api.StatusCode.OK)
-                    async for response in await wrapped(*args, **kwargs):
+                    async for response in wrapped(*args, **kwargs):
                         yield response
+                    run_response = agent.run_response
+                    span.set_status(trace_api.StatusCode.OK)
+                    span.set_attribute(OUTPUT_VALUE, run_response.to_json())
                 else:
-                    run_response = await wrapped(*args, **kwargs)
-                    span.set_attribute(
```
dirkbrnd, line 220: This should be on the LLM, but at the moment we don't have a good way to parse this on the LLM layer. Will improve this on Agno

mikeldking: composition with the other instrumentors (e.g. OpenAIInstrumentor) really helps as you can replay the invocations and get token counts from those.
Screenshot 2025-05-13 at 9 21 51 AM

PR #1634, opened 2025-05-14
fix(mcp): support mcp>=1.8.0 with new SessionMessage structure

### LINKED ISSUE #1616: [bug] MCP instrumentation and context propagation is not working
**Describe the bug**
Unable to reproduce this:
https://github.com/Arize-ai/phoenix/tree/main/tutorials/mcp/tracing_between_mcp_client_and_server
https://www.youtube.com/watch?v=dhK2pWQdGjk
The traces are disjointed.
Could someone validate this issue, I could look into solving it

**To Reproduce**
https://github.com/Arize-ai/phoenix/tree/main/tutorials/mcp/tracing_between_mcp_client_and_server
https://www.youtube.com/watch?v=dhK2pWQdGjk

**Expected behavior**
The traces should be joined with the MCP trace as a child of the general trace.

**Screenshots**
2 separate traces: mcp and general
Image
general trace
Image
mcp trace
Image

 - OS: Ubuntu
 - Version: 24.1

rajephon:
fixes #1616 

## Description

This PR updates openinference-instrumentation-mcp to be compatible with mcp>=1.8.0, which introduced a breaking change in the message format. The new version wraps the `JSONRPCRequest` inside a `SessionMessage`, requiring instrumentation logic to be updated accordingly.

## Summary of changes:

- mcp is upgraded from 1.6.0 → 1.8.0 in both pyproject.toml and test-requirements.txt.
- Wrapped messages are now unwrapped via SessionMessage.message.root to extract the original `JSONRPCRequest`.
- Compatibility logic updated in both the `__aiter__` and `send()` paths.

## Why

Without this change, context injection and extraction fail because the instrumentation attempts to unwrap the wrong type (JSONRPCMessage instead of SessionMessage). This causes runtime errors in tracing environments when used with mcp >= 1.8.0.

Relevant upstream change in MCP:
https://github.com/modelcontextprotocol/python-sdk/compare/v1.7.1...v1.8.0#diff-acedd8b44e9fb1a288cc95191fc773e99af0b75edaa11c78295dad3147cb7eeaR39

## Breaking Change Notice

This change drops compatibility with mcp<1.8.0, as SessionMessage does not exist in those versions and cannot be safely polyfilled. Supporting both formats simultaneously would require complex conditional unpacking and dual-type handling, which is out of scope for a lightweight instrumentation layer.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "mcp >= 1.6.0",
+  "mcp >= 1.8.0",
 ]
 
 [project.entry-points.opentelemetry_instrumentor]
--- a/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
@@ -91,10 +91,12 @@ class InstrumentedStreamReader(ObjectProxy):  # type: ignore
         return await self.__wrapped__.__aexit__(exc_type, exc_value, traceback)
 
     async def __aiter__(self) -> AsyncGenerator[Any, None]:
-        from mcp.types import JSONRPCMessage, JSONRPCRequest
+        from mcp.types import JSONRPCRequest
+        from mcp.shared.message import SessionMessage
 
         async for item in self.__wrapped__:
-            request = cast(JSONRPCMessage, item).root
+            session_message = cast(SessionMessage, item)
+            request = session_message.message.root
 
             if not isinstance(request, JSONRPCRequest):
                 yield item
@@ -122,9 +124,11 @@ class InstrumentedStreamWriter(ObjectProxy):  # type: ignore
         return await self.__wrapped__.__aexit__(exc_type, exc_value, traceback)
 
     async def send(self, item: Any) -> Any:
-        from mcp.types import JSONRPCMessage, JSONRPCRequest
+        from mcp.types import JSONRPCRequest
+        from mcp.shared.message import SessionMessage
 
-        request = cast(JSONRPCMessage, item).root
+        session_message = cast(SessionMessage, item)
+        request = session_message.message.root
         if not isinstance(request, JSONRPCRequest):
             return await self.__wrapped__.send(item)
         meta = None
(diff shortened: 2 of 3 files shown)


rajephon:

axiomofjoy (approved):

axiomofjoy:
Thanks for the contribution @rajephon!

CakeCrusher:
Great job @rajephon !
I was also working on this issue and I finally got around to figuring out that `JSONRPCMessage` was the problem. That discovery was hard fought. How did you arrive at the solution?

rajephon:
> Great job @rajephon ! I was also working on this issue and I finally got around to figuring out that `JSONRPCMessage` was the problem. That discovery was hard fought though. I would love to learn how went about solving this?

Thanks! Glad it helped — and happy to share how I approached it.

I was working on building a client/server system that follows the [Model Context Protocol](https://modelcontextprotocol.io/) spec, and I needed telemetry for the communication layer. That’s when I found openinference-instrumentation-mcp, which looked like exactly what I needed.

Initially, I tried setting up the instrumentation as early as possible, per the comment [here](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-mcp/tests/test_instrumenter.py#L24), but I wasn’t seeing any spans being collected. So I started digging into the source.

Looking through the code, I saw that the instrumentation uses a function wrapper internally. I put in some breakpoints and noticed that the wrapper was being bypassed because the condition if not isinstance(request, JSONRPCRequest) was evaluating to true — which caused an early return before any span could be created.

That led me to inspect the actual request objects being passed by python-sdk (the MCP client library I was using). I compared the __aiter__ and send methods and eventually saw the mismatch in how messages were typed. Once I aligned them to ensure JSONRPCRequest was correctly recognized, everything started working as expected.

Appreciate the kind words — debugging this was definitely a fun (and tricky) challenge. ☺️

CakeCrusher:
Awesome thanks for the explanation. I need to use breakpoints more often. I was instead using a logfile, clearly not nearly as efficient.

> Once I aligned them to ensure JSONRPCRequest was correctly recognized, everything started working as expected.

Do you mean "ensure SessionMessage was correctly..", it appears as if the problem was MCP replacing the JSONRPCRequest with SessionMessage

rajephon:
> Awesome thanks for the explanation. I need to use breakpoints more often. I was instead using a logfile, clearly not nearly as efficient.
> 
> > Once I aligned them to ensure JSONRPCRequest was correctly recognized, everything started working as expected.
> 
> Do you mean "ensure SessionMessage was correctly..", it appears as if the problem was MCP replacing the JSONRPCRequest with SessionMessage

Great catch — and you’re absolutely right.

To clarify, the core issue wasn’t with `JSONRPCRequest` itself, but with the MCP SDK introducing a layer of indirection by wrapping it inside a `SessionMessage`. As a result, the instrumentation logic was no longer receiving `JSONRPCRequest` instances directly, and was silently skipping over them due to the `isinstance` check failing.

The fix was to adapt the wrapper to unpack the `SessionMessage` and access the underlying .message.root, which restored compatibility with the span creation logic.

Issue #1348, opened 2025-03-12
☔ [feature request] Pydantic AI instrumentaiton


mikeldking:
Pydantic AI already has native OTel tracing so it might make sense to ingest these as is or wrap this capability in some way.

Jeffrey-Peng:
Clio could use this - they're currently doing
```
Agent(
instrumentation=True)
```
but this doesn't capture certain semantic attributes (and traces don't look as nice)

samuzaffar99:
Bump, Langfuse already does this (traces are shown properly, including tool calls):
https://langfuse.com/docs/integrations/pydantic-ai
https://langfuse.com/docs/opentelemetry/get-started

PR #1639, opened 2025-05-16
feat: pydanticai instrumentation

davidgmonical:
resolves: https://github.com/Arize-ai/openinference/issues/1348

Adds instrumentation for PydanticAI. Captures and converts `gen_ai` attributes from running an Agent with `instrument=True`.

Screenshot 2025-05-16 at 12 50 05 PM


python/instrumentation/openinference-instrumentation-pydanticai/MIGRATION.md
```diff
@@ -0,0 +1 @@
+# Migrations
```
mikeldking, line 1: I'm not sure you need this?

python/instrumentation/openinference-instrumentation-pydanticai/README.md
```diff
@@ -0,0 +1,3 @@
+# OpenInference PydanticAI Instrumentation
+
+Python auto-instrumentation library for PydanticAI.
```
mikeldking, line 3: can you flesh this out more? it gets scaffolded into pypi and is important for various reasons like search.

python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py
```diff
@@ -0,0 +1,45 @@
+import os
```
mikeldking, line 1: nit: name it according to the use case. it's easier to grep for

python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py
```diff
@@ -0,0 +1,45 @@
...
+
+from openinference.instrumentation.pydanticai import (
+    OpenInferenceSpanExporter,
+)
+
+# Get the secret key from environment variables
+os.environ["OPENAI_API_KEY"] = "YOUR_OPENAI_API_KEY"
+
+# Launch Phoenix app
+session = px.launch_app()
```
mikeldking, line 21: I would probably avoid this pattern - it encourages bad practices. Just assume phoenix is running.

python/instrumentation/openinference-instrumentation-pydanticai/pyproject.toml
```diff
@@ -0,0 +1,105 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "openinference-instrumentation-pydanticai"
```
mikeldking, line 6: I think if this is not actually an instrumentor, it might be worth naming this opeininference-pydanticai rather than instrumentation since it's not an instrumentor

mikeldking: though if this just makes the symlinking confusing I think it's okay. Defer to @axiomofjoy

python/instrumentation/openinference-instrumentation-pydanticai/pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+  "openinference-instrumentation>=0.1.27",
+  "openinference-semantic-conventions>=0.1.17",
+  "typing-extensions",
+  "wrapt",
+]
+
+[project.optional-dependencies]
+instruments = [
+  "pydantic>=2.10.0",
+  "pydantic-ai>=0.2.0",
```
mikeldking, line 40: is the pydantic necessary here?

python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py
```diff
@@ -0,0 +1,45 @@
+import os
+
+import phoenix as px
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from pydantic import BaseModel
```
mikeldking, line 8: I get a circular import on this line:

```
Traceback (most recent call last):
  File "/Users/mikeldking/work/openinference/python/instrumentation/openinference-instrumentation-pydanticai/./examples/pydantic.py", line 3, in <module>
    import phoenix as px
  File "/Users/mikeldking/work/phoenix/src/phoenix/__init__.py", line 10, in <module>
    from .session.client import Client
  File "/Users/mikeldking/work/phoenix/src/phoenix/session/client.py", line 31, in <module>
    from phoenix.db.insertion.dataset import DatasetKeys
  File "/Users/mikeldking/work/phoenix/src/phoenix/db/__init__.py", line 1, in <module>
    from .engines import get_printable_db_url
  File "/Users/mikeldking/work/phoenix/src/phoenix/db/engines.py", line 21, in <module>
    from phoenix.db.helpers import SupportedSQLDialect
  File "/Users/mikeldking/work/phoenix/src/phoenix/db/helpers.py", line 23, in <module>
    from phoenix.db import models
  File "/Users/mikeldking/work/phoenix/src/phoenix/db/models.py", line 48, in <module>
    from phoenix.db.types.annotation_configs import (
  File "/Users/mikeldking/work/phoenix/src/phoenix/db/types/annotation_configs.py", line 4, in <module>
    from pydantic import AfterValidator, Field, RootModel, model_validator
  File "/Users/mikeldking/work/openinference/python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py", line 8, in <module>
ImportError: cannot import name 'BaseModel' from partially initialized module 'pydantic' (most likely due to a circular import) (/Users/mikeldking/work/openinference/python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py)
```

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/span_exporter.py
```diff
@@ -0,0 +1,59 @@
...
+        # Export the openinference spans
+        result = self._base_exporter.export(openinference_spans)
+        return result
+
+    def shutdown(self) -> None:
+        self._base_exporter.shutdown()
+
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
+        result = self._base_exporter.force_flush(timeout_millis)
+        return bool(result)
```
mikeldking, line 59: why not just return the base force_flush?

python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py
mikeldking: can you also add a more interesting example?

mikeldking (commented):
This is awesome - though I wonder if we should re-think this category of integration. It's not an instrumentor after all.

On the js side we just do `openinference-xxxx`

Note you are missing :
- the changes to the base readme
- release please manifest changes
- symlinking support for a single interactive install

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/semantic_conventions.py
```diff
@@ -0,0 +1,648 @@
...
+    LENGTH = "length"
+    STOP = "stop"
+    TOOL_CALLS = "tool_calls"
+
+
+class GenAIToolType(Enum):
+    FUNCTION = "function"
+
+
+class GenAICommonAttributes:
```
mikeldking, line 74: is there no package that has these experimental attributes?

davidgmonical: Good call out - looks like I can get most of them from the [opentelemetry-python package](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/gen_ai_attributes.py). 

There seem to be some left out, specifically the [body fields for events](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aichoice) (the docs link an issue to migrate them which might be why). I will keep those defined in our own constants but use the imported ones wherever I can

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/__init__.py
```diff
@@ -0,0 +1,9 @@
+from openinference.instrumentation.pydanticai.span_exporter import OpenInferenceSpanExporter
+from openinference.instrumentation.pydanticai.utils import is_openinference_span
+from openinference.instrumentation.pydanticai.version import __version__
+
+__all__ = [
+    "OpenInferenceSpanExporter",
+    "is_openinference_span",
+    "__version__",
+]
```
axiomofjoy, line 9:
```suggestion
from openinference.instrumentation.pydanticai.span_exporter import OpenInferenceSpanExporter
from openinference.instrumentation.pydanticai.version import __version__

__all__ = [
    "OpenInferenceSpanExporter",
    "__version__",
]
```

This looks like a helper function we are using in tests only rather than something we want to expose to end users?

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+# Define types for span filtering
+SpanFilter = Callable[[ReadableSpan], bool]
+
+
+def is_openinference_span(span: ReadableSpan) -> bool:
+    """Check if a span is an OpenInference span."""
+    if span.attributes is None:
+        return False
+    return SpanAttributes.OPENINFERENCE_SPAN_KIND in span.attributes
```
axiomofjoy, line 15: This looks like something we only use in tests. Can I be in the test file rather than shipped with the package?

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/span_exporter.py
```diff
@@ -0,0 +1,59 @@
...
+from opentelemetry.sdk.trace import ReadableSpan
+from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
+
+from openinference.instrumentation.pydanticai.semantic_conventions import (
+    OpenInferenceAttributesExtractor,
+)
+from openinference.instrumentation.pydanticai.utils import SpanFilter, should_export_span
+
+
+class OpenInferenceSpanExporter(SpanExporter):
```
axiomofjoy, line 12: Curious why use an exporter rather than a span processor?

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/semantic_conventions.py
```diff
@@ -0,0 +1,648 @@
...
+
+class PydanticMessageRoleUser:
+    USER = "user"
+
+
+class OpenInferenceAttributesExtractor:
+    """Extracts OpenInference attributes from GenAI attributes."""
+
+    def __init__(self) -> None:
+        pass
```
axiomofjoy, line 189: Strikes me that all of the methods on this class are pure functions and the class has no state. Maybe they can just be functions?

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/semantic_conventions.py
axiomofjoy: Agree here. The more we can leverage types/ enums/ constants from OTel or Pydantic itself, the better.

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/__init__.py
davidgmonical: Yeah this was kind of copying the vercel js implementation. I assumed it's because you end up with some spans that don't map to openinference ones (something I saw as well in pydantic). I was thinking it is a convenience function for a user to block those from getting to phoenix, but probably need more context (maybe @Parker-Stafford can chime in) on what that was for in vercel

python/instrumentation/openinference-instrumentation-pydanticai/examples/pydantic.py
davidgmonical: should be fixed once I rename the script

mikeldking:
That trace looks pretty great - I wish we could figure out the tool spans!

python/instrumentation/openinference-instrumentation-pydantic-ai/README.md
```diff
@@ -0,0 +1,83 @@
...
+Install required packages.
+
+```shell
+pip install pydantic-ai arize-phoenix opentelemetry-sdk opentelemetry-exporter-otlp
+```
+
+Start Phoenix in the background as a collector. By default, it listens on `http://localhost:6006`. You can visit the app via a browser at the same address. (Phoenix does not send data over the internet. It only operates locally on your machine.)
+
+```shell
+python -m phoenix.server.main serve
```
mikeldking, line 26:
```suggestion
phoenix serve
```

python/instrumentation/openinference-instrumentation-pydantic-ai/src/openinference/instrumentation/pydantic_ai/semantic_conventions.py
```diff
@@ -0,0 +1,560 @@
+import json
+import logging
+from typing import Any, Dict, Iterator, List, Mapping, Optional, Tuple, cast
+
+from opentelemetry.semconv._incubating.attributes.gen_ai_attributes import (
```
mikeldking, line 5: Nice

python/instrumentation/openinference-instrumentation-pydantic-ai/pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+]
+dependencies = [
+  "opentelemetry-api",
+  "opentelemetry-instrumentation",
+  "opentelemetry-semantic-conventions",
```
mikeldking, line 30: I think we need a minimum version here. I get an import error:

```
ImportError: cannot import name 'GEN_AI_REQUEST_SEED' from 'opentelemetry.semconv._incubating.attributes.gen_ai_attributes' (/Users/mikeldking/anaconda3/envs/phoenix/lib/python3.12/site-packages/opentelemetry/semconv/_incubating/attributes/gen_ai_attributes.py). Did you mean: 'GEN_AI_REQUEST_MODEL'?
```

mikeldking:
I hit a road block with the import of the conventions. Will dogfood tomorrow.

python/instrumentation/openinference-instrumentation-pydanticai/pyproject.toml
davidgmonical: removed

python/instrumentation/openinference-instrumentation-pydanticai/src/openinference/instrumentation/pydanticai/span_exporter.py
davidgmonical: Switched to use span processor

mikeldking (approved):

axiomofjoy (approved):

PR #1640, opened 2025-05-18
feat(mcp): add instrumentation support for Streamable-HTTP transport and bump mcp dependency to ≥1.8.1

rajephon:
## Summary

Prior to this PR, OpenInference’s MCP instrumentation only covered STDIO and SSE transports, leaving calls over the newer Streamable-HTTP unmonitored. To enable context propagation for Streamable-HTTP, we’ve added a dedicated wrapper that correctly handles its three-tuple API. While developing tests against mcp v1.8.0, we uncovered a deadlock in the Streamable-HTTP implementation; upgrading the dependency to mcp ≥ 1.8.1 pulls in the upstream fix ([v1.8.1 release notes](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v1.8.1)) and ensures our new wrapper operates without hanging.

## Changes

1. Upgrade mcp requirement from >=1.8.0 to >=1.8.1
2. Instrumentation hooks (`__init__.py`)
  - **Streamable-HTTP client**: wrap `mcp.client.streamable_http.streamablehttp_client` with `_wrap_transport_with_callback`.
  - **Streamable-HTTP server**: wrap `mcp.server.streamable_http.StreamableHTTPServerTransport.connect` with `_wrap_plain_transport`.
  - **Refactor**: rename the original `_transport_wrapper` to `_wrap_plain_transport` (used for STDIO/SSE) and introduce `_wrap_transport_with_callback` to support the extra callback argument.
3. Tests
  - In `tests/mcpserver.py`, accept `MCP_TRANSPORT=streamable-http`.
  - In `tests/test_instrumenter.py`, add "streamable-http" to the transport parameter list and import the new client wrapper.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-mcp/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "mcp >= 1.8.0",
+  "mcp >= 1.8.1",
 ]
 
 [project.entry-points.opentelemetry_instrumentor]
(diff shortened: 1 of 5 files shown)


mikeldking (approved):
Thanks @rajephon !

Issue #1641, opened 2025-05-19
[feature request] Show reasoning_content returned by LiteLLM model calls issued by smolagents in the UI.


bruffridge:
**Is your feature request related to a problem? Please describe.**
I want to see the reasoning_content returned from LiteLLM model calls issued by smolagents in the Phoenix UI.

**Describe the solution you'd like**
![Image](https://github.com/user-attachments/assets/f946006a-34d7-4391-b6ff-a4990586a089)

PR #1642, opened 2025-05-19
fix(smolagents): adds reasoning_content

bruffridge:
resolves: #1641 
CC: @albertvillanova , @nate-mar 

If `reasoning_content` is present (it should be if a reasoning LiteLLM model is called), add it to the output message.

Screenshot 2025-05-19 at 4 39 57 PM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/examples/env.example
+++ b/python/instrumentation/openinference-instrumentation-smolagents/examples/env.example
@@ -2,6 +2,8 @@
 OPENAI_API_KEY=sk-YOUR_API_KEY
 # API Key from https://e2b.dev/docs/legacy/getting-started/api-key
 E2B_API_KEY=e2b_YOUR_API_KEY
+# API Key from https://console.anthropic.com/
+ANTHROPIC_API_KEY=YOUR_API_KEY
 
 # Phoenix listens on the default gRPC port 4317, so you don't need to change
 # exporter settings. If you prefer to export via HTTP, uncomment this:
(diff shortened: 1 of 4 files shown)


bruffridge:

axiomofjoy:
@bruffridge Can you add a code snippet to the examples directory I can run to verify that this appears as expected in the Phoenix UI? A unit test would also be helpful, I can help there if needed.

bruffridge:
I added an example. If you have trouble connecting to anthropic, I can add a different thinking model such as o3 to the example file.

axiomofjoy:
> I added an example. If you have trouble connecting to anthropic, I can add a different thinking model such as o3 to the example file.

Thanks @bruffridge! Will take a look.

bruffridge:
I added a unit test. I didn't run it though, so you should check to see if I missed anything.

mikeldking:
hey @bruffridge - we're working through how we capture reasoning blocks[ uniformly across all our instrumentation](https://github.com/Arize-ai/openinference/issues/1649) so we can leverage it for downstream use-cases like evals, datasets, and replay. We shouldn't be that far off from getting that in and pushing your PR through.

Thanks for your patience! We'll get it sorted

mikeldking:
@bruffridge I believe @axiomofjoy added reasoning as a message and you should be able to see it in the messages. I'm gonna close this PR for now but please raise another issue or PR if it's not to your satisfaction. Thanks for PR'ing - we appreciate it!

PR #1644, opened 2025-05-20
fix: propagate context at MCP server

CakeCrusher:
TLDR:
The context passed by the `InstrumentedStreamWriter.send` was being mutated by processes preceeding `InstrumentedStreamReader` resulting in disjointed MCP traces.
image
By adding instrumentation at `_handle_request` (an earlier state in the program flow) the original message composed by `InstrumentedStreamWriter.send` is preserved. This context is then propagated at the server level.
image

the context is detached by the same method it was being detached before, through `InstrumentedStreamReader`.


CakeCrusher:
Have not tested multiplexing, so it may be vulnerable to that.
(Testing this seems pretty complex which is why I have not written any tests yet)

mikeldking:
hey @CakeCrusher - thanks for working on the MCP instrumentation! I think this branch is not in good shape however due to merges from main. Can you drop the commits you added via an interactive rebase so the PR just contains your changes? Thanks!

CakeCrusher:
@mikeldking fixed

CakeCrusher:
for whatever reason this PR is not updating to the squashed commit history so im closing this in favor of #1677

Issue #1633, opened 2025-05-14
[bug] Agno instrumentation outputs an error log when instrumenting


dreverri:
**Describe the bug**
When instrumenting agno an error log is output:
```
ERROR    `boto3` not installed. Please install it via `pip install boto3`.
ERROR    `boto3` not installed. Please install it via `pip install boto3`.
```

**To Reproduce**

main.py
```
# /// script
# dependencies = [
#   "agno",
#   "openinference-instrumentation-agno",
# ]
# ///

from openinference.instrumentation.agno import AgnoInstrumentor

AgnoInstrumentor().instrument()
```

```
uv run main.py
Installed 41 packages in 31ms
ERROR    `boto3` not installed. Please install it via `pip install boto3`.
ERROR    `boto3` not installed. Please install it via `pip install boto3`.
```

**Expected behavior**
No errors

**Additional context**
The error is logged by the bedrock model in agno:
https://github.com/agno-agi/agno/blob/969c86274bc8e6d29a6466c92cba60425b4f673a/libs/agno/agno/models/aws/bedrock.py#L19

The instrumentor is importing all subclasses of the agno base Model here (which includes the bedrock model):
https://github.com/Arize-ai/openinference/blob/cb0f60be83bc7b1b3f9ec6f85c098ff492e7f033/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/__init__.py#L43

mikeldking:
I also observed this when testing. I chalked it up to Agno itself. Does the error go away if instrumentation is removed?

It might be just that all model base classes are imported?

dirkbrnd:
This will be fixed in the next iteration

mikeldking:
hey @dreverri - as dirk mentioned this should be fixed in the latest version of agno and the instrumentor. See https://github.com/Arize-ai/openinference/releases/tag/python-openinference-instrumentation-agno-v0.1.4

Gonna close this out but feel free to re-open!

PR #1647, opened 2025-05-20
feat: Updates for new Agno verison

dirkbrnd:
This updates the run wrapper logic for the new internal Agno structure

Also addresses #1633


mikeldking:
@dirkbrnd I think the test failures are valid - I'm no longer getting tracing when I run this. Can you confirm?

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/__init__.py
```diff
@@ -15,7 +15,7 @@
 )
 from openinference.instrumentation.agno.version import __version__
 
-_instruments = ("agno >= 1.4.5",)
+_instruments = ("agno >= 1.5.2",)
```
caroger, line 18: the latest Agno version on pypi as of now is 1.5.1 (released May 16, 2025).

dirkbrnd: Apologies, just release 1.5.2

dirkbrnd:
> @dirkbrnd I think the test failures are valid - I'm no longer getting tracing when I run this. Can you confirm?

This should work now with the latest 1.5.2. Can you confirm please? Sorry, I should have raised this after we released 1.5.2

mikeldking (approved):

mikeldking:
> > @dirkbrnd I think the test failures are valid - I'm no longer getting tracing when I run this. Can you confirm?
> 
> This should work now with the latest 1.5.2. Can you confirm please? Sorry, I should have raised this after we released 1.5.2

I bumped the versions in the examples as well as in the pyproject.toml and confirmed that the trace composed correctly with openai. No more boto errors. Thanks for the fix!

Screenshot 2025-05-20 at 12 40 39 PM

PR #1656, opened 2025-05-21
feat(bedrock): Multi Agent Support, capturing time metrics from metadata

satya-turing:
# Amazon Bedrock Multi-Agent Support & Time Metrics

## Overview

This MR adds two key enhancements to the Amazon Bedrock instrumentation:

1. **Multi-Agent Support**: Adds support for tracing interactions between multiple collaborating Bedrock agents
2. **Time Metrics Capture**: Implements extraction of accurate timing information from Bedrock metadata

### Multi-Agent Support

- Added capability to trace agent collaborator invocations
- Implemented hierarchical span creation for nested agent calls
- Enhanced trace collection to properly organize multi-agent interaction data

### Time Metrics Capture

- Improved span timing accuracy by extracting timestamps from Bedrock metadata
- Added start and end time extraction for all span types
- Enhanced performance metrics for better observability

## Impact

These enhancements provide:
- Better visibility into complex multi-agent workflows
- More accurate performance metrics for agent interactions
- Improved debugging capabilities for agent collaborations

## Testing

The implementation includes tests for:
- Multi-agent trace collection
- Timing metadata extraction
- Hierarchical span creation with nested agents


mikeldking (changes requested):
Hey @satya-turing - we're not fully comprehending all these abstractions and honestly not comfortable at the current moment to have this go into the main bedrock instrumentor since we will have to maintain it. @axiomofjoy and I will come up with a reasonable next step. Possibly a fork. But as it stands it's not a true instrumentation from what I can tell - it's taking signals from bedrock and emitting spans - which will have issues with context propagation. 

Will review more in depth tomorrow and come up with a strategy.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
```diff
@@ -0,0 +1,847 @@
...
+            for message in input_messages.get("messages", []):
+                role = message.get("role", "")
+                if content := message.get("content"):
+                    parsed_contents = fix_loose_json_string(content) or [content]
+                    for parsed_content in parsed_contents:
+                        message_content = content
+                        if isinstance(parsed_content, dict):
+                            if parsed_content_type := parsed_content.get("type"):
+                                message_content = parsed_content.get(parsed_content_type, "")
+                        messages.append(Message(content=message_content, role=role))
```
mikeldking, line 74: this might not accommodate multi-part content

mikeldking (approved):
@satya-turing - we went through this - to be honest I don't fully understand it and I think it lacks a fair amount of extra use-cases.

I think what I would like as follow ups are:
- more examples that are somewhat easy to run
- more testing on things other than anthropic (this seems risky)
- Less abstractions that mirror existing nomenclature like spans and traces (e.g.  why do we need a TraceNode)

Over all though I do appreciate the test coverage and the complete nature of the PR.

Issue #1477, opened 2025-04-03
[instrumentation][mcp][python] mcp python instrumentation


mikeldking:

mikeldking:
**Triage notes**

This appears fully shipped. The Python MCP instrumentor exists at `python/instrumentation/openinference-instrumentation-mcp/` (entry point `src/openinference/instrumentation/mcp/__init__.py`, with `_instrument` at line 113 and trace-context inject/extract at lines 205 and 167). Its `CHANGELOG.md` starts at 1.1.0 (2025-04-17) and is currently at 2.0.10 (2026-09-09).

Since the body is empty I cannot tell whether a specific sub-feature was intended beyond "an MCP Python instrumentor". Flagging for a maintainer to close or narrow.

Issue #1475, opened 2025-04-03
🗺 MCP Client Instrumentation


mikeldking:
Instrument both JS and Python clients to understand tool, resource, prompt

jasonmeverett:
+1

mikeldking:
Context propagation is done. Which MCP specific operations are people wanting to introspect? MCP semcov is also in the works.

CakeCrusher:
What is the state of this? I know the tool instrumentation and context propagation is complete.

PR #1675, opened 2025-05-23
feat: MCP instrumentation for `tool`, `resources`, and `prompt` clean

CakeCrusher:
- Addresses #1477 , #1475 
- Depends on #1644 

Provides instrumentation for MCP's `tools`, `resources`, and `prompts` through the highest level methods (`call_tool`, `read_resource`, `get_prompt`) on the lowest level mcp package (`mcp`).

- Supports derrivative packages like `fastmcp` that depend on `mcp`
- Resources instrumentation not thuroughly tested  due to https://github.com/Arize-ai/phoenix/issues/7687 
- A couple of TODOs in there to be addressed once theis is merged

Will be adding a complimentary tutorial on Phoenix soon

### DIFF
--- a/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
@@ -272,10 +272,27 @@ class MCPInstrumentor(BaseInstrumentor):  # type: ignore
                 result = await wrapped(*args, **kwargs)
                 # Add output attributes
                 if hasattr(result, "contents") and result.contents:
+                    for i, content in enumerate(result.contents):
+                        if content.text:
+                            span.set_attribute(
+                                f"retrieval.documents.{i}.document.content", content.text
+                            )
+                            span.set_attribute(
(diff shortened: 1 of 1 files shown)


CakeCrusher:
Have to revert type hints because of #1676

srikhi:
@CakeCrusher @mikeldking - This looks very interesting. Any particular reason why `get_tools` should not be included in this set?
Also, any chance we could get this feature in the upcoming release?

CakeCrusher:
No reason @srikhi 
Up to @mikeldking

logwiser:
Thank You @CakeCrusher for making this change.

mikeldking:
Thanks for this PR - In general generating spans inside of the MCP instrumentation does make some level of sense for tools, and possibly prompts. However some of this we will need an opt-in mechanism and also I think we are conflating the idea of a session to mean a request or a session from MCP session management perspective, which won't play nice with sessions being driven by agents or user flows. For that reason I'd like to close this PR and figure out a set of MCP instrumentation conventions that ultimately doesn't bloat the overall trace tree un-necessarily.

Thanks again for the PR. Let's discuss scope in an issue first.

Issue #1679, opened 2025-05-24
[ci] crewAI tests failing on latest


mikeldking:
3. py310-ci-crewai-latest
 4. py312-ci-crewai-latest

The following tests are failing CI - meaning some regression has happended.

PR #1682, opened 2025-05-26
fix(crewai): crewai default empty tasks

satyadevai:
- Adjusted task instrumentation to support explicit empty context lists introduced in CrewAI [0.120.0](https://github.com/crewAIInc/crewAI/releases/tag/0.120.0), [Crewai MR](https://github.com/crewAIInc/crewAI/pull/2793).
- Incorporated CrewAI fix treating all falsy context values as “no context.”
- Re-recorded tests due to changes in response behavior.
- Upgraded CrewAI from 0.95 to 0.121.

Closes https://github.com/Arize-ai/openinference/issues/1679

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-crewai/pyproject.toml
@@ -34,10 +34,10 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "crewai >= 0.121.0",
+  "crewai >= 0.119.0",
 ]
 test = [
-  "crewai == 0.121.0",
+  "crewai == 0.119.0",
   "crewai-tools == 0.45.0",
   "opentelemetry-sdk",
   "responses",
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-crewai/pyproject.toml
```diff
@@ -34,11 +34,11 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "crewai >= 0.95.0",
+  "crewai >= 0.121.0",
```
mikeldking, line 37: this seems backwards compatible change. I think maybe leaving the dependency as low as we can support makes sense here.

satya-turing: Yes, but crewAI is currently throwing exceptions in Python 3.12. This issue has already been reported in the CrewAI Tools repository: https://github.com/crewAIInc/crewAI-tools/issues/288.

Also, if we upgrade only the crewai-tools package to the latest version, it requires a minimum crewAI version of 0.119.

mikeldking: Okay sounds good. Thanks!

Issue #1648, opened 2025-05-20
[feature request] add tool definition schemas for langchain.js


mikeldking:
add tool definition schemas that are available on the python side.

PR #1700, opened 2025-05-28
feat(langchain-js): capture tool json schemas from LLM invocation parameters for llm calls with bound tools

caroger:
resolves: #1648
Before:
Screenshot 2025-05-27 at 11 13 11 PM

After:
Screenshot 2025-05-27 at 11 13 22 PM
Screenshot 2025-05-27 at 11 13 46 PM


js/packages/openinference-instrumentation-langchain/src/types.ts
```diff
@@ -36,6 +36,7 @@ export type RetrievalDocument = {
 export type LLMParameterAttributes = {
   [SemanticConventions.LLM_MODEL_NAME]?: string;
   [SemanticConventions.LLM_INVOCATION_PARAMETERS]?: string;
+  [key: string]: string | undefined;
```
caroger, line 39: pretty sure this is not the way to go about it. 

Let me know the proper way to define the Union type for 
```
const key = `${SemanticConventions.LLM_TOOLS}.${index}.${SemanticConventions.TOOL_JSON_SCHEMA}`
```

js/packages/openinference-instrumentation-langchain/src/types.ts
```diff
@@ -36,6 +36,7 @@ export type RetrievalDocument = {
...
 };
```
cephalization, line 40:
```suggestion
export type LLMParameterAttributes = {
  [SemanticConventions.LLM_MODEL_NAME]?: string;
  [SemanticConventions.LLM_INVOCATION_PARAMETERS]?: string;
  [
    key: `${typeof SemanticConventions.LLM_TOOLS}.${number}.${typeof SemanticConventions.TOOL_JSON_SCHEMA}`
  ]: string;
};
```

Typescript lets you define patterns for string literal types called "template literal types" https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

This type says that the key must start with "llm.tools." then have any number followed by "." and finally end with "tool.json_schema".

See my other comment for when you insert the key you must declare the key string as `const` (unchanging) so typescript can analyze its shape.

We can either explicitly define template literals patterns like this as we go, adding more indexed types, or we can do a loose pattern of `key: \`llm.${string}\`` To say any keyname starting with `llm.` is a valid key for LLMParameterAttributes

cephalization (approved):
Approved, pending type changes

Issue #1678, opened 2025-05-23
[bug] Agno exception due to OpenAIResponses.get_request_params() signature mismatch


rosielarose:
**Describe the bug**
I'm seeing this stack trace when using OpenAI after auto-instrumenting with the Agno instrumentor:

```
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/agno/agent/agent.py", line 1218, in run                                          
             return next(resp)                                                                                                                                                
                    ^^^^^^^^^^                                                                                                                                                
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/openinference/instrumentation/agno/_wrappers.py", line 152, in                   
         run                                                                                                                                                                  
             for run_response in response:                                                                                                                                    
                                 ^^^^^^^^                                                                                                                                     
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/agno/agent/agent.py", line 828, in _run                                          
             model_response = self.model.response(messages=run_messages.messages)                                                                                             
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                             
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/agno/models/base.py", line 187, in response                                      
             assistant_message, has_tool_calls = self._process_model_response(                                                                                                
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/agno/models/base.py", line 323, in _process_model_response                       
             response = self.invoke(messages=messages)                                                                                                                        
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                        
           File "/Users/adam/dev/GitHub/alyx_proto_ai_api/.venv/lib/python3.12/site-packages/openinference/instrumentation/agno/_wrappers.py", line 310, in                   
         run                                                                                                                                                                  
             **dict(_llm_invocation_parameters(model)),                                                                                                                       
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                        
[...]

PR #1701, opened 2025-05-28
fix: handle agnoinstrumentor failing with OpenAIResponses

kausmeows:
## What does this PR do?

- fixes: https://github.com/Arize-ai/openinference/issues/1678
For the `OpenAIResponses` model class the `get_request_params()` expects a non-optional list of Messages which was not being passed. This PR fixes that.

- To be noted that in future with Agno 2.0 major version bump, we'll revisit this to refactor some things and handle things in a more consistent way.

image


kausmeows:

mikeldking (approved):
Thanks @kausmeows !

PR #1730, opened 2025-05-30
feat: Add model name and provider fields to DSPy for better model tracking/instrumentation #1730

### LINKED ISSUE #1609: [dspy] add model_name and provider to DSPy LM calls
See screenshot

Image

GeLi2001:
resolves #1609 
Screenshot 2025-05-30 at 2 27 24 AM

## Summary by Sourcery

Include LLM provider and model name fields in DSPy instrumentation with string parsing fallbacks, update tests to validate the new attributes, and add pytest as a development dependency.

New Features:
- Add LLM_PROVIDER span attribute to capture the LLM provider for DSPy models
- Enhance _llm_model_name to extract model names from provider/model strings when model_name is unavailable

Build:
- Add pytest as a dev dependency group in pyproject.toml

Tests:
- Update tests to assert LLM_PROVIDER and LLM_MODEL_NAME attributes and adjust span output assertions


GeLi2001:
> can you add resolves in the pr? Also adding a bit more color in the PR title helps with our LLMs that use RAG.

updated

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
 def _llm_model_name(lm: "LM") -> Iterator[Tuple[str, Any]]:
     if (model_name := getattr(lm, "model_name", None)) is not None:
         yield LLM_MODEL_NAME, model_name
+    elif (model := getattr(lm, "model", None)) is not None:
+        # Extract model name from "provider/model" format or use as-is
+        model_str = str(model)
+        if "/" in model_str:
+            # Extract model from "provider/model" format (e.g., "openai/gpt-4" -> "gpt-4")
+            model_name = model_str.split("/", 1)[1]
```
mikeldking, line 1001: what if there is a trailing slash. Is this actually safe? I know this might be how litellm does things but might be worth double checking

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
...
+    # Fallback to extracting provider from model string
+    if (model := getattr(lm, "model", None)) is not None:
+        model_str = str(model)
+        if "/" in model_str:
+            # Extract provider from "provider/model" format
+            provider_name = model_str.split("/")[0].lower()
+            # Handle cases like "text-completion-openai" -> "openai"
+            if "-" in provider_name:
+                provider_name = provider_name.split("-")[-1]
+            yield LLM_PROVIDER, provider_name
```
mikeldking, line 1037: tthere's quite a bit of overlap with the model name code. Can you consolidate the parsing?

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
...
+        elif hasattr(provider, "__name__"):
+            yield LLM_PROVIDER, provider.__name__.lower()
+        else:
+            # Extract provider name from class name (e.g., OpenAIProvider -> openai)
+            provider_class_name = provider.__class__.__name__.lower()
+            if "provider" in provider_class_name and provider_class_name != "provider":
+                provider_name = provider_class_name.replace("provider", "")
+                yield LLM_PROVIDER, provider_name
+                return
+            # If it's just a generic Provider class, fall through to model string extraction
```
mikeldking, line 1026: can you add a test for this function? It's not clear that it works on all cases from just reading it. It also has a fair amount of string matching, which makes it quite hard to validate.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
 def _llm_model_name(lm: "LM") -> Iterator[Tuple[str, Any]]:
     if (model_name := getattr(lm, "model_name", None)) is not None:
         yield LLM_MODEL_NAME, model_name
+    elif (model := getattr(lm, "model", None)) is not None:
```
mikeldking, line 996: this probably needs to be tested based on the lm classes available in dspy

mikeldking:
@sroucery-ai review

python/instrumentation/openinference-instrumentation-dspy/uv.lock
axiomofjoy, line 1: Let's nix this file from version control. Since this package is a library rather than application, we want to leave our dependencies as loosely pinned as possible to allow users maximum flexibility to use the versions appropriate for their application, and the presence of a lockfile could suggest to users that they should use those exact versions (plus it becomes additional work to recompile and test on the versions described in the lockfile). You can add `uv.lock` to `.gitignore` if you want to keep your lockfile locally.

python/instrumentation/openinference-instrumentation-dspy/pyproject.toml
```diff
@@ -100,3 +100,8 @@ select = ["E", "F", "W", "I"]
 
 [tool.ruff.lint.isort]
 force-single-line = false
+
+[dependency-groups]
+dev = [
+    "pytest>=8.3.5",
+]
```
axiomofjoy: Looks like you want a quick extra to set up a dev env, but this one is missing DSPy and OTel dependencies.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
...
+        if "/" in model_str:
+            # Extract model from "provider/model" format (e.g., "openai/gpt-4" -> "gpt-4")
+            model_name = model_str.split("/", 1)[1]
+            yield LLM_MODEL_NAME, model_name
+        else:
+            # Use the model string as-is
+            yield LLM_MODEL_NAME, model_str
+
+
+def _llm_provider(lm: "LM") -> Iterator[Tuple[str, Any]]:
```
axiomofjoy, line 1022: It looks like it's currently possible for this iterator to yield the same convention more than once.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
```diff
@@ -991,6 +993,48 @@ def _output_value_and_mime_type(response: Any) -> Iterator[Tuple[str, Any]]:
...
+    # Fallback to extracting provider from model string
+    if (model := getattr(lm, "model", None)) is not None:
+        model_str = str(model)
+        if "/" in model_str:
+            # Extract provider from "provider/model" format
+            provider_name = model_str.split("/")[0].lower()
+            # Handle cases like "text-completion-openai" -> "openai"
+            if "-" in provider_name:
+                provider_name = provider_name.split("-")[-1]
+            yield LLM_PROVIDER, provider_name
```
axiomofjoy: I'm a little unclear on what the different cases here represent. There doesn't seem to be a `name` attribute on the base `Provider` class. Is that a property that is present on some but not all providers? I assume the same for `__name__`? I assume the last case is for parsing LiteLLM-style model strings if all else fails?

axiomofjoy: Agree with the comment on adding some tests to cover the various cases here.

python/instrumentation/openinference-instrumentation-dspy/pyproject.toml
axiomofjoy: Our typical pattern has been to include such dependencies in `examples/requirements.txt` in case that helps.

GeLi2001: got it, removed the temp dep

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: thanks! added return

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: yep yep added tests in at bottom of `python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py`

GeLi2001: looked into it further:
The name and __name__ attributes don't actually exist on DSPy provider instances - those were incorrect assumptions in the original code
The only working extraction method is from the provider's class name (e.g., OpenAIProvider → "openai")

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
```diff
@@ -854,8 +866,8 @@ def add(x: int, y: int) -> int:
     output_value = json.loads(output_value)
     assert isinstance(output_value, list)
     assert len(output_value) == 1
-    assert output_value[0]["next_tool_name"] == "finish"
-    assert output_value[0]["next_tool_args"] == {}
+    print(output_value[0])
```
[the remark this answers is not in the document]
GeLi2001: resolved

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: added more tests, referred to this https://dspy.ai/learn/programming/language_models/#__tabbed_1_5

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: added comprehensive tests for the _llm_provider function:
Unit tests in test_instrumentor.py that test the extraction logic with mock objects
Integration tests in test_real_lm_classes.py that test with actual DSPy LM classes

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: onsolidated the parsing logic into a shared parse_provider_and_model() function that handles all the string parsing in one place:

Both _llm_provider and _llm_model_name now use this shared function, eliminating the code duplication.

python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
GeLi2001: the parse_provider_and_model() function strips trailing slashes before parsing:
`model_str = model_str.strip().rstrip("/")  # Removes trailing slashes
`We also have a test case specifically for this: "openai/gpt-4/" → provider="openai", model="gpt-4"
So strings like "openai/gpt-4/" are safely handled.

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
```diff
@@ -1175,14 +1212,13 @@ def forward(self, question: str) -> dspy.Prediction:
...
 TOOL = OpenInferenceSpanKindValues.TOOL.value
-LLM = OpenInferenceSpanKindValues.LLM.value
 OPENINFERENCE_SPAN_KIND = SpanAttributes.OPENINFERENCE_SPAN_KIND
 INPUT_VALUE = SpanAttributes.INPUT_VALUE
 INPUT_MIME_TYPE = SpanAttributes.INPUT_MIME_TYPE
 OUTPUT_VALUE = SpanAttributes.OUTPUT_VALUE
 OUTPUT_MIME_TYPE = SpanAttributes.OUTPUT_MIME_TYPE
 LLM_INVOCATION_PARAMETERS = SpanAttributes.LLM_INVOCATION_PARAMETERS
-LLM_MODEL_NAME = SpanAttributes.LLM_MODEL_NAME
+# LLM_MODEL_NAME and LLM_PROVIDER are already imported from the module
```
axiomofjoy:
```suggestion
```

Stray Cursor comment?

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+if __name__ == "__main__":
+    print("Testing with real DSPy LM classes...\n")
+    test_real_dspy_lm_classes()
+    print("\nTesting edge cases...")
+    test_edge_cases_with_real_lm()
+    print("\nTesting provider class name extraction...")
+    test_provider_class_name_extraction()
+    print("\nAll tests completed!")
```
axiomofjoy: Let's nix the `main` function.

```suggestion
```

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
```diff
@@ -0,0 +1,129 @@
...
+
+# Set dummy API keys to avoid errors
+os.environ["OPENAI_API_KEY"] = "dummy-key"
+os.environ["ANTHROPIC_API_KEY"] = "dummy-key"
+os.environ["COHERE_API_KEY"] = "dummy-key"
+
+
+def test_real_dspy_lm_classes() -> None:
+    """Test provider and model extraction with actual DSPy LM instances."""
+    import dspy
```
axiomofjoy: Let's hoist the imports to the top.

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
```diff
@@ -1209,3 +1245,105 @@ def forward(self, question: str) -> dspy.Prediction:
...
+    lm = DummyLM(model_name="gpt-4", model="openai/gpt-3.5-turbo")
+    model_results = list(_llm_model_name(lm))
+    assert len(model_results) == 1
+    assert model_results[0] == (LLM_MODEL_NAME, "gpt-4")
+
+    # Test case where LM has both provider object and model string
+    lm = DummyLM(provider=OpenAIProvider(), model="anthropic/claude-3")
+    provider_results = list(_llm_provider(lm))
+    assert len(provider_results) == 1
+    assert provider_results[0] == (LLM_PROVIDER, "openai")
```
axiomofjoy: These tests feel a little redundant given the ones you've added in `test_real_lm_classes.py`. Safe to nix given your other tests that use the real DSPy classes?

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
```diff
@@ -0,0 +1,129 @@
...
+from openinference.instrumentation.dspy import (
+    _llm_model_name,
+    _llm_provider,
+    parse_provider_and_model,
+)
+
+# Set dummy API keys to avoid errors
+os.environ["OPENAI_API_KEY"] = "dummy-key"
+os.environ["ANTHROPIC_API_KEY"] = "dummy-key"
+os.environ["COHERE_API_KEY"] = "dummy-key"
```
axiomofjoy: Can you convert these into fixtures using `monkeypatch.setenv` in a `conftest.py` instead of directly modifying `os.environ`? (See `openai_api_key` in `test_instrumentor` for an example). The reason is that globally modifying environment variables in a test suite can interfere with other tests that depend on that environment variable, while using `monkeypatch.setenv` will scope modifications to an environment variable to a particular test.

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
```diff
@@ -0,0 +1,129 @@
...
+            "databricks/databricks-meta-llama-3-1-70b-instruct",
+            "databricks",
+            "databricks-meta-llama-3-1-70b-instruct",
+        ),
+        (
+            "together/meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo",
+            "together",
+            "meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo",
+        ),
+    ]
```
axiomofjoy: Non-blocking suggestion: You can ask Cursor to turn these into `pytest.params`, which helps you know which test case in particular is failing when reading logs in CI.

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
axiomofjoy, line 1: I think we're overusing `try`/`except` a bit in these tests. Makes it difficult to know for which tests the except clause is being hit. Probably an indication that some of the tests could be broken up into separate tests for `parse_provider_and_model` on the one hand vs. `_llm_provider` and `_llm_model_name` on the other.

axiomofjoy (approved):

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
GeLi2001: ohh yeah didn't notice about this

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_instrumentor.py
GeLi2001: agree, just nixed

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
GeLi2001: updated

python/instrumentation/openinference-instrumentation-dspy/tests/openinference/instrumentation/dspy/test_real_lm_classes.py
GeLi2001: agree, refactored

PR #1737, opened 2025-05-30
fix: filter sensitive model params

kausmeows:
## What does this PR do?

- fixes: https://github.com/agno-agi/agno/issues/3273
- if we pass sensitive params like `api_key`, `api_base` etc in the Agno `Model` config, its being tracked and shown in otel platforms.
- This PR marks such sensitive info as `REDACTED` 

#### Before:

image

#### After:

image


axiomofjoy:
@kausmeows Good to merge here?

kausmeows:
> @kausmeows Good to merge here?

Yes good to go from my side 🙌

Issue #1636, opened 2025-05-15
[bug] SmolagentsInstrumentor does not create model spans when using CodeAgent


igor-gorohovsky:
**Describe the bug**
`opentelemetry-instrumentation-smolagents` instruments the model layer by monkey-patching `Model.__call__`.
The `CodeAgent` (and any logic that reuses its `_step_stream` / `step` implementation) never calls `__call__`; it uses `model.generate(...)` or `model.generate_stream(...)` directly.
Because of this, no model spans are produced when a `CodeAgent` is executed – only the agent-level and tool-level spans appear.

**To Reproduce**

1. Create python script:
```python
from smolagents import CodeAgent, tool, OpenAIServerModel

model = OpenAIServerModel(model_id="gpt-3.5-turbo")

@tool
def get_tasks() -> dict:
    """Returns list of tasks"""
    return {"title": "some work"}
agent = CodeAgent(tools=[get_tasks], model=model)

agent.run("What tasks I have?")
```
2. Run script with opentelemetry-instrumentary
```bash
opentelemetry-instrumentary python3 script.py
```

**Expected behavior**
Every call to the underlying LLM (generate / generate_stream) should be wrapped in a span so that prompt, completion, latency, token counts, etc. are recorded.

**Versions / Environment**
- smolagents: 1.15.0
- opentelemetry-instrumentation-smolagents: 0.1.11
- opentelemetry-sdk/api: 1.33.0
- Python 3.11.12
- OS: Ubuntu
- Version: 24.04

**Additional context**
The instrumentor only wraps:
```python3
wrap_function_wrapper("smolagents", name=model_subclass.__name__ + ".__call__", ...)
```
but CodeAgent._step_stream uses:
```python3
chat_message = self.model.generate(...)
# or
output_stream = self.model.generate_stream(...)
```

**Possible fix**
- Wrap `Model.generate` and `Model.generate_stream` instead of `Model.__call__`.
(`__call__` simply delegates to generate)

victorlearned:
Any update on this? Basically makes are codeagent traces worthless

mikeldking:
Hey @victorlearned @igor-gorohovsky - sorry for the delay. So I believe if you add `openinference-instrumentation-litellm` in addition to smalagents you should get all the details. I have not tested this however so if you could give that a try and let me know.

Apologize for the delay. Will work with @Arize-ai/dev-rel to see if we can reproduce.

axiomofjoy:
[...]

PR #1744, opened 2025-06-04
fix(smolagents): instrument both __call__ and generate methods for complete model tracing

igor-gorohovsky:
#1636

## Summary by Sourcery

Extend Smolagents instrumentation to wrap and trace model ‘generate’ methods alongside ‘__call__’, storing and restoring original methods for clean uninstrumentation.

Enhancements:
- Introduce tracking and wrapping of model ‘generate’ methods for tracing.
- Restore original ‘generate’ methods during uninstrumentation to revert instrumentation.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
@@ -279,6 +279,13 @@ class TestModels:
             )
             == "get_weather"
         )
+        assert isinstance(
+            tool_call_arguments_json := attributes.pop(
+                f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_TOOL_CALLS}.0.{TOOL_CALL_FUNCTION_ARGUMENTS_JSON}"
+            ),
+            str,
+        )
+        assert json.loads(tool_call_arguments_json) == {"location": "Paris"}
         assert not attributes
 
     @pytest.mark.vcr(
--- a/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
@@ -279,13 +279,6 @@ class TestModels:
             )
             == "get_weather"
         )
-        assert isinstance(
-            tool_call_arguments_json := attributes.pop(
-                f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_TOOL_CALLS}.0.{TOOL_CALL_FUNCTION_ARGUMENTS_JSON}"
-            ),
-            str,
-        )
-        assert json.loads(tool_call_arguments_json) == {"location": "Paris"}
         assert not attributes
 
     @pytest.mark.vcr(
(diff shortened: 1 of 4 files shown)


victorlearned:
There an ETA on getting this merged? We use smolagents and currently using the LiteLLM as a stop gap but it's not as good.

axiomofjoy:
@igor-gorohovsky Thanks for the contribution. Can you follow the instructions from the bot above to sign the CLA?

https://github.com/Arize-ai/openinference/pull/1744#issuecomment-2939396897

axiomofjoy:
@igor-gorohovsky Thanks for the contribution. Correct me if I am wrong, but since we are instrumenting both the `generate` and `__call__` methods, I expect this change to produce duplicate LLM spans when a model is invoked via a direct call. Can we remove the wrapper for `__call__` and keep the wrapper for `generate` to avoid this issue?

It looks like formatters need to be run and tests also need to be updated. You can see find a guide on `tox` setup [here](https://github.com/Arize-ai/openinference/blob/main/python/DEVELOPMENT.md#introduction-to-tox).

Let us know if you need help getting this PR over the line. Unfortunately, I can't push to `main` of your fork since it is a protected branch.

igor-gorohovsky:
@axiomofjoy You're right about the duplicate spans - didn't catch that. Thanks for the tox guide link!
I'll push the changes soon.

igor-gorohovsky:
@axiomofjoy I noticed `openinference-instrumentation-smolagents` pins `smolagents==1.12.0`, which doesn't have the `generate()` method on models yet.
Should we maintain backward compatibility with 1.12.0? We could support >=1.12.0 by wrapping `__call__` for older versions (when `generate` doesn't exist) and `generate` for newer ones or we could just handle this by checking the smolagents version.
What do you think?

igor-gorohovsky:

igor-gorohovsky:
I've run `tox` and encountered only one issue with mypy related to importing `Resource` from opentelemetry-sdk (see https://github.com/open-telemetry/opentelemetry-python/issues/4615). This appears to be unrelated to my changes and is likely an environment-specific issue on my end. Could you please run the CI pipeline to verify that everything passes correctly?

axiomofjoy:
> I've run `tox` and encountered only one issue with mypy related to importing `Resource` from opentelemetry-sdk (see [open-telemetry/opentelemetry-python#4615](https://github.com/open-telemetry/opentelemetry-python/issues/4615)). This appears to be unrelated to my changes and is likely an environment-specific issue on my end. Could you please run the CI pipeline to verify that everything passes correctly?

Hey @igor-gorohovsky, looks like CI is still failing above. You should be able to add a `# type: ignore[attr-defined]` to resolve.

axiomofjoy:
> @axiomofjoy I noticed `openinference-instrumentation-smolagents` pins `smolagents==1.12.0`, which doesn't have the `generate()` method on models yet. Should we maintain backward compatibility with 1.12.0? We could support >=1.12.0 by wrapping `__call__` for older versions (when `generate` doesn't exist) and `generate` for newer ones or we could just handle this by checking the smolagents version. What do you think?

Feel free to increase the lower bound on `smolagents` to an appropriate version.

igor-gorohovsky:
@axiomofjoy  Hi! I've updated this PR based on the previous feedback. When you have a chance, could you please take another look? Thanks!

python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
axiomofjoy, line 1: Hey @igor-gorohovsky, overall the PR looks great. Looking at the tests, it seems that some of the tool call attributes have changed or disappeared. Do you have a sense of why that is?

igor-gorohovsky: Hi @axiomofjoy . Thanks for catching that!

I initially thought this was an issue with the `smolagents` library, but after investigating further, I realized the problem was actually in our result parsing logic. I've now fixed the parsing implementation and restored all the assertions in the tests. The tool call attributes should now match the expected behavior.

PR #1748, opened 2025-06-04
feat: Add semantic conventions for agent and graph attributes

generalzhou:
We want to add graph and agent attributes to semantic conventions. This PR adds the following conventions:

| name | type | example | description |
| -- | -- | -- | -- |
| `agent.name`                           | String                      | `researcher`                                                               | The name of the agent that this span represents.                                                                                     |
| `graph.node.id`                        | String                      | `search_api_0`                                                             | The id of the node in the execution graph. This along with graph.node.parent_id are used to visualize the execution graph.           |
| `graph.node.name`                      | String                      | `Search API`                                                               | The name of the node in the execution graph. Use this to present a human readable name for the node. Optional                        |
| `graph.node.parent_id`                 | String                      | `router_0`                                                                 | This references the id of the parent node. Leaving this unset or set as empty string implies that the current span is the root node. |

### DIFF
--- a/js/.changeset/tough-badgers-occur.md
+++ b/js/.changeset/tough-badgers-occur.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/openinference-semantic-conventions": minor
+---
+
+Adding agent and graph semantic conventions


generalzhou:

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -121,6 +123,18 @@ export const AudioAttributesPostfixes = {
   transcript: "transcript",
 } as const;
 
+export const AgentPostfixes = {
+  name: "name",
+  step: "step",
+  next_agent: "next_agent",
```
mikeldking: is this a hand-off? as in you would have

```json
{
     "name": "customer_support"
     "step": 1
     "next_agent": "billing_agent"
}
```
if there's no `next` does that mean it terminates? Also if the `next` is not an agent, then this might be better to be just `next`?

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -121,6 +123,18 @@ export const AudioAttributesPostfixes = {
...
+} as const;
+
+export const GraphPostfixes = {
+  node_name: "node.name",
+  node_step: "node.step",
+  node_next_node: "node.next_node",
+} as const;
```
mikeldking: this might be me, but just adding "graph" semcov seems like the right idea. It might be difficult to try to mix the two.
```suggestion
export const GraphPostfixes = {
  node_name: "node.name",
  node_step: "node.step",
  node_next: "node.next",
} as const;
```

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -452,6 +466,37 @@ export const AUDIO_MIME_TYPE =
...
+/**
+ * The next agent to call
+ */
+export const AGENT_NEXT_AGENT =
+  `${SemanticAttributePrefixes.agent}.${AgentPostfixes.next_agent}` as const;
+/**
+ * The name of the node in the execution graph. Nodes are like agents but are for general purpose graphs.
+ */
+export const GRAPH_NODE_NAME =
+  `${SemanticAttributePrefixes.graph}.${GraphPostfixes.node_name}` as const;
```
mikeldking: what do you think of leaving a comment here that these need to be somewhat unique, otherwise the graph is broken

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
```diff
@@ -452,6 +466,37 @@ export const AUDIO_MIME_TYPE =
...
+/**
+ * The name of the agent.
+ */
+export const AGENT_NAME =
+  `${SemanticAttributePrefixes.agent}.${AgentPostfixes.name}` as const;
+/**
+ * The step number of the current agent in the execution graph.
+ */
+export const AGENT_STEP =
+  `${SemanticAttributePrefixes.agent}.${AgentPostfixes.step}` as const;
```
mikeldking: What number does step start at?

generalzhou: For langgraph it's 1, I think for crew it's a list of tasks so technically 0 indexed. It doesn't really matter for arize since we don't look at the absolute step number but rather the progression of steps. So I hesitate to put down guidance in case this makes instrumentation clunky because someone has to translate 0 to 1 indexed or vice versa

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
generalzhou: yep 100%, something like 

```
Nodes that perform the same function should have the same name. Avoid adding any high cardinality information such as identifiers in the name
```

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
generalzhou: correct, empty string/not set means leaf node. `next` has to be an agent, as we join on `agent.next_agent == agent.name` to get additional properties. 

Separate question, how do you recommend we handle multiple next_agents? Either:
Make `next_agent` comma delimited
`next_agent = "researcher,summary_agent"`

Or add a new semantic conv
`next_agents = ["researcher", "summary_agent"]`

js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts
axiomofjoy: Not sure I understand the difference between the graph and agent semconv. I assume the graph is intended to capture the set of nodes and edges, regardless of whether those edges were actually traversed in a particular trajectory. Are agent conventions meant to capture actual trajectories?

PR #1781, opened 2025-06-13
feat: add context manager for capturing openinference spans

anuraaga:
Working with @codefromthecrypt on providing user feedback for gen AI operations, we found that we need a way to get the span ID of a previous LLM operation to be able to apply feedback to it through a downstream feedback UI. This adds a context manager that can be used to capture all openinference spans within its scope - hopefully it makes sense as a useful feature.

### DIFF
--- a/python/openinference-instrumentation/pyproject.toml
+++ b/python/openinference-instrumentation/pyproject.toml
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "OpenInference instrumentation utilities"
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.8, <3.14"
+requires-python = ">=3.9, <3.14"
 authors = [
   { name = "OpenInference Authors", email = "oss@arize.com" },
 ]
--- a/python/openinference-instrumentation/src/openinference/instrumentation/capture.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/capture.py
@@ -1,0 +1,63 @@
+from contextvars import ContextVar, Token
+from types import TracebackType
+from typing import Optional, Type
+
+from opentelemetry.trace import SpanContext
+
+_current_capture_span_context = ContextVar["capture_span_context"]("current_capture_span_context")
+
+
+def _capture_span_context(span_context: SpanContext) -> None:
+    capture = _current_capture_span_context.get(None)
+    if capture is not None:
+        capture._contexts.append(span_context)
+
+
+class capture_span_context:
+    _contexts: list[SpanContext]
+    _token: Token["capture_span_context"]
+
+    """
+    Context manager for capturing OpenInference span context.
+
+    Examples:
+        with capture_span_context() as capture:
+            response = openai_client.chat.completions.create(...)
+            span_context = capture.get_last_span_context()
+            if span_context:
+                phoenix_client.annotations.add_span_annotation(
+                    span_id=span_context.span_id,
+                    annotation_name="feedback",
+                    ...
+                )
+    """
+
+    def __init__(self) -> None:
+        self._contexts = []
+
+    def __enter__(self) -> "capture_span_context":
+        self._token = _current_capture_span_context.set(self)
+        return self
+
+    def __exit__(
+        self,
+        _exc_type: Optional[Type[BaseException]],
+        _exc_value: Optional[BaseException],
+        _traceback: Optional[TracebackType],
+    ) -> None:
+        _current_capture_span_context.reset(self._token)
+        self._contexts.clear()
+
+    def get_last_span_context(self) -> Optional[SpanContext]:
+        """
+        Returns the last captured span context, or None if no spans were captured.
+        """
+        if self._contexts:
+            return self._contexts[-1]
+        return None
+
+    def get_span_contexts(self) -> list[SpanContext]:
+        """
+        Returns a list of all captured span contexts.
+        """
+        return self._contexts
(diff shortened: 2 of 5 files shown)


python/openinference-instrumentation/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "OpenInference instrumentation utilities"
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.8, <3.14"
+requires-python = ">=3.9, <3.14"
```
anuraaga, line 11: FWICT, the project already dropped support for 3.8 and this was probably missed. I wasn't able to resolve the project without this change due to the dependency on `semconv` which already has been bumped

axiomofjoy:
Thanks @anuraaga, we'll take a look!

python/openinference-instrumentation/src/openinference/instrumentation/capture.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __exit__(
+        self,
+        _exc_type: Optional[Type[BaseException]],
+        _exc_value: Optional[BaseException],
+        _traceback: Optional[TracebackType],
+    ) -> None:
+        if self._token:
+            _current_capture_span_context.reset(self._token)
+        self._contexts.clear()
```
RogerHYang: reset may [fail](https://github.com/open-telemetry/opentelemetry-python/issues/2606) 
```suggestion
        self._contexts.clear()
        if self._token:
            try:
                _current_capture_span_context.reset(self._token)
            except Exception:
                pass
```

python/openinference-instrumentation/src/openinference/instrumentation/capture.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_last_span_context(self) -> Optional[SpanContext]:
+        """
+        Returns the last captured span context, or None if no spans were captured.
+        """
+        return self._contexts[-1] if self._contexts else None
+
+    def get_span_contexts(self) -> Iterable[SpanContext]:
```
codefromthecrypt: right I can see how in the same block, there could be many calls to openai. It could be a mistake to have had many, but at least this warns that it is possible since we don't control the closure.

python/openinference-instrumentation/src/openinference/instrumentation/capture.py
```diff
@@ -0,0 +1,63 @@
...
+    if capture is not None:
+        capture._contexts.append(span_context)
+
+
+class capture_span_context:
+    """
+    Context manager for capturing OpenInference span context.
+
+    Examples:
+        with capture_span_context() as capture:
```
codefromthecrypt: thanks for putting an example here! I'm not sure if it is normal to have the code that adds the annotation in the same file or not. otherwise, it could be like.. 

```python
# Assume this function returns a span ID for user feedback annotation
annotation_span_id = None

# Wrap your OpenAI request in a capture block. It doesn't need to include response parsing.
with capture_span_context() as capture:
    response = openai_client.chat.completions.create(...)
    span_context = capture.get_last_span_context()
    if span_context is not None:
        annotation_span_id = format_span_id(span_context.span_id)
```

I might even add a helper for `capture.get_last_span_id()` as I'm not sure the use cases we have for context except doing the conversion. I see you are following this guide, but same could apply to langchain also. If we had a helper to skip the grunt work, the caller's code would be less and then also less imports in potentially weird places. My 2p https://arize.com/docs/phoenix/tracing/how-to-tracing/feedback-and-annotations/capture-feedback

codefromthecrypt (approved):
I made a style suggestion though I know where you are going with this. Mainly to hint at what could be improved in general, not just here specifically. Thanks, @anuraaga!

python/openinference-instrumentation/src/openinference/instrumentation/capture.py
codefromthecrypt: one reason to not do span ID could be that other systems may not be able to look up data given only a span ID. For example, often indexes are partitioned by trace ID. So it might be special-cased to phoenix. I think that's a judgement call, basically, but it would mildly help in pheonix as the user doesn't need to know about otel types like span_context and just use a string.

anuraaga: Thanks - yeah for phoenix users getting the span ID should be the important use case, and while if it was just as simple as `.span_id`, I realized the string conversion makes it quite awkward. I changed `get_last_span_context` -> `get_last_span_id`, and left `get_span_contexts` which should still not be too bad for any other use case

codefromthecrypt: Excellent

codefromthecrypt (approved):

python/openinference-instrumentation/src/openinference/instrumentation/capture.py
axiomofjoy, line 1: Can we rename this file to `_capture.py` to make it clear that the import should be `from openinference.instrumentation import capture_span_contexts`?

axiomofjoy (approved):

anuraaga (commented):
Thanks - applied the cleanups

RogerHYang:
This is released in openinference-instrumentation v0.1.34. Thank you for your contribution!

codefromthecrypt:
Thanks so much! I can now use this in my pycon Singapore workshop Friday. Good job @anuraaga and excellent support, team!

PR #1782, opened 2025-06-13
fix: Enable sampling when using OpenInference tracer

anticorrelator:
Resolves https://github.com/Arize-ai/phoenix/issues/7914

`sampler.should_sample` accepts attributes passed in at span-creation time, however this process is hijacked by the OpenInference wrapper in order defer attribute-setting in order to enable masking and attribute prioritization.

## Summary by Sourcery

Fix sampling for the OpenInference tracer by passing all relevant attributes to the underlying tracer at span creation instead of deferring them.

Bug Fixes:
- Ensure sampler.should_sample receives user, span-kind, and context attributes by passing them when initiating spans.
- Restore correct sampling behavior when using the OpenInference tracer.

Enhancements:
- Merge user, span-kind, and context attributes into a single dict for start_span.
- Streamline OpenInferenceSpan attribute application after the underlying span is created.

### DIFF
--- a/python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
@@ -152,25 +152,24 @@ class OITracer(wrapt.ObjectProxy):  # type: ignore[misc]
         *,
         openinference_span_kind: Optional["OpenInferenceSpanKind"] = None,
     ) -> OpenInferenceSpan:
-        user_attributes = dict(attributes) if attributes else {}
-        span_kind_attributes = (
-            get_span_kind_attributes(openinference_span_kind)
-            if openinference_span_kind is not None
-            else {}
-        )
-        context_attributes = dict(get_attributes_from_context())
-
         otel_span: Span
         if get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             otel_span = INVALID_SPAN
         else:
             # Apply masking to attributes before passing to sampler to ensure
             # samplers don't see sensitive data that should be masked
+            user_attributes = dict(attributes) if attributes else {}
+            span_kind_attributes = (
+                get_span_kind_attributes(openinference_span_kind)
+                if openinference_span_kind is not None
+                else {}
+            )
+            context_attributes = dict(get_attributes_from_context())
             combined_attributes = self._get_masked_attributes_for_sampling(
                 {
+                    **context_attributes,
                     **user_attributes,
                     **span_kind_attributes,
-                    **context_attributes,
                 }
             )
 
--- a/python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
@@ -164,11 +164,15 @@ class OITracer(wrapt.ObjectProxy):  # type: ignore[misc]
         if get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             otel_span = INVALID_SPAN
         else:
-            combined_attributes = {
-                **user_attributes,
-                **span_kind_attributes,
-                **context_attributes,
-            }
+            # Apply masking to attributes before passing to sampler to ensure
+            # samplers don't see sensitive data that should be masked
+            combined_attributes = self._get_masked_attributes_for_sampling(
+                {
+                    **user_attributes,
+                    **span_kind_attributes,
+                    **context_attributes,
+                }
+            )
 
             tracer = cast(Tracer, self.__wrapped__)
             otel_span = tracer.__class__.start_span(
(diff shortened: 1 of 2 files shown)


python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
```diff
@@ -152,30 +152,69 @@ def start_span(
...
-            openinference_span.set_attributes(get_span_kind_attributes(openinference_span_kind))
-        openinference_span.set_attributes(dict(get_attributes_from_context()))
+
+        # Use OpenInferenceSpan wrapper's attribute handling
+        if user_attributes:
+            openinference_span.set_attributes(user_attributes)
+        if span_kind_attributes:
+            openinference_span.set_attributes(span_kind_attributes)
+        if context_attributes:
+            openinference_span.set_attributes(context_attributes)
```
RogerHYang, line 197: Is it necessary to set the attributes a second time?
```suggestion
```

python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
```diff
@@ -152,30 +152,69 @@ def start_span(
...
         if get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             otel_span = INVALID_SPAN
         else:
+            # Apply masking to attributes before passing to sampler to ensure
+            # samplers don't see sensitive data that should be masked
+            combined_attributes = self._get_masked_attributes_for_sampling(
+                {
+                    **user_attributes,
+                    **span_kind_attributes,
+                    **context_attributes,
```
RogerHYang: context attributes probably should go first otherwise users won't be able to override it
```suggestion
                    **context_attributes,
                    **user_attributes,
                    **span_kind_attributes,
```

RogerHYang (approved):
Should we add some tests or do you think that's not necessary?

python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
anticorrelator: I'm not 100% positive because a little bit of this depends on Otel behavior but *I think yes* because we do some special handling to make sure important attributes are prioritized via the `set_attributes` method

RogerHYang: do we have a test to show that the actual precedence is indeed what we want?

python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
```diff
@@ -152,28 +152,47 @@ def start_span(
         *,
         openinference_span_kind: Optional["OpenInferenceSpanKind"] = None,
     ) -> OpenInferenceSpan:
+        user_attributes = dict(attributes) if attributes else {}
+        span_kind_attributes = (
+            get_span_kind_attributes(openinference_span_kind)
+            if openinference_span_kind is not None
+            else {}
+        )
+        context_attributes = dict(get_attributes_from_context())
```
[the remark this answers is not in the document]
anticorrelator: Not addressing as these attributes were set regardless of the suppression state prior to this PR (implementing this causes downstream test failures)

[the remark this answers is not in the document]
RogerHYang: Are the downstream tests that are failing actually necessary and correct?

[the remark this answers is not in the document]
RogerHYang: On its face, having to set attributes on an INVALID_SPAN does not appear necessary or correct.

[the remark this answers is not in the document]
RogerHYang: For what it's worth, the set_attribute method of INVALID_SPAN is a [no op](https://github.com/open-telemetry/opentelemetry-python/blob/6ed676a1b2d99dacf3b3597340b8ea6baef7db3c/opentelemetry-api/src/opentelemetry/trace/span.py#L531-L537) function.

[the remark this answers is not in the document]
RogerHYang: Yes, please file a follow-up

python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
RogerHYang: So it looks like the only purpose of this code block is to set attributes on the INVALID_SPAN. For normal spans this is setting the same attributes twice.

RogerHYang: OK, that assertion was not 100% correct because the overridden `set_attributes` method actually buffers and segregates the attributes. Yes, at the end all the attributes are set twice, but part of the intent of the `set_attributes` method is attribute sequestration.

PR #1794, opened 2025-06-20
feat(crewai): capture graph.node.id and graph.node.parent_id semantics

generalzhou:
- Log new `graph.node.id` and `graph.node.parent_id` attributes. 
- The `graph.node.id` corresponds to the agent name
- The `graph.node.parent_id` corresponds to the previous agent's name

## Summary by Sourcery

Log new agent attributes in CrewAI spans and update tests and dependencies accordingly

New Features:
- Log agent role as graph.node.id span attribute
- Log parent agent role as graph.node.parent_id span attribute

Build:
- Bump openinference-semantic-conventions dependency to >=0.1.21

Tests:
- Add assertions in CrewAI instrumentation tests for graph.node.id and graph.node.parent_id attributes


python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
```diff
@@ -118,12 +118,26 @@ def __call__(
...
             crew = agent.crew if agent else None
             task = instance
 
-            if crew:
+            if crew and agent:
                 span.set_attribute("crew_key", crew.key)
                 span.set_attribute("crew_id", str(crew.id))
+
+                # Find the current agent, the previous agent is the parent node
+                for i, a in enumerate(crew.agents):
```
mikeldking:
```suggestion
            if crew:
                span.set_attribute("crew_key", crew.key)
                span.set_attribute("crew_id", str(crew.id))

            if crew and agent:
                # Find the current agent, the previous agent is the parent node
                for i, a in enumerate(crew.agents):
```
Isn't this what you need?

generalzhou: Sorry just getting back into this. This one is a little bit weird because of the above line:
```
crew = agent.crew if agent else None
```
So if `crew` is defined then `agent` is defined as well. I added `and agent` for type reasons

generalzhou: I can re-order the code a little bit which I think will make this flow a little more obvious

caroger (approved):

PR #1803, opened 2025-06-26
fix: agno instrumentation arun

manuhortet:
Small fix for this arun method. It was raising for agents with name=None


manuhortet:

PR #1809, opened 2025-06-27
feat: Prioritize subclasses method wrapping to prevent multiple spans

notTyche:
This commit introduces a sorting mechanism for `agno` model subclasses based on their Method Resolution Order (MRO) length. This ensures that more specific subclasses are processed and wrapped before their superclasses.

Previously, if a base class's method was wrapped before a method of a subclass that inherits it, `wrap_function_wrapper` could apply the wrapper a second time to the subclass's method. This led to incorrect behavior, such as duplicated OpenInference spans or metrics, as the instrumentation was applied multiple times to the same underlying operation.

By sorting from the most specific class to the most general, we guarantee that each method is wrapped only once, at the level of the most-derived class in its inheritance hierarchy. This resolves the issue of duplicated instrumentation and ensures accurate tracing and metric collection.

## Summary by Sourcery

Sort agno model subclasses by their method resolution order length so that more specific subclasses are wrapped before their superclasses, preventing duplicate instrumentation.

Bug Fixes:
- Prevent duplicate OpenInference spans and metrics by ensuring each method is wrapped only once at the most-derived class level.

Enhancements:
- Prioritize subclass method wrapping by sorting model subclasses by MRO length in find_model_subclasses().


notTyche:

notTyche:
recheck

mikeldking:
thanks @notTyche ! Looks like CI is failing but this looks good to me!

notTyche:
@mikeldking I think i have solved the problem on CI

RogerHYang (approved):

Issue #1808, opened 2025-06-26
[feature request] handle tool output in response API better


zyxue:
**Is your feature request related to a problem? Please describe.**

I'm using openai-agents sdk with `openinference-instrumentation-openai-agents`
for a tool that just returns text, but the agents sdk would wrap it inside `{"type": .., "text": ..., "annotation": ..}`, which makes it hard to read on phoenix. I assume it's due to the response API as it matches the structure from response example on https://platform.openai.com/docs/api-reference/responses/create.

I'm already using `agents.OpenAIChatCompletionsModel` in my code, but it looks the agents SDK converts the chat completion format to response format somehow internally.

![Image](https://github.com/user-attachments/assets/6ea7c382-ce35-46a6-8531-81b99b912314)

**Describe the solution you'd like**
detect if it's response api output and only show the text part on phoenix 

**Describe alternatives you've considered**
NA

**Additional context**

discussion on https://arize-ai.slack.com/archives/C04R3GXC8HK/p1750959843027929

mikeldking:
thanks @zyxue - I think for now we can figure out why the output is not properly JSON parsable. Will get @satyadevai to check the serialization.

satyadevai:
@zyxue can you provide me sameple snippiet of code? I will reproduce this issue at my side.

zyxue:
It turns out this issue is specific to MCP servers. Here is a small reproducible example.

The MCP server:
```python
"""A simple MCP server."""

from mcp.server.fastmcp import FastMCP  # pylint: disable=no-name-in-module

mcp_server = FastMCP(
    "Demo",
    host="0.0.0.0",
    port=8000,
)


@mcp_server.tool()
async def add(a: int, b: int) -> int:
    """Adds two numbers together."""
    return a + b


if __name__ == "__main__":
    mcp_server.run(transport="streamable-http")
```

The tracing code:
```python
import asyncio
import os
from typing import Any

from agents import (
    Agent,
    OpenAIChatCompletionsModel,
    RunContextWrapper,
    Runner,
    function_tool,
    gen_trace_id,
    trace,
)
from agents.mcp import server
from openai import AsyncAzureOpenAI
from openinference.instrumentation.openai_agents import OpenAIAgentsInstrumentor
from phoenix.otel import register

BASE_URL = os.environ["OPENAI_API_BASE"]
API_KEY = os.environ["OPENAI_API_KEY"]
MODEL_NAME = "gpt-4o-mini"

[...]

PR #1821, opened 2025-07-01
fix(openai-agents): Added Serializer for tool response

satyadevai:
## Summary by Sourcery

Add JSON-based tool response serializer and integrate it into message content and span data extraction to ensure correct handling of wrapped tool outputs.

Enhancements:
- Introduce _get_tool_response helper to parse and extract text from JSON-wrapped tool responses
- Apply tool response serializer in chat message content attributes
- Apply tool response serializer in function span output attributes
- Surpressed mypy warnings

Closes https://github.com/Arize-ai/openinference/issues/1808

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
@@ -389,9 +389,10 @@ def _get_attributes_from_chat_completions_message_dicts(
 
 def _get_tool_response(content: str) -> Any:
     try:
-        tool_output = json.loads(content)
-        if tool_output.get("type") == "text":
-            return tool_output.get("text", content)
+        if isinstance(content, str) and content[0] == "{" and content[-1] == "}":
+            tool_output = json.loads(content)
+            if isinstance(tool_output, dict) and tool_output.get("type") == "text":
+                return tool_output.get("text", content)
     except json.JSONDecodeError:
         return content
     return content
--- a/python/instrumentation/openinference-instrumentation-openai-agents/tests/test_span_attribute_helpers.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/tests/test_span_attribute_helpers.py
@@ -197,7 +197,7 @@ from openinference.instrumentation.openai_agents._processor import (
         ),
         pytest.param(
             [
-                ResponseFunctionWebSearchParam(  # type: ignore
+                ResponseFunctionWebSearchParam(
                     type="web_search_call",
                     id="web-123",
                     status="searching",


python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
```diff
@@ -386,12 +387,23 @@ def _get_attributes_from_chat_completions_message_dicts(
         msg_idx += 1
 
 
+def _get_tool_response(content: str) -> Any:
+    try:
+        tool_output = json.loads(content)
+        if tool_output.get("type") == "text":
+            return tool_output.get("text", content)
```
mikeldking: I think the problem is not just the text but the MIME type and all the newlines. Does this solve this problem? Because technically I feel the response is the entire payload. We could just return the JSON payload?

satyadevai: If I am not added json parser, then the response will be dict, added screenshot for your reference.
image

The tool response from the MCP server is a dictionary that includes additional context. However, when using the tools directly, they return only the raw response(text). So, we check the response type—if it's text, we extract the text content from the mcp tool response object accordingly.

RogerHYang:
One consideration for this issue is that the instrumentor is actually working as it should: it's accurately reflecting the output of the tool, which in this case is in the standardized [format](https://github.com/modelcontextprotocol/python-sdk/blob/6f43d1f17022d9d45b51955304e39ad43ca48014/src/mcp/types.py#L658) of the MCP protocol. Without a reliable way to identify when we're dealing with an MCP tool specifically, this PR might inadvertently modify the output of other tools that shouldn't be affected.

python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
RogerHYang: @mikeldking The UI does the json interpretation. Below you can see the before and after effect of the current change.

https://github.com/user-attachments/assets/251e27e9-6c1b-4526-afef-4983d07f6061

mikeldking:
> One consideration for this issue is that the instrumentor is actually working as it should: it's accurately reflecting the output of the tool, which in this case is in the standardized [format](https://github.com/modelcontextprotocol/python-sdk/blob/6f43d1f17022d9d45b51955304e39ad43ca48014/src/mcp/types.py#L658) of the MCP protocol. Without a reliable way to identify when we're dealing with an MCP tool specifically, this PR might inadvertently modify the output of other tools that shouldn't be affected.

I think your right. I thought it was a serialization issue but it's really JSON over text.

Maybe we leave it as is

satyadevai:
@mikeldking I am closing this MR and issue for now. I will update the ticket.

PR #1841, opened 2025-07-07
fix(openinference-vercel): Bump package ranges

cephalization:
Our vercel instrumentation sometimes would work with next js projects or vercel/otel packages for some project setups but would mysteriously not work for others.

It turns out that some package managers would not be able to resolve the same version of opentelemetry/api in this package that the vercel ai package hard pins, meaning that the ai sdk would sometimes send traces into the ether, using the wrong global tracer provider.

Opening up the version range to the same that ai sdk pins, seems to resolve to a single API package instance, ensuring that there is only one tracer provider registered globally.

### DIFF
--- a/js/packages/openinference-vercel/package.json
+++ b/js/packages/openinference-vercel/package.json
@@ -63,6 +63,6 @@
     "@types/jest": "^29.5.12",
     "jest": "^29.7.0",
     "@opentelemetry/sdk-trace-base": ">=1.19.0 <2.0.0",
-    "@opentelemetry/api": ">=1.0.0 <1.9.0"
+    "@opentelemetry/api": ">=1.7.0 <2.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


js/packages/openinference-vercel/package.json
```diff
@@ -57,12 +57,12 @@
...
   "peerDependencies": {
-    "@opentelemetry/api": ">=1.0.0 <1.9.0"
+    "@opentelemetry/api": ">=1.7.0 <2.0.0"
   },
   "devDependencies": {
     "@types/jest": "^29.5.12",
     "jest": "^29.7.0",
-    "@opentelemetry/sdk-trace-base": "^1.30.1",
+    "@opentelemetry/sdk-trace-base": ">=1.19.0 <2.0.0",
     "@opentelemetry/api": ">=1.0.0 <1.9.0"
```
cephalization:
```suggestion
    "@opentelemetry/api": ">=1.7.0 <2.0.0"
```

PR #1845, opened 2025-07-08
feat(smolagents): Improve Display Of Handled Errors

### LINKED ISSUE #1198: improve display of handled errors in smolagents

ialisaleh:
This PR improves observability of handled (non-fatal) errors in multi-step `smolagents` execution by adding a minimal structured event (`agent.step_recovery`) to the span when an `AgentError` occurs in `_StepWrapper`. Instead of recording the `exception` or marking the span as failed, it attaches the error `type` & `message` using `step_log.error.dict()`, enabling Arize & Phoenix to display soft failures clearly without triggering red-flash-bad-error states. This ensures better trace visibility for recoverable agent behavior while maintaining clean, successful span statuses.

Closes #1198

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
@@ -165,10 +165,7 @@ class _StepWrapper:
             step_log = args[0]  # ActionStep
             span.set_attribute(OUTPUT_VALUE, step_log.observations)
             if step_log.error is not None:
-                span.add_event(
-                    name="agent.step_recovery",
-                    attributes=step_log.error.dict()
-                )
+                span.add_event(name="agent.step_recovery", attributes=step_log.error.dict())
             span.set_status(trace_api.StatusCode.OK)
         return result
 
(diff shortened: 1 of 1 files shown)


python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -16,6 +16,10 @@
     ToolAttributes,
     ToolCallAttributes,
 )
+from smolagents.utils import (  # type: ignore[import-untyped]
+    AgentToolCallError,
+    AgentToolExecutionError,
+)
```
axiomofjoy: Let's make these dynamic imports inside of `__call__` to avoid making `smolagents` a dependency of our instrumentation library.

axiomofjoy (approved):

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @axiomofjoy - I've replaced imports with string comparison to avoid adding `smolagents` as a hard dependency. Thanks!

PR #1851, opened 2025-07-10
feat: JS Bedrock Instrumentation for Invoke Model

Aemrox:
Largely complete instrumentation for Bedrock JS SDK.

I know there are still some type issues in the linter - will continue to work through those over the weekend, but wanted to get this up asap.

Covers the InvokeModel API and the ConverseAPI - but does not touch the knowledge base work, or advanced streaming features. 

I have some open questions around the proper implementation of conventions: 
- Stop_reason
    - I saw this as something I could return, but not in the conventions. Should I remove this?
- input.value and output.value
    - I'm stingifying the input and output json and placing it here whole, however, in the python instrumentation we are extracting the text where possible. What is the best practice? 
 - Just want to double check that tool_use is a valid message content type

Otherwise here are some screenshots of the spans in phoenix
Screenshot 2025-07-18 at 5 32 15 PM
Screenshot 2025-07-18 at 5 32 06 PM
Screenshot 2025-07-18 at 5 31 59 PM
Screenshot 2025-07-18 at 5 31 53 PM

Screenshot 2025-07-18 at 5 31 35 PM
Screenshot 2025-07-18 at 5 32 20 PM


js/packages/openinference-instrumentation-bedrock/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "files": [
+    "dist",
+    "src"
+  ],
+  "author": "oss-devs@arize.com",
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "@aws-sdk/client-bedrock-runtime": "^3.693.0",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.46.0",
+    "@opentelemetry/instrumentation": "^0.46.0",
```
Parker-Stafford: you've got this in deps, don't need to add it in dev deps also

js/packages/openinference-instrumentation-bedrock/scripts/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### What It Validates
+
+- ✅ Instrumentation is properly applied to the Bedrock client
+- ✅ Spans are created for InvokeModel API calls
+- ✅ Request attributes are correctly extracted
+- ✅ Response attributes are correctly extracted
+- ✅ Tool calling scenarios work properly
+- ✅ Multi-modal content is processed correctly
+- ✅ Traces are exported to Phoenix (when configured)
```
Parker-Stafford: nit: can probably pare some of this readme down, although it is helpful to have the scripts and testing but just these sections that describe what it does and output are probably overkill

js/packages/openinference-instrumentation-bedrock/scripts/README.md
```diff
@@ -0,0 +1,105 @@
...
+4. **Phoenix connection**: Check your Phoenix endpoint and API key configuration
+
+#### Local Development
+
+When working on the instrumentation locally, this script helps verify:
+
+- Changes are working correctly
+- Local package linking is functioning
+- All test scenarios pass in real-world conditions
+- Traces appear correctly in Phoenix for manual verification
```
Parker-Stafford: same here

js/packages/openinference-instrumentation-bedrock/scripts/clear-recordings.js
```diff
@@ -0,0 +1,439 @@
...
+  } else {
+    // No specific options, use interactive mode
+    interactiveMode();
+  }
+}
+
+/**
+ * Main execution
+ */
+function main() {
```
Parker-Stafford: i think this could be helpful for managing the nock recordings but i would maybe just question whether its fully necessary, if this is just like deleting a directory of files in a directory we probably don't need 400 lines of code to do it

js/packages/openinference-instrumentation-bedrock/scripts/record-helper.js
```diff
@@ -0,0 +1,309 @@
...
+      console.log(
+        `${colors.yellow}⚠️  Recording mode requested but credentials missing. Tests will likely fail.${colors.reset}\n`,
+      );
+      process.exit(1);
+    }
+
+    process.exit(0);
+  } catch (error) {
+    console.error(`${colors.red}❌ Error: ${error.message}${colors.reset}`);
+    process.exit(1);
```
Parker-Stafford: same with this, i don't know how they mnage the recordings on the python side - if this is helpful we can keep for sure, i'm not familiar enough to know

js/packages/openinference-instrumentation-bedrock/scripts/validate-invoke-model.ts
```diff
@@ -0,0 +1,722 @@
...
+    const responseBody = JSON.parse(new TextDecoder().decode(response.body));
+
+    console.log(
+      "   🖼️ Multi-modal response:",
+      responseBody.content[0].text.substring(0, 50) + "...",
+    );
+    return true;
+  }
+
+  private async runToolResultsScenario(): Promise<boolean> {
```
Parker-Stafford, line 421: probably something similar here, very comprehensive at the cost of maybe being a bit overly robust. I think for validating and testing this is helpful for sure theoretically that can be captured in unit tests. As an example this is probably not very helpful because its very non standard in terms of setup i.e., this won't help someone actually learn how to use the instrumentation in their application code. If we follow the current pattern it would be

- test these things via unit tests
- add a simple example that could be copy and pasted into an application and work for users

that being said there might be room for stuff like this

Aemrox: Good comments, I'll dramatically trim this down post-development. It's been useful for me to validate how things look in phoenix in a tighter loop, but probably doesn't belong in the final version (or if it does, it should be dramatically slimmed down)

SESSION_SUMMARY.md
```diff
@@ -0,0 +1,374 @@
+# OpenInference Bedrock Instrumentation - Complete Session Summary
```
Parker-Stafford: remove

REMAINING_ENHANCEMENTS.md
```diff
@@ -0,0 +1,215 @@
+# Bedrock Instrumentation - Remaining Enhancements
+
+## Overview
```
Parker-Stafford: remove

CLAUDE.md
```diff
@@ -0,0 +1,195 @@
...
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Additional Memory Files
+
+This project maintains additional memory files that should be referenced for complete context:
+
+- **SESSION_SUMMARY.md**: Detailed session history, current project state, and TDD progress for AWS Bedrock instrumentation
+- **BEST_PRACTICES.md**: Code review standards, development methodology preferences, and critical architectural decisions
+
+These files contain essential context about ongoing work, established patterns, and user preferences that inform development decisions.
```
Parker-Stafford: oh nice, claude.md probably makes sense in here but remove references to the current project

CLAUDE.md
```diff
@@ -0,0 +1,195 @@
...
+pnpm install                 # Install dependencies
+pnpm test                    # Run all tests
+pnpm run lint                # Lint entire workspace
+pnpm run type:check          # Type check all packages
+pnpm run prettier:write      # Fix formatting
+
+# From individual package directory
+npm test                     # Run package tests
+npm run type:check           # Type check single package
+```
```
Parker-Stafford: there are some build commands too

pnpm run -r prebuild 
pnpm run -r build
also stuff at the top level might need to be prefixed with -r? or maybe it does it automatically

for install we also typically do pnpm install --frozen-lockfile to prevent unintended bumps / changes although thats maybe less of a concern with pinned pnpm version

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Sets a span attribute only if the value is not null, undefined, or empty string
+ * Provides null-safe attribute setting for OpenTelemetry spans
+ */
+export function setSpanAttribute(
+  span: Span,
+  key: string,
+  value: AttributeValue | null | undefined,
+): void {
+  if (value !== undefined && value !== null && value !== "") {
```
Parker-Stafford: nit: can combine the first two checks unless we have deep equality lint in this repo


```suggestion
  if (value != null && value !== "") {
```

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+  span: Span,
+  attributes: Record<string, AttributeValue | null | undefined>,
+) {
+  Object.entries(attributes).forEach(([key, value]) => {
+    setSpanAttribute(span, key, value);
+  });
+}
+
+// ========================================================================
+// CONVERSE API HELPER FUNCTIONS
```
Parker-Stafford: nit: like the doc comments on functions like this form a library, don't think comments like this are necessary as they often don't get updated when code does and become meaningless at best and confusing at worst

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+// ========================================================================
+
+import {
+  SystemPrompt,
+  ConverseMessage,
+  ConverseContentBlock,
```
Parker-Stafford: nit move imports to top of file

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+  isConverseToolResultContent,
+} from "../types/bedrock-types";
+
+/**
+ * Aggregates multiple system prompts into a single string with space separation
+ */
+export function aggregateSystemPrompts(systemPrompts: SystemPrompt[]): string {
+  return systemPrompts
+    .map((prompt) => prompt.text || "")
+    .join(" ")
```
Parker-Stafford: weird edge case but i think this will cause there to be double spaces if we fallback to "" here. Looking at the types though this doesn't look nullable so this fallback is unecessary

Image

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+      content: [{ text: aggregateSystemPrompts(systemPrompts) }],
+    });
+  }
+
+  aggregated.push(...messages);
+  return aggregated;
+}
+
+/**
+ * Generator function to extract attributes from a single message
```
Parker-Stafford: wow, i basically never see nor use generators in js, i'm not opposed but might reconsider just to be consistent with the code base

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Generator function to extract attributes from message content
+ */
+export function* getAttributesFromMessageContent(
+  content: ConverseContentBlock,
+): Generator<[string, AttributeValue]> {
+  if (isConverseTextContent(content)) {
+    yield ["message_content.type", "text"];
+    yield ["message_content.text", content.text];
```
Parker-Stafford: would make use of the semantic conventions library for these as opposed to the raw strings (same goes for all functions in this file)

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+  } else if (isConverseImageContent(content)) {
+    yield ["message_content.type", "image"];
+    if (content.image.format) {
+      yield ["message_content.image.image.format", content.image.format];
```
Parker-Stafford: not in semantic conventions so i probably would jus tnot capture, not sure the general stance on this like we discussed

Parker-Stafford: same for others, would definitely just hinge off of that file and if there is something missing go from there

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+      yield [
+        "message_content.image.image.url",
+        `data:${mimeType};base64,${base64}`,
+      ];
+    }
+  } else if (isConverseToolUseContent(content)) {
+    yield ["message_content.type", "tool_use"];
+    yield ["message_content.tool_use.id", content.toolUse.toolUseId];
+    yield ["message_content.tool_use.name", content.toolUse.name];
+    if (content.toolUse.input) {
```
Parker-Stafford: in the spirit of the other comments, looks like this is tool_call not tool_use

overall spec here: https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md
tool_call specifically: https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts#L84

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+      yield [
+        "message_content.tool_use.input",
```
Parker-Stafford: see above this is function_call_arguments_json i believe

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+/**
+ * Processes multiple messages and sets attributes on span with proper indexing
+ */
+export function processMessages(
+  span: Span,
+  messages: ConverseMessage[],
+  baseKey: string,
```
Parker-Stafford: this can be more strongly typed, it should only be typeof INPUT or OUTPUT mesages from sem conv package

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+  span: Span,
+  command: ConverseCommand,
+): void {
+  const input = command.input as ConverseRequestBody;
+  if (!input) return;
+
+  // Extract base request attributes
+  extractBaseRequestAttributes(span, input);
+
+  // Extract input messages attributes
```
Parker-Stafford: nit: these comments are unnecessary as they are self evident if you've named things clearly and written clear code (which you have!) imo and prone to being wrong when code changes would remove

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+} from "../types/bedrock-types";
+
+/**
+ * Extracts request attributes from Converse command and sets them on the span
+ */
+export function extractConverseRequestAttributes(
+  span: Span,
+  command: ConverseCommand,
+): void {
+  const input = command.input as ConverseRequestBody;
```
Parker-Stafford: i haven't gotten tot he other functions that consume this but this is probably too dangerous of a cast to sit in instrumentation as the cost of throwing in here can be really dramatic (sitting in production calls). I would make a typeguared to assert this type.

in addition to that take a look at the withSafety util in openinference core and the langchain instrumentation. Probably worth wrappign at least some if not all of these with withSafety

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+  extractInputMessagesAttributes(span, input);
+
+  // Extract tool configuration attributes
+  extractInputToolAttributes(span, input);
+}
+
+/**
+ * Extracts base request attributes: model, system, provider, parameters
+ */
+function extractBaseRequestAttributes(
```
Parker-Stafford: nit: sorry would probably put these above where they are called just so the file is easier to read top to bottom (personal preference)

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+): void {
+  // Model identification
+  if (input.modelId) {
+    setSpanAttribute(span, SemanticConventions.LLM_MODEL_NAME, input.modelId);
+  }
+
+  // System attribute for provider identification
+  setSpanAttribute(span, SemanticConventions.LLM_SYSTEM, "bedrock");
+
+  // Inference configuration as invocation parameters (always set, empty object if none)
```
Parker-Stafford: same note on comments, for me applies throughout but just my personal preference

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,121 @@
...
+function extractInputToolAttributes(
+  span: Span,
+  input: ConverseRequestBody,
+): void {
+  const toolConfig = input.toolConfig;
+  if (!toolConfig?.tools) return;
+
+  // Process each tool definition
+  toolConfig.tools.forEach((tool, index: number) => {
+    if (tool.toolSpec) {
```
Parker-Stafford: yeah this is probably too dangerous (meaning without an upstream typeguard this is too dangerous) see note above about safety, we need to have typeguards for everything here

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  // Find tool use content blocks
+  const toolUseBlocks = outputMessage.content.filter(isConverseToolUseContent);
+
+  toolUseBlocks.forEach((content, toolCallIndex: number) => {
+    const toolUse = content.toolUse;
+    if (toolUse) {
+      setSpanAttribute(
+        span,
+        `${SemanticConventions.LLM_OUTPUT_MESSAGES}.0.${SemanticConventions.MESSAGE_TOOL_CALLS}.${toolCallIndex}.${SemanticConventions.TOOL_CALL_FUNCTION_NAME}`,
```
Parker-Stafford, line 119: this is more how i would expect the file that did the message processing to look, with the semantic conventiosn everywhere isntead of the raw strings

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+function extractUsageAttributes(
+  span: Span,
+  response: ConverseResponseBody,
+): void {
+  const usage = response.usage;
+  if (!usage) return;
+
+  // Set only token counts that are available (null-safe)
+  setSpanAttribute(
```
Parker-Stafford, line 154: any cache token info available here

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+ * Request attribute extraction for AWS Bedrock instrumentation
+ *
+ * Handles extraction of semantic convention attributes from InvokeModel requests including:
+ * - Base model and system attributes
+ * - Input message processing
+ * - Tool definition processing
+ * - Invocation parameters
+ */
+
+import {
```
Parker-Stafford: probably would rename this file so that its clear this is just for invoke, and the otherone is just for converse, if i'm understanding correctly

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+import { setSpanAttribute } from "./attribute-helpers";
+import { Span } from "@opentelemetry/api";
+
+/**
+ * Extracts base request attributes (model, system, provider, invocation parameters)
+ */
+export function extractBaseRequestAttributes(
+  span: Span,
+  command: InvokeModelCommand,
+): void {
```
Parker-Stafford: oh one thing i've missed on all these functions typically we prefer "named" parameters over positional since its harder to mess up for other devs. 

like so:


```suggestion
export function extractBaseRequestAttributes({span, command}: {
  span: Span,
  command: InvokeModelCommand,
}): void {
```

for anything with 2 params or more i would say would just go with that format

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+/**
+ * Extracts input messages attributes from request body
+ */
+export function extractInputMessagesAttributes(
+  span: Span,
+  requestBody: InvokeModelRequestBody,
+): void {
+  // Extract user's message text as primary input value
+  const inputValue = extractPrimaryInputValue(requestBody);
+  setSpanAttribute(span, SemanticConventions.INPUT_VALUE, inputValue);
```
Parker-Stafford: so this is the stuff you were talking about with python instrumentation? with invoke just doing text with converse doing all? i think this makes sense if there are no messages in invoke

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+  if (requestBody.tools && Array.isArray(requestBody.tools)) {
+    requestBody.tools.forEach((tool, index) => {
+      // Convert Bedrock tool format to function schema for consistent format
+      const toolFormat = {
+        type: "function",
+        function: {
+          name: tool.name,
+          description: tool.description,
+          parameters: tool.input_schema,
+        },
```
Parker-Stafford: this looks right but what is this informed by? the normalization of tools is a big topic all around but curious where this is comng from

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+    // Handle both string (test format) and Uint8Array (SDK format)
+    let bodyString: string;
+    if (typeof command.input.body === "string") {
+      bodyString = command.input.body;
+    } else if (command.input.body instanceof Uint8Array) {
+      bodyString = new TextDecoder().decode(command.input.body);
+    } else {
+      // Handle other blob types by converting to Uint8Array first
+      bodyString = new TextDecoder().decode(
+        new Uint8Array(command.input.body as ArrayBuffer),
```
Parker-Stafford: hmm, this might be a semi sketchy cast here, it looks like its typed as any sadly... but i guess we are catching so maybe okay.

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+function extractInvocationParameters(
+  requestBody: InvokeModelRequestBody,
+): ExtractedInvocationParameters {
+  const invocationParams: ExtractedInvocationParameters = {};
+
+  // Standard AWS SDK parameters (using camelCase naming)
+  if (requestBody.max_tokens) {
+    invocationParams.maxTokens = requestBody.max_tokens;
+  }
+  if (requestBody.temperature !== undefined) {
```
Parker-Stafford: would change these to != null just so you catch if they were null or undefined since we have no idea what the api might do at any given time or version

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  // Handle complex content arrays (tool results, etc.)
+  if (Array.isArray(message.content)) {
+    const toolResultBlocks = extractToolResultBlocks(message.content);
+    toolResultBlocks.forEach((contentBlock, contentIndex: number) => {
+      // Extract tool result attributes
+      setSpanAttribute(
+        span,
+        `${SemanticConventions.LLM_INPUT_MESSAGES}.${index}.${SemanticConventions.MESSAGE_TOOL_CALLS}.${contentIndex}.${SemanticConventions.TOOL_CALL_ID}`,
```
Parker-Stafford: same comment here re the first file using different conventions than these, would use these if possible

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,402 @@
...
+        }
+      } else if (content.type === "tool_use") {
+        setSpanAttribute(
+          span,
+          `${contentPrefix}.${SemanticConventions.MESSAGE_CONTENT_TYPE}`,
+          "tool_use",
+        );
+        setSpanAttribute(
+          span,
+          `${contentPrefix}.message_content.tool_use.name`,
```
Parker-Stafford: same as above about tool_use, the attributes are admittedly confusing but these ones should go with the messages i think

https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts#L343

js/packages/openinference-instrumentation-bedrock/src/attributes/response-attributes.ts
```diff
@@ -0,0 +1,243 @@
...
+  );
+
+  // Set only token counts provided in the response
+  // If the response includes total tokens, we could add:
+  // setSpanAttribute(span, SemanticConventions.LLM_TOKEN_COUNT_TOTAL, responseBody.usage.total_tokens);
+
+  // Add cache-related token attributes
+  setSpanAttribute(
+    span,
+    `${SemanticConventions.LLM_TOKEN_COUNT_PROMPT}.cache_read`,
```
Parker-Stafford: we have cahce conventions

js/packages/openinference-instrumentation-bedrock/src/attributes/response-attributes.ts
```diff
@@ -0,0 +1,243 @@
...
+      );
+    } else if (isToolUseContent(content)) {
+      setSpanAttribute(
+        span,
+        `${contentPrefix}.${SemanticConventions.MESSAGE_CONTENT_TYPE}`,
+        "tool_use",
+      );
+      setSpanAttribute(
+        span,
+        `${contentPrefix}.message_content.tool_use.name`,
```
Parker-Stafford: same

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+
+/**
+ * Check if Bedrock instrumentation is enabled/disabled
+ */
+export function isPatched() {
+  return _isBedrockPatched;
+}
+
+export interface BedrockInstrumentationConfig extends InstrumentationConfig {
+  // Remove traceConfig since we'll use direct OpenTelemetry API
```
Parker-Stafford: remove? also we need to use our oiTracer from core and config so that we can support attributes masking, 

tracer from core: https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-core/src/trace/trace-config/OITracer.ts

implementation example:https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-langchain/src/instrumentation.ts#L38-L78

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+        },
+      );
+
+      _isBedrockPatched = true;
+    }
+
+    return moduleExports;
+  }
+
+  private _handleInvokeModelCommand(
```
Parker-Stafford: nit no need for _ with the private keyword

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+        }
+      }
+    }
+
+    // Create response body structure for consistency with non-streaming
+    const responseBody = {
+      id: "bedrock-streaming-response",
+      type: "message",
+      role: "assistant",
+      model: "bedrock-streaming", // This would need to be extracted from request
```
Parker-Stafford: hmm not sure if this is necessary... feels a little weird but not familiar with all the nuances of bedrock stuff here

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+        usage.input_tokens,
+      );
+      setSpanAttribute(
+        span,
+        SemanticConventions.LLM_TOKEN_COUNT_COMPLETION,
+        usage.output_tokens,
+      );
+      // Set only token counts provided in the response
+      // If the response includes total tokens, we could add:
+      // setSpanAttribute(span, SemanticConventions.LLM_TOKEN_COUNT_TOTAL, usage.total_tokens);
```
Parker-Stafford: remove?

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -0,0 +1,304 @@
...
+  ConverseCommandOutput,
+  InferenceConfiguration,
+  SystemContentBlock,
+  Message as AwsConverseMessage,
+  ContentBlock,
+  ToolConfiguration,
+} from "@aws-sdk/client-bedrock-runtime";
+
+// Core content types
+export interface TextContent {
```
Parker-Stafford, line 75: man it would be great if they exposed these types somewhere an we dind't have to write them all ourselves, makes me a little uneasy

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+  ConverseCommand,
+  BedrockRuntimeClient,
+} from "@aws-sdk/client-bedrock-runtime";
+import { extractInvokeModelRequestAttributes } from "./attributes/request-attributes";
+import { extractInvokeModelResponseAttributes } from "./attributes/response-attributes";
+import { extractConverseRequestAttributes } from "./attributes/converse-request-attributes";
+import { extractConverseResponseAttributes } from "./attributes/converse-response-attributes";
+import { setSpanAttribute } from "./attributes/attribute-helpers";
+import { isToolUseContent } from "./types/bedrock-types";
+import { splitStream } from "@smithy/util-stream";
```
Parker-Stafford: asuming this is the backbone of not consuming the stream here in instrumentation so that the client can still consume it? that's a concern with streaming responses and geting their output

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+            return { ...response, body: userStream };
+          } catch (splitError: any) {
+            // If stream splitting fails, fall back to original behavior
+            diag.warn(
+              "Stream splitting failed, falling back to direct consumption:",
+              splitError,
+            );
+            try {
+              await this._consumeBedrockStreamChunks(response.body, span);
+            } catch (streamError: any) {
```
Parker-Stafford: would make errors unknown type not any

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,548 @@
...
+                span.setStatus({
+                  code: SpanStatusCode.ERROR,
+                  message: streamError.message,
+                });
+                span.end();
+              });
+
+            // Return response with user stream immediately
+            return { ...response, body: userStream };
+          } catch (splitError: any) {
```
Parker-Stafford: if this fails is the stream to the user properly preserved?

js/packages/openinference-instrumentation-bedrock/src/utils/content-processing.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (Array.isArray(content)) {
+    const textParts: string[] = [];
+
+    content.forEach((block) => {
+      if (isTextContent(block)) {
+        textParts.push(block.text);
+      } else if (isToolResultContent(block)) {
+        textParts.push(`[Tool Result: ${block.content}]`);
```
Parker-Stafford: what are these string prefixes? seem non standard if we need them for later parsing should put them in constants

SESSION_SUMMARY.md
Aemrox: yup - tried to remove these in my last commit, but i added them wrong plan on taking out all of the docs here, including Claude.md

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: thanks for catching this! would have never realized

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: good call, was trying to hew to python too closely here, refactored this along the openai instrumentation lines.

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: Thank you, I totally missed this - doing so now

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
Aemrox: Thanks for this, I totally missed this method - refactored all of converse request and response attributes to use this paradigm and skip on potentially dangerous type checks

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
Aemrox: good note!

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
Aemrox: There should be but it doesn't look like I'm getting any in my recordings even in situations where I should. I even tried writing a new test for this  using the cachePoint argument and trying to get as many cache tokens as I could, but I ran into AWS JS SDK serialization issues trying to use that argument - looks like this is a known issue with the AWS JS SDK: https://github.com/langchain-ai/langchain-aws/issues/326, it works for langchain but they are wrapping the call in their own function. There has to be a way to solve this serialization issue, but I'm not sure what my next step should be here. 

Including the error I get when I attempt to use cachepoint in the manner outline in this ticket and Committing the test anyway. 

> 
> TypeError: Cannot read properties of undefined (reading '0')
> 
>       at Object.visit (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:701:36)
>       at se_ContentBlock (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1823:23)
>       at ../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1836:12
>           at Array.map (<anonymous>)
>       at se_ContentBlocks (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1835:41)
>       at content (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1917:21)
>       at applyInstruction (../../node_modules/.pnpm/@smithy+smithy-client@4.2.0/node_modules/@smithy/smithy-client/dist-cjs/index.js:1115:27)
>       at take (../../node_modules/.pnpm/@smithy+smithy-client@4.2.0/node_modules/@smithy/smithy-client/dist-cjs/index.js:1083:5)
>       at se_Message (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1916:40)
>       at ../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1923:12
>           at Array.map (<anonymous>)
>       at se_Messages (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1922:41)
>       at messages (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1168:24)
>       at applyInstruction (../../node_modules/.pnpm/@smithy+smithy-client@4.2.0/node_modules/@smithy/smithy-client/dist-cjs/index.js:1115:27)
>       at take (../../node_modules/.pnpm/@smithy+smithy-client@4.2.0/node_modules/@smithy/smithy-client/dist-cjs/index.js:1083:5)
>       at se_ConverseCommand (../../node_modules/.pnpm/@aws-sdk+client-bedrock-runtime@3.750.0/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1163:35)
>       at ../../node_modules/.pnpm/@smithy+middleware-serde@4.0.3/node_modules/@smithy/middleware-serde/dist-cjs/index.js:71:25
>       at ../../node_modules/.pnpm/@smithy+middleware-endpoint@4.1.0/node_modules/@smithy/middleware-endpoint/dist-cjs/index.js:216:12
>       at ../../node_modules/.pnpm/@aws-sdk+middleware-logger@3.734.0/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:33:22

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
Aemrox: Good callout, will fix

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
Aemrox: Based on our conversation, I'm going to change this to keep consistency on spans across both APIs. I went with what I did with invoke model, having the input just be the entire raw format. Going to leave this comment unresolved as an open question for other reviewers.

js/packages/openinference-instrumentation-bedrock/src/attributes/request-attributes.ts
Aemrox: Tried my best to rework the checks so that I'm checking all possible types this could be - between API and SDK calls, this is the ultimate union type it can be: 

```
/**
 * @public
 *
 * A union of types that can be used as inputs for the service model
 * "blob" type when it represents the request's entire payload or body.
 *
 * For example, in Lambda::invoke, the payload is modeled as a blob type
 * and this union applies to it.
 * In contrast, in Lambda::createFunction the Zip file option is a blob type,
 * but is not the (entire) payload and this union does not apply.
 *
 * Note: not all types are signable by the standard SignatureV4 signer when
 * used as the request body. For example, in Node.js a Readable stream
 * is not signable by the default signer.
 * They are included in the union because it may work in some cases,
 * but the expected types are primarily string and Uint8Array.
 *
 * Additional details may be found in the internal
 * function "getPayloadHash" in the SignatureV4 module.
 */
export type BlobPayloadInputTypes = string | ArrayBuffer | ArrayBufferView | Uint8Array | NodeJsRuntimeBlobTypes | BrowserRuntimeBlobTypes;
```

So reworked this to check for almost each type

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: Refactored this a bit, this was my mistake - leftover dummy data from my first implementation that didn't impact the ultimate instrumentation.

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: yeah, this gives me the unmutated stream to pass through and a disposable split off one to work through the insturmentation

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
Aemrox: They do for a lot of them - I was looking in the wrong places - deleting a ton of these and falling back on the types they provie

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: yup!

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: It in fact did not, I reworked this entirely to use the native ReadableStream from node, which I had missed in my initial pass on how to handle this. I feel more confident in this now and have done my best to test through failures.

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
Aemrox: They are, I missed them - using them now

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,184 @@
...
+  if (message.role) {
+    attributes[SemanticConventions.MESSAGE_ROLE] = message.role;
+  }
+
+  if (message.content) {
+    let toolCallIndex = 0;
+
+    for (const [index, content] of message.content.entries()) {
+      // Process content as our custom types for attribute extraction
+      const contentAttributes = getAttributesFromMessageContent(content);
```
Parker-Stafford, line 166: nit: its easier for me at least to read top to bottom, wasn't sure where this function was coming from as its not in imports so would just move it above this one

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,184 @@
...
+          content.toolUse.toolUseId;
+        attributes[
+          `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_NAME}`
+        ] = content.toolUse.name;
+        if (content.toolUse.input) {
+          attributes[
+            `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}`
+          ] = JSON.stringify(content.toolUse.input);
+        }
+        toolCallIndex++;
```
Parker-Stafford, line 183: so just so i'm understanding the tool index incrementsat different rate than the message content index? cause there could be multiple tool calls per message content?

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,184 @@
...
+      } else if (isConverseToolResultContent(content)) {
+        attributes[SemanticConventions.MESSAGE_TOOL_CALL_ID] =
```
Parker-Stafford, line 185: is there anything else here for tool results? or is that just captured in output.value or something?

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -0,0 +1,184 @@
...
+ * Extracts OpenInference semantic convention attributes from a single Bedrock message
+ * Handles role, content, tool calls, and tool results following the OpenInference specification
+ *
+ * @param message The Bedrock message to extract attributes from
+ * @returns {Record<string, AttributeValue>} Object containing semantic convention attributes
+ */
+export function getAttributesFromMessage(
+  message: Message,
+): Record<string, AttributeValue> {
+  const attributes: Record<string, AttributeValue> = {};
```
Parker-Stafford: don't have to change it but I _think_ there is a type called Attributes from otel

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,216 @@
...
+ * @param input The input object to validate
+ * @returns {boolean} True if input is a valid ConverseRequest, false otherwise
+ */
+function isConverseRequest(input: unknown): input is ConverseRequest {
+  if (!input || typeof input !== "object" || input === null) {
+    return false;
+  }
+
+  const obj = input as Record<string, unknown>;
+  return (
```
Parker-Stafford: we have a typeguard in the utils of openinference core https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-core/src/utils/typeUtils.ts#L29 for string keyed objects

see others in there as well it might be helpful

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+function getSystemFromModelId(modelId: string): string {
+  if (modelId.includes("anthropic")) return "anthropic";
+  if (modelId.includes("ai21")) return "ai21";
+  if (modelId.includes("amazon")) return "amazon";
+  if (modelId.includes("cohere")) return "cohere";
+  if (modelId.includes("meta")) return "meta";
+  if (modelId.includes("mistral")) return "mistral";
+  return "bedrock";
+}
```
Parker-Stafford: there is an enum for this: https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts#L733

it doesn't have all of these but we should add to it and use that enum since the downstream apps (arize and phoenix) may be doing checks using that enum to determine system / provider etc.

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,216 @@
...
+    SemanticConventions.LLM_SYSTEM,
+    getSystemFromModelId(modelId),
+  );
+  setSpanAttribute(
+    span,
+    SemanticConventions.LLM_MODEL_NAME,
+    extractModelName(modelId),
+  );
+
+  // Use AWS SDK InferenceConfiguration directly - no conversion needed!
```
Parker-Stafford: nit don't have to do it but would presonally remove comments like this

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
```diff
@@ -0,0 +1,216 @@
...
+ * Processes:
+ * - Base model and system attributes
+ * - Input messages with system prompt aggregation
+ * - Tool configuration definitions
+ *
+ * @param params Object containing extraction parameters
+ * @param params.span The OpenTelemetry span to set attributes on
+ * @param params.command The ConverseCommand to extract attributes from
+ */
+export const extractConverseRequestAttributes = withSafety({
```
Parker-Stafford, line 150: this works, else where have just written the functions and wrapped them so they can be unit testable in and of themselves if needed but you have all the helpers here anyways.

On our other conversation wrapping at this level means that if extractBaseRequestAttributes fails then you wont get the other two. if you wrap each individually you get as much as possible, this is probably fine though

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Type guard to safely validate ConverseResponse structure
+ * Ensures the response object has the expected output structure for processing
+ *
+ * @param response The response object to validate
+ * @returns {boolean} True if response is a valid ConverseResponse, false otherwise
+ */
+function isConverseResponse(response: unknown): response is ConverseResponse {
+  if (!response || typeof response !== "object" || response === null) {
```
Parker-Stafford: see comment above re core type utils

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
Parker-Stafford: alright no worries, if they exist on the object in a known location you could try to get them here even if you don't have a way to reliably test for them i would say but also can just be omitted probably

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-response-attributes.ts
```diff
@@ -0,0 +1,203 @@
...
+ */
+export const extractConverseResponseAttributes = withSafety({
+  fn: ({
+    span,
+    response,
+  }: {
+    span: Span;
+    response: ConverseResponse;
+  }): void => {
+    if (!response || !isConverseResponse(response)) {
```
Parker-Stafford: same as above

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-request-attributes.ts
```diff
@@ -0,0 +1,492 @@
...
+  );
+  setSpanAttribute(
+    span,
+    SemanticConventions.LLM_MODEL_NAME,
+    extractModelName(modelId),
+  );
+  setSpanAttribute(span, SemanticConventions.INPUT_MIME_TYPE, MimeType.JSON);
+  setSpanAttribute(span, SemanticConventions.LLM_PROVIDER, LLMProvider.AWS);
+
+  const invocationParams = extractInvocationParameters(requestBody);
```
Parker-Stafford: same comment as above just for readability move these functions above where they are used

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-request-attributes.ts
```diff
@@ -0,0 +1,492 @@
...
+ * @returns {string} The system provider name (e.g., "anthropic", "meta", "mistral") or "bedrock" as fallback
+ */
+function getSystemFromModelId(modelId: string): string {
+  if (modelId.includes("anthropic")) return "anthropic";
+  if (modelId.includes("ai21")) return "ai21";
+  if (modelId.includes("amazon")) return "amazon";
+  if (modelId.includes("cohere")) return "cohere";
+  if (modelId.includes("meta")) return "meta";
+  if (modelId.includes("mistral")) return "mistral";
+  return "bedrock";
```
Parker-Stafford: this looks the same as the other function, can we make a set of utils that work across both api's for stuff like this

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-request-attributes.ts
```diff
@@ -0,0 +1,492 @@
...
+  }
+  if (requestBody.stop_sequences) {
+    invocationParams.stopSequences = requestBody.stop_sequences;
+  }
+
+  if (requestBody.top_k != null) {
+    invocationParams.top_k = requestBody.top_k;
+  }
+  if (requestBody.anthropic_version) {
+    invocationParams.anthropic_version = requestBody.anthropic_version;
```
Parker-Stafford: i don't have a problem with this but just curious if there was a reason in this case we grab specific things fromt eh request body where as for invoke we just stringified everything?

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-request-attributes.ts
```diff
@@ -0,0 +1,492 @@
...
+    setSpanAttribute(
+      span,
+      `${SemanticConventions.LLM_INPUT_MESSAGES}.${messageIndex}.${SemanticConventions.MESSAGE_CONTENTS}.0.${SemanticConventions.MESSAGE_CONTENT_TYPE}`,
+      "text",
+    );
+    setSpanAttribute(
+      span,
+      `${SemanticConventions.LLM_INPUT_MESSAGES}.${messageIndex}.${SemanticConventions.MESSAGE_CONTENTS}.0.${SemanticConventions.MESSAGE_CONTENT_TEXT}`,
+      message.content,
+    );
```
Parker-Stafford: this looks very similar / duplicated from lines 331+ above, seems liek they are handling the same case but in slightly different ways, i don't think this will ever be hit with that logic. My instinct would be to just put it all in here

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
```diff
@@ -0,0 +1,286 @@
...
+  // Determine if this is a simple text response or complex multimodal response
+  const isSimpleTextResponse =
+    responseBody.content &&
+    Array.isArray(responseBody.content) &&
+    responseBody.content.length === 1 &&
+    isTextContent(responseBody.content[0]);
+
+  if (isSimpleTextResponse) {
+    // For simple text responses, use the message content
+    const textBlock = responseBody.content[0] as TextContent;
```
Parker-Stafford: ahh i see if you wrap the above check in a typeguard you wn't have to cast here which makes this slightly more safe going forward not a big deal though, or maybe you can remove the cast? typscript probably isn't smar tenough though

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
```diff
@@ -0,0 +1,286 @@
...
+  // Standard token counts
+  if (usage.input_tokens !== undefined) {
+    setSpanAttribute(
+      span,
+      SemanticConventions.LLM_TOKEN_COUNT_PROMPT,
+      usage.input_tokens,
+    );
+  }
+
+  if (usage.output_tokens !== undefined) {
```
Parker-Stafford, line 180: do you need these checks given what your setSpanAttribute function do?

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
```diff
@@ -0,0 +1,286 @@
...
+  if (!responseBody.content || !Array.isArray(responseBody.content)) {
+    return;
+  }
+
+  // Process each content block in the response
+  responseBody.content.forEach((content, contentIndex) => {
+    const contentPrefix = `${SemanticConventions.LLM_OUTPUT_MESSAGES}.0.${SemanticConventions.MESSAGE_CONTENTS}.${contentIndex}`;
+
+    if (isTextContent(content)) {
+      const textContent = content as TextContent;
```
Parker-Stafford: this shouldn't be necessary if the above typeguard is written properly

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (typeof response.body === "string") {
+      responseText = response.body;
+    } else if (response.body instanceof Uint8Array) {
+      responseText = new TextDecoder().decode(response.body);
+    } else {
+      // Handle other potential types
+      responseText = new TextDecoder().decode(response.body as Uint8Array);
+    }
+
+    return JSON.parse(responseText) as InvokeModelResponseBody;
```
Parker-Stafford: i don't know what the shape of this InvokeModelResponseBody is supposed to be but this is definitely not a safe cast.

This will not fail the with safety check but almost certainly will not be the correct type for example:

```typescript
JSON.parse('{"parker": 1}')
```

any down stream function of this will alow you to act on the return of this function as if its a `InvokeModelResponseBody` which will cause errors

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
```diff
@@ -0,0 +1,327 @@
...
+ * Streaming event data that matches Bedrock's event stream format
+ * Using AWS SDK types where possible, extending for missing fields
+ */
+interface StreamEventData {
+  type:
+    | "message_start"
+    | "content_block_start"
+    | "content_block_delta"
+    | "message_delta";
+  message?: {
```
Parker-Stafford: i'm abit surprised these aren't exposed in some way but if they aren't no biggie, i know a lot of their types are super generic and don't include specifics about the api your actually calling

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
```diff
@@ -0,0 +1,327 @@
...
+  const outputValue = {
+    text: outputText || "",
+    tool_calls: contentBlocks.filter(isToolUseContent).map((content) => ({
+      id: content.id,
+      name: content.name,
+      input: content.input,
+    })),
+    usage: usage || {},
+    streaming: true,
+  };
```
Parker-Stafford: feels a little strange to build this object to put in output_value but if there's nothing else to use i suppose we can

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
```diff
@@ -0,0 +1,327 @@
...
+ * blocking the user's stream consumption.
+ *
+ * @param stream - The Bedrock response stream (AsyncIterable), typically from splitStream()
+ * @param span - The OpenTelemetry span to set attributes on
+ * @throws {Error} If critical stream processing errors occur
+ *
+ * @example
+ * ```typescript
+ * // Background processing after stream splitting
+ * const [instrumentationStream, userStream] = await splitStream(response.body);
```
Parker-Stafford, line 547: this line is wrong now given the changes above

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
```diff
@@ -0,0 +1,327 @@
...
+                } else {
+                  contentBlocks.push({
+                    type: "text",
+                    text: data.delta.text,
+                  });
+                }
+              }
+
+              if (data.type === "message_delta" && data.usage) {
+                usage = { ...usage, ...data.usage };
```
Parker-Stafford: ahh so this usage is cumulative in the chunks meaning
chunk 1
{usage: {response: 1}}
chunk 2
{usage: {response: 2}}
chunk 3
{usage: {response: 3}}

not
chunk 1
{usage: {response: 1}}
chunk 2
{usage: {response: 1}}
chunk 3
{usage: {response: 1}}

correct?

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
```diff
@@ -0,0 +1,327 @@
...
+ * ```typescript
+ * // Background processing after stream splitting
+ * const [instrumentationStream, userStream] = await splitStream(response.body);
+ * consumeBedrockStreamChunks(instrumentationStream, span)
+ *   .then(() => span.end())
+ *   .catch(error => { span.recordException(error); span.end(); });
+ * ```
+ */
+export const consumeBedrockStreamChunks = withSafety({
+  fn: async ({
```
Parker-Stafford, line 554: same comments as above re withSafety

js/packages/openinference-instrumentation-bedrock/src/utils/content-processing.ts
```diff
@@ -0,0 +1,132 @@
...
+  isToolUseContent,
+  isToolResultContent,
+} from "../types/bedrock-types";
+
+// Content formatting constants
+const TOOL_RESULT_PREFIX = "[Tool Result: ";
+const TOOL_CALL_PREFIX = "[Tool Call: ";
+const CONTENT_SUFFIX = "]";
+
+/**
```
Parker-Stafford: this feels a little funky, what is the purpose of this and the function below? like do we reparse this for attributes? do we put this in some big text block on some other attribute?

js/packages/openinference-instrumentation-bedrock/src/utils/content-processing.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Formats Bedrock image source data into OpenInference data URL format
+ * Converts Bedrock image source to standard data URL: data:{media_type};base64,{data}
+ *
+ * @param source The Bedrock image source containing type, data, and media type
+ * @returns {string} Formatted data URL or empty string if source is invalid
+ */
+export function formatImageUrl(source: ImageSource): string {
+  if (source.type === "base64" && source.data && source.media_type) {
+    return `data:${source.media_type};base64,${source.data}`;
```
Parker-Stafford: i think i saw this being done not by using this function somewhere up above we should probably use this everywhere, if i'm wrong ignore

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,439 @@
...
+ * - InvokeModel commands (synchronous)
+ * - InvokeModelWithResponseStream commands (streaming)
+ * - Converse commands (conversation API)
+ *
+ * @param instrumentationConfig The config for the instrumentation @see {@link InstrumentationConfig}
+ * @param traceConfig The OpenInference trace configuration. Can be used to mask or redact sensitive information on spans. @see {@link TraceConfigOptions}
+ */
+export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExports> {
+  static readonly COMPONENT = "@arizeai/openinference-instrumentation-bedrock";
+  static readonly VERSION = VERSION;
```
Parker-Stafford: is this standard? our others just have this as constants in the file not as properties on the instrumentation, no biggie either way i think

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,439 @@
...
+      ]) as Promise<InvokeModelResponse>;
+
+      // AWS SDK v3 send() method always returns a Promise
+      return result
+        .then((response: InvokeModelResponse) => {
+          extractInvokeModelResponseAttributes({ span, response });
+          span.setStatus({ code: SpanStatusCode.OK });
+          span.end();
+          return response;
+        })
```
Parker-Stafford: any reason to not just await th e result on line 214? i guess we'd have to make this method async, seems fine as is

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -0,0 +1,439 @@
...
+    moduleVersion?: string,
+  ): void {
+    diag.debug(`Removing patch for ${MODULE_NAME}@${moduleVersion}`);
+
+    if (moduleExports?.BedrockRuntimeClient) {
+      this._unwrap(moduleExports.BedrockRuntimeClient.prototype, "send");
+      _isBedrockPatched = false;
+    }
+  }
+
```
Parker-Stafford: tiny nit: just for readability maybe move above the handlers next to patch

js/packages/openinference-instrumentation-bedrock/test/helpers/test-helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+  body: AsyncIterable<unknown> | undefined;
+}) => {
+  if (!result.body) {
+    return;
+  }
+
+  // Consume the stream to trigger instrumentation
+  for await (const _chunk of result.body) {
+    // Just consume chunks, don't need to do anything with them
+  }
```
Parker-Stafford, line 30: is this necessary? given the split stream stuff?

js/packages/openinference-instrumentation-bedrock/test/recordings/should-comprehensively-test-all-token-count-types.json
```diff
@@ -0,0 +1,221 @@
...
+            {
+              "text": "What is the capital of France?"
+            }
+          ],
+          "role": "user"
+        }
+      ],
+      "system": [
+        {
+          "text": "You are an expert geography assistant. You have extensive knowledge about world capitals, countries, and their historical backgrounds. Please provide accurate and detailed information about geographical questions. Always include interesting historical context in your responses when relevant."
```
Parker-Stafford, line 23: the other bedrock pr is not using nock, i gues we need to standardize on it

Parker-Stafford (approved):

Parker-Stafford (changes requested):
accidentally approved adding request changes to block

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: yup! There is a test that shows what this looks like "should handle multiple tools in single request" - do want to double check that this is right though

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: There were other things that I extracted here previously: 
- content.toolResult.status
- content.toolResult.content.entries() (this was output) 

but I didn't see those being recorded elsewhere or semantic conventions for them so I stripped out everything but the toolcall id. Should I put them back in?

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
Aemrox: thanks for showing me this - adding it wherever I can

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: got it and used it, thanks!

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-request-attributes.ts
Aemrox: good call - will do

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-request-attributes.ts
Aemrox: I think part of it is that the converse request type has a more specific inference config type that neatly contains all the pieces which we can stringify, whereas the InvokeAPI is a bit more jank - having just a requestBody field typed as a Uint8Array where they say in the comment it needs to be json that adheres to this https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters.html  - here is the code: https://github.com/aws/aws-sdk-js-v3/blob/9c6a29eb79cf6ddb6dc46c68c1621e2cfe3410a7/clients/client-bedrock-runtime/src/models/models_0.ts#L4522

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
Aemrox: nope - removing them across the insturmentation. good call

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts
Aemrox: It's because of how we intercept it, the types they expose are for parsed events. left a comment explaining more - but let me put it here because I want to make sure I made the right call here: 

 * Note: We define this custom interface because we're parsing the raw JSON lines
 * from the stream bytes directly, before the AWS SDK processes them into its
 * own event types (like ContentBlockDeltaEvent). The SDK types represent the
 * final processed events, but we need to work with the intermediate JSON structure
 * to avoid interfering with the user's stream consumption.

js/packages/openinference-instrumentation-bedrock/test/recordings/should-comprehensively-test-all-token-count-types.json
Aemrox: Yeah this may be above my paygrade to solve. FWIW I tried using Polly and ran into a bunch of trouble, ultimately deciding the hoops I had to jump through to get the tests to work wasn't worth it. But I agree that coming to some kind of consensus is desirable.

js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-response-attributes.ts
Aemrox: Yeah, I think I referenced above, the issue is this is a very weak type with 'body' being Uint8Array that gets cast into an object that ends up being very different for each model type. I totally get what you're saying about the unsafe cast. 

I feel like my options right now are to make a bunch of specialized types for models and cast based on the model type (or use something like a discriminated union), or just do a very basic "is this an object with string keys" validation and just access it as safely as possible down the road. 

Not sure which is best - if we go with path one I feel like I'll want tests for each model type. I'll address the rest of the comments and come back to this

js/packages/openinference-instrumentation-bedrock/test/helpers/test-helpers.ts
Aemrox: yes, it seems to be - fails without it

Parker-Stafford (approved):

Parker-Stafford:
some follow ups

- [ ] update typeguards in invoke streaming file
- [ ] normalize usage while processing it like we're doing with text etc.
- [ ] maybe add diag.warn for mlformed json in invoke streaming file

PR #1853, opened 2025-07-11
fix: vercel orphan span not captured when applied with filter

GeLi2001:
WARNING: not the correct solution, current fix will put the trace to be the parent span whenever it sees there's a parentSpan and spanFilter is applied:
Screenshot 2025-07-24 at 10 34 46 AM

### DIFF
--- a/js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
+++ b/js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
@@ -30,7 +30,7 @@ import { Context } from "@opentelemetry/api";
  * ```
  */
 export class OpenInferenceSimpleSpanProcessor extends SimpleSpanProcessor {
-  private spanFilter?: SpanFilter;
+  private readonly spanFilter?: SpanFilter;
 
   constructor({
     exporter,
--- a/js/packages/openinference-vercel/src/utils.ts
+++ b/js/packages/openinference-vercel/src/utils.ts
@@ -444,13 +444,11 @@ const shouldPromoteToRootSpan = (
     return false;
   }
 
-  // First check if it's one of the top-level AI operations we care about
   const isTopLevelOperation = TOP_LEVEL_AI_OPERATIONS.some(
     (op) =>
       operationName.startsWith(op) && !operationName.includes(".", op.length),
   );
 
-  // Only promote top-level AI operations that might genuinely be orphaned
   return isTopLevelOperation;
 };
 


js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
```diff
@@ -30,35 +30,43 @@ import { Context } from "@opentelemetry/api";
...
-     */
-    readonly exporter: SpanExporter;
-    /**
-     * A filter to apply to spans before exporting. If it returns true for a given span, that span will be exported.
-     */
-    readonly spanFilter?: SpanFilter;
-
-    config?: BufferConfig;
+    exporter: SpanExporter;
+    spanFilter?: SpanFilter;
```
mikeldking: you lost the docstrings for some reason

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
```diff
@@ -30,35 +30,43 @@ import { Context } from "@opentelemetry/api";
...
   }) {
     super(exporter);
     this.spanFilter = spanFilter;
+    // eslint-disable-next-line no-console
+    console.error(
+      `🚨 PROCESSOR DEBUG: OpenInferenceSimpleSpanProcessor created with spanFilter: ${spanFilter ? "enabled" : "disabled"}`,
+    );
```
mikeldking: this feels like AI... can you remove?

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -548,10 +565,18 @@ const getOpenInferenceAttributes = (attributes: Attributes): Attributes => {
...
  * {@link getOpenInferenceAttributes} wrapped in {@link withSafety} which will return null if any error is thrown
  */
-export const safelyGetOpenInferenceAttributes = withSafety({
-  fn: getOpenInferenceAttributes,
-  onError: onErrorCallback(""),
-});
+export const safelyGetOpenInferenceAttributes = (
+  attributes: Attributes,
+  span?: ReadableSpan,
+  spanFilter?: SpanFilter,
```
mikeldking:
```suggestion
  spanFilter?: SpanFilter,
  isRootSpanFn?: (span: ReadableSpan) -> boolean,
```

One possible solution is to make this configurable

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
GeLi2001: removed

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
GeLi2001: added back

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
```diff
@@ -30,28 +30,28 @@ import { Context } from "@opentelemetry/api";
...
+
   constructor({
     exporter,
     spanFilter,
   }: {
     /**
      * The exporter to pass spans to.
      */
-    readonly exporter: SpanExporter;
+    exporter: SpanExporter;
```
Parker-Stafford: why remove readonly?

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -409,18 +410,84 @@ const safelyGetMetadataAttributes = withSafety({
...
+ */
+const shouldPromoteToRootSpan = (
+  span: ReadableSpan,
+  spanFilter?: SpanFilter,
+): boolean => {
+  if (!span.parentSpanId || !spanFilter) {
+    return false;
+  }
+
+  const operationName = span.attributes["operation.name"] as string;
```
Parker-Stafford: i dont think we should cast here, just doa quick 

```typescript
if (typeof operationName !== "string") {
 return false // or whatever should happen here
}
```

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -409,18 +410,84 @@ const safelyGetMetadataAttributes = withSafety({
...
+  // Only promote top-level AI operations that might genuinely be orphaned
+  // Examples: ai.generateText, ai.streamText, ai.generateObject, ai.streamObject, ai.embed, ai.embedMany
+  const topLevelAIOperations = [
+    "ai.generateText",
+    "ai.streamText",
+    "ai.generateObject",
+    "ai.streamObject",
+    "ai.embed",
+    "ai.embedMany",
+  ];
```
Parker-Stafford: nit: make this a constant outside of this function

also is there a way to invert this, just check if its one of these things at the top and other wise return false? given this check i think it would work

js/packages/openinference-vercel/src/utils.ts
Parker-Stafford: why remove with safety here? if  you need to update doc comments

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -587,15 +662,40 @@ export const shouldExportSpan = ({
...
+
+  // Check if this span should be made a root span (based on attributes set by getOpenInferenceAttributes)
+  const shouldBeRootSpan =
+    newAttributes[`${SemanticConventions.METADATA}._should_be_root_span`];
+  if (shouldBeRootSpan && span.parentSpanId) {
+    // Cast to ReadWriteSpan to allow modification
+    const writableSpan = span as ReadWriteSpan;
+
+    // Clear the parent span ID to make this a root span
+    writableSpan.parentSpanId = undefined;
```
Parker-Stafford: might want to try catch this, in some environments sthis might actually be locked and not writeable and could throw

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
GeLi2001: sry its accidentally removed, added back

js/packages/openinference-vercel/src/utils.ts
GeLi2001: updated

js/packages/openinference-vercel/src/utils.ts
GeLi2001: got it:
- moved outside as const
- inverted to only promote if its top-level ops

js/packages/openinference-vercel/src/utils.ts
GeLi2001: got it thanks for the note, added try-catch

GeLi2001:
Screenshot 2025-07-24 at 2 52 41 PM

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -409,18 +410,79 @@ const safelyGetMetadataAttributes = withSafety({
...
+  if (!span.parentSpanId || !spanFilter) {
+    return false;
+  }
+
+  const operationName = span.attributes["operation.name"];
+  if (typeof operationName !== "string") {
+    return false;
+  }
+
+  // First check if it's one of the top-level AI operations we care about
```
Parker-Stafford: tiny nit - this is the only check so the comment here is probably not worth it, would opt for any additional context be added to the doc comment above the function, same with the comment on line 453.

js/packages/openinference-vercel/src/OpenInferenceSpanProcessor.ts
```diff
@@ -30,7 +30,8 @@ import { Context } from "@opentelemetry/api";
  * ```
  */
 export class OpenInferenceSimpleSpanProcessor extends SimpleSpanProcessor {
-  private readonly spanFilter?: SpanFilter;
+  private spanFilter?: SpanFilter;
```
Parker-Stafford: also can retain readonly

js/packages/openinference-vercel/src/utils.ts
```diff
@@ -409,18 +410,77 @@ const safelyGetMetadataAttributes = withSafety({
   onError: onErrorCallback("metadata"),
 });
 
+/**
+ * Top-level AI operations that might genuinely be orphaned and should be promoted to root spans.
+ * These are operations that can have their parent spans filtered out.
+ */
+const TOP_LEVEL_AI_OPERATIONS = [
```
mikeldking, line 417: In general this doesn't work for me. Promoting things based on vercel AI sdk TOP level objects is not correct in a multitude of scenarios:
- AI sdk is used with langchain / llama-index
- there are wrapping spans created manually

This is why the only place we can determine if things should be root spans is in the export process where the whole tree is available.

PR #1854, opened 2025-07-11
feat: Update OpenAI agents instrumentation to log agent attributes

generalzhou:
Log new graph.node.id and graph.node.parent_id attributes:
- The graph.node.id corresponds to the agent name when the span is `AgentSpanData`
- When there is a handoff - we capture the the to and from agents
- When the handoff is complete and the next agent span starts, we log the `graph.node.parent_id`


axiomofjoy:
@generalzhou it looks like CI is red, can you resolve?

axiomofjoy (approved):

Issue #1852, opened 2025-07-11
[bug] GoogleGenAIInstrumentor doesn't capture spans when config options are passed in the client


trevorlaviale:
**Describe the bug**
The below code doesn't print any spans out to the console, but if you comment out the `config` parameter, it works fine. 

**To Reproduce**
```
import os

from arize.otel import register
from google import genai
from google.genai import types
from openinference.instrumentation.google_genai import GoogleGenAIInstrumentor


tracer_provider = register(
    space_id = "SPACE ID", # in app space settings page
    api_key = "API KEY", # in app space settings page
    project_name = "PROJECT NAME", # name this to whatever you would like,
    log_to_console = True,
    batch=False
)

# Instrument the Google GenAI client
GoogleGenAIInstrumentor().instrument(tracer_provider=tracer_provider)

client = genai.Client(
    vertexai=True, project="project-id-here", location='us-central1'
)

chat = client.models.generate_content(
    model="gemini-2.5-flash",
    contents=["What is the capital of France?"],
    config={
        "temperature": 0.5,
        "top_p": 0.95,
        "top_k": 40,
        "candidate_count": 1,
        "thinking_config": {
            "thinking_budget": 100
        }
    }
)

print("Chat --------------------------------")
print(chat)
print("--------------------------------")
```

**Expected behavior**
Spans are printed to the console with the config options set. 

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: MacOS Sequoia 15.5
 - Versions: 
 ```
openinference-instrumentation              0.1.34
openinference-instrumentation-google-genai 0.1.4
openinference-semantic-conventions         0.1.21
```

**Additional context**
https://arizeai.slack.com/archives/C04UM3125PC/p1752198549840679

trevorlaviale:
Looks like it may be happening here because we assume the `config` object is a pydantic model, but it fails when a regular dictionary is passed - https://github.com/Arize-ai/openinference/blob/8ceb1e1e1feaefa1270697427d0171a67c183c3a/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py#L111 

Seems like `config` can either be a pydantic model or a typeddict
- https://github.com/googleapis/python-genai/blob/d06254ccd94d6d90a20769bb003f8eb5dfb62553/google/genai/models.py#L4654
[...]

PR #1855, opened 2025-07-11
fix: fix config_json serialization for GoogleGenAIInstrumentor

trevorlaviale:
resolves: #1852 

Looks we assume the `config` object is a pydantic model, but it fails when a regular dictionary is passed - https://github.com/Arize-ai/openinference/blob/8ceb1e1e1feaefa1270697427d0171a67c183c3a/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py#L111 

Looks like `config` can either be a pydantic model or a typeddict so we need to account for both scenarios so it doesn't fail silently for dictionaries 
- https://github.com/googleapis/python-genai/blob/d06254ccd94d6d90a20769bb003f8eb5dfb62553/google/genai/models.py#L4654
- https://github.com/googleapis/python-genai/blob/d06254ccd94d6d90a20769bb003f8eb5dfb62553/google/genai/types.py#L4151
- https://github.com/googleapis/python-genai/blob/d06254ccd94d6d90a20769bb003f8eb5dfb62553/google/genai/types.py#L4151


trevorlaviale:

trevorlaviale:
recheck

harrisonchu (approved):

PR #1857, opened 2025-07-12
fix(vercel): span processor should not override existing span kind

### LINKED ISSUE #1856: [bug] openinfernce-vercel whipes out span_kind if set already
create a span that has the `openinference.span.kind` key set. this will get overwritten by the vercel OpenInferenceSpanProcessor.

Image

mikeldking:
Prevent `openinference.span.kind` from being overwritten if already set.

This fixes #1856 where the `openinference-vercel` package would incorrectly overwrite an existing `openinference.span.kind` attribute.


mikeldking:
recheck

PR #1872, opened 2025-07-16
fix(smolagents): Getting Multiple Traces While Streaming Code Agent

### LINKED ISSUE #1322: Instrumentation splits transactions into steps when streaming=True in codeAgent
My app writes instrumentations via Opentelemtry-sdk==1.29.0, opentelemtry-exporter-otlp==1.29.0 and openinference-instrumentation-smolagents==0.1.4
My smolagents version is 1.9.2

my_agent = CodeAgent(model=my_model, tools=my_tools_list)
my_agent.run("say hello", stream=True) causes each span to be its own step instead of all step being gathered under the same trace

Thank you for your help

ialisaleh:
This PR fixes an issue where `CodeAgent.run(stream=True)` in `smolagents` would generate separate traces for each step instead of grouping them under a single parent trace. The problem stemmed from Python `generators` not preserving OpenTelemetry `context` across `yield` statements. To address this, a new `wrapped_generator()` was introduced that ensures the tracing `context` is reattached during streaming and the final output is collected for proper span annotation. This change preserves existing behavior for non-streaming runs and ensures consistent, complete tracing data regardless of the stream flag.

Closes #1322

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
@@ -360,7 +360,6 @@ class TestModels:
 
 
 class TestRuns:
-    @pytest.mark.xfail
     def test_streaming_and_non_streaming_code_agent_runs(
         self, in_memory_span_exporter: InMemorySpanExporter
     ) -> None:
@@ -501,7 +500,6 @@ final_answer("Test result from CodeAgent")
         assert main_span_non_stream.status.status_code == trace_api.StatusCode.OK
         assert main_span_stream.status.status_code == trace_api.StatusCode.OK
 
-    @pytest.mark.xfail
     def test_multiagents(self) -> None:
         class FakeModelMultiagentsManagerAgent:
             def __call__(


python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -103,13 +104,17 @@ def __call__(
         instance: Any,
         args: Tuple[Any, ...],
         kwargs: Mapping[str, Any],
-    ) -> Any:
+    ) -> Union[Any, Generator[str, None, None]]:
```
axiomofjoy, line 108: Not sure the union with any buys us more than just `Any`.

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
axiomofjoy, line 1: Rather than handling both the generator and non-generator cases in the same try-except block, can we separate them into separate blocks for readability? For the non-generator case, it's probably sufficient to use `start_as_current_span`.

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @axiomofjoy - Using `Any` will work fine, but it will hide what the function actually returns. By using `Union[Any, Generator[str, None, None]]`, we make it clear that the function might return a final result or a streaming output, which helps tools like `mypy` check the code better. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @axiomofjoy - I have separated the blocks for generator & non-generator cases. I’m already handling the `context` properly with `attach` and `detach`, so this setup ensures consistent span tracking across both streaming and non-streaming executions. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
```diff
@@ -484,6 +486,16 @@ def __call__(
...
+        # Handle streaming (generator) run
+        report_chunks = manager_code_agent.run("Fake question.", stream=True)
+        final_result = "".join([chunk for chunk in report_chunks])
+        assert final_result == "Final report."
+
+        # Handle streaming (generator) run
+        report_chunks = manager_toolcalling_agent.run("Fake question.", stream=True)
+        final_result = "".join([chunk for chunk in report_chunks])
+        assert final_result == "Final report."
+
```
axiomofjoy: This test looks like it is being xfailed and doesn't seem to test the emitted traces at all. Can we nix altogether and add a more minimal test for the generator?

ialisaleh: Hi @axiomofjoy - You're right. `test_multiagents` is currently marked as `XFAIL` and doesn’t verify any emitted traces. Since I can't use `CodeAgent` directly in this test (due to the way instrumentation is triggered), I replaced it with a minimal test using a dummy agent to cover both the simple & generator cases in `_RunWrapper`. However, this dummy agent setup doesn't emit spans, so it can't validate the traces either. I will try to fix the xfailed test separately. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/tests/openinference/instrumentation/smolagents/test_instrumentor.py
```diff
@@ -360,129 +355,68 @@ def test_litellm_reasoning_model_has_expected_attributes(
...
+        wrapped_method = DummySimpleAgent.run
+        result = run_wrapper(wrapped_method, agent, ("Hi",), {})
+
+        assert isinstance(result, str)
+        assert not isinstance(result, GeneratorType), (
+            "Expected result to be a string, not generator"
+        )
+        assert result == "Final Simple Response."
+        assert agent.monitor.total_input_token_count == 10
+        assert agent.monitor.total_output_token_count == 20
```
axiomofjoy: We shouldn't need to manually wrap in the test case because of the `instrument` fixture above.

Let's check the content of the actual spans as in the tests below, otherwise, we are only testing that the instrumentation does not break the application code, not that we actually get the expected attributes.

ialisaleh: Hi @axiomofjoy - I have added a testcase to verify the emitted traces for streaming & non-streaming `CodeAgent` runs, along with fixing the `XFAILED` testcase for the `test_multiagents` run. Thanks!

axiomofjoy (approved):

PR #1873, opened 2025-07-16
feat: add support for non global trace provider in instrumentations

### LINKED ISSUE #1837: [instrumentation][js] Support non-global OpenTelemetry providers in @arizeai/openinference-instrumentation-langchain
_Is your feature request related to a problem? Please describe._
When instrumenting OpenTelemetry with [`@arizeai/openinference-instrumentation-langchain`](https://github.com/Arize-ai/openinference/tree/main/js/packages/openinference-instrumentation-langchain), the library forces the usage of the OpenTelemetry global provider (e.g., `provider.register();`). My code already depends on the global provider, and I'm unable to find a way to use the library as intended.
I am frustrated that I cannot use `LangChainInstrumentation.manuallyInstrument()` with a non-global OpenTelemetry provider.

_Describe the solution you'd like_
I would like the ability to use `LangChainInstrumentation.manuallyInstrument()` with a custom, non-global OpenTelemetry provider, rather than being forced to rely on the global provider registered by the library.

_Describe alternatives you've considered_
I have tried patching in a non-global provider using the following code. No errors occur, and the span processor is getting called, but the metrics are not being sent.

```ts
private createArizeLangChainInstrumentation(provider: NodeTracerProvider): LangChainInstrumentation {
  const lcInstrumentation = new LangChainInstrumentation();

  // hack: Opentelemetry double global registration
  (provider as any).getDelegateTracer = (name: string, version?: string, options?: any): any => provider.getTracer(name, version, options);
  const arizeTracer = new ProxyTracer(provider as any, 'arize-instrumentation-patch', '1.0.0');

  // hack: Opentelemetry double global registration
  (lcInstrumentation as any).trace = arizeTracer;
  (lcInstrumentation as any).oiTracer = new OITracer({
    tracer: arizeTracer as any,
    traceConfig: (lcInstrumentation as any).oiTracer.traceConfig,
  });

  return lcInstrumentation;
}
```

Any help will be greatly appreciated,
Thank you!

Parker-Stafford:
resolve #1837 

- adds options traceProvider param to langchain, beeai, and openai instrumentations to allow the tracer to be pulled from the passed in trace provider instead of the global one

### DIFF
--- a/js/packages/openinference-instrumentation-beeai/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-beeai/src/instrumentation.ts
@@ -68,7 +68,10 @@ export class BeeAIInstrumentation extends InstrumentationBase {
     this.traceConfig = traceConfig;
     this.oiTracer = new OITracer({
       tracer:
-        this.tracerProvider?.getTracer(INSTRUMENTATION_NAME) ?? this.tracer,
+        this.tracerProvider?.getTracer(
+          INSTRUMENTATION_NAME,
+          this.instrumentationVersion,
+        ) ?? this.tracer,
       traceConfig,
     });
   }
--- a/js/packages/openinference-instrumentation-beeai/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-beeai/src/instrumentation.ts
@@ -3,7 +3,7 @@ import {
   InstrumentationNodeModuleDefinition,
   InstrumentationConfig,
 } from "@opentelemetry/instrumentation";
-import { diag } from "@opentelemetry/api";
+import { diag, TracerProvider } from "@opentelemetry/api";
 import { Version } from "beeai-framework";
 import * as bee from "beeai-framework";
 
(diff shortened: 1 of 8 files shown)


js/packages/openinference-instrumentation-beeai/src/instrumentation.ts
```diff
@@ -85,6 +103,26 @@ export class BeeAIInstrumentation extends InstrumentationBase {
...
+    this.tracerProvider = tracerProvider;
+    this.updateOITracer();
+  }
+
+  private updateOITracer(): void {
+    this.oiTracer = new OITracer({
+      tracer: this.tracer,
+      traceConfig: this.traceConfig,
+    });
+  }
```
mikeldking: very very minor nit but doesn't need to be a method if it's only called on L116

Parker-Stafford: good point, will update

PR #1874, opened 2025-07-16
feat(js-bedrock-agent-runtime): AWS Bedrock Agent Instumentaion for JS

satyadevai:
This MR introduces foundational support for AWS Bedrock Agent instrumentation within the JavaScript runtime. It enables observability capabilities by wrapping Bedrock invocations and capturing key request and response attributes.

- Adds a new package: openinference-instrumentation-bedrock-agent-runtime containing instrumentation logic. 

- Implements instrumentation via an initializer that wraps Bedrock agent invocations in OpenTelemetry modules. 

-  Extracts and records invocation parameters, converting non-string values to JSON where needed, and supports validation using isAttributeValue. 

- Includes example script examples/run-invoke-agent.js (renamed from a validation script) that demonstrates how to instrument Bedrock usage. 

Current implementation captures only input and output. Trace streaming and span tree assembly will be pursued in a follow-up MR.


js/packages/openinference-instrumentation-bedrock-agent-runtime/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "dependencies": {
+    "@arizeai/openinference-core": "workspace:*",
+    "@arizeai/openinference-semantic-conventions": "workspace:*",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/core": "^1.25.1",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@pollyjs/adapter-node-http": "^6.0.6",
+    "@pollyjs/core": "^6.0.6",
+    "@pollyjs/persister-fs": "^6.0.6",
+    "@smithy/util-stream": "^4.2.0"
```
mikeldking: These should be dev dependancies.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,167 @@
...
+
+let _isBedrockAgentPatched = false;
+
+export function isPatched() {
+  return _isBedrockAgentPatched;
+}
+
+export interface BedrockAgentInstrumentationConfig
+  extends InstrumentationConfig {
+  traceConfig?: TraceConfigOptions;
```
mikeldking: can you add the ability to add a custom tracer provider as per https://github.com/Arize-ai/openinference/pull/1873

js/packages/openinference-instrumentation-bedrock-agent-runtime/scripts/validate-invoke-agent.ts
```diff
@@ -0,0 +1,69 @@
+import "./agent-instrumentation";
```
mikeldking: this feels more like an example. Can we make it a example directory?

js/packages/openinference-instrumentation-bedrock-agent-runtime/scripts/validate-invoke-agent.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import {
+  BedrockAgentRuntimeClient,
+  InvokeAgentCommand,
+} from "@aws-sdk/client-bedrock-agent-runtime";
+import * as console from "node:console";
```
mikeldking: is this needed? Also looks like bun syntax?

js/packages/openinference-instrumentation-bedrock-agent-runtime/scripts/validate-invoke-agent.ts
```diff
@@ -0,0 +1,69 @@
...
+
+async function run() {
+  const agentId = "3EL4X42BSO";
+  const agentAliasId = "LSIZXQMZDN";
```
mikeldking: should we remove these hard-codings.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/response-handler.ts
```diff
@@ -0,0 +1,48 @@
+import * as console from "node:console";
```
mikeldking: Can we just use the built-in console here.

mikeldking: nit. Can we name the file camel case as this is more common in JS

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/stream-utils.ts
```diff
@@ -0,0 +1,25 @@
+import { ResponseHandler } from "./response-handler";
+
+export function interceptAgentResponse<
+  T extends { chunk?: { bytes?: Uint8Array } },
+>(
+  originalStream: AsyncIterable<T>,
+  callback: ResponseHandler,
```
mikeldking: callback params typically are actual callbacks (e.g. lambdas) - I would name it more appropriately

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,167 @@
...
+    const result = original.apply(client, [command]);
+    return result
+      .then((response: InvokeAgentCommandOutput) => {
+        const callback = new ResponseHandler(span);
+        if (
+          response.completion &&
+          Symbol.asyncIterator in response.completion
+        ) {
+          response.completion = interceptAgentResponse(
+            response.completion,
```
mikeldking, line 157: are we sure this doesn't consume the original stream?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  private _handleInvokeAgentCommand(
+    command: InvokeAgentCommand,
+    original: (
+      command: InvokeAgentCommand,
+    ) => Promise<InvokeAgentCommandOutput>,
+    client: unknown,
+  ) {
+    const span = this.oiTracer.startSpan("bedrock.invoke_agent", {
+      kind: SpanKind.CLIENT,
```
mikeldking: I think since this isn't at the network boundary we might want this to be INTERNAL

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/response-handler.ts
```diff
@@ -0,0 +1,48 @@
...
+import { SpanStatusCode, Span } from "@opentelemetry/api";
+import {
+  MimeType,
+  SemanticConventions,
+} from "@arizeai/openinference-semantic-conventions";
+
+export class ResponseHandler {
```
mikeldking: This doesn't feel like a generic class. It's super specific. Can be name it different?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: yes, we are yieling he response in interceptAgentResponse

js/packages/openinference-instrumentation-bedrock-agent-runtime/examples/validate-invoke-agent.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import {
+  BedrockAgentRuntimeClient,
+  InvokeAgentCommand,
+} from "@aws-sdk/client-bedrock-agent-runtime";
+import { diag } from "@opentelemetry/api";
+
+async function run() {
+  const agentId = "<AgentId>"; // Replace with your actual agent ID
+  const agentAliasId = "<AgentAliasId>"; // Replace with your actual agent alias ID
```
Parker-Stafford: could move to env variables and have an env.example or something, don't have to but might be more consistent with other instrumentation examples

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (requestBody.agentId) {
+    invocationParams.agentId = requestBody.agentId;
+  }
+  if (requestBody.agentAliasId) {
+    invocationParams.agentAliasId = requestBody.agentAliasId;
+  }
+  if (requestBody.sessionId) {
+    invocationParams.sessionId = requestBody.sessionId;
+  }
```
Parker-Stafford: just curious what makes these three invocation params important here? like should we just spread the command.input into invocation params? what are we missing if we just take these three?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,47 @@
...
+  return invocationParams;
+}
+
+export function extractBaseRequestAttributes(
+  command: InvokeAgentCommand,
+): Record<string, string> {
+  const attributes: Record<string, string> = {
```
Parker-Stafford: there is an attributes type from open inference would use it

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,47 @@
...
+    [SemanticConventions.OPENINFERENCE_SPAN_KIND]: OpenInferenceSpanKind.LLM,
+    [SemanticConventions.LLM_SYSTEM]: "bedrock",
+    [SemanticConventions.INPUT_MIME_TYPE]: MimeType.JSON,
```
Parker-Stafford: is this the case? looks like down below you are just adding text as the input, which would make this mimeType.text

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,47 @@
...
+    invocationParams.sessionId = requestBody.sessionId;
+  }
+  return invocationParams;
+}
+
+export function extractBaseRequestAttributes(
+  command: InvokeAgentCommand,
+): Record<string, string> {
+  const attributes: Record<string, string> = {
+    [SemanticConventions.OPENINFERENCE_SPAN_KIND]: OpenInferenceSpanKind.LLM,
```
Parker-Stafford: don't know enough about this package, is this not span kind of agent?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,47 @@
...
+    [SemanticConventions.LLM_SYSTEM]: "bedrock",
+    [SemanticConventions.INPUT_MIME_TYPE]: MimeType.JSON,
+    [SemanticConventions.LLM_PROVIDER]: LLMProvider.AWS,
```
Parker-Stafford, line 18: model is missing here llm_model

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.span = span;
+  }
+
+  consumeResponse(chunk: Uint8Array) {
+    const text = Buffer.from(chunk).toString("utf8");
+    this.outputChunks.push(text);
+  }
+  consumeTrace(trace: object) {
+    diag.info(JSON.stringify(trace));
+  }
```
Parker-Stafford: is this necessary? seems pretty strange

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
```diff
@@ -0,0 +1,49 @@
...
+        "assistant",
+      [`${SemanticConventions.LLM_OUTPUT_MESSAGES}.0.${SemanticConventions.MESSAGE_CONTENT}`]:
+        finalOutput,
+    });
+    this.span.setStatus({ code: SpanStatusCode.OK });
+    this.span.end();
+  }
+
+  onError(error: unknown): void {
+    this.span.recordException(error as Error);
```
Parker-Stafford: do we need a type check here over a cast? 


```suggestion
if (error instanceof Error) {
    this.span.recordException(error);
    }
```

Parker-Stafford: that also might simplify the logic below

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,210 @@
...
+  private getBedrockAgentModule(moduleExports: unknown):
+    | {
+        BedrockAgentRuntimeClient: {
+          prototype: {
+            send: (
+              command: InvokeAgentCommand,
+            ) => Promise<InvokeAgentCommandOutput>;
+          };
+        };
+      }
```
Parker-Stafford: do they not have a type for this?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.tracerProvider = tracerProvider;
+    this.traceConfig = traceConfig;
+    this.oiTracer = new OITracer({
+      tracer:
+        this.tracerProvider?.getTracer(INSTRUMENTATION_NAME) ?? this.tracer,
+      traceConfig: traceConfig,
+    });
+  }
+
+  protected init(): InstrumentationModuleDefinition<unknown>[] {
```
Parker-Stafford: i think if we type these things with the type of the BedrockModule not unknown we can simplify some of the checks and stuff dodwn below liek the getBedrockAgentModule and also provide type saftey for clients. If not possible no problem but maybe look at the openai instrumentation for how it handles these types

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,210 @@
...
+          if (command?.constructor?.name === "InvokeAgentCommand") {
+            return instrumentationInstance._handleInvokeAgentCommand(
+              command,
+              original,
+              this,
+            );
+          }
+          return original.apply(this, [command]);
+        };
+        /* eslint-disable @typescript-eslint/no-this-alias */
```
Parker-Stafford: remove

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,210 @@
...
+          command: InvokeAgentCommand,
+        ) {
+          if (command?.constructor?.name === "InvokeAgentCommand") {
+            return instrumentationInstance._handleInvokeAgentCommand(
+              command,
+              original,
+              this,
+            );
+          }
+          return original.apply(this, [command]);
```
Parker-Stafford, line 129: just for my own understanding why is this wrapped in an array when the original type does not seem to do so?

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/integration.instrumentation.agent-without-traces.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  // Global cleanup
+  afterAll(async () => {
+    instrumentation.disable();
+    await provider.shutdown();
+  });
+
+  it("should record agent attributes and API response in span", async () => {
+    // instrumentation = create_instrumentation(memoryExporter);
```
Parker-Stafford: remove

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/recordings/bedrock-agent-without-traces-bedrockagentinstrumentation-integration-agent-attributes-and_3073832749/recording.har
```diff
@@ -0,0 +1,127 @@
...
+          },
+          "queryString": [],
+          "url": "https://bedrock-agent-runtime.ap-south-1.amazonaws.com/agents/3EL4X42BSO/agentAliases/LSIZXQMZDN/sessions/default-session1_1234567890/text"
+        },
+        "response": {
+          "bodySize": 480,
+          "content": {
+            "encoding": "base64",
+            "mimeType": "application/vnd.amazon.eventstream",
+            "size": 480,
```
Parker-Stafford, line 71: on the comment above on consuming the stream it doesn't look like any of the recordings have stream chunks is it possible to do so or no?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: You're right — I initially wanted to explicitly set the exact parameters used for invocation. The original object included other fields like input, memory, and some complex objects that weren't relevant for tracing. I've now updated the code to extract and set everything except inputText as invocationParams.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: Implemented as suggested. Thanks for the input!

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: Implemented as suggested.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: Yes, It is AGENT. Modified as suggested

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: Model details won't be available in the Input itself, since the models are configured at the Bedrock agent level. The model-related information will be captured in the agent traces, which will be implemented in a follow-up MR.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
satyadevai: This is a placeholder for capturing agent traces. It will be used in a follow-up MR where we’ll implement the trace tree creation logic.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: Done

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/integration.instrumentation.agent-without-traces.test.ts
satyadevai: Done

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/recordings/bedrock-agent-without-traces-bedrockagentinstrumentation-integration-agent-attributes-and_3073832749/recording.har
satyadevai: The array wrapping is necessary because the init() method must return an array of InstrumentationModuleDefinition objects, even if there is only one module definition. This is required by the OpenTelemetry instrumentation API, which expects an array to support multiple modules. The original type for a single module is not an array, but the method signature and framework usage require returning an array.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: The array wrapping is necessary because the init() method must return an array of InstrumentationModuleDefinition objects, even if there is only one module definition. This is required by the OpenTelemetry instrumentation API, which expects an array to support multiple modules. The original type for a single module is not an array, but the method signature and framework usage require returning an array.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: This section is not required any more

js/packages/openinference-instrumentation-bedrock-agent-runtime/examples/validate-invoke-agent.ts
```diff
@@ -0,0 +1,62 @@
+import "./agent-instrumentation";
+
+import {
```
Parker-Stafford: i think @mikeldking made a comment on examples vs. validation scripts, i think this is a fine example but would maybe just rename it as such.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
Parker-Stafford: ahh gotcha, okay ty for the context!

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+    [SemanticConventions.INPUT_MIME_TYPE]: MimeType.TEXT,
+    [SemanticConventions.LLM_PROVIDER]: LLMProvider.AWS,
+  };
+
+  // Add invocation parameters for model configuration
+  const { inputText: _inputText, ...invocationParams } = command.input;
+  const jsonParams = Object.fromEntries(
+    Object.entries(invocationParams).map(([key, value]) => [
+      key,
+      typeof value === "string" ? value : JSON.stringify(value),
```
Parker-Stafford: i think otel exposes a function isAttributeValue which we use here: https://github.com/Arize-ai/openinference/blob/1b9219d7c250044dab23b42c0b51a59eb4205f09/js/packages/openinference-core/src/utils/typeUtils.ts#L43

i would use that check here, since we don't need to stringify everything that's not a string, just things that aren't attribute values

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.span = span;
+  }
+
+  consumeResponse(chunk: Uint8Array) {
+    const text = Buffer.from(chunk).toString("utf8");
+    this.outputChunks.push(text);
+  }
+  consumeTrace(trace: object) {
+    diag.info(JSON.stringify(trace));
+  }
```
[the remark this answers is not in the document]
Parker-Stafford: I agree with cursor on this an i think i left a comment on this, don't see the reason for having this in here

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
Parker-Stafford: Hmm I'm a little ocnfused this doesn't seem related to the instrumentationModuleDefinition, nor otel instrumentation api, you're just applying arguments to the original function see here in the openai instrumentation for example: https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-openai/src/instrumentation.ts#L234.

I think this has to do with the signature of the original function... let me know if this doesn't make sense

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/recordings/bedrock-agent-without-traces-bedrockagentinstrumentation-integration-agent-attributes-and_3073832749/recording.har
Parker-Stafford: this doesn't seem like a real response to my comment

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,184 @@
...
+          command: InvokeAgentCommand,
+        ) => Promise<InvokeAgentCommandOutput>,
+      ) => {
+        /* eslint-disable @typescript-eslint/no-this-alias */
+        const instrumentationInstance = this;
+        return function patchedSend(
+          this: typeof moduleExports.BedrockAgentRuntimeClient.prototype,
+          command: InvokeAgentCommand,
+        ) {
+          if (command?.constructor?.name === "InvokeAgentCommand") {
```
Parker-Stafford: i would probably heed cursor's advice here, i don't think constructor names are very reliable https://github.com/Arize-ai/openinference/pull/1874#pullrequestreview-3039318745

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: You are correct, original method(patched send method) will be applied here and apply function will take the list of arguments. So we need to pass it as list of args.

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/recordings/bedrock-agent-without-traces-bedrockagentinstrumentation-integration-agent-attributes-and_3073832749/recording.har
satyadevai: we have two versions of invoke_agent method, one is enabling trace - which will stream the traces and will build the span tree for the traces. another one is simple input and ouput - it returns the simple text as response. 

current implementation is base version which contains only input & output. thats the reason you don;t see the stream chunks in the recording. 

Once it is approved, will create another follow up MR with the trace collection and building the Span Tree with traces. there you can see the trace chunks for stream in recording.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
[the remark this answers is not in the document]
satyadevai: Yes, I too agree, but this is base version of implementation. we have followup implementation to this MR where collecting traces from aws bedrock, collecting span tree with the traces data. This debug line will not be there in followup MR.

js/packages/openinference-instrumentation-bedrock-agent-runtime/examples/validate-invoke-agent.ts
satyadevai: Renamed file name to `examples/run-invoke-agent.js`

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
satyadevai: Yes, Now i am using isAttributeValue method for checking the value is valid or not.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: NOw I am using `instanceof` insted constructor name

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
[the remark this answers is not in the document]
Parker-Stafford: Still feel it should just be removed if its not used now

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
Parker-Stafford: i see ty ty

Parker-Stafford (approved):

js/packages/openinference-instrumentation-bedrock-agent-runtime/examples/agent-instrumentation.ts
```diff
@@ -0,0 +1,33 @@
...
+} from "@opentelemetry/sdk-trace-node";
+import { diag, DiagConsoleLogger, DiagLogLevel } from "@opentelemetry/api";
+import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
+import { BedrockAgentInstrumentation } from "../src";
+import { registerInstrumentations } from "@opentelemetry/instrumentation";
+
+// For troubleshooting, set the log level to DiagLogLevel.DEBUG
+diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
+/////////////////// instrumentation ///////////////////
```
mikeldking, line 14: nit, this comment is useless

js/packages/openinference-instrumentation-bedrock-agent-runtime/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "type": "git",
+    "url": "https://github.com/Arize-ai/openinference.git"
+  },
+  "scripts": {
+    "prebuild": "rimraf dist && pnpm run version:update",
+    "build": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json && tsc-alias -p tsconfig.esm.json",
+    "postbuild": "echo '{\"type\": \"module\"}' > ./dist/esm/package.json && rimraf dist/test",
+    "version:update": "../../scripts/version-update.js",
+    "type:check": "tsc --noEmit",
+    "test": "jest .",
```
mikeldking, line 20: nit: can we move to vite in a follow-up?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
mikeldking: If this is the case I don't think we need the LLM provider on the agent. It's not gonna add any info. Let's only add the provider on the llm spans.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
Parker-Stafford: hmm i thought i saw that you weren't but looks like it didn't get added would change to be


```suggestion
      isAttributeValue(value) ? value : JSON.stringify(value),
```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/request-attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { inputText: _inputText, ...invocationParams } = command.input;
+  const jsonParams = Object.fromEntries(
+    Object.entries(invocationParams).filter(([, value]) =>
+      isAttributeValue(value),
+    ),
+  );
+  if (Object.keys(jsonParams).length > 0) {
+    attributes[SemanticConventions.LLM_INVOCATION_PARAMETERS] =
+      JSON.stringify(jsonParams);
+  }
```
mikeldking, line 31: also think you can leave this off from the agent for now

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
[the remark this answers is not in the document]
mikeldking: just comment it out?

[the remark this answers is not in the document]
satyadevai: Giving linting issue, thats the reason i added as debug log

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callback-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.span = span;
+  }
+
+  consumeResponse(chunk: Uint8Array) {
+    const text = Buffer.from(chunk).toString("utf8");
+    this.outputChunks.push(text);
+  }
+  consumeTrace(trace: object) {
+    diag.debug(JSON.stringify(trace));
+  }
```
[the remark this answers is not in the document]
Parker-Stafford: Ignoring the first issue here since multibyte characters split across chunk boundaries is very unlikely since these chunks need to be consumable and displayable for streaming UI's if it comes up we can add a text decoder to this class but don't want to block now as there are some considerations on when to instantiate a new decoder and when / how to end it.

Removed the consumeTrace method as its not used

cc @satyadevai

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,184 @@
...
+            response.completion,
+            callback,
+          );
+        } else {
+          // End the span if response.completion is not a stream
+          span.setStatus({ code: SpanStatusCode.OK });
+          span.end();
+        }
+        return response;
+      })
```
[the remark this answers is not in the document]
Parker-Stafford: @satyadevai if you could addres this in a followup

PR #1877, opened 2025-07-16
fix: exclude java from python release path, read version from gradle.…

ArcticFaded:
…properties

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,11 @@ jobs:
       fail-fast: false
       matrix:
         path: ${{ fromJSON(needs.release.outputs.paths_released) }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -28,6 +28,11 @@ jobs:
       fail-fast: false
       matrix:
         path: ${{ fromJSON(needs.release.outputs.paths_released) }}
+        exclude:
+          # Exclude Java packages from Python publishing
+          - path: java/openinference-semantic-conventions
+          - path: java/openinference-instrumentation
```
mikeldking, line 34: can we exclude all of java?

mikeldking (approved):
approving to unblock

PR #1878, opened 2025-07-16
fix: fix IndexError for empty function output

luke-moehlenbrock:
If a function returns an empty string, `_get_attributes_from_function_span_data` will raise an IndexError - https://github.com/Arize-ai/openinference/blob/4645932b68f7ed5ab3ecd8818ddad9e1011c027e/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py#L459


caroger:
thanks @luke-moehlenbrock !

PR #1883, opened 2025-07-16
Fix snapshot publish

ArcticFaded:


java/build.gradle
```diff
@@ -6,9 +6,12 @@ plugins {
     id "com.diffplug.spotless" version "6.8.0" apply false
 }
 
+// Root project version for JReleaser
+version = '0.1.1'
```
mikeldking, line 10: this doesn't need to be auto-updated via release please?

java/openinference-semantic-conventions/build.gradle
```diff
@@ -1,5 +1,6 @@
 description = 'OpenInference Semantic Conventions for Java'
 
+// thirrd test release please
```
mikeldking, line 3:
```suggestion
// third test release please
```

java/build.gradle
ArcticFaded: this is just the version of the root project but the root itself has no published artifacts, its only here because JReleaser wants one

java/openinference-semantic-conventions/build.gradle
ArcticFaded: will be removing this part soon

PR #1897, opened 2025-07-17
fix(Langchain): Custom data handling when using additional kwargs

ArnavJain23:


axiomofjoy:
Looks like CI is red @ArnavJain23

PR #2053, opened 2025-08-07
feat(Agno): Capture graph attributes for agent visual

ArnavJain23:
Capture graph.node.id, graph.node.name, and graph.node.parent_id to display the agent visual.

Agent visual for an Agno agent team using coordinate mode:
coordinate-mode-graph

Agent visual for an Agno agent team using collaborate mode:
collaborate-mode-graph

Agent visual for an Agno agent team using route mode:
route-mode-graph


python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -74,17 +80,57 @@ def _strip_method_args(arguments: Mapping[str, Any]) -> dict[str, Any]:
     return {key: value for key, value in arguments.items() if key not in ("self", "cls")}
 
 
+def _generate_node_id(path: str) -> str:
```
generalzhou, line 83: why do this vs just using the path? Wouldn't the path be unique?

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -74,17 +80,57 @@ def _strip_method_args(arguments: Mapping[str, Any]) -> dict[str, Any]:
...
+            yield GRAPH_NODE_NAME, agent.name
+
+        # Use context parent instead of structural parent
+        if context_parent_id:
+            yield GRAPH_NODE_PARENT_ID, cast(str, context_parent_id)
+
     elif isinstance(agent, Agent):
+        if current_path is None:
+            current_path = agent.name or "agent"
+        node_id = _generate_node_id(cast(str, current_path))
```
generalzhou, line 118: we can just yse the agent name as the node id - no need to have 2 different ones

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -173,33 +247,61 @@ def run_stream(
...
+        current_path = f"{parent_path}.{agent_name}" if parent_path else agent_name
+        node_id = _generate_node_id(current_path)
+
+        # Set current path in context before span creation so _agent_run_attributes can use it
+        current_path_ctx = context_api.set_value(
+            _AGNO_CURRENT_PATH_CONTEXT_KEY, current_path, context_api.get_current()
+        )
+        current_path_token = context_api.attach(current_path_ctx)
+
+        try:
```
generalzhou, line 275: why put the try so high up?

generalzhou (commented):
can you add a description (in a comment) of how the context and parent path stuff works?

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
ArnavJain23: Because there is a finally on line 303

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
ArnavJain23: I think it's necessary to generate node ids because theoretically its possible for a team to delegate multiple tasks to the same agent which would have the same path each time it's called and that would cause collisions. Also a 16 character hex string is more efficient if teams contain other teams as members which would result in long hierarchical path strings.

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
ArnavJain23: Does my answer to the previous comment address this or are you suggesting something else here?

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
generalzhou: Wouldn't the same path generate the same ids?

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -105,10 +151,21 @@ def _agent_run_attributes(
                     tool_names.append(tool.__name__)
                 else:
                     tool_names.append(str(tool))
-            yield "agno{key_suffix}.tools", tool_names
+            yield f"agno{key_suffix}.tools", tool_names
 
 
 class _RunWrapper:
+    """
+    The hierarchical path for each agent or team is constructed using context propagation.
```
generalzhou: uhh can you make this less of a paragraph? Just say something like:
```
We need to keep track of parent/child relationships for agent logging. The parent is logged on the child span. We do this by:
1. set blah blah
2. read blah from context
3. ...
```

caroger (approved):

PR #2058, opened 2025-08-08
feat(bedrock): Add instrumentation for guardrailTrace

luke-moehlenbrock:
This PR adds instrumentation for Bedrock Guardrails. By doing so it also fixes an issue where if streaming is enabled with guardrails, it would create a long chain of spans.

Tested with:
`streamingConfigurations={
        "applyGuardrailInterval": 3,
        "streamFinalResponse": True
 }`

Trace without guardrail isntrumentation:
Screenshot 2025-08-07 at 6 28 01 PM

Trace with guardrail instrumentation:
Screenshot 2025-08-07 at 6 27 45 PM

Example of intervening guardrail 
Screenshot 2025-08-07 at 7 19 11 PM


python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
```diff
@@ -861,9 +862,35 @@ def extract_trace_id(cls, trace_data: dict[str, Any]) -> Any:
             if "traceId" in rationale:
                 return rationale["traceId"]
 
+        # For guardrail traces
+        if "traceId" in event_data:
```
nate-mar: Doesn't this logic already exist up at 835?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
```diff
@@ -411,8 +421,22 @@ def _create_spans(self, parent_span: Span, trace_node: TraceNode) -> None:
...
+            status_code = StatusCode.OK
+            if attributes.span_kind == OpenInferenceSpanKindValues.GUARDRAIL:
+                guardrail_actions = attributes.metadata.get("guardrails", [])
+                try:
+                    if any(
+                        (action.get("action") == "INTERVENED")
+                        for action in guardrail_actions
+                        if isinstance(action, dict)
+                    ):
+                        status_code = StatusCode.ERROR
```
nate-mar: Is it intentional that a guardrail action is equivalent to an error?  And we also don't care about any of the other contents in the guardrail?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
luke-moehlenbrock: Yeah you're right I missed that

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
```diff
@@ -861,9 +862,35 @@ def extract_trace_id(cls, trace_data: dict[str, Any]) -> Any:
...
+        # Extract client_request_id from the guardrail metadata if present
+        metadata_attributes = cls.get_metadata_attributes(guardrail_trace.get("metadata", {}))
+        if client_request_id := metadata_attributes.get("client_request_id"):
+            guardrail_trace_data["client_request_id"] = client_request_id
+
+        # Normalize keys to snake_case for consistency in metadata
+        if "action" in guardrail_trace:
+            guardrail_trace_data["action"] = guardrail_trace["action"]
+        if "inputAssessments" in guardrail_trace:
+            guardrail_trace_data["input_assessments"] = guardrail_trace["inputAssessments"]
```
ArcticFaded: do we have to also normalize the keys of the assessments themselves https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_GuardrailAssessment.html

inputAssessments
The details of the input assessments used in the Guardrail Trace.

Type: Array of [GuardrailAssessment](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_GuardrailAssessment.html) objects

outputAssessments
The details of the output assessments used in the Guardrail Trace.

Type: Array of [GuardrailAssessment](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_GuardrailAssessment.html) objects

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
```diff
@@ -411,8 +421,22 @@ def _create_spans(self, parent_span: Span, trace_node: TraceNode) -> None:
...
             start_time = self._fetch_span_start_time(attributes, trace_span)
+
             span = self._create_chain_span(parent_span, attributes, start_time)
-            span.set_status(Status(StatusCode.OK))
+            status_code = StatusCode.OK
+            if attributes.span_kind == OpenInferenceSpanKindValues.GUARDRAIL:
+                guardrail_actions = attributes.metadata.get("guardrails", [])
+                try:
+                    if any(
+                        (action.get("action") == "INTERVENED")
```
nate-mar: Is there a reference for this? I've only found GUARDRAIL_INTERVENED as a response type, but not sure if I'm looking at the right documentation

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
luke-moehlenbrock: Yeah there's two possibilities for action, NONE or INTERVENED. If the action is NONE then there's really not much information included in the guardrailTrace, just the clientRequestId. If it's INTERVENED then the guardrail blocked the rest of the execution and will include the outputAssessments that contributed to the intervention. 

As for setting the status to error, it provides a visual distinction that the guardrail was triggered and blocked the request

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
nate-mar: oh okay, I see here; it's under events. not totally sure I understand the difference between event and action here. seems like in the documentation they refer to different things? https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_GuardrailEvent.html

luke-moehlenbrock: Yeah I based it off of this - https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_GuardrailTrace.html

nate-mar: ah got it

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
luke-moehlenbrock: That's a good point, there are also some other areas in this instrumentation where we preserve the camel case notations from bedrock:
Screenshot 2025-08-13 at 3 33 54 PM

Would it be preferable to normalize all of the attributes or revert this normalization and preserve the bedrock naming conventions?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
luke-moehlenbrock: Removed this check

nate-mar:
Two things:
- Just curious, what is the mechanism by which the streaming issue gets fixed? i.e. which part of your PR?
- Any chance we can update an existing unit test to happy path test this?

luke-moehlenbrock:
> Two things:
> 
> * Just curious, what is the mechanism by which the streaming issue gets fixed? i.e. which part of your PR?

The streaming issue actually happens because the guardrailTrace isn't part of trace_events. So every time we see a guardrailTrace, the TraceCollector creates a new LLM span by default, and since we don't have a trace event to create the node_trace_id and match it to the previous guardrailTraces, each individual guardrailTrace gets created as a new TraceNode that gets nested under the previous one. This is also kind of amplified by the fact that each streamed output chunk has a guardrail applied to it.

So this is the easiest change to point to that resolves the streaming issue, but mostly it's the entire PR 
https://github.com/Arize-ai/openinference/blob/4c1a2225b8cdd8ee161c6cb51147b22bc5444e5e/python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py#L539-L545

> * Any chance we can update an existing unit test to happy path test this?

Yeah absolutely, I'll work on adding some additional testing tonight

nate-mar:
> > Two things:
> > 
> > * Just curious, what is the mechanism by which the streaming issue gets fixed? i.e. which part of your PR?
> 
> The streaming issue actually happens because the guardrailTrace isn't part of trace_events. So every time we see a guardrailTrace, the TraceCollector creates a new LLM span by default, and since we don't have a trace event to create the node_trace_id and match it to the previous guardrailTraces, each individual guardrailTrace gets created as a new TraceNode that gets nested under the previous one. This is also kind of amplified by the fact that each streamed output chunk has a guardrail applied to it.
> 
> So this is the easiest change to point to that resolves the streaming issue, but mostly it's the entire PR
> 
> https://github.com/Arize-ai/openinference/blob/4c1a2225b8cdd8ee161c6cb51147b22bc5444e5e/python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py#L539-L545
> 
> > * Any chance we can update an existing unit test to happy path test this?
> 
> Yeah absolutely, I'll work on adding some additional testing tonight

awesome, sounds good. And thanks for the detailed explanation.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
nate-mar: Ah okay. I see, gotcha.

I don't really think the try is necessary based on your code here, you have a guard for the type, so I would not expect there to be an exception.

luke-moehlenbrock:
@nate-mar Added some tests cases for streaming with guardrails and for guardrail intervention.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_trace_collector.py
```diff
@@ -245,6 +257,11 @@ def _handle_new_trace_node(
         elif event_type == "failureTrace":
             trace_node = TraceNode(node_trace_id, event_type)
             trace_node.chunks.append(trace_data)
+        elif event_type == "guardrailTrace":
+            trace_node = TraceNode(node_trace_id, event_type)
+            trace_span = TraceSpan(chunk_type)
+            trace_span.add_chunk(trace_data)
+            trace_node.add_span(trace_span)
```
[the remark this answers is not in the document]
luke-moehlenbrock: Double checked the indentation, no issues there. Also the trace_span.parent_node property doesn't appear to be used anywhere in the instrumentor and therefore shouldn't have an impact on trace hierarchy

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
luke-moehlenbrock: Agreed, I was trying to be too careful there, I took out the try/except

python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
```diff
@@ -638,3 +640,104 @@ def test_multi_agent_collaborator(
...
+        "orchestrationTrace",
+        "bedrock_agent.invoke_agent",
+    ]
+    guardrail_span = [span for span in spans if span.name == "Guardrails"][-1]
+    guardrail_span_attributes = dict(guardrail_span.attributes or {})
+    assert guardrail_span_attributes.pop("openinference.span.kind") == "GUARDRAIL"
+    guardrail_metadata = guardrail_span_attributes.pop("metadata")
+    assert isinstance(guardrail_metadata, str)
+    guardrail_metadata = json.loads(guardrail_metadata)
+    assert isinstance(guardrail_metadata, dict)
```
nate-mar, line 685: would it be good to also check that the metadata appears correctly here?

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
nate-mar: @ArcticFaded any thought on this? 

I think it'd probably be better to just preserve what bedrock passes, but let me know if you think differently.

python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
luke-moehlenbrock: Thanks for the suggestion @nate-mar, I added some additional checks for the metadata structure for the guardrails

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
```diff
@@ -864,6 +865,65 @@ def extract_trace_id(cls, trace_data: dict[str, Any]) -> Any:
...
+            guardrail_trace_data["action"] = guardrail_trace["action"]
+        if "inputAssessments" in guardrail_trace:
+            guardrail_trace_data["inputAssessments"] = guardrail_trace["inputAssessments"]
+        if "outputAssessments" in guardrail_trace:
+            guardrail_trace_data["outputAssessments"] = guardrail_trace["outputAssessments"]
+
+        return guardrail_trace_data
+
+    @classmethod
+    def get_blocked_guardrail_status(cls, guardrails: List[dict[str, Any]]) -> bool:
```
nate-mar: It's generally more common to name functions like this with an "is_" prefix as that commonly communicates that it'll return true/false. This currently reads like it'll return the actual status of a blocked guardrail.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_trace_collector.py
```diff
@@ -198,6 +198,18 @@ def __init__(self) -> None:
         self.trace_nodes: dict[str, TraceNode] = {default_trace_id: self.initial_node}
         self.trace_ids: List[str] = [self.initial_node.node_trace_id]
 
+    def find_node_by_trace_id(self, trace_id: str) -> Optional[TraceNode]:
```
nate-mar: This is a nit but the trace_nodes.get(trace_id) is pretty self explanatory as it is and doesn't really benefit that much from a separate function here.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_attribute_extractor.py
```diff
@@ -864,6 +865,65 @@ def extract_trace_id(cls, trace_data: dict[str, Any]) -> Any:
...
+                        if filter.get("action") == blocked:
+                            return True
+                if policy := assessment.get("topicPolicy"):
+                    for topic in policy.get("topics", []):
+                        if topic.get("action") == blocked:
+                            return True
+                if policy := assessment.get("wordPolicy"):
+                    words = policy.get("customWords", []) + policy.get("managedWordLists", [])
+                    for word in words:
+                        if word.get("action") == blocked:
```
nate-mar: the deep nesting is a little bit hard to follow; would recommend extracting out the inner policy checks. pseudocode:

```
            for assessment in assessments:
               if  is_blocked_assessment(assessment, "contentPolicy", ["filters"]):
                   return True
``` 

```
def is_assessment_blocked(assessment: Assessment, policyType: str, policyChecks: List): <-- or whatever an appropriate name would be
```

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_response_accumulator.py
```diff
@@ -116,7 +117,8 @@ def __call__(self, obj: _AnyT) -> _AnyT:
                 if "chunk" in obj:
                     if "bytes" in obj["chunk"]:
                         output_text = obj["chunk"]["bytes"].decode("utf-8")
-                        self._span.set_attributes(get_output_attributes(output_text))
+                        self._final_response += output_text
```
nate-mar, line 120: if I understand correctly, each chunk of the stream is accumulated here along with all the guardrails for the entire stream in one span instead of across multiple spans?  i.e. will someone be able to see which chunk was associated with which guardrail check?

luke-moehlenbrock: Yeah that's correct, each chunk is accumulated and used as the output value of the root span for the trace. Each chunk of the final response also comes through as a separate event from the associated guardrail so there isn't really a way to associate each chunk with a guardrail check. Also in the case of "blocked" guardrails, we would never get the associated chunk in the first place.

nate-mar (approved):

Issue #2064, opened 2025-08-08
[instruemntation][js] Bedrock Agents - add support for consuming bedrock traces to build openinference trace tree and spans


Parker-Stafford:

sammyrw234:
Hey @satyadevai do we have an eta on this?

satyadevai:
@sammyrw234 I have completed the development and opened PR for review. If any review comments, I will work on Tomorrow and will try to fix them by Wednesday.

PR #2074, opened 2025-08-11
feat(bedrock-agent-runtime): Js bedrock agent instrumentation trace tree Building

satyadevai:
This package provides hierarchical trace collection and OpenInference-compatible span processing for Bedrock Agent Runtime. It is designed to aggregate streaming trace payloads, build a tree of trace nodes and spans, and extract rich semantic attributes for each span.

---

### Trace Collection & Storage

**Data Structures:**

- **AgentTraceNode**: Tree node representing an agent or model invocation. Each node can own multiple spans (`AgentChunkSpan`), child nodes, and loose chunks (e.g., failure traces). Nodes have a stable ID, semantic type, and a back-pointer to their parent.
- **AgentChunkSpan**: Logical work-unit grouping sequential chunks of the same type. Each span contains ordered raw chunks, a span type, child nodes (for nested agent collaborators), and a back-pointer to its parent node.
- **TraceNodeStack**: Lightweight LIFO stack to track the current node during a streaming trace collection.
- **AgentTraceAggregator**: Orchestrates the construction of the trace tree. It unwraps incoming trace payloads, routes chunks to the correct node/span, detects agent collaboration events, and maintains a map of nodes by ID and a set of seen IDs.

---

### Trace Processing

- **SpanCreator**: Recursively traverses the trace tree to create OpenTelemetry spans. For each node/span, it prepares attributes using `AttributeExtractor`, handles parent/child relationships, sets input/output attributes and metadata, and extracts timing information for accurate span start/end.

---

### Attribute Extraction

- **attribute-extractor.ts**: Centralizes logic for parsing raw trace payloads and extracting key fields (event type, chunk type, trace ID, input/output messages, model invocation parameters, token counts, metadata, etc.). Specialized extractors exist for action groups, code interpreter, knowledge base, and agent collaborators.
- **attribute-utils.ts**: Utility functions for normalizing and formatting attributes (input/output value extraction, MIME type normalization, LLM system/model/provider attributes, message and tool attributes, document attributes for retrieval events).
- **types.ts**: TypeScript interfaces for core trace concepts (Message, TokenCount, DocumentReference, etc.).

---

### Example Flow

1. **Trace Collection**
   - Incoming trace payloads are passed to `AgentTraceAggregator.collect()`.
   - The aggregator unwraps and routes each chunk, building/updating the trace tree.
2. **Trace Processing**
   - After collection, `SpanCreator.createSpans()` traverses the tree.
   - For each node/span, attributes are extracted and OpenTelemetry spans are created.
3. **Attribute Extraction**
   - `AttributeExtractor` and `attribute-utils` are used to parse and format attributes for each span, including semantic conventions for LLM, tool, agent, and retriever spans.

---

image
image
image
image
image
image
image

Closes https://github.com/Arize-ai/openinference/issues/2064


js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+        chunkType in (traceData[eventType] as Record<string, unknown>)
+      ) {
+        const chunkObj = (traceData[eventType] as Record<string, unknown>)[
+          chunkType
+        ];
+        if (chunkObj && typeof chunkObj === "object" && "traceId" in chunkObj) {
+          return (
+            ((chunkObj as Record<string, unknown>)["traceId"] as string) ??
+            "unknown-trace-id"
+          );
```
Parker-Stafford: this is unsafe and can be cleaned up abit with some helpers and typeguards so that you don't have to cast over and over again. Also the cast to Record<string,unknown> is unsafe so i would use our helper from openinference/core to do this `isStringKeyedObject`

for example

```typescript
function getObjectDataFromUnknown({data, key}: {data: unknown, key: string}) {
  if (!isStringKeyedObject(data) {
    return null
  }
  const maybeData = data[key] ?? null
  if (isStringKeyedObject(maybeData)) {
    return maybeData
  }

  return null;

}
extractTraceId(traceData: unkown): string {
  const eventType = this.getEventType(traceData);
  const eventData = getObjectDataFromUnknown({data: traceData, key: eventType})
  if (eventData == null) {
    return "unknownTraceId"
  }
  for (const chunkType of this.CHUNK_TYPES) {
   const chunkData = getObjectDataFromUnknown({data: eventData, key: chunkType})
   if (chunkData == null) {
     continue
   } 
   const maybeTraceId = chunkData["traceId"]
   if (typeof maybeTraceId === "string") {
    return maybeTraceId
   }
  }


 }

```

also if eventType is unknownTrace then we don't need to do this at all

Parker-Stafford: also is the string "unknown-trace-id" doing anything for us or should we just return null here?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+    "rationale",
+    "observation",
+  ] as const;
+
+  /**
+   * Return the first matching event type key discovered in {@link traceData}.
+   * Defaults to `'unknownTrace'` when not found.
+   */
+  static getEventType(traceData: Record<string, unknown>): string {
+    for (const eventType of this.EVENT_TYPES) {
```
Parker-Stafford: for this and the function below why is the first matching one good?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+  }
+
+  /**
+   * Extracts messages an object from a text.
+   * Parses input text into a list of Message objects.
+   */
+  public static getMessagesObject(text: string): Message[] {
+    const messages: Message[] = [];
+    try {
+      const inputMessages: ParsedInput = safeJsonParse(text);
```
Parker-Stafford: we already have a safe json parse in openinference core, you should use that also whatever this parsedInput type is its probably not safe to cast here like this, you need to typeguard to make sure it is of the type you expect

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+          const role = message.role || "";
+          if (message.content) {
+            const parsedContents = fixLooseJsonString(message.content) || [
+              message.content,
+            ];
+            for (const parsedContent of parsedContents) {
+              let messageContent: string = message.content;
+              if (typeof parsedContent === "object" && parsedContent !== null) {
+                if (parsedContent.type && parsedContent[parsedContent.type]) {
+                  messageContent = parsedContent[parsedContent.type] as string;
```
Parker-Stafford: this is not a safe cast, i would do something more akin to what i laid out above to clean these up, in general if you are going to say something is a type in typescript its best to determine that ahead of time. This whole function suffers because we unsafely asserted that json parsing text at the top was a type called ParsedInput which we have no business asserting, everything from there is possibly wrong. I would make functions to determine those types a typeguard looks  like this

```typescript
type MyType = {
text: string,
maybeNumber?: number
}
function isMyType(something: unknown): something is MyType {
return isStringKeyedObject(something) && typeof something.text === "string" && (something.maybeNumber === undefined || typeof something.maybeNumber === "number)
}

```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+   * (action group, code interpreter, knowledge base lookup, agent collaborator)
+   * to be set on the parent span.
+   */
+  public static getParentInputAttributesFromInvocationInput(
+    invocationInput: Record<string, unknown>,
+  ): Record<string, unknown> | undefined {
+    if (!invocationInput || typeof invocationInput !== "object") return {};
+
+    const actionGroup =
+      (invocationInput as Record<string, unknown>)[
```
Parker-Stafford: these casts are unnecessary because the input type enforces this must be the case

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+  /**
+   * Extracts parent input attributes from invocation input.
+   * Extracts input attributes from various types of invocation inputs
+   * (action group, code interpreter, knowledge base lookup, agent collaborator)
+   * to be set on the parent span.
+   */
+  public static getParentInputAttributesFromInvocationInput(
+    invocationInput: Record<string, unknown>,
+  ): Record<string, unknown> | undefined {
+    if (!invocationInput || typeof invocationInput !== "object") return {};
```
Parker-Stafford: this is unnecessary as the input type enforces this is the case

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+
+    const codeInterpreter =
+      (invocationInput as Record<string, unknown>)[
+        "codeInterpreterInvocationInput"
+      ] || {};
+    if (codeInterpreter && typeof codeInterpreter === "object") {
+      const inputValue =
+        (codeInterpreter as Record<string, unknown>)["code"] || "";
+      if (inputValue) {
+        return getInputAttributes(inputValue as string);
```
Parker-Stafford: unsafe

Parker-Stafford: all of the comments above go for the rest of the file, going to stope here but we will need to apply those checks everywhere

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-extractor.ts
```diff
@@ -0,0 +1,974 @@
...
+/**
+ * @module AttributeExtractor
+ * @description
+ * Utility helpers for pulling key fields (event type, trace-ID, chunk type)
+ * from raw Bedrock Agent Runtime trace payloads. Centralising this logic keeps
+ * parsing rules consistent across the code-base.
+ */
+export class AttributeExtractor {
+  /** Ordered a list of known event wrapper keys inside a trace object. */
+  private static readonly EVENT_TYPES = [
```
Parker-Stafford: in general while nothing absolutely wrong with putting all of this in a class, its a bit non standard and has a few drawbacks
- inconsistent with the rest of the code base (we don't use classes for utility functions and constants like this)
- these functions while somewhat related in scope are not actually related in any way to the internals of this class but are unnecessary coupled to it
- it makes it difficult to use them with our utilities from openinference core like withSafety which we should be using


I would just remove

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attributes.ts
```diff
@@ -0,0 +1,37 @@
+import { OpenInferenceSpanKind } from "@arizeai/openinference-semantic-conventions";
+
+/**
+ * Container for span attributes and metadata.
+ * Stores various attributes related to a span, including name, type, request/output attributes, metadata, and span kind.
+ */
+export class Attributes {
```
Parker-Stafford: same there is no reason this needs to be a class

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/collector/agent-chunk-span.ts
```diff
@@ -0,0 +1,56 @@
+import { AgentTraceNode } from "./agent-trace-node";
+
+/**
+ * Logical work-unit inside a node. A span groups sequential chunks that share
+ * the same spanType. Spans can be nested under nodes but never under each other.
+ */
+export class AgentChunkSpan {
```
Parker-Stafford: arguably more class like but also not entirely necessary/ could be achieved with some functions won't block on this one necessarily but i do think the ocnsistency is important

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+/**
+ * Normalize MIME type from text or object.
+ * Returns MimeType.JSON for JSON-like strings or objects, otherwise MimeType.TEXT.
+ * @param input Input string or object to check.
+ * @returns MimeType for the input.
+ */
+function normalizeMimeType(input: unknown): MimeType {
+  if (typeof input === "string") {
+    const trimmed = input.trim();
+    if (trimmed.startsWith("{") || trimmed.startsWith("[")) {
```
Parker-Stafford: i don't think this is right? we should just parse it then check if its an array or object.

For example: this "[INTERNAL] Name of my document" passes as json but its not

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+
+/**
+ * Extract input attributes from a text or object.
+ * Adds value and MIME type to the attributes.
+ * @param input Text or object to extract attributes from.
+ * @returns Attributes an object with input value and MIME type.
+ */
+export function getInputAttributes(input: unknown): Attributes {
+  const attributes: Attributes = {};
+  if (input === undefined || input === null) return attributes;
```
Parker-Stafford: can squash this check into input == null

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+ * @param messages - Array of input message objects.
+ * @returns Attributes an object for input messages.
+ */
+export function getLLMInputMessageAttributes(
+  messages: Message[] | undefined,
+): Attributes {
+  return {
+    ...llmMessagesAttributes(messages, "input"),
+  };
+}
```
Parker-Stafford: this funciton is not necessary

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+ * Extracts output message attributes for LLM tracing.
+ * Returns an Attributes object with output message attributes.
+ * @param messages - Array of output message objects.
+ * @returns Attributes an object for output messages.
+ */
+export function getLLMOutputMessageAttributes(
+  messages: Message[] | undefined,
+): Attributes {
+  return {
+    ...llmMessagesAttributes(messages, "output"),
```
Parker-Stafford: same

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+    attributes[SemanticConventions.TOOL_DESCRIPTION] = description;
+  }
+  return attributes;
+}
+
+export function generateUniqueTraceId(
+  eventType: string,
+  traceId: string,
+): string {
+  return `${eventType}_${traceId.split("-").slice(0, 5).join("-")}`;
```
Parker-Stafford: use named vs. positional arguments, it makes it less error prone to switch params around

also just curious why are we generating trace ids?


```suggestion
export function generateUniqueTraceId({eventType, traceId}: {
  eventType: string,
  traceId: string,
}): string {
  return `${eventType}_${traceId.split("-").slice(0, 5).join("-")}`;
```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attribute-utils.ts
```diff
@@ -0,0 +1,460 @@
...
+}
+
+/**
+ * Extracts provider attributes for LLM tracing.
+ * Accepts either an object with a 'value' property or a string provider name.
```
Parker-Stafford: this seems fine but why? might be nice to put that info here

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/collector/agent-trace-aggregator.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  /**
+   * Extracts trace data from nested Bedrock trace structure.
+   * @param obj - The object potentially containing wrapped trace data
+   * @returns The unwrapped trace data
+   */
+  private unwrapTrace(
+    obj: Record<string, object>,
+  ): Record<string, Record<string, unknown>> {
+    return (obj["trace"] as Record<string, Record<string, unknown>>) ?? obj;
```
Parker-Stafford: this is not a safe cast

js/packages/openinference-core/src/utils/index.ts
```diff
@@ -25,6 +25,10 @@ export function withSafety<T extends GenericFunction>({
   };
 }
 
+const unknownJSONParse = (...args: Parameters<typeof JSON.parse>): unknown => {
+  return JSON.parse(...args);
```
mikeldking, line 29: I think this is fine, but in general if we want to make it safe across all calls we can just change the type

```typescript
declare global {
  interface JSON {
    parse(text: string, reviver?: (this: any, key: string, value: any) => any): unknown;
  }
}

```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/attributeExtractionUtils.ts
```diff
@@ -0,0 +1,955 @@
...
+  TRACE_EVENT_TYPES,
+  TraceEventType,
+} from "./constants";
+import { isArrayOfObjectWithStringKeys } from "../utils/typeUtils";
+import { Attributes } from "@opentelemetry/api";
+import { isAttributeValue } from "@opentelemetry/core";
+
+/**
+ * Return the first matching event type key discovered in {@link traceData}.
+ * Defaults to `'unknownTrace'` when not found.
```
mikeldking: this is not true, it returns a null

mikeldking (approved):
Approving to unblock

PR #2077, opened 2025-08-11
fix: fix build for bedrock agent instrumentation

Parker-Stafford:
fixes issue where module cannot be found for bedrock agent runtime instrumentation

image

### DIFF
--- a/js/packages/openinference-instrumentation-bedrock-agent-runtime/tsconfig.esm.json
+++ b/js/packages/openinference-instrumentation-bedrock-agent-runtime/tsconfig.esm.json
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.base.esm.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "./dist/esm"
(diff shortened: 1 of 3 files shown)


mikeldking (approved):

js/packages/openinference-instrumentation-bedrock/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "./dist/src"
+    "outDir": "./dist"
   },
-  "include": ["src/**/*"]
+  "include": ["src/**/*.ts"]
```
mikeldking:
```suggestion
  "include": ["src/**/*"]
```
should be fine to include all?

Parker-Stafford: yeah should be, this is consistent with the other packages so was just converging on that but can remove

PR #2080, opened 2025-08-12
feat: openai support custom tool calls

davidgmonical:


python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_attributes/_responses_api.py
```diff
@@ -265,6 +265,23 @@ def _get_attributes_from_response_function_tool_call(
...
+        prefix: str = "",
+    ) -> Iterator[Tuple[str, AttributeValue]]:
+        if (call_id := obj.call_id) is not None:
+            yield f"{prefix}{ToolCallAttributes.TOOL_CALL_ID}", call_id
+        if (name := obj.name) is not None:
+            yield f"{prefix}{ToolCallAttributes.TOOL_CALL_FUNCTION_NAME}", name
+        if (input_data := obj.input) is not None:
+            yield (
+                f"{prefix}{ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}",
+                safe_json_dumps({"input": input_data}),
```
davidgmonical, line 282: Might open a conversation about adding `tool_call.function.input` as a potential semantic convention. Not sure we want to add it into the arguments like this

PR #2090, opened 2025-08-13
feat(Agno): Capture agent graph attributes 

ArnavJain23:


python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -173,33 +262,61 @@ def run_stream(
...
+        current_path = f"{parent_path}.{agent_name}" if parent_path else agent_name
+        node_id = _generate_node_id(current_path)
+
+        # Set current path in context before span creation so _agent_run_attributes can use it
+        current_path_ctx = context_api.set_value(
+            _AGNO_CURRENT_PATH_CONTEXT_KEY, current_path, context_api.get_current()
+        )
+        current_path_token = context_api.attach(current_path_ctx)
+
+        try:
```
caroger: is wrapping the context manager `with self._tracer.start_as_current_span` inside the new try-finally block a requirement to capture the agent graph attributes?

caroger: @RogerHYang  @axiomofjoy , does this pattern look right? they way context detaches are called in the finally clause below

ArnavJain23: Ok so did some testing and looks like I don't actually need to detach context for the proper parent-child relationships to work. I think the _AGNO_PARENT_PATH_CONTEXT_KEY, _AGNO_PARENT_NODE_CONTEXT_KEY, and _AGNO_CURRENT_PATH_CONTEXT_KEY take care of that.

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -74,20 +80,60 @@ def _strip_method_args(arguments: Mapping[str, Any]) -> dict[str, Any]:
     return {key: value for key, value in arguments.items() if key not in ("self", "cls")}
 
 
+def _generate_node_id(path: str) -> str:
```
RogerHYang: nit. unnecessary argument
```suggestion
def _generate_node_id() -> str:
```

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -128,33 +188,64 @@ def run(
...
-                span.set_attribute(OUTPUT_MIME_TYPE, JSON)
-                return run_response
+        # Build hierarchical path for node ID generation
+        parent_path = context_api.get_value(_AGNO_PARENT_PATH_CONTEXT_KEY) or ""
+        current_path = f"{parent_path}.{agent_name}" if parent_path else agent_name
+        node_id = _generate_node_id(current_path)
+
+        # Set current path and node ID in context before span creation
+        ctx = context_api.get_current()
+        ctx = context_api.set_value(_AGNO_CURRENT_NODE_CONTEXT_KEY, node_id, ctx)
```
RogerHYang: nit. current is the default
```suggestion
        ctx = context_api.set_value(_AGNO_CURRENT_NODE_CONTEXT_KEY, node_id)
```

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -128,33 +188,64 @@ def run(
...
+                    )
+                ),
+            ) as span:
+                # Set up context for team executions
+                if isinstance(agent, Team):
+                    # Set both node ID and path for children to use
+                    team_ctx = context_api.get_current()
+                    team_ctx = context_api.set_value(
+                        _AGNO_PARENT_NODE_CONTEXT_KEY, node_id, team_ctx
+                    )
```
RogerHYang: nit. is this redundant, since it's same value as line 198?
```suggestion
```

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -33,6 +35,10 @@
     ToolCallAttributes,
 )
 
+_AGNO_PARENT_NODE_CONTEXT_KEY = context_api.create_key("agno_parent_node_id")
+_AGNO_PARENT_PATH_CONTEXT_KEY = context_api.create_key("agno_parent_path")
+_AGNO_CURRENT_NODE_CONTEXT_KEY = context_api.create_key("agno_current_node_id")
```
RogerHYang: This variable is not needed: set the value directly on the span instead.
```suggestion
```

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -128,33 +187,64 @@ def run(
...
+                                **kwargs,
+                            ),
+                            **dict(_agent_run_attributes(agent)),
+                            **dict(get_attributes_from_context()),
+                        }
+                    )
+                ),
+            ) as span:
+                # Set up context for team executions
+                if isinstance(agent, Team):
```
RogerHYang: add a helper function for this logic

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py
```diff
@@ -74,20 +80,60 @@ def _strip_method_args(arguments: Mapping[str, Any]) -> dict[str, Any]:
...
     if agent.user_id:
         yield USER_ID, agent.user_id
 
     if isinstance(agent, Team):
+        # Use the node ID from context (generated in run() method)
+        if context_node_id and isinstance(context_node_id, str):
+            node_id = context_node_id
+
+        # Set graph attributes for team
+        yield GRAPH_NODE_ID, node_id
```
RogerHYang: `node_id` is also potentially undefined here, but his is moot if the value is set on the span directly instead

Issue #1864, opened 2025-07-14
[feature request] [mastra] session and metadata support


mikeldking:
Add support for mastra sessions and metadata.

tylerschoppe:
Sorry for the delay on adding some context here, I needed to wrap my head around a few things first and get my day job done!

The main thing here is utilizing the mastra threadId to be used as a sessionId in phoenix. I was able to accomplish this with some agent method wrapping to ensure all spans will use the threadId passed in as invocation options. A little hacky so I had a short discussion with some folks at mastra and opened up a [pr to add the threadId to traces](https://github.com/mastra-ai/mastra/pull/6273)! Now it will be easy to grab the threadId in the mastra span processor and add the sessionId semantic convention tag. With that alone spans will be grouped into nice agent invocation traces.

Mastra allows adding metadata via telemetry options on invocations so we can just pull from there.

Lastly we need to parse spans for user input and LLM output so in a session we get the nice conversation thread.

tylerschoppe:
I am happy to take on some or all of this, maybe with a little guidance on the last note. I've started getting fairly in the weeds here so I've got some good context!

mikeldking:
> I am happy to take on some or all of this, maybe with a little guidance on the last note. I've started getting fairly in the weeds here so I've got some good context!

@tylerschoppe thanks for digging into this end-to-end. That would be amazing if you could adjust here to pick up metadata and sessions as you probably have the perfect use-case. Thank you!

tylerschoppe:
Once the mastra pr moves forward I will pick this up

tylerschoppe:
Should be able to jump on the first parts of this in the next few days!

mikeldking:
> Should be able to jump on the first parts of this in the next few days!

Very exciting!

tylerschoppe:
I went to move the ticket into in progress but it appears I cannot.

I've got the sessions part working, just cleaning up the code. Metadata needs a little bit of thought and maybe some discussion here but I will keep them separate so things move and the reviews are broken up.

tylerschoppe:
For the metadata component, I was thinking something along the lines of piggybacking on to the telemetry metadata options that can be passed in to the agent stream/generate calls as so
```typescript
await weatherAgent.generate(
  'What is the weather in Tokyo?',
  {
    telemetry: {
      metadata: {
        userId: 'test-user-id',
      },
    },
  }
);
```

[...]

PR #2092, opened 2025-08-14
feat: mastra session support

tylerschoppe:
Resolves session component of https://github.com/Arize-ai/openinference/issues/1864 by adding automatic root span discovery for agent traces, mastra threadId to OpenInference sessionId extraction and user input/LLM output capture on child spans for top level conversation tracking.

**Description**
The implementation addresses the session requirements component of the feature request by leveraging [Mastra tracings threadId](https://github.com/mastra-ai/mastra/pull/6273) to group related agent invocations into cohesive sessions in Phoenix. The solution includes span analysis logic that automatically identifies root spans in agent workflows, extracts session context from Mastra threads, and captures user input/LLM output on child spans to enable conversation-style tracking in Phoenix.

The changes provide Mastra users with a much more rich experience in Phoenix out of the box regardless of configuration or use case.

**Key Changes**
- **Session Tracking**: Extracts Mastra threadId from span attributes and maps it to OpenInference sessionId semantic convention, enabling automatic grouping of related agent invocations. _Note: direct agent.generate.stream calls need to manage adding a threadId via args_ 
- **Root Span Discovery**: Implements intelligent span analysis to identify the root span of agent workflows and marks them with appropriate span kinds for Phoenix filtering. This was previously unsupported out of the box for routes outside of those prefixed with `post /agents`, leaving the Phoenix experience to be limited to just `All Spans`.
- **Input/Output Capture**: Captures user input and LLM output from child spans and propagates them to root spans for conversation thread visualization in Phoenix sessions view.

**Whats left**
Changes here do not address the metadata component of the issue. This is mainly due to keeping the PR small but there are also some implementation details that should be ironed out.

### DIFF
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -434,19 +434,19 @@ importers:
         specifier: workspace:*
         version: link:../openinference-vercel
       '@opentelemetry/api':
-        specifier: ^1.9.0
+        specifier: '>=1.9.0'
         version: 1.9.0
       '@opentelemetry/core':
-        specifier: ^2.0.1
+        specifier: '>=2.0.1'
         version: 2.0.1(@opentelemetry/api@1.9.0)
       '@opentelemetry/exporter-trace-otlp-proto':
-        specifier: ^0.201.1
+        specifier: '>=0.201.1'
         version: 0.201.1(@opentelemetry/api@1.9.0)
       '@opentelemetry/sdk-trace-base':
-        specifier: ^2.0.1
+        specifier: '>=2.0.1'
         version: 2.0.1(@opentelemetry/api@1.9.0)
       '@opentelemetry/semantic-conventions':
-        specifier: ^1.34.0
+        specifier: '>=1.34.0'
         version: 1.34.0
     devDependencies:
       vitest:


tylerschoppe:

js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts
```diff
@@ -75,23 +77,21 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
...
-        ...span,
-        // backwards compatibility with older versions of sdk-trace-base
-        instrumentationLibrary: {
-          name: "@arizeai/openinference-mastra",
-        },
-      });
-      // add OpenInference attributes to the span based on Mastra span attributes
-      addOpenInferenceAttributesToMastraSpan(span);
+      // Process span with all standard Mastra OpenInference attributes
+      processMastraSpanAttributes(span);
```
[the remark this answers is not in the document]
tylerschoppe: This is an interesting point but in practice does not prove to be true. Removing the filter did however reveal a small bug with sessions 🎉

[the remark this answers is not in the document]
tylerschoppe: Ah this does not appear to be a bug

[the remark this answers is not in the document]
tylerschoppe: This is an interesting point but is not true in practice. The top level `'post /api/agents'` span will be found as a root to agent spans and get the `AGENT` kind added

js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts
```diff
@@ -75,23 +77,21 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
...
       filteredSpans = filteredSpans.filter(this.spanFilter);
     }
+    // Add missing root spans for traces with agent operations.
+    // Consider batching for large sets of spans for performance and lower memory usage.
+    const missingRoots = findMissingAgentRootSpans(spans, filteredSpans);
+    if (missingRoots.length > 0) {
+      filteredSpans.push(...missingRoots);
+    }
+    // Add user input and agent output to root spans for Phoenix session I/O
+    addIOToRootSpans(filteredSpans);
```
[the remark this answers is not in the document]
tylerschoppe: This is not correct as `processMastraSpanAttributes` will only be run on spans found to be roots. Those found roots will be added to the spans list if not already present where IO will then be added

mikeldking:
hey @tylerschoppe I'm super excited to dive deep into this PR. I'd love to sit down on Monday to digest this when I have the time to give it the attention it deserves. Hope that works for you. LMK if you would like a review sooner.

Appreciate you contributing upstream. Amazing work.

tylerschoppe:
@mikeldking that is no problem, I've got more than enough other tasks to work on! If you have any questions feel free to DM me in the community slack so we can chat or huddle!

js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts
```diff
@@ -75,23 +77,21 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
...
     });
     if (this.spanFilter) {
       filteredSpans = filteredSpans.filter(this.spanFilter);
     }
+    // Add missing root spans for traces with agent operations.
+    // Consider batching for large sets of spans for performance and lower memory usage.
+    const missingRoots = findMissingAgentRootSpans(spans, filteredSpans);
+    if (missingRoots.length > 0) {
+      filteredSpans.push(...missingRoots);
+    }
```
mikeldking: given that the filter is meant to be user-defined, I'm planning on hoisting this to happen somewhat before the filtering.

tylerschoppe: That seems to be the safer choice. My rationale was to perform the root span search on a smaller set of spans.

tylerschoppe: Also, feel free to request changes, I am happy to address things in my own work :)

mikeldking: Okay great @tylerschoppe - I'll do a pure review then. Sorry was just trying to be efficient with smaller issues.

tylerschoppe: you're helping me out, no need to be sorry!

mikeldking: @tylerschoppe I think it's not quite clear to me the cases when we think the "root" ends up being missing. I think from glancing at the code it probably makes more sense for me to get the agent spans properly figured out in `processMastraSpanAttributes` - this way you get all the data in place to pass into the filter if needed.

js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts
mikeldking: is it possible to add IO during the span processing? I think I see some nested search so maybe traversing the tree is necessary - however I do think it should happen pre-filtering so that everything is available before the filter.

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -3,12 +3,22 @@ import {
...
+const AGENT_PATTERNS = {
+  GET_RECENT_MESSAGE: "agent.getMostRecentUserMessage",
+  GET_RECENT_MESSAGE_RESULT: "agent.getMostRecentUserMessage.result",
+  STREAM: "agent.stream",
+  STREAM_ARGUMENT: "agent.stream.argument.0",
+  GENERATE: "agent.generate",
+  GENERATE_ARGUMENT: "agent.generate.argument.0",
+  MASTRA_PREFIX: "mastra.",
+  AGENT_PREFIX: "agent.",
+};
```
mikeldking: could use a doctring here explaining the rationale for these prefixes being agents

mikeldking: is it worth for all of these just to be prefixes?

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+
+/**
+ * Gets the span ID from a ReadableSpan.
+ *
+ * @param span - The span to get the ID from.
+ * @returns The span ID or "unknown" if not available.
+ */
+export const getSpanId = (span: ReadableSpan): string => {
+  return span.spanContext?.()?.spanId || "unknown";
+};
```
mikeldking:
```suggestion
export const getSpanId = (span: ReadableSpan): string | null => {
  return span.spanContext?.()?.spanId || null;
};
```

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+
+/**
+ * Gets the trace ID from a ReadableSpan.
+ *
+ * @param span - The span to get the trace ID from.
+ * @returns The trace ID or "unknown" if not available.
+ */
+export const getTraceId = (span: ReadableSpan): string => {
+  return span.spanContext?.()?.traceId || "unknown";
+};
```
mikeldking:
```suggestion
/**
 * Gets the trace ID from a ReadableSpan.
 *
 * @param span - The span to get the trace ID from.
 * @returns The trace ID or "unknown" if not available.
 */
export const getTraceId = (span: ReadableSpan): string | null => {
  return span.spanContext?.()?.traceId | null
};
```
Feels unsafe to have these fallbacks here. Could lead to errors downstram if we start using "unknown" as traceIds

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+ * @param spans - Array of spans to search through.
+ * @returns The extracted user input or undefined if not found.
+ */
+export const extractMastraUserInput = (
+  spans: ReadableSpan[],
+): string | undefined => {
+  // Look for the most recent user message from agent.getMostRecentUserMessage.result
+  for (const span of spans) {
+    if (span.name === AGENT_PATTERNS.GET_RECENT_MESSAGE) {
+      const result = span.attributes[AGENT_PATTERNS.GET_RECENT_MESSAGE_RESULT];
```
mikeldking: the pattern CONST seems to contain span names and attribute values. The mixture of these two seems particularly difficult to figure out why they need to be grouped in a object. I think it might make more sense to just use constants?

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+          if (messageData.content && typeof messageData.content === "string") {
+            return messageData.content;
+          }
+        } catch (error) {
+          diag.warn("Failed to parse agent.getMostRecentUserMessage.result", {
+            error: error instanceof Error ? error.message : String(error),
+            rawResult: result,
+            spanId: span.spanContext?.()?.spanId,
+          });
+        }
```
mikeldking: We have a utility to do this in openinfernce-core - https://github.com/Arize-ai/openinference/blob/7047abd270ab3c52e3dcfa21f7fb4175a0d52fe0/js/packages/openinference-core/src/utils/index.ts#L34

I would leverage that here

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+      }
+    }
+  }
+
+  // Look for direct user input from agent.generate.argument.0
+  for (const span of spans) {
+    if (span.name === AGENT_PATTERNS.GENERATE) {
```
mikeldking: this if statement seems to be forming a switch statement that could be in one loop as the loop above

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+      const argument = span.attributes[AGENT_PATTERNS.GENERATE_ARGUMENT];
+      if (typeof argument === "string") {
+        try {
+          // Try to parse as JSON first in case it's a complex structure
+          const parsed = JSON.parse(argument);
+          if (typeof parsed === "string") {
+            return parsed;
+          }
+        } catch {
```
mikeldking: Same comment as above, probably just use openinference-core's safeJSONParse

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+          // If JSON parsing fails, treat as plain string (remove quotes if present)
+          return argument.replace(/^"(.*)"$/, "$1");
```
mikeldking: What's the rationale for removing the quotes? Feels like we want to keep it as is?

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+            // Find the last user message
+            for (let i = messages.length - 1; i >= 0; i--) {
+              const message = messages[i];
+              if (message.role === "user" && message.content) {
+                return typeof message.content === "string"
+                  ? message.content
+                  : JSON.stringify(message.content);
+              }
+            }
+          }
```
mikeldking: I would put this in a single type-safe utils that grabs the last user message. That way it will be easier to test and extend

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+            error: error instanceof Error ? error.message : String(error),
+            rawArgument: argument,
+            spanId: span.spanContext?.()?.spanId,
+          });
+        }
+      }
+    }
+  }
+
+  return undefined;
```
mikeldking:
```suggestion
```
Techniclly not needed

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+/**
+ * Extracts agent output from Mastra spans for session I/O.
+ *
+ * Looks for output.value in any span (typically ai.streamText or similar).
+ *
+ * @param spans - Array of spans to search through.
+ * @returns The extracted agent output or undefined if not found.
+ */
+export const extractMastraAgentOutput = (
+  spans: ReadableSpan[],
```
mikeldking, line 298: how do we know all these spans belong to a single trace?

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+ * to any root spans that don't already have I/O attributes set.
+ *
+ * @param spans - Array of spans to process.
+ */
+export const addIOToRootSpans = (spans: ReadableSpan[]): void => {
+  const rootSpans = spans.filter(
+    (span) => span.parentSpanContext === undefined,
+  );
+
+  if (rootSpans.length === 0) return;
```
mikeldking:
```suggestion
  if (!rootSpans.length) return;
```

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+
+  const userInput = extractMastraUserInput(spans);
+  const agentOutput = extractMastraAgentOutput(spans);
```
mikeldking: I don't see the spans being grouped to the rootSpan's traceId here?

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+
+  // Add input and output to root spans
+  for (const rootSpan of rootSpans) {
+    if (userInput && !rootSpan.attributes[SemanticConventions.INPUT_VALUE]) {
+      rootSpan.attributes[SemanticConventions.INPUT_VALUE] = userInput;
+      rootSpan.attributes[SemanticConventions.INPUT_MIME_TYPE] = "text/plain";
```
mikeldking: this is implicit so not needed

js/packages/openinference-mastra/src/attributes.ts
```diff
@@ -66,11 +110,241 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+ *
+ * @param allSpans - All spans before filtering.
+ * @param filteredSpans - Spans after filtering.
+ * @returns Array of missing root spans that should be included.
+ */
+export const findMissingAgentRootSpans = (
+  allSpans: ReadableSpan[],
+  filteredSpans: ReadableSpan[],
+): ReadableSpan[] => {
+  const filteredSpanIds = new Set(filteredSpans.map((s) => getSpanId(s)));
```
mikeldking: I would just perform everything before filtering and nix this code path as I think you might end up with a strange topology when re-adding spans that were asked to be removed in the filter.

tylerschoppe:
> @tylerschoppe took a quick scan. I wonder if it would be possilbe to create an example directory where we could quickly test an agent? Would be extremely helpful in testing.
> 
> Thanks again for the PR. Happy to sync on this if my comments don't make sense.

After a skim I think I should be able to address everything async with maybe a question or two in comments. Appreciate a thorough pass.

An example directory would be nice, hints at improving documentation for this integration so users can have a better understanding of options given their use case

js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts
tylerschoppe: Good question. I should I have called this out in more detail than just

> This was previously unsupported out of the box for routes outside of those prefixed with post /agents, leaving the Phoenix experience to be limited to just All Spans.

**Current state**

If a filter is not used or the filter is fairly permissive, the root will be found without issue other than not being marked as an agent kind. This case also causes a lot of likely unwanted roots to be captured from mastras internal processes, server handling non agent requests such as options, token refresh, etc.

When the `isOpenInferenceSpan` filter is used, which I would imagine would be a majority of the use cases as its in the docs so people getting started will copy/paste, the roots will be filtered out as they do not align with open inference span kinds. The exception to this is spans whose names start with the prefixes:

```javascript
const MASTRA_AGENT_SPAN_NAME_PREFIXES = [
  "agent",
  "mastra.getAgent",
  "post /api/agents",
];
````
While this covers a fair share of use cases, ultimately finding root spans that would otherwise be filtered out, it makes a sizable assumption on how the user has implemented their application. In my case I was using an endpoint that does not start with the prefix `post /api/agents` which left all of my spans orphaned. IMO this assumption of use is too restrictive.

**Proposed solution**
I took a bit of a heavy handed/complex/opinionated approach by choosing to look for roots even if they have been previously filtered out with the caveat that they are only related to agent calls, leaving out the unwanted roots mentioned above. The rationale behind this decision is that it will provide _any_ user with root span capture out of the box without making assumptions or needing configuration. Since the root span is a fundamental building block to sessions functionality in phoenix I felt this decision was well grounded.

**Alternatives**
* wider set of prefixes
* user defined prefixes
* probably something I am not thinking of with my head in the implementation details

Comment (tylerschoppe) [reply js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts]: I think we can find the agent root spans earlier in the process, where it was before, so filtering can still be applied if needed. I was likely overemphasizing the performance impact of looking for roots after the fact.

I think there is something nice about not allowing the user to shoot themselves in the root by filtering out a root span and loosing functionality in phoenix but if we find it earlier and they still filter it out, at that point its on them.

File: js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts (hunk 2 above)
Comment (tylerschoppe) [reply js/packages/openinference-mastra/src/OpenInferenceTraceExporter.ts]: This is a good principle to abide by. Similar to the root span logic I think I was putting too much thought into performance impact

File: js/packages/openinference-mastra/src/attributes.ts (hunk 14 above)
Comment (tylerschoppe) [reply js/packages/openinference-mastra/src/attributes.ts:298]: Great callout. A bit narrow minded in my approach here

File: js/packages/openinference-mastra/src/attributes.ts [hunk 19]
```diff
@@ -66,11 +108,302 @@ const getOpenInferenceSpanKindFromMastraSpan = (
...
+
+/**
+ * Gets the trace ID from a ReadableSpan.
+ *
+ * @param span - The span to get the trace ID from.
+ * @returns The trace ID or "unknown" if not available.
+ */
+export const getTraceId = (span: ReadableSpan): string | null => {
+  return span.spanContext?.()?.traceId || null;
+};
```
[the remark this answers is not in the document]
tylerschoppe: the comment on `'post /api/agents'` is invalid and I believe the comment on `span.spanContext` is too

tylerschoppe:
@mikeldking thanks for the great feedback, feeling much better about the shape of things now!

A bit short on time to come up with full working examples in the project directory but to provide a little insight into my testing I've used:
- calls to agent.generate/stream. without passing threadId/resourceId or using memory you will not have sessions of course but the root span is still captured
```javascript
await agent.generate('what is the capital of france?');
await agent.stream('what is the weather going to be like today?, { threadId: 'abc-123', resourceId: 'weatherAgent' });
```
- invoking agents via the api endpoints created from the dev server. I just use the UI for this but you could curl. _runId, threadId and resourceId are handled by the dev UI_
```bash
curl -X POST "http://localhost:4111/api/agents/weatherAgent/stream" \
  -H "Content-Type: application/json" \
  -d '{
    "messages": [
      {
        "role": "user",
        "content": "what is the weather in Tokyo?"
      }
    ],
    "runId": "weatherAgent",
    "runtimeContext": {},
    "threadId": "72766221-59f6-4577-97ab-c20b4036262b",
    "resourceId": "weatherAgent"
  }'
```
- Using copilotkit. I use this in my org and it is fairly popular within the mastra community. https://mastra.ai/en/docs/frameworks/agentic-uis/copilotkit#configure-mastra-server

tylerschoppe:
Just wanted to poke here to see if there is anything else I can do to help move this along. No worries if there hasn't been time to give it a look, just trying to keep the strong momentum!

mikeldking:
> Just wanted to poke here to see if there is anything else I can do to help move this along. No worries if there hasn't been time to give it a look, just trying to keep the strong momentum!

Hey @tylerschoppe thanks for adjusting to the PR comments. I'm gonna pull down this branch and make some adjustments as I review. Hope that's okay. Will aim to get this over the line. Thanks for getting it this far.

tylerschoppe:
> > Just wanted to poke here to see if there is anything else I can do to help move this along. No worries if there hasn't been time to give it a look, just trying to keep the strong momentum!
> 
> Hey @tylerschoppe thanks for adjusting to the PR comments. I'm gonna pull down this branch and make some adjustments as I review. Hope that's okay. Will aim to get this over the line. Thanks for getting it this far.

If that works for you its fine by me. Hoping there isn't too much remaining.

js/packages/openinference-mastra/test/__snapshots__/OpenInferenceTraceExporter.test.ts.export.json
```diff
@@ -773,7 +773,7 @@
           "net.peer.ip": "::1",
           "net.peer.port": 51258,
           "net.transport": "ip_tcp",
-          "openinference.span.kind": "AGENT",
+          "openinference.span.kind": undefined,
```
mikeldking, line 776: I wonder if this causes spans to be detached strangely. Haven't double checked.

mikeldking:
@tylerschoppe I did a big sweep but admittedly don't have time today to test. There's some pre-releases on this branch if you want to try it out. If it looks good on your end I can merge and get this published.

Thanks again for the PR!

js/packages/openinference-mastra/src/constants.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Constants for Mastra agent span names and attributes
+ */
+
+/**
+ * Prefix for Mastra agent span names. Examples: "agent.generate", "mastra.getAgent"
+ */
+export const MASTRA_AGENT_SPAN_NAME_PREFIXES = ["agent", "mastra.getAgent"];
```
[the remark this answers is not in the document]
mikeldking: this is true, and not sure if it causes orphaning in a way we don't want.

mikeldking:
I gave this a spin this morning. I see some quirks in some traces but overall I feel it's much improved. It feels like the treadId might only show up in the telemetry later once a tread is created so this might miss the first interaction. Even so I think this is worth shipping and we can maybe look into how to tie together the sessions more cleanly in a follow-up.

https://github.com/user-attachments/assets/2ddc9132-662c-41d2-9128-24e730505d18

tylerschoppe:
> I gave this a spin this morning. I see some quirks in some traces but overall I feel it's much improved. It feels like the treadId might only show up in the telemetry later once a tread is created so this might miss the first interaction. Even so I think this is worth shipping and we can maybe look into how to tie together the sessions more cleanly in a follow-up.

I'm wondering if the threadId appearing later is something that is out of our hands due to upstream behavior. I haven't seen this myself but I can pull all the latest mastra packages and poke around some. In general I think this is a good place to add some general documentation just to make users aware of how things work. Ideally they don't need to care because its drop in and go but its nice to have an understanding :)

tylerschoppe:
> > I gave this a spin this morning. I see some quirks in some traces but overall I feel it's much improved. It feels like the treadId might only show up in the telemetry later once a tread is created so this might miss the first interaction. Even so I think this is worth shipping and we can maybe look into how to tie together the sessions more cleanly in a follow-up.
> 
> I'm wondering if the threadId appearing later is something that is out of our hands due to upstream behavior. I haven't seen this myself but I can pull all the latest mastra packages and poke around some. In general I think this is a good place to add some general documentation just to make users aware of how things work. Ideally they don't need to care because its drop in and go but its nice to have an understanding :)

Following up as I finally got some time to poke around here. I was able to reproduce this behavior in instances where we export the traces from a turn in more than one call to `super.export`. We've got a few options but I think it is important to find the right balance of how much we should reasonably do in this component.

PR #2094, opened 2025-08-15
feat(litellm): add llm provider attributes

### LINKED ISSUE #1659: [litellm] Add provider to litellm
parse out litellm provider and add it to the attributes

axiomofjoy:
resolves #1659

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
@@ -58,36 +58,28 @@ def _set_span_attribute(span: trace_api.Span, name: str, value: AttributeValue) 
         span.set_attribute(name, value)
 
 
-def _get_provider_from_model(model: str) -> Optional[OpenInferenceLLMProviderValues]:
-    """
-    Extract the LLM provider from the model name using LiteLLM's get_llm_provider utility.
-    Maps LiteLLM providers to OpenInference provider values.
-    Returns the OpenInference provider enum value or None if not supported.
-    """
-    # Mapping from LiteLLM provider names to OpenInference provider values
-    LITELLM_TO_OPENINFERENCE_PROVIDER = {
-        "openai": OpenInferenceLLMProviderValues.OPENAI,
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
```diff
@@ -55,6 +58,38 @@ def _set_span_attribute(span: trace_api.Span, name: str, value: AttributeValue)
...
+        "vertex_ai": OpenInferenceLLMProviderValues.GOOGLE,
+        "gemini": OpenInferenceLLMProviderValues.GOOGLE,
+        "azure": OpenInferenceLLMProviderValues.AZURE,
+        "bedrock": OpenInferenceLLMProviderValues.AWS,
+        "xai": OpenInferenceLLMProviderValues.XAI,
+        "deepseek": OpenInferenceLLMProviderValues.DEEPSEEK,
+    }
+
+    try:
+        provider_info = litellm.get_llm_provider(model=model)  # type: ignore
```
GeLi2001: wondering why we need to type ignore this line, does pytest want us to type the `provider_info`?

axiomofjoy: This is actually `mypy` that is running type checks rather than `pytest`.

It looks like `mypy` is complaining about an implicit import, meaning that the `get_llm_provider` is in the `litellm.*` namespace but isn't explicitly exported.

```
ci-litellm: commands[2]> mypy .
src/openinference/instrumentation/litellm/__init__.py:67: error: Module "litellm" does not explicitly export attribute "get_llm_provider"  [attr-defined]
```

It looks like the [LiteLLM docs](https://docs.litellm.ai/docs/providers/custom_llm_server#quick-start) actually recommend this import path, so this is a situation where it is probably okay to use a type ignore, although we should scope the ignore to the particular type error described in the message above using `# type: ignore[attr-defined]`.

GeLi2001: ahh I see, thanks for explaining!

PR #2103, opened 2025-08-19
feat: bedrock js provider

### LINKED ISSUE #1778: [providers] add providers where necessary
- [ ] [provider] openai under azure gets different provider
- [ ] [provider] bedrock gets provider `aws`

GeLi2001:
resolves #1778

### DIFF
--- a/js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
+++ b/js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
@@ -71,7 +71,11 @@ export function getProviderFromModelId(modelId: string): LLMProvider {
   return LLMProvider.AWS;
 }
 
-export function setBasicSpanAttributes(span: Span, llm_system: LLMSystem, llm_provider: LLMProvider) {
+export function setBasicSpanAttributes(
+  span: Span,
+  llm_system: LLMSystem,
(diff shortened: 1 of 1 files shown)


js/packages/openinference-instrumentation-bedrock/test/instrumentation.test.ts
```diff
@@ -577,7 +577,7 @@ The key things are to dress for the warm temperatures and have layers you can",
 
 Honeybees can recognize human faces.",
   "llm.output_messages.0.message.role": "assistant",
-  "llm.provider": "aws",
+  "llm.provider": "anthropic",
```
Parker-Stafford, line 580: i think this was actually correct before, this is bedrock so the provider is aws the system is anthropic

Parker-Stafford: looks like provider is where the models are hosted which in this case will always be aws

Image

GeLi2001: ah I see, should I just close this pr as the provider should just be static 'aws'

GeLi2001: closing as bedrock js should always be aws provider

Issue #2018, opened 2025-07-31
[feature request] JS Bedrock - add instrumentation for retrieve_and_generate and retrieve


sammyrw234:
A continuation of #1776 

**Describe the solution you'd like**

The linked ticket includes support for InvokeModel and InvokeAgent but it would be great to also instrument [retrieve](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agent-runtime/client/retrieve.html) and [retrieve_and_generate](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agent-runtime/client/retrieve_and_generate.html) calls as well, similar to the Python version.

PR #2110, opened 2025-08-20
feat(bedrock-agent-runtime): Rag Instrumentation

satyadevai:


js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @param input Bedrock RetrieveAndGenerateCommandInput containing configuration and session info.
+ * @returns OpenTelemetry attributes for invocation parameters.
+ */
+export function extractRagInvocationParams(
+  input: bedrockAgentRunTime.RetrieveAndGenerateCommandInput,
+): Attributes {
+  const invocationParams: Attributes = {};
+  if (input?.retrieveAndGenerateConfiguration) {
+    invocationParams["retrieveAndGenerateConfiguration"] = JSON.stringify(
```
Parker-Stafford: use safe stringify

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+      input.retrieveAndGenerateConfiguration,
+    );
+  }
+  if (input?.sessionConfiguration) {
+    invocationParams["sessionConfiguration"] = JSON.stringify(
+      input.sessionConfiguration,
+    );
+  }
```
Parker-Stafford: same for this one

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+      citation?.retrievedReferences,
+    )
+      ? citation.retrievedReferences
+      : [];
+    for (const document of documents) {
+      const ragDocument: DocumentReference = {
+        metadata: document.metadata,
+        content: document.content,
+        location:
+          getObjectDataFromUnknown({ data: document, key: "location" }) ?? {},
```
Parker-Stafford: i think we should just not include this if it doesn't exist, same with in `getDocumentAttributes` called down below

Image

just don't add these keys if they aren't there the empty object is just ocnfusing i think

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+  let index = 0;
+  let attributes: Attributes = {};
+  citations = Array.isArray(citations) ? citations : [];
+  for (const citation of citations) {
+    const documents: bedrockAgentRunTime.RetrievedReference[] = Array.isArray(
+      citation?.retrievedReferences,
+    )
+      ? citation.retrievedReferences
+      : [];
+    for (const document of documents) {
```
Parker-Stafford, line 107: nit: instead of manually controlling the index (line 86 and 109) just doa documents.forEach((docuemnt, index) => //...

less error prone

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+ * This function processes the response from a Bedrock retrieve_and_generate operation
+ * and extracts both the generated output and the retrieved document citations.
+ * It handles the complex structure of RAG responses that include both generation
+ * results and retrieval citations.
+ *
+ * @param response Response object from the retrieve_and_generate operation, containing:
+ *                 - output: Generated text output
+ *                 - citations: List of citations with retrieved references
+ *                 - sessionId: Session identifier (optional)
+ * @returns A dictionary of OpenTelemetry attributes containing:
```
Parker-Stafford: Tiny nit

```suggestion
 * @returns OpenTelemetry attributes containing:
```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+      location:
+        getObjectDataFromUnknown({ data: document, key: "location" }) ?? {},
+    };
+    attributes = {
+      ...attributes,
+      ...getDocumentAttributes(index, ragDocument),
+    };
+    index += 1;
+  }
+  return attributes;
```
Parker-Stafford: whats the difference between this and the function on line 83 above, seems very similar if not the same, might be able to just reuse the same one and change the input outside of the function...

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/requestAttributes.ts
```diff
@@ -32,6 +47,61 @@ function extractBaseRequestAttributes(command: InvokeAgentCommand): Attributes {
...
+ * @param command The RetrieveAndGenerateStreamCommand or RetrieveAndGenerateCommand instance.
+ * @returns Attributes for OpenTelemetry span instrumentation.
+ */
+function extractRagBaseRequestAttributes(
+  command: RetrieveAndGenerateStreamCommand | RetrieveAndGenerateCommand,
+): Attributes {
+  return {
+    [SemanticConventions.OPENINFERENCE_SPAN_KIND]:
+      OpenInferenceSpanKind.RETRIEVER,
+    [SemanticConventions.LLM_SYSTEM]: LLMSystem.AMAZON,
```
Parker-Stafford: is this correct? i think it should be based on teh model right? anthropic if anthropic etc. etc.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/requestAttributes.ts
```diff
@@ -32,6 +47,61 @@ function extractBaseRequestAttributes(command: InvokeAgentCommand): Attributes {
...
+  if (input?.nextToken) {
+    invocationParams["nextToken"] = input.nextToken;
+  }
+  if (input?.retrievalConfiguration) {
+    invocationParams["retrievalConfiguration"] = input.retrievalConfiguration;
+  }
+  return {
+    [SemanticConventions.OPENINFERENCE_SPAN_KIND]:
+      OpenInferenceSpanKind.RETRIEVER,
+    [SemanticConventions.LLM_SYSTEM]: LLMSystem.AMAZON,
```
Parker-Stafford: same here also maybe we don't even need this on these retrieval spans? might only need it on llm spans

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callbackHandler.ts
```diff
@@ -67,3 +70,61 @@ export class CallbackHandler {
...
+  private readonly span: Span;
+
+  /**
+   * Callback handler for processing agent responses and traces.
+   * @param span {Span} - The OpenTelemetry span to associate with the response.
+   */
+  constructor(span: Span) {
+    this.span = span;
+    this.output = "";
+    this.citation = [];
```
Parker-Stafford: nit:


```suggestion
    this.citations = [];
```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callbackHandler.ts
```diff
@@ -67,3 +70,61 @@ export class CallbackHandler {
...
+    if (error instanceof Error) {
+      this.span.recordException(error);
+    }
+    const message =
+      error &&
+      typeof error === "object" &&
+      "message" in error &&
+      typeof error.message === "string"
+        ? error.message
+        : String(error);
```
Parker-Stafford: safely stringify here instead of String()


```suggestion
        : safelyJSONStringify(error);
```

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callbackHandler.ts
```diff
@@ -67,3 +70,61 @@ export class CallbackHandler {
...
+  }
+  handleCitation(citation: Record<string, unknown>) {
+    if (citation && Object.keys(citation).length > 0) {
+      this.citation.push(citation);
+    }
+  }
+  onComplete(): void {
+    this.span.setAttributes({
+      ...getOutputAttributes(this.output),
+      ...extractRetrievedReferencesAttributes(this.citation),
```
Parker-Stafford: why not use safe versions of these as well?

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -210,6 +248,171 @@ export class BedrockAgentInstrumentation extends InstrumentationBase<Instrumenta
...
+        attributes: safelyExtractBaseRagAttributes(command) ?? undefined,
+      },
+    );
+    const result = original.apply(client, args);
+    return result
+      .then(
+        (
+          response: bedrockAgentRunTime.RetrieveAndGenerateStreamCommandOutput,
+        ) => {
+          const callback = new RagCallbackHandler(span);
```
Parker-Stafford, line 272: just for my understanign are you only using this callback pattern when it is a streaming use case? Just trying to understand why we have a callback handler which extracts attributes, then multiple functions which also extract attributes and when we would use one over the other

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -210,6 +248,171 @@ export class BedrockAgentInstrumentation extends InstrumentationBase<Instrumenta
...
+    client: unknown,
+  ) {
+    const span = this.oiTracer.startSpan("bedrock.retrieve", {
+      kind: SpanKind.INTERNAL,
+      attributes: safelyExtractBaseRetrieveAttributes(command) ?? undefined,
+    });
+    const result = original.apply(client, args);
+    return result
+      .then((response: bedrockAgentRunTime.RetrieveCommandOutput) => {
+        span.setAttributes(extractBedrockRetrieveResponseAttributes(response));
```
Parker-Stafford: this needs to be safe too otherwise we might wind up throwing an instrumentation error to client, same goes for everything in this file as well, we should always catch and mayb e warn with diag.warn but ultimately swallow errors from attribute extraction, otherwise we might throw it to the user in their app somethign like

```typescript
try {
      span.setAttributes(extractBedrockRetrieveResponseAttributes(response));
      span.setStatus({ code: SpanStatusCode.OK });
    } catch (attributeError) {
      // Handle any errors in attribute extraction separately
      span.recordException(attributeError);
    } finally {
      span.end();
    }

```

js/packages/openinference-instrumentation-bedrock-agent-runtime/test/integration.instrumentation.rag.test.ts
```diff
@@ -0,0 +1,335 @@
...
+        const metadata = attrs[metadataKey];
+        expect(metadata).toContain('"customDocumentLocation":{"id":"2222"}');
+        delete attrs[metadataKey];
+      }
+    }
+
+    // Validate invocation parameters
+    const invocation = attrs["llm.invocation_parameters"];
+    expect(invocation).toContain('"knowledgeBaseId":"SSGLURQ9A5"');
+    delete attrs["llm.invocation_parameters"];
```
Parker-Stafford: could we instead just do a expect(Object.keys(attrs).sort()).toEqual(expectedkeys.sort()) instead of deleting keys from the attributes, feels a bit weird

Parker-Stafford (commented):
nice work ty!

gabe0912:
for clarity, this PR resolves: https://github.com/Arize-ai/openinference/issues/2018

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
satyadevai: We cannot do this, as we are running a nested loop and each iteration increments the index of the documents in our instrumentation. If we remove this logic, the document indices will no longer align with the actual iteration order, which would cause inconsistencies in how the documents are tracked and reported.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
satyadevai: The inputs differ for each method. RetrieveCommand only handles document retrieval, while RetrieveAndGenerate performs both retrieval and generation. I’ve now removed the duplicate code and improved the implementation.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/requestAttributes.ts
satyadevai: My intention was to use AMAZON since all knowledge bases originate from Amazon. I’ve now removed this attribute.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
satyadevai: Yes, the callback pattern is primarily used for streaming cases. In streaming, the callback handler progressively collects data as chunks are received and then extracts attributes once the response is complete. For non-streaming scenarios, we extract attributes directly within the respective functions since the full response is available.

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @param input Bedrock RetrieveAndGenerateCommandInput containing configuration and session info.
+ * @returns OpenTelemetry attributes for invocation parameters.
+ */
+export function extractRagInvocationParams(
+  input: bedrockAgentRunTime.RetrieveAndGenerateCommandInput,
+): Attributes {
+  const invocationParams: Attributes = {};
+  if (input?.retrieveAndGenerateConfiguration) {
+    invocationParams["retrieveAndGenerateConfiguration"] =
+      safelyJSONStringify(input.retrieveAndGenerateConfiguration) || "";
```
Parker-Stafford: just fallback to undefiend here, we shouldn't set empty string attributes for things that don't exist or are unparseable, same with line 32 below

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/attributes/ragAttributeExtractionUtils.ts
Parker-Stafford: ahh i see i missed the outer citations loop

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/callbackHandler.ts
```diff
@@ -67,3 +70,73 @@ export class CallbackHandler {
...
+    try {
+      this.span.setAttributes({
+        ...getOutputAttributes(this.output),
+        ...extractRetrievedReferencesAttributes(this.citations),
+      });
+      this.span.setStatus({ code: SpanStatusCode.OK });
+    } catch (error: unknown) {
+      if (error instanceof Error) {
+        this.span.recordException(error);
+      }
```
Parker-Stafford: i guess this pattern feels a little weird to me, i'm not sure what to do so we can keep it like this. But having an error on the span because our instrumentation was not able to get attributes feels a bit off. It feels like we should diag.warn or debug but the span should be okay since the unit of work the span was tracking (the actual call to aws) was fine we just couldn't capture attributes

Parker-Stafford: yeah confirmed with mikyo we should not error the span when instrumentation fails, just do diag.debug with the error and end the span with status code okay

Parker-Stafford: same goes for everywhere this is happening in this package if you could take a look and patch

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
Parker-Stafford: thanks ty

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -210,6 +252,200 @@ export class BedrockAgentInstrumentation extends InstrumentationBase<Instrumenta
...
+                errorMessage = err instanceof Error ? err.message : err;
+              } else {
+                errorMessage =
+                  "Unknown error occurred in Openinference Bedrock Agent Instrumentation while trying to process rag stream response.";
+              }
+              span.setStatus({
+                code: SpanStatusCode.ERROR,
+                message: errorMessage,
+              });
+              span.end();
```
Parker-Stafford: same here as the comment above, if there is no error in the actual call (handled on 320 below we shouldn't error the span

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -210,6 +252,200 @@ export class BedrockAgentInstrumentation extends InstrumentationBase<Instrumenta
...
+            span.setAttributes(extractBedrockRagResponseAttributes(response));
+            span.setStatus({ code: SpanStatusCode.OK });
+          } catch (error: unknown) {
+            if (error instanceof Error) {
+              span.recordException(error);
+            }
+            const errorMessage =
+              error instanceof Error
+                ? error.message
+                : (safelyJSONStringify(error) ?? undefined);
```
Parker-Stafford: same

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts
```diff
@@ -210,6 +252,200 @@ export class BedrockAgentInstrumentation extends InstrumentationBase<Instrumenta
...
+          span.setAttributes(
+            extractBedrockRetrieveResponseAttributes(response),
+          );
+          span.setStatus({ code: SpanStatusCode.OK });
+        } catch (error: unknown) {
+          if (error instanceof Error) {
+            span.recordException(error);
+          }
+          const errorMessage =
+            error instanceof Error
```
Parker-Stafford: same

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/streamUtils.ts
```diff
@@ -25,3 +27,48 @@ export function interceptAgentResponse<
...
+          if (item?.citation) {
+            callback.handleCitation(
+              getObjectDataFromUnknown({ data: item, key: "citation" }) || {},
+            );
+          }
+          yield item;
+        }
+        callback.onComplete();
+      } catch (err) {
+        callback.onError(err);
```
Parker-Stafford, line 86: same here, should make sure to patch the callback errors, also i think we need to yield items to not consume the original stream no matter what so we should always yield then try catch the callback handler stuff only

Parker-Stafford (commented):
lets do one more pass

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/streamUtils.ts
```diff
@@ -10,14 +13,75 @@ export function interceptAgentResponse<
...
+          } catch (err: unknown) {
+            diag.debug("Error in interceptRagResponse stream", err);
+          }
+          yield item;
+        }
+        try {
+          callback.onComplete();
+        } catch (err: unknown) {
+          diag.debug("Error in interceptRagResponse:", err);
+        }
```
Parker-Stafford, line 84: nit: could we just wrap all the instrumentation stuff in one try catch, we lose a slight amount of granularity of error messages (should be captured in teh error anyways) for a decent bit of readability

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/streamUtils.ts
```diff
@@ -10,14 +13,75 @@ export function interceptAgentResponse<
...
+            diag.debug("Error in interceptAgentResponse Stream:", err);
           }
           yield item;
         }
-        callback.onComplete();
+        try {
+          callback.onComplete();
+        } catch (err: unknown) {
+          diag.debug("Error in interceptAgentResponse:", err);
+        }
```
Parker-Stafford, line 31: same thing here, thanks for pathcing this but lets just do one try catch for all instruemntation stuff here

js/packages/openinference-instrumentation-bedrock-agent-runtime/src/streamUtils.ts
satyadevai: I don’t think we can merge them into a single try/catch. The first try/catch is scoped to handling exceptions while building the citations list for each event/chunk, while the second one is responsible for processing all the citations. If we combine them, we risk losing the ability to recover gracefully — e.g. failing to deliver either individual citations or completing the loop properly. 

Keeping them separate ensures both steps are handled reliably.

Parker-Stafford: okay, i see the onComplete also sets output attributes, we can keep them separate maybe there is a way to make it more readable in the future but can punt fo rnow

Parker-Stafford: we could for example wrap the handle citation and also both of the function calls in handle output in withSafety allowing them to gracefully fail and have specific error messages for each. This would allow us to:
1. always get the most possible attributes on the span
2. have specific error messages for debugging
3. avoid multiple nested try catch blocks

satyadevai: Each stream item can contain an output, a citation, or both. That’s why handling them together within one try block makes sense — after processing each iteration, we still need to yield the chunk back to the caller.

At the end of the loop, we take the accumulated data (citations + output attributes) and upload it to the SpanProcessor. This separation ensures we don’t block streaming progress while still reliably persisting everything once the stream completes.

Parker-Stafford: ah yeah i see the yield now, ideally like our other instrumentors we would split the stream [bedrock](https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-bedrock/src/attributes/invoke-model-streaming-response-attributes.ts#L481) [openai](https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-openai/src/instrumentation.ts#L334) to not  couple the yielding of the stream for the user to isntrumentation as well. That way we can cleanly do the three thigns i said above

satyadevai: yeah, I understood now, we can make this change later, its not blocking our release.

Issue #2029, opened 2025-08-04
[feature request] JS Bedrock - Add instrumentation for converse_stream


sammyrw234:
A continuation of #1776 and #2018 

Describe the solution you'd like

The linked tickets above includes support for InvokeModel InvokeAgent, retrieve and retrieve and generate but it would be great to also instrument [invoke_model_with_response_stream](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model_with_response_stream.html), [converse](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html) and [converse_stream](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse_stream.html) as well

caroger:
[invoke_model_with_response_stream](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model_with_response_stream.html), [converse](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html) should already be supported and covered by #1851

nate-mar:
@Aemrox will get this out this week. Thanks!

PR #2111, opened 2025-08-20
feat(js-bedrock-client): Add converse-stream support

Aemrox:
resolves: #2029 

Add converse-stream support - in the end this is much simpler than InvokeModel stream. There are better types from the client, as well as the client normalizing the calls across different model providers. 

Come things I did not cover explicitly that exist as input parameters - from https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/bedrock-runtime/command/ConverseStreamCommand/ 
- additionalModelRequestFields
- additionalModelResponseFieldPaths
- guardrailConfig
- performanceConfig

These will show up as a part of the full call in input.value, but will not have explicit fields for them, as there are no semantic conventions that exist. 

Screenshots below
Screenshot 2025-08-20 at 1 27 40 PM
Screenshot 2025-08-20 at 1 27 34 PM
Screenshot 2025-08-20 at 1 27 29 PM
Screenshot 2025-08-20 at 1 27 22 PM
Screenshot 2025-08-20 at 1 27 14 PM

### DIFF
--- a/js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
+++ b/js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
@@ -113,7 +113,7 @@ function appendToolInputChunk(
 
   const parsed = safelyJSONParse(tool.partialJsonInput);
   if (parsed != null && isObjectWithStringKeys(parsed)) {
-    tool.input = parsed 
+    tool.input = parsed;
   }
 }
 
--- a/js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
@@ -40,6 +40,7 @@ import {
   getSystemFromModelId,
   setBasicSpanAttributes,
 } from "./attributes/attribute-helpers";
+import { isPromise } from "util/types";
 
 const MODULE_NAME = "@aws-sdk/client-bedrock-runtime";
 const INSTRUMENTATION_NAME = "@arizeai/openinference-instrumentation-bedrock";
@@ -290,13 +291,23 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
     extractInvokeModelRequestAttributes({ span, command, system });
 
     try {
-      const result = original.apply(
-        client,
-        args,
-      ) as unknown as Promise<InvokeModelResponse>;
+      const result = original.apply(client, args);
 
-      // AWS SDK v3 send() method always returns a Promise
-      return result
+      // AWS SDK v3 send() method should always return a Promise
+      if (!isPromise(result)) {
+        diag.warn(
+          "Expected Promise from AWS SDK send method, got:",
+          typeof result,
+        );
+        span.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: "Unexpected return type from AWS SDK",
+        });
+        span.end();
+        return result as Promise<InvokeModelResponse>;
+      }
+
+      return (result as Promise<InvokeModelResponse>)
         .then((response: InvokeModelResponse) => {
           extractInvokeModelResponseAttributes({
             span,
@@ -366,11 +377,23 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
 
     // Execute AWS SDK call and handle stream splitting outside error boundaries
     // This ensures the user stream is ALWAYS returned, regardless of instrumentation failures
-    const result = original.apply(client, args) as unknown as Promise<{
-      body: AsyncIterable<unknown>;
-    }>;
+    const result = original.apply(client, args);
 
-    return result
+    // Validate that we got a Promise as expected
+    if (!isPromise(result)) {
+      diag.warn(
+        "Expected Promise from AWS SDK send method, got:",
+        typeof result,
+      );
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: "Unexpected return type from AWS SDK",
+      });
+      span.end();
+      return result as Promise<{ body: AsyncIterable<unknown> }>;
+    }
+
+    return (result as Promise<{ body: AsyncIterable<unknown> }>)
       .then((response: { body: AsyncIterable<unknown> }) => {
         // Guard against missing response body - return original response
         if (!response.body) {
@@ -414,7 +437,9 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
               });
           }
         } else {
-          diag.debug("No instrumentation stream available, ending span cleanly");
+          diag.debug(
+            "No instrumentation stream available, ending span cleanly",
+          );
           span.setStatus({ code: SpanStatusCode.OK });
           span.end();
         }
@@ -463,13 +488,23 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
     extractConverseRequestAttributes({ span, command });
 
     try {
-      const result = original.apply(
-        client,
-        args,
-      ) as unknown as Promise<ConverseResponse>;
+      const result = original.apply(client, args);
 
-      // AWS SDK v3 send() method always returns a Promise
-      return result
+      // AWS SDK v3 send() method should always return a Promise
+      if (!isPromise(result)) {
+        diag.warn(
+          "Expected Promise from AWS SDK send method, got:",
+          typeof result,
+        );
+        span.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: "Unexpected return type from AWS SDK",
+        });
+        span.end();
+        return result as Promise<ConverseResponse>;
+      }
+
+      return (result as Promise<ConverseResponse>)
         .then((response: ConverseResponse) => {
           extractConverseResponseAttributes({ span, response });
           span.setStatus({ code: SpanStatusCode.OK });
@@ -534,11 +569,23 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
 
     // Execute AWS SDK call and handle stream splitting outside error boundaries
     // This ensures the user stream is ALWAYS returned, regardless of instrumentation failures
-    const result = original.apply(client, args) as unknown as Promise<{
-      stream: AsyncIterable<unknown>;
-    }>;
+    const result = original.apply(client, args);
 
-    return result
+    // Validate that we got a Promise as expected
+    if (!isPromise(result)) {
+      diag.warn(
+        "Expected Promise from AWS SDK send method, got:",
+        typeof result,
+      );
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: "Unexpected return type from AWS SDK",
+      });
+      span.end();
+      return result as Promise<{ stream: AsyncIterable<unknown> }>;
+    }
+
+    return (result as Promise<{ stream: AsyncIterable<unknown> }>)
       .then((response: { stream: AsyncIterable<unknown> }) => {
         // Guard against missing response stream - return original response
         if (!response.stream) {
@@ -576,7 +623,9 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
               span.end();
             });
         } else {
-          diag.debug("No instrumentation stream available, ending span cleanly");
+          diag.debug(
+            "No instrumentation stream available, ending span cleanly",
+          );
           span.setStatus({ code: SpanStatusCode.OK });
           span.end();
         }


js/packages/openinference-instrumentation-bedrock/scripts/validate-converse-stream-phoenix.ts
```diff
@@ -0,0 +1,851 @@
+#!/usr/bin/env tsx
+
+/**
```
Parker-Stafford: nice, maybe just remove phoenix from the name of this? i don't know probably doesn't matter

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -127,8 +127,39 @@ export function getAttributesFromMessageContent(
...
+      let base64: string;
+      if (content.image.source.bytes instanceof Uint8Array) {
+        // Live execution: convert Uint8Array to base64
+        base64 = Buffer.from(content.image.source.bytes).toString("base64");
+      } else if (Buffer.isBuffer(content.image.source.bytes)) {
+        // Direct Buffer: convert to base64
+        base64 = (content.image.source.bytes as Buffer).toString("base64");
+      } else if (typeof content.image.source.bytes === "string") {
+        // Nock playback: already a base64 string
+        base64 = content.image.source.bytes as string;
```
Parker-Stafford: even if this only ever happens with nock i would remove this comment since it feels weird to have testing infrastructure referenced in teh actual package

also shouldn't need to cast here? but maybe typescript isn't smart enough, could use typeguards but no biggy

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -127,8 +127,39 @@ export function getAttributesFromMessageContent(
...
+      } else if (
+        typeof content.image.source.bytes === "object" &&
+        content.image.source.bytes !== null &&
+        "type" in content.image.source.bytes &&
+        (content.image.source.bytes as { type: string; data: number[] })
+          .type === "Buffer" &&
+        "data" in content.image.source.bytes &&
+        Array.isArray(
+          (content.image.source.bytes as { type: string; data: number[] }).data,
+        )
```
Parker-Stafford: okay here defintiely use a typeguard for this, too complicated to be inline

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Parker-Stafford: would also remove the other self evident comments

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -127,8 +127,39 @@ export function getAttributesFromMessageContent(
...
+      ) {
+        // Buffer object format: convert data array to Buffer then to base64
+        base64 = Buffer.from(
+          (content.image.source.bytes as { type: string; data: number[] }).data,
+        ).toString("base64");
+      } else {
+        // Fallback: try to convert as-is
+        base64 = Buffer.from(content.image.source.bytes as Uint8Array).toString(
+          "base64",
+        );
```
Parker-Stafford: is this whole thing try caught? otherwise this is probably not safe

Parker-Stafford: if buffer.from only takes in a u8intarray, then there is probably no reason to do this, should just throw maybe or diag.warn and exit since this will always fail. if there are things it can take in that you haven't captured above then we can do this maybe

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -139,6 +170,9 @@ export function getAttributesFromMessageContent(
         media_type: mimeType,
       });
     }
+    // Add format attribute for image content
+    attributes[`${SemanticConventions.MESSAGE_CONTENT_IMAGE}.format`] =
+      content.image.format;
   }
```
Parker-Stafford, line 165: fine with adding this although it doesn't look like its in the openinference spec (no need to remove)

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -159,31 +193,42 @@ export function getAttributesFromMessage(message: Message): Attributes {
...
+      let toolCallIndex = 0;
+
+      for (const [index, content] of message.content.entries()) {
+        // Process content as our custom types for attribute extraction
+        const contentAttributes = getAttributesFromMessageContent(content);
+        for (const [key, value] of Object.entries(contentAttributes)) {
+          attributes[
+            `${SemanticConventions.MESSAGE_CONTENTS}.${index}.${key}`
+          ] = value as AttributeValue;
+        }
```
Parker-Stafford: not safe to as Attributevalue here, `getAttributesFromMessageContent` should return attribute values

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -159,31 +193,42 @@ export function getAttributesFromMessage(message: Message): Attributes {
...
+        if (isConverseToolUseContent(content)) {
+          const toolCallPrefix = `${SemanticConventions.MESSAGE_TOOL_CALLS}.${toolCallIndex}`;
+          attributes[`${toolCallPrefix}.${SemanticConventions.TOOL_CALL_ID}`] =
+            content.toolUse.toolUseId;
+          attributes[
+            `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_NAME}`
+          ] = content.toolUse.name;
+          attributes[
+            `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}`
+          ] = JSON.stringify(content.toolUse.input);
```
Parker-Stafford: we have a safeJsonStringify in core you can use here

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+} from "../types/bedrock-types";
+
+/**
+ * Processes Converse stream chunks and updates processing state
+ * Handles messageStart, contentBlockDelta, toolUse events, and metadata with proper accumulation
+ *
+ * @param data The parsed Converse stream event data
+ * @param state Current processing state to update
+ * @returns Updated processing state
+ */
```
Parker-Stafford: one note on how these comments work in js, this will be attached to the function directly below it, although it looks like it was intended for the file itself, i would just write individual ones for each function rather than file level ones

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+function startToolCall(
+  state: ConverseStreamProcessingState,
+  {
+    id,
+    name,
+    contentBlockIndex,
+  }: { id: string; name: string; contentBlockIndex?: number },
+) {
+  if (contentBlockIndex !== undefined) {
+    state.toolUseIdByIndex ??= {};
```
Parker-Stafford, line 76: whoa, i don't think i've ever seen this nullish coalescing =, cool

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+  const resolveId = () => {
+    if (id) return id;
+    if (
+      contentBlockIndex !== undefined &&
+      state.toolUseIdByIndex?.[contentBlockIndex]
+    ) {
+      return state.toolUseIdByIndex[contentBlockIndex];
+    }
+  };
+  const targetId = resolveId();
```
Parker-Stafford: making this function inside of this other function feels a little strange to me would probably just define this function outside of this other one or not wrap this

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+  if (!targetId) return;
+  const tool =
+    state.toolCalls.find((t) => t.id === targetId) ??
+    state.toolCalls[state.toolCalls.length - 1];
```
Parker-Stafford: this doesn't seem like a reasonable fallback, if we can find it i don't think we should go to the last element in the array?

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+  if (!tool) return;
+
+  tool.partialJsonInput = (tool.partialJsonInput ?? "") + chunk;
+
+  try {
+    const parsed = JSON.parse(tool.partialJsonInput);
+    tool.input = parsed as Record<string, unknown>;
```
Parker-Stafford: can use safelyParseJson here

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+    case "toolUseInputChunk":
+      appendToolInputChunk(state, ev);
+      return;
+    case "metadata":
+      state.usage = {
+        inputTokens: ev.usage.inputTokens ?? state.usage.inputTokens,
+        outputTokens: ev.usage.outputTokens ?? state.usage.outputTokens,
+        totalTokens: ev.usage.totalTokens ?? state.usage.totalTokens,
+      };
+      return;
```
Parker-Stafford, line 156: hmm my first instinct is to maake this swithc exhaustive like:

```suggestion
    case "metadata":
      state.usage = {
        inputTokens: ev.usage.inputTokens ?? state.usage.inputTokens,
        outputTokens: ev.usage.outputTokens ?? state.usage.outputTokens,
        totalTokens: ev.usage.totalTokens ?? state.usage.totalTokens,
      };
      return;
      default:
         assertUnreachable(ev.kind)
```

but i guess we can't rely on kind probably, the nic ething about exhaustive is that we know we cover all the kinds but it will blow up if they chang ethe types on us, also if kind is just a string its not possible. so this is fine

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+    usage: normalizedUsage,
+    streaming: true,
+    ...(stopReason && { stop_reason: stopReason }),
+  };
+
+  // Set output value as JSON (matching converse response behavior)
+  setSpanAttribute(
+    span,
+    SemanticConventions.OUTPUT_VALUE,
+    JSON.stringify(outputValue),
```
Parker-Stafford: safe stringify

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+  );
+  setSpanAttribute(
+    span,
+    SemanticConventions.OUTPUT_MIME_TYPE,
+    "application/json",
```
Parker-Stafford: use enum from semantic conventions package

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+    );
+    setSpanAttribute(
+      span,
+      `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_NAME}`,
+      toolCall.name,
+    );
+    setSpanAttribute(
+      span,
+      `${toolCallPrefix}.${SemanticConventions.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}`,
+      JSON.stringify(toolCall.input),
```
Parker-Stafford: same

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+    const state: ConverseStreamProcessingState = {
+      outputText: "",
+      toolCalls: [],
+      usage: {},
+    };
+
+    for await (const chunk of stream) {
+      // ConverseStream returns events directly as objects, no need to decode bytes
+      if (isValidConverseStreamEventData(chunk)) {
+        const data: ConverseStreamEventData = chunk;
```
Parker-Stafford: nit but shouldnt need to do this?

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,291 @@
...
+ */
+export const consumeConverseStreamChunks = withSafety({
+  fn: async ({
+    stream,
+    span,
+  }: {
+    stream: AsyncIterable<unknown>;
+    span: Span;
+  }): Promise<void> => {
+    // Initialize processing state for Converse streaming
```
Parker-Stafford: remove these types of comments

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -282,3 +370,98 @@ export function isConverseToolResultContent(
     typeof content.toolResult.toolUseId === "string"
   );
 }
+
+// Type guards for Converse streaming types
+export function isValidConverseStreamEventData(
+  data: unknown,
+): data is ConverseStreamEventData {
+  return isObjectWithStringKeys(data);
+}
```
Parker-Stafford: oof with the shape of the `ConversStreamEventData` interface and the looseness of Record<string, unknown> what `isObjectWithStringKeys` actually checks for this feels pretty sketchy, we have to be very vigilant in anything down stream that consumes this or else it will blow up, anyways we can improve this a bit?

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -58,6 +66,80 @@ export type InvokeModelRequestBody = Record<string, unknown>;
 export type InvokeModelResponseBody = Record<string, unknown>;
 export type InvocationParameters = Record<string, unknown>;
 
+// Converse Stream event data types using AWS SDK types where possible
+// This is a discriminated union of all possible streaming event types
```
Parker-Stafford: this doesn't seem right, this is the union of all possible things in one interface, the discriminated union is your normalized type down below, also if you update this comment to be a doc comment /** */ it will show up in ide on hover

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -282,3 +370,98 @@ export function isConverseToolResultContent(
...
 }
+
+// Type guards for Converse streaming types
+export function isValidConverseStreamEventData(
+  data: unknown,
+): data is ConverseStreamEventData {
+  return isObjectWithStringKeys(data);
+}
+
+export type NormalizedConverseStreamEvent =
```
Parker-Stafford, line 427: this is the discriminated union see comment above

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -282,3 +370,98 @@ export function isConverseToolResultContent(
...
+  | {
+      kind: "metadata";
+      usage: {
+        inputTokens?: number;
+        outputTokens?: number;
+        totalTokens?: number;
+      };
+    };
+
+export function toNormalizedConverseStreamEvent(
```
Parker-Stafford, line 456: this helps with the sketchyness but as a step forward with the sketchyness of the typeguard above, maybe we could just make that typeguard say this is a Record<string,unknown> (what it checks for now but not what it says the type is)

then use this funciton to safely normalize everything wdyt, it will make this function quite a bit more beastly but much safer

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -209,6 +212,14 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
               );
             }
 
+            if (command?.constructor?.name === "ConverseStreamCommand") {
```
Parker-Stafford: constructor .name is not a very reliable check for a variety of reasons can we do an instance of check here instead

Parker-Stafford: you will probably have to import the actual class to o instance of which may give some warnings about order of import for instrumentation so you should add a manually instrement method as well so that you can instrument no matter what, its too hard to control import order

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -472,4 +483,103 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
+    const modelId = command.input.modelId;
+    const system = getSystemFromModelId(modelId ?? "");
+    setBasicSpanAttributes(span, system);
+
+    const contextAttributes = getAttributesFromContext(context.active());
+    span.setAttributes(contextAttributes);
+
+    extractConverseRequestAttributes({
+      span,
+      command: command as unknown as ConverseCommand,
```
Parker-Stafford: shouldn't need to do this? can you make this take in a union of ConversCommand | ConversStreamCommand? if this cast is working then the function should be able to take in either no problem

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -472,4 +483,103 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
+    });
+
+    // Execute AWS SDK call and handle stream splitting outside error boundaries
+    // This ensures the user stream is ALWAYS returned, regardless of instrumentation failures
+    const result = original.apply(client, [command]) as Promise<{
```
Parker-Stafford: i think we eed this to take in ...args if the user passes in something other than the command we will swallow their arguments and not pass them to the original function, we need to do this anywhere we are calling original.apply. We can't assume the we know the args. this will also forc eyou to have better types for your original. Actually this s already broken in this instrumentation and needs to be fixed everywhere: here is how we did it for agents

https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-bedrock-agent-runtime/src/instrumentation.ts#L126-L139

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: Yup! meant to do this later but forgot

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: just gonna remove this - I think I'm being too defensive - AWS is pretty explicit about uint8array types, I'll include buffer to be safe, but I'll remove the weird object format and string. I think I was over-complicating because I was worried about the nock mock, but it runs fine without

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: simplified as per above

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
Aemrox: good callout

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
Aemrox: thank GPT-5 - had to look it up when it suggested it, used to use something very similar in Ruby though

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
Aemrox: You sure you don't want to use the fancy coalescing bit again? This is super readable to me 

`const targetId =
  id ?? (contentBlockIndex !== undefined ? state.toolUseIdByIndex?.[contentBlockIndex] : undefined);`
  
  /s 
  
  I'll pop it in another function, good call out

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
Aemrox: yeah, you're right - seems sketchy. will delete

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
Aemrox: Maybe loudly blowing up has an appeal? I dunno, I have to learn to think differently with instrumentation and this critical path nonsense. Unfortunately, kind is a string though - I did build toNormalizedConverseStreamEvent to be exhaustive according to the docs.

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
Aemrox: It's tough, if I go too far I end up duplicating the code in toNormalizedConverseStreamEvent - maybe there is something fancy I can do by attempting to cast it as I normalize it? Since I'm not that familiar, I'm attempting a middle ground of just checking for the first pieces of the message as a secondary check

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
Aemrox: yup should have read all your comments first - tell me what you think of my approach above, but I'm ready to change it to make this even more exhaustive and gnarly if you think it needs it.

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: Took a swing at copying the bedrock agent one, but I'm gonna leave this comment unresolved as it's important and I'm not 100% I did it right. Tests pass though

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -106,6 +107,27 @@ export function aggregateMessages(
   return [...aggregated, ...messages];
 }
 
+/**
+ * Safely converts supported byte-like inputs to base64.
+ * Handles: Uint8Array, Buffer, base64 string passthrough, and {type:'Buffer', data:number[]}.
```
Parker-Stafford: nice, tis comment looks slightly wrong though no string passthrough, probably fine to expand the input type to include string and do tha tthough if that is possible from bedrock, if not this looks good

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
```diff
@@ -0,0 +1,331 @@
...
+  });
+  if (!targetId) return;
+  const tool = state.toolCalls.find((t) => t.id === targetId);
+  if (!tool) return;
+
+  tool.partialJsonInput = (tool.partialJsonInput ?? "") + chunk;
+
+  const parsed = safelyJSONParse(tool.partialJsonInput);
+  if (parsed != null && isObjectWithStringKeys(parsed)) {
+    tool.input = parsed as Record<string, unknown>;
```
Parker-Stafford: nit: shouldn't have to cast here, line 114 shoulkd be smart enough

js/packages/openinference-instrumentation-bedrock/src/attributes/converse-streaming-response-attributes.ts
Parker-Stafford: okay so actually you are the creator of this normalized event type, so we should make the kind not a string and make it a string literal and we should do an exhaustive siwtch here so if someone adds a kind to that union, this will say "hey you need to do something with that kind here"

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -282,3 +394,124 @@ export function isConverseToolResultContent(
...
+    "type" in obj
+  );
+}
+
+/**
+ * Discriminated union of normalized Converse streaming events.
+ * Produced by toNormalizedConverseStreamEvent for safe downstream handling.
+ */
+export type NormalizedConverseStreamEvent =
+  | { kind: "messageStart" }
```
Parker-Stafford, line 428: re teh comment on the exhaustive switch on kind, we have control over the kind so we should do an exhaustive switch, typescript should understand these kind types as string literals

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
```diff
@@ -282,3 +394,124 @@ export function isConverseToolResultContent(
...
+ * Ensures we have an object and at least one known event marker.
+ * Exhaustive discrimination happens in toNormalizedConverseStreamEvent.
+ */
+export function isValidConverseStreamEventData(
+  data: unknown,
+): data is ConverseStreamEventData {
+  if (!isObjectWithStringKeys(data)) {
+    return false;
+  }
+  const obj = data as Record<string, unknown>;
```
Parker-Stafford: shouldn't need to do this

js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
Parker-Stafford: i think that can be good as long as we are defensive in our withSafety / try catch stuff, nice

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -178,39 +188,50 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
-        (original: any) => {
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          return function patchedSend(this: unknown, command: any) {
-            if (command?.constructor?.name === "InvokeModelCommand") {
+        (original: SendMethod) => {
+          return function patchedSend(
+            this: BedrockClient,
+            ...args: Parameters<SendMethod>
+          ) {
+            const [command] = args as [unknown, ...unknown[]];
```
Parker-Stafford: just const command = args[0]

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -472,4 +492,102 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
+        }
+
+        const { instrumentationStream, userStream } = safelySplitStream({
+          originalStream: response.stream,
+        });
+
+        // Start background instrumentation processing (non-blocking)
+        // Only the instrumentation is wrapped in error handling - the user stream is protected
+        if (instrumentationStream) {
+          // @ts-expect-error - instrumentationStream is guaranteed non-null by the if check above
```
Parker-Stafford, line 608: hmm? this seems off to me, as seem weird typescript is confused here

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -472,4 +492,102 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
+              span.setStatus({
+                code: SpanStatusCode.ERROR,
+                message: error.message,
+              });
+              span.end();
+            });
+        } else {
+          // No instrumentation stream available, end span cleanly
+          span.setStatus({ code: SpanStatusCode.OK });
+          span.end();
```
Parker-Stafford: maybe warn or debug here? probably not necessary but could be nice, debug is probaby better than warn

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -472,4 +492,102 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
...
+    extractConverseRequestAttributes({
+      span,
+      command: command,
+    });
+
+    // Execute AWS SDK call and handle stream splitting outside error boundaries
+    // This ensures the user stream is ALWAYS returned, regardless of instrumentation failures
+    const result = original.apply(client, args) as unknown as Promise<{
+      stream: AsyncIterable<unknown>;
+    }>;
```
Parker-Stafford: i'm trying to think of a better way to do this than the as unknown as promise, does ts actually not mind that? in either case we could probably do a typecheck, if it passes do below if not just return the result

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
```diff
@@ -439,9 +458,10 @@ export class BedrockInstrumentation extends InstrumentationBase<BedrockModuleExp
     extractConverseRequestAttributes({ span, command });
 
     try {
-      const result = original.apply(client, [
-        command,
-      ]) as Promise<ConverseResponse>;
+      const result = original.apply(
+        client,
+        args,
+      ) as unknown as Promise<ConverseResponse>;
```
Parker-Stafford: comment below goes for all of these result casts its a pain but a little guard might make sense, if its too crazy it may be fine

Parker-Stafford (approved):

js/packages/openinference-instrumentation-bedrock/src/instrumentation.ts
Aemrox: It was super weird, the error I was excepting was "Object might be null" referring to instrumentationStream but the if block should stop that. I even did an explicit null check `if (instrumentationStream != null)` I still get the error. Every code assist spun out on this, but I think it boils down to the fact that safelySplitStream allows instrumentationStream to return null. Somehow that overrides this guard. I can change stuff around, but ultimately, it would come down to building an empty iterator and returning that.

Aemrox: Ok nope, tried that, still gives me an error - things I've tried: 
- explicit null check in the if block
- Altering safelysplitstream to return an empty iterator so it's never an optional return + null check 
- Create a type check for AsyncIterable and use that in the if block 
- alter safelysplitstream to return an explicit false value for instrumentation stream when it fails
- Convert safelysplitstream from try/catch to withSafety

Aemrox: Ok, I found something that works but it's hideous - it boils down to Typescript thinking anything wrapped with withSafety could return null (i.e. consumeBedrockChunks and consumeConverseStreamChunks). so if I wrap the function like so: 

```
          const stream: AsyncIterable<unknown> = instrumentationStream;
          const consumeConverse = (consumeConverseStreamChunks ?? (async () => {})) as (args: {
            stream: AsyncIterable<unknown>;
            span: Span;
          }) => Promise<void>;
          consumeConverse({
            stream,
            span,
          })
```

It works...

Parker-Stafford: fixed, this was cause the promise from consume stream was maybe null not instrumentation stream itself

js/packages/openinference-instrumentation-bedrock/src/attributes/attribute-helpers.ts
```diff
@@ -106,6 +107,31 @@ export function aggregateMessages(
...
+    if (bytes instanceof Uint8Array) {
+      return Buffer.from(bytes).toString("base64");
+    }
+    diag.warn("Unsupported image bytes type encountered");
+    return undefined;
+  },
+  onError: (error) => {
+    diag.warn("Failed to convert image bytes to base64", error as Error);
+  },
+});
```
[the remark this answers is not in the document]
Parker-Stafford: the type is pretty weird here from aws so cursor might be irght here but we can keep it like this for now and if we see issues patch

PR #2137, opened 2025-08-27
docs: converge spec to Python OpenAI practice

codefromthecrypt:
I updated [Envoy AI Gateway to use OpenInference for its tracing](https://aigateway.envoyproxy.io/docs/0.3/capabilities/observability/tracing/) format. As a part of that, I captured the behavior of the latest Python OpenAI instrumentation and ported it to Go (the gateway process is Go).

While there are a lot of changes here, I think it is best to update as it will make it easier for the next person or agent who tries to implement this with the spec. I'm very flexible in the amount of change, so if you want things removed or re-organized, just say it!

If you want to read about how this works, consider [this](https://tetrate.io/blog/opentelemetry-tracing-arrives-in-envoy-ai-gateway) click bait ;)

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -22,9 +22,18 @@ A trace is made of one or more spans. The first span represents the root span. E
 
 ## Specifications
 
-- [Traces](./traces.md)
-- [Semantic Conventions](./semantic_conventions.md)
-- [Configuration](./configuration.md)
+### Core Specifications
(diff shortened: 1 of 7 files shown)


mikeldking:
@codefromthecrypt thanks so much for detailing this. This is going to be a huge lift for implementors.

mikeldking (approved):

codefromthecrypt:
oh wow thanks for merging!

PR #2170, opened 2025-09-03
fix(mcp): handle ValidationError in stdio stream instrumentation

codefromthecrypt:
MCP stdio transport can pass ValidationError objects through the stream when invalid JSON-RPC messages are received. The instrumentation was incorrectly assuming all items were SessionMessage objects, causing AttributeError when trying to access .message attribute on ValidationError.

Added isinstance check before accessing message attributes and test to verify ValidationError handling works correctly.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
@@ -124,8 +124,13 @@ class InstrumentedStreamReader(ObjectProxy):  # type: ignore
         from mcp.types import JSONRPCRequest
 
         async for item in self.__wrapped__:
-            session_message = cast(SessionMessage, item)
-            request = session_message.message.root
+            # Handle exceptions and other non-SessionMessage items
+            # MCP can pass ValidationError or other exceptions through the stream
+            if not isinstance(item, SessionMessage):
+                yield item
+                continue
+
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
cc @anuraaga in case you have style notes on this

python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py
```diff
@@ -124,8 +124,13 @@ async def __aiter__(self) -> AsyncGenerator[Any, None]:
         from mcp.types import JSONRPCRequest
 
         async for item in self.__wrapped__:
-            session_message = cast(SessionMessage, item)
-            request = session_message.message.root
+            # Handle exceptions and other non-SessionMessage items
+            # MCP can pass ValidationError or other exceptions through the stream
+            if not isinstance(item, SessionMessage):
```
anuraaga, line 129: Ouch, yeah casting without `isinstance` on any non-internal type was bad, thanks.

How about adding the check for the other spot too? Even if we can't find a way to trigger it in a unit test it's probably ok to err on caution.

https://github.com/Arize-ai/openinference/blob/daa143d9f86178b1d74383f37947ac23ae8c0518/python/instrumentation/openinference-instrumentation-mcp/src/openinference/instrumentation/mcp/__init__.py#L164

codefromthecrypt:
ok fix is consistent now and tests for both branches

PR #2171, opened 2025-09-04
fix(smolagents): Not Getting Step Spans From Latest Smolagents

### LINKED ISSUE #2000: [bug] Smolagents instrumentor is not instrumenting steps
**Describe the bug**
After updating smolagents, the instrumentation is no longer instrumenting the steps.

Image

After changing in the __init__.py file with the following (replace `step` with `_step_stream`):

```
for step_cls in [CodeAgent, ToolCallingAgent]:
            self._original_step_methods[step_cls] = getattr(step_cls, "_step_stream", None)
            wrap_function_wrapper(
                module="smolagents",
                name=f"{step_cls.__name__}._step_stream",
                wrapper=step_wrapper,
            )
```
This improved the behaviour a bit:

Image

The `step` function is no longer used, so the wrapper isn't being actually applied. Also, maybe because it is a generator, it is not adding the LLM and tool call inside of the Step span.

**To Reproduce**
Try running anything with `smolagents == 1.20.0` and `openinference-instrumentation-smolagents==0.1.14`. You will notice that steps are nowhere to be seen.

**Expected behavior**
Steps are being correctly wrapped as they were in previous versions of smolagents.

**Additional context**

ialisaleh:
This PR fixes step-level instrumentation due to breaking changes in the latest `smolagents` package, where the `step` method was replaced with `_step_stream` returning a `generator` instead of a direct result. The old `_StepWrapper` wasn’t generator-aware, causing spans to break and lose context. The update adds `yield from` delegation with proper error handling to preserve span context and attributes, restoring correct step instrumentation.

**Arize Trace**

Smolagents_Step_Spans_Arize

**Phoenix Trace**

Smolagents_Step_Spans_Phoenix

Closes #2000


python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -139,6 +139,55 @@ def __call__(
         return agent_output
 
 
+def _finalize_step_span(
+    span: trace_api.Span,
+    step_log: Any,
+    span_status_already_set: bool,
+) -> None:
+    """Record step results & set final span status."""
```
mikeldking: nit: these doctrings are not very useful. Can you flesh out more

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -139,6 +139,55 @@ def __call__(
...
+    try:
+        observations = getattr(step_log, "observations", None)
+        if observations is not None:
+            span.set_attribute(OUTPUT_VALUE, str(observations))
+
+        if not span_status_already_set:
```
mikeldking: I'm not sure I follow this logic. Even if it was previously set, don't you want to capture errors? 

In general this `span_status_already_set` param is very brittle and hard to understand because it forces me to understand control flow upstream. Can you think of a way to change the control flow so these functions are not as imparative? In general functions should aim to be pure. Having a boolean flag changing the behavior of a function is very difficult.

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -139,6 +139,55 @@ def __call__(
...
+        error_attrs: dict[str, Any]
+        if hasattr(error, "dict") and callable(getattr(error, "dict")):
+            error_attrs = error.dict()
+        else:
+            error_attrs = {"message": str(error)}
+
+        span.add_event(
+            name="agent.step_recovery",
+            attributes={**error_attrs, "severity": "expected"},
+        )
```
mikeldking, line 270: is there a precedent to recording these as events? I just don't want to record events because we can - it bloats the payload un-necessarily.

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -151,7 +200,9 @@ def __call__(
         kwargs: Mapping[str, Any],
     ) -> Any:
         if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
-            return wrapped(*args, **kwargs)
+            yield from wrapped(*args, **kwargs)
```
mikeldking, line 297: I'm unfamiliar with python generators. Can we double check with @Arize-ai/oss-python that this doesn't un-expetedly consume streams?

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
mikeldking: I think more concretely you want errors to always set the status to error I believe. I think this re-work. 

Is the span's status not available on itself? The parameterization seems un-necessary when you have the span.

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @mikeldking - I have updated the docstrings in these functions. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @mikeldking - I have replaced the usage of `span_status_already_set` param with `span.status.status_code` to manage the logic. It is used to avoid overwriting the execution errors. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @mikeldking - We need to record the events in this case to capture the expected tool errors. They were added to resolve the following issue earlier. Thanks!

Ticket: https://github.com/Arize-ai/openinference/issues/1198

PR: https://github.com/Arize-ai/openinference/pull/1845

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @mikeldking - It does not consume streams unexpectedly. The wrapper is generator-transparent. The stream is only advanced when the caller iterates it. Errors & span finalization still happen in sync with the stream’s lifecycle. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
```diff
@@ -57,10 +57,10 @@ def _instrument(self, **kwargs: Any) -> None:
         self._original_step_methods: Optional[dict[type, Optional[Callable[..., Any]]]] = {}
         step_wrapper = _StepWrapper(tracer=self._tracer)
         for step_cls in [CodeAgent, ToolCallingAgent]:
-            self._original_step_methods[step_cls] = getattr(step_cls, "step", None)
+            self._original_step_methods[step_cls] = getattr(step_cls, "_step_stream", None)
```
axiomofjoy:
```suggestion
            self._original_step_stream_methods[step_cls] = getattr(step_cls, "_step_stream", None)
```

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -139,7 +139,75 @@ def __call__(
...
+            span.set_attribute(OUTPUT_VALUE, str(observations))
+
+        if span.status.status_code != trace_api.StatusCode.ERROR:  # type: ignore[attr-defined]
+            error = getattr(step_log, "error", None)
+            if error is None:
+                span.set_status(trace_api.StatusCode.OK)
+            else:
+                _record_step_error(span, error)
+
+    except Exception as finalization_error:
```
axiomofjoy: Why do we need this try/ except here? Can we just nix?

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
ialisaleh: Hi @axiomofjoy - I have renamed it from `_original_step_methods` to `_original_step_stream_methods` in this module. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
ialisaleh: Hi @axiomofjoy - I have removed this try-except block as it is no longer needed. Thanks!

python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
```diff
@@ -223,7 +223,66 @@ def wrapped_generator() -> Generator[str, None, None]:
...
+    Record error details for the step span.
+
+    - Marks expected tool errors as recoverable (status = OK).
+    - Adds structured error details as span events for expected tool errors.
+    - Marks unexpected errors as failures & records the exception.
+    """
+    error_type = error.__class__.__name__
+    expected_error_types = {"AgentToolCallError", "AgentToolExecutionError"}
+
+    if error_type in expected_error_types:
```
axiomofjoy, line 260: Can we import these errors inside this function and do an `isinstance` check so we know if those errors are removed in future versions of the package?

ialisaleh: Hi @axiomofjoy - I've replaced imports with string comparison to avoid adding `smolagents` as a hard dependency. Thanks!

PR #2191, opened 2025-09-10
chore(crewai): fix pytest fixtures to be less flakey require less auth

codefromthecrypt:
I noticed that this test was flakey and running live when it shouldn't. I restructured it to match others, and also changed the agent kickoff to prevent it from making many commands to the same resource or sending telemetry to crewai. Finally, I removed serper for just normal scraping as this isn't a test about serper specifically and that way we don't need yet another key to auth with.


python/instrumentation/openinference-instrumentation-crewai/tests/openinference/instrumentation/crewai/fixtures/crew_session.yaml
```diff
@@ -1,1692 +0,0 @@
...
-      Connection:
-      - keep-alive
-      Content-Length:
-      - '2100'
-      Content-Type:
-      - application/x-protobuf
-      User-Agent:
-      - OTel-OTLP-Exporter-Python/1.31.1
-    method: POST
-    uri: https://telemetry.crewai.com:4319/v1/traces
```
codefromthecrypt, line 304: here's an example which bloated the yaml.. by default crewai adds its own anonymous tracing and sends stuff out like this, which ends up recorded in VCR unless you disable it!

python/instrumentation/openinference-instrumentation-crewai/tests/openinference/instrumentation/crewai/fixtures/crew_session.yaml
```diff
@@ -1,1692 +0,0 @@
...
-      accept-encoding:
-      - gzip, deflate, zstd
-      connection:
-      - keep-alive
-      content-length:
-      - '3371'
-      content-type:
-      - application/json
-      cookie:
-      - __cf_bm=pyXRwWW0EZ.MBiNjEhiGsZQiBSvyA0JGiO7VmKJvxmQ-1743187725-1.0.1.1-QArMazupZLVhgCTABbQ0oPO57r.IB9m1f.O9Ja1btlzmFmJ4VRslaXhWmQrzOSd.w7U4IlWgiH6n_uFnrO9YziG4L2gQZSRPICxrSGNpOlk;
```
codefromthecrypt, line 727: also it was misconfigured and leaked cookies

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
```diff
@@ -0,0 +1,231 @@
...
+        model="gpt-4.1-nano", api_key=openai_api_key, temperature=0
+    )  # Use a smaller model for tests
+    scraper_agent = Agent(
+        role="Website Scraper",
+        goal="Scrape the content from a given website URL",
+        backstory="You are an expert at extracting text content from websites using scraping tools",
+        tools=[ScrapeWebsiteTool()],
+        llm=llm,
+        max_iterations=1,
+        max_retry_limit=0,
```
codefromthecrypt, line 74: unless you add things like this it might randomly do the search a couple times

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
```diff
@@ -0,0 +1,231 @@
...
+
+from openinference.instrumentation import OITracer, using_attributes
+from openinference.instrumentation.crewai import CrewAIInstrumentor
+from openinference.semconv.trace import (
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+)
+
+# Don't record or send telemetry to CrewAI during tests
+os.environ["CREWAI_DISABLE_TELEMETRY"] = "true"
```
mikeldking, line 21: 👍🏻

PR #2202, opened 2025-09-15
feat(crewai): enhance span naming to include crew, agent, and tool context

### LINKED ISSUE #2185: [feature request] CrewAI AutoInstrumentor - Renaming spans to include Crew Name, Agent Name, Task Name
For CrewAI instrumentation, the spans are named as Crew.kickoff and Task._execute_core. 

When there are a bunch of different crews interacting with each other in a multi-agent setup, it is very hard to understand which crew gets kicked off when and which task or agent gets called. Checking the attributes section for crew name or agent name gets very manual and hard to troubleshoot in production. 

It would be great if spans are renamed so that it includes the crew name and task name itself. 

For example, instead of Crew.kickoff, it could be DataEnhancerCrew.kickoff

We have considered manually instrumenting certain parts in order to add crew name as an attribute or creating another span for crew name, but the solution above makes the most sense. 

**Additional context**

Image

caroger:
resolves #2185

before:
Screenshot 2025-09-15 at 1 15 12 AM

after:
Screenshot 2025-09-15 at 1 15 39 AM

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -164,7 +164,7 @@ class _ExecuteCoreWrapper:
         if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
             return wrapped(*args, **kwargs)
         # task naming - include agent role and task context
-        agent = args[0] if args else None
+        agent = args[0] if args else kwargs.get("agent")
         span_name = _get_execute_core_span_name(instance, wrapped, agent)
         with self._tracer.start_as_current_span(
             span_name,
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -84,6 +84,63 @@ def _get_input_value(method: Callable[..., Any], *args: Any, **kwargs: Any) -> s
     )
 
 
+def _get_crew_name(crew: Any) -> str:
+    """Generate a meaningful crew name for span naming."""
+    # First try to use crew.name attribute (user-friendly name)
+    if hasattr(crew, "name") and crew.name and crew.name.strip():
+        crew_name = str(crew.name).strip()
+        # If it's not just the default "crew", use it
+        if crew_name.lower() != "crew":
+            return crew_name
+
+    # Fallback to Crew with ID
+    if hasattr(crew, "id") and crew.id is not None:
+        return f"Crew_{str(crew.id)}"
+
+    # Final fallback
+    return "Crew"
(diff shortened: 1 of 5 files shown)


python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
```diff
@@ -192,7 +195,16 @@ def _verify_agent_span(
         attributes.get(SpanAttributes.OPENINFERENCE_SPAN_KIND)
         == OpenInferenceSpanKindValues.AGENT.value
     )
-    assert span.name == expected_name
+    # Enhanced naming: expect agent role in span name
+    assert span.name.endswith("._execute_core"), (
+        f"Expected span name to end with '._execute_core', got: {span.name}"
+    )
+    # Verify agent role is part of the span name
+    agent_role = attributes.get(SpanAttributes.GRAPH_NODE_ID)
```
RogerHYang: fwiw
```suggestion
    graph_node_id = attributes.get(SpanAttributes.GRAPH_NODE_ID)
```

RogerHYang (approved):

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
RogerHYang: since this is fallback, the assertion wouldn't work if `crew.name` exists, correct?

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
```diff
@@ -106,10 +163,9 @@ def __call__(
     ) -> Any:
         if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
             return wrapped(*args, **kwargs)
-        if instance:
-            span_name = f"{instance.__class__.__name__}.{wrapped.__name__}"
-        else:
-            span_name = wrapped.__name__
+        # task naming - include agent role and task context
+        agent = args[0] if args else None
```
RogerHYang: any risk of this being specified as keyword argument instead of positional, i.e. found in `kwargs` and not in `args`?

caroger: good call.
addressed

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
caroger: The variable should be renamed to graph_node_id for clarity - done

However, I think the assertion logic should work correctly even with fallbacks because:

If agent exists with role → GRAPH_NODE_ID = agent.role, span name = "{agent.role}._execute_core" → assertion passes ✓
If no agent/role → GRAPH_NODE_ID = None, the if condition skips the assertion ✓

The crew.name fallback is used in crew span naming (_get_crew_name function), but that's for kickoff spans, not the agent spans being tested here.
Suggested fix:

PR #2210, opened 2025-09-17
feat(openai): align embedding instrumentation with pending spec

codefromthecrypt:
### Impact

This PR aligns openai with the specification changes around embeddings in #2162 

**Spec changes from 2162**
- Consistent span name: `"CreateEmbeddings"`
- Standardized attribute structure: `embedding.embeddings.N.embedding.{text|vector}`
- Unified invocation parameter tracking: `embedding.invocation_parameters`
- Proper `llm.system` attribute for provider identification

**Code improvements:**
- Full batch embedding support with indexed attributes
- Separated invocation parameters from input data
- Improved handling of token IDs vs text inputs

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
+++ b/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
@@ -302,8 +302,12 @@ class _Request(_WithTracer, _WithOpenAI):
             return wrapped(*args, **kwargs)
         try:
             cast_to, request_parameters = _parse_request_args(args)
-            # E.g. cast_to = openai.types.chat.ChatCompletion => span_name = "ChatCompletion"
-            span_name: str = cast_to.__name__.split(".")[-1]
+            # Use consistent span names: "CreateEmbeddings" for embeddings, class name for others
+            if cast_to is self._openai.types.CreateEmbeddingResponse:
+                span_name = "CreateEmbeddings"
+            else:
+                # E.g. cast_to = openai.types.chat.ChatCompletion => span_name = "ChatCompletion"
+                span_name = cast_to.__name__.split(".")[-1]
         except Exception:
             logger.exception("Failed to parse request args")
             return wrapped(*args, **kwargs)
@@ -359,8 +363,12 @@ class _AsyncRequest(_WithTracer, _WithOpenAI):
             return await wrapped(*args, **kwargs)
         try:
             cast_to, request_parameters = _parse_request_args(args)
-            # E.g. cast_to = openai.types.chat.ChatCompletion => span_name = "ChatCompletion"
-            span_name: str = cast_to.__name__.split(".")[-1]
+            # Use consistent span names: "CreateEmbeddings" for embeddings, class name for others
+            if cast_to is self._openai.types.CreateEmbeddingResponse:
+                span_name = "CreateEmbeddings"
+            else:
+                # E.g. cast_to = openai.types.chat.ChatCompletion => span_name = "ChatCompletion"
+                span_name = cast_to.__name__.split(".")[-1]
         except Exception:
             logger.exception("Failed to parse request args")
             return await wrapped(*args, **kwargs)
--- a/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
+++ b/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
@@ -22,12 +22,16 @@ from openinference.instrumentation import safe_json_dumps
 from openinference.instrumentation.openai._attributes._responses_api import _ResponsesApiAttributes
 from openinference.instrumentation.openai._utils import _get_openai_version
 from openinference.semconv.trace import (
+    EmbeddingAttributes,
     ImageAttributes,
     MessageAttributes,
     MessageContentAttributes,
     SpanAttributes,
     ToolCallAttributes,
 )
+
+# TODO: Update to use SpanAttributes.EMBEDDING_INVOCATION_PARAMETERS when released in semconv
+_EMBEDDING_INVOCATION_PARAMETERS = "embedding.invocation_parameters"
 
 if TYPE_CHECKING:
     from openai.types import Completion, CreateEmbeddingResponse
@@ -212,6 +216,16 @@ def _get_attributes_from_completion_create_param(
     invocation_params.pop("prompt", None)
     yield SpanAttributes.LLM_INVOCATION_PARAMETERS, safe_json_dumps(invocation_params)
 
+    model_prompt = params.get("prompt")
+    if isinstance(model_prompt, str):
+        yield SpanAttributes.LLM_PROMPTS, [model_prompt]
+    elif (
+        isinstance(model_prompt, list)
+        and model_prompt
+        and all(isinstance(item, str) for item in model_prompt)
+    ):
+        yield SpanAttributes.LLM_PROMPTS, model_prompt
+
 
 def _get_attributes_from_embedding_create_param(
     params: Mapping[str, Any],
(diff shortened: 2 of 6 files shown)


python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
```diff
@@ -302,8 +302,12 @@ def __call__(
             return wrapped(*args, **kwargs)
         try:
             cast_to, request_parameters = _parse_request_args(args)
-            # E.g. cast_to = openai.types.chat.ChatCompletion => span_name = "ChatCompletion"
-            span_name: str = cast_to.__name__.split(".")[-1]
+            # Use consistent span names: "CreateEmbeddings" for embeddings, class name for others
+            if cast_to is self._openai.types.CreateEmbeddingResponse:
+                span_name = "CreateEmbeddings"
```
mikeldking, line 307: just for my understanding here - what does the consistent name buy us rather than maybe naming based on function name.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
```diff
@@ -212,6 +216,16 @@ def _get_attributes_from_completion_create_param(
     invocation_params.pop("prompt", None)
     yield SpanAttributes.LLM_INVOCATION_PARAMETERS, safe_json_dumps(invocation_params)
 
+    model_prompt = params.get("prompt")
+    if isinstance(model_prompt, str):
+        yield SpanAttributes.LLM_PROMPTS, [model_prompt]
```
mikeldking, line 229: Oh interesting, this is openAI's prompts I'm guesssing.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
codefromthecrypt: backends typically index the span name field for query and sometimes people do aggregates/auto-complete on the span name. For example, zipkin does auto-complete on span name, but not arbitrary attributes. Some tools dereive metrics from spans and need to aggregate on something, typically that's a span name. So, if something that is in a spec leaves the span name out, it ends up unable to be usable for things like this.

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
mikeldking: is this yielding a list? below it seems to be yielding a single prompt.

codefromthecrypt: yeah refactoring from here which was done at response time, but we can actually record it in request time https://github.com/Arize-ai/openinference/pull/2210/files/12b937130030f888cc85ecbf1d3cb2d0f2fb9e50

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_response_attributes_extractor.py
```diff
@@ -37,11 +35,6 @@
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
-try:
-    _NUMPY: Optional[ModuleType] = import_module("numpy")
-except ImportError:
-    _NUMPY = None
-
```
mikeldking, line 44: I honestly don't remember why this was needed.

mikeldking (approved):

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
codefromthecrypt: here's an oldie example of this, which is subverted when span names are subtly different for the same operation https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml#L44

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_response_attributes_extractor.py
codefromthecrypt: originally it was doing this

```                # See https://github.com/openai/openai-python/blob/f1c7d714914e3321ca2e72839fe2d132a8646e7f/src/openai/resources/embeddings.py#L100  # noqa: E501
                vector = _NUMPY.frombuffer(base64.b64decode(_vector), dtype="float32").tolist()
```

but we have test cases to prove we can get the vectors at the moment without it. If we need to re-introduce it we probably should with a failing test

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
codefromthecrypt: I'll make sure there are test coverage and whatever the spec allows is here and no more!

codefromthecrypt: so the attribute allows for a list I think, but the call site should never permit one due to chat completions being only having a single prompt. that said, I'm not sure if since this type is shared between legacy and non-legacy if someone could force multiple in anyway... looking deeper and currently I'm thinking to just skip if it isn't a single prompt and also I assume since it is prompts not prompt, the openinference attribute should always be a list..

codefromthecrypt: ok added a commit hopefully clarifies it. this is tricky because the legacy completions api is weird

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request_attributes_extractor.py
```diff
@@ -204,14 +208,32 @@ def _get_attributes_from_image(
 def _get_attributes_from_completion_create_param(
     params: Mapping[str, Any],
 ) -> Iterator[Tuple[str, AttributeValue]]:
-    # openai.types.completion_create_params.CompletionCreateParamsBase
-    # See https://github.com/openai/openai-python/blob/f1c7d714914e3321ca2e72839fe2d132a8646e7f/src/openai/types/completion_create_params.py#L11  # noqa: E501
+    """
+    Extract attributes from parameters for the LEGACY completions API.
```
codefromthecrypt, line 212: here's a rewording of the comments which I think better explains the prompt union weirdness

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
mikeldking: I see. I mainly ask because I tend to face a fair amount of pressure to cram things like agent operations in names because there's a need for operators to groc the control flow.

With embedding's generation, though I think this makes a lot of sense

axiomofjoy (approved):

python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_request.py
codefromthecrypt: Agreed. I don't think agent ops are commodity yet maybe not for a long while. Wont go trying to normalize those ;)

codefromthecrypt:
gonna be away this weekend, so lemme know if there's anything else to do here. regardless, I'll chase up the others next week.

codefromthecrypt:
thanks, put to action here! https://github.com/envoyproxy/ai-gateway/pull/1232

PR #2212, opened 2025-09-18
fix(instructor): JSON Serialization Error for max_retries in Span Attributes

### LINKED ISSUE #1569: [bug] instructor max retries invocation params issue
**Describe the bug**
When `max_retries` is set as a Tenacity `Retrying` object, `Type Error: Object of type Retrying is not JSON serializable` is raised by [_wrapper.py (line 160)](https://github.com/Arize-ai/openinference/blob/ba21b40ada7197c7c3bd0fe9bc04e64343cf7993/python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py#L160)

**To Reproduce**
Assign `max_retries` as a `Retrying` object as seen here in the [Instructor docs](https://github.com/instructor-ai/instructor/blob/main/docs/concepts/retrying.md#advanced-retry-logic). 

**Expected behavior**
Retrying object to be excluded from `SpanAttributes.LLM_INVOCATION_PARAMETERS` so the Type Error is not raised.  

 - OS: MacOS
 - Version 15.1.1
 - Python: 3.13

**Additional context**

Example code:

```
import instructor
from openai import OpenAI
from tenacity import Retrying, stop_after_attempt, wait_fixed

client = instructor.from_openai(OpenAI())

class UserInfo(BaseModel):
    name: str
    age: int

response = client.chat.completions.create(
    model="gpt-4o-mini",
    response_model=UserInfo,
    messages=[
        {"role": "user", "content": "John Doe is 30 years old."},
    ],
    max_retries=Retrying(
        stop=stop_after_attempt(2), 
        wait=wait_fixed(1), 
    )
)
```

ialisaleh:
This PR fixes an issue where setting `max_retries` as a `tenacity.Retrying` object caused `JSON` serialization errors in span attributes. The update ensures non-serializable objects are handled safely so traces can be sent without breaking. It also adds explicit `StatusCode.OK` on successful runs, making traces clearer and easier to follow in Arize & Phoenix.

**Instructor Trace**

Instructor_Max_Retries_Error

Closes #1569

### DIFF
--- a/python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
@@ -205,7 +205,8 @@ class _PatchWrapper:
                     if resp is not None and hasattr(resp, "dict"):
                         span.set_attribute(OUTPUT_VALUE, json.dumps(resp.dict()))
                         span.set_attribute(OUTPUT_MIME_TYPE, "application/json")
-                        span.set_status(trace_api.StatusCode.OK)
+
+                    span.set_status(trace_api.StatusCode.OK)
                     return resp
                 except Exception as e:
                     span.set_status(trace_api.Status(trace_api.StatusCode.ERROR, str(e)))
@@ -239,7 +240,8 @@ class _PatchWrapper:
                     if resp is not None and hasattr(resp, "dict"):
                         span.set_attribute(OUTPUT_VALUE, json.dumps(resp.dict()))
                         span.set_attribute(OUTPUT_MIME_TYPE, "application/json")
-                        span.set_status(trace_api.StatusCode.OK)
+
+                    span.set_status(trace_api.StatusCode.OK)
                     return resp
                 except Exception as e:
                     span.set_status(trace_api.Status(trace_api.StatusCode.ERROR, str(e)))
--- a/python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
@@ -117,9 +117,42 @@ class _PatchWrapper:
 
     @classmethod
     def _get_invocation_params(cls, request_params: Any) -> Dict[str, Any]:
+        """
+        Clean up invocation parameters for span attributes.
+        - Removes messages (can be large / sensitive).
+        - Handles tenacity.Retrying objects safely without a hard dependency.
+        - Ensures all values are JSON serializable.
+        """
         attributes = dict(request_params)
+
+        # Drop messages (too big / sensitive)
         if "messages" in request_params:
-            attributes.pop("messages")
+            attributes.pop("messages", None)
+
+        if "max_retries" in request_params:
+            max_retries = attributes.get("max_retries")
+
+            # Handle tenacity.Retrying safely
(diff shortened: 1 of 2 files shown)


python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
```diff
@@ -117,9 +117,42 @@ def _clean_request_params(cls, attributes: Dict[str, Any]) -> Dict[str, Any]:
...
+                    attributes["max_retries"] = {
+                        "stop": repr(getattr(max_retries, "stop", None)),
+                        "wait": repr(getattr(max_retries, "wait", None)),
+                        "sleep": repr(getattr(max_retries, "sleep", None)),
+                        "retry": repr(getattr(max_retries, "retry", None)),
+                        "before": repr(getattr(max_retries, "before", None)),
+                        "after": repr(getattr(max_retries, "after", None)),
+                    }
+                except Exception:
+                    attributes.pop("max_retries", None)
```
mikeldking, line 148: can you add a quick test for these.

In general there seems to be a mismatch between the attributes being llm specific but the span_kind being a tool so things don't visualize well.

Wondering if we should follow-up and fix this. cc @caroger

python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
```diff
@@ -117,9 +117,42 @@ def _clean_request_params(cls, attributes: Dict[str, Any]) -> Dict[str, Any]:
...
+            else:
+                # Ensure JSON-serializability for other types
+                try:
+                    json.dumps(max_retries)
```
mikeldking, line 152: shouldn't this be assigning the dumped value to the attribute?

ialisaleh: Hi @mikeldking - In this case, the `json.dumps(...)` call is only being used as a quick check to see whether the value is JSON-serializable for other types. If it passes, we keep the original value since it can be serialized later when the full attribute dict is dumped. If we directly assign the dumped string here, it would get double-encoded downstream. Thanks!

python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
ialisaleh: Hi @mikeldking - We shouldn’t add `tenacity` as a hard dependency just for testing, since invocation attributes can vary across integrations. I can add a `FakeRetrying` test, which will only be useful to guard our own serialization logic here.

On the `span_kind`, I agree there’s a mismatch. Right now, we’re treating LLM calls as tool spans, which makes the visualization less clear. I think it makes sense to follow up with a separate PR to align the span kind with LLM-specific attributes so things show up correctly. Thanks!

mikeldking: Okay yeah let's do that as a follow-up

python/instrumentation/openinference-instrumentation-instructor/src/openinference/instrumentation/instructor/_wrappers.py
mikeldking: I see

PR #2238, opened 2025-09-28
feat(litellm): align embedding instrumentation with pending spec

codefromthecrypt:
This PR aligns litellm with the specification changes around embeddings in #2162

**Spec changes from 2162**
- Consistent span name: `"CreateEmbeddings"`
- Standardized attribute structure: `embedding.embeddings.N.embedding.{text|vector}`
- Unified invocation parameter tracking: `embedding.invocation_parameters`
- Proper `llm.system` attribute for provider identification

**Code improvements:**
- Full batch embedding support with indexed attributes
- Separated invocation parameters from input data
- Improved handling of token IDs vs text inputs
- Vectors stored as tuples instead of JSON strings

This is the same as #2210, except litellm.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
@@ -1,4 +1,3 @@
-import json
 from enum import Enum
 from functools import wraps
 from typing import (
@@ -47,6 +46,9 @@ from openinference.semconv.trace import (
     ToolAttributes,
     ToolCallAttributes,
 )
+
+# TODO: Update to use SpanAttributes.EMBEDDING_INVOCATION_PARAMETERS when released in semconv
+_EMBEDDING_INVOCATION_PARAMETERS = "embedding.invocation_parameters"
 
 # Skip capture
 KEYS_TO_REDACT = ["api_key", "messages"]
@@ -226,10 +228,39 @@ def _instrument_func_type_embedding(span: trace_api.Span, kwargs: Dict[str, Any]
         SpanAttributes.OPENINFERENCE_SPAN_KIND,
         OpenInferenceSpanKindValues.EMBEDDING.value,
     )
+    _set_span_attribute(span, SpanAttributes.LLM_SYSTEM, "litellm")
     _set_span_attribute(
         span, SpanAttributes.EMBEDDING_MODEL_NAME, kwargs.get("model", "unknown_model")
     )
-    _set_span_attribute(span, EmbeddingAttributes.EMBEDDING_TEXT, str(kwargs.get("input")))
+
+    # Extract invocation parameters (everything except input)
+    invocation_params = {k: v for k, v in kwargs.items() if k != "input"}
+    if invocation_params:
+        _set_span_attribute(
+            span, _EMBEDDING_INVOCATION_PARAMETERS, safe_json_dumps(invocation_params)
+        )
+
+    # Extract text from embedding input - only records text, not token IDs
+    embedding_input = kwargs.get("input")
+    if embedding_input is not None:
+        if isinstance(embedding_input, str):
+            # Single string input
+            _set_span_attribute(
+                span,
+                f"{SpanAttributes.EMBEDDING_EMBEDDINGS}.0.{EmbeddingAttributes.EMBEDDING_TEXT}",
+                embedding_input,
+            )
+        elif isinstance(embedding_input, list) and embedding_input:
+            # Check if it's a list of strings (not tokens)
+            if all(isinstance(item, str) for item in embedding_input):
+                # List of strings
+                for index, text in enumerate(embedding_input):
+                    _set_span_attribute(
+                        span,
+                        f"{SpanAttributes.EMBEDDING_EMBEDDINGS}.{index}.{EmbeddingAttributes.EMBEDDING_TEXT}",
+                        text,
+                    )
+
     _set_span_attribute(span, SpanAttributes.INPUT_VALUE, str(kwargs.get("input")))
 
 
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
so this is the same as #2210, and a FAQ might be why return an indexed list? The reason is that if you have a mixed batch request, you'll get multiple embedding vectors back, index correlated with the input text. one example recorded from the openai instrumentation is this one https://github.com/envoyproxy/ai-gateway/blob/main/tests/internal/testopeninference/spans/embeddings-mixed-batch.json

python/instrumentation/openinference-instrumentation-litellm/tests/test_batch_embedding.py
axiomofjoy, line 1: Any reason not to use vcrpy here? I assume you are mainly trying to avoid bloating the repo. We might be able to mitigate this with some of the newer models that allow us to choose the embedding dimension?

python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
```diff
@@ -226,10 +228,39 @@ def _instrument_func_type_embedding(span: trace_api.Span, kwargs: Dict[str, Any]
         SpanAttributes.OPENINFERENCE_SPAN_KIND,
         OpenInferenceSpanKindValues.EMBEDDING.value,
     )
+    _set_span_attribute(span, SpanAttributes.LLM_SYSTEM, "litellm")
```
axiomofjoy: I think we ought to nix the LLM system here for now. Our formal definition is ambiguous currently.

```
The AI product as identified by the client or server instrumentation.
```

We have so far reserved system for API providers, examples of which can be seen [here](https://github.com/Arize-ai/openinference/blob/40b60e13bdb7f64b2c45d307089506c2061007a0/python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py#L494).

In terms of how we conceptualize this, we think of the LLM system as describing the shape of the API and the [LLM provider](https://github.com/Arize-ai/openinference/blob/40b60e13bdb7f64b2c45d307089506c2061007a0/python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py#L502) as describing the owner of the physical hardware that runs those APIs. For Arize and Phoenix, the main place these particular conventions are consumed is in playground, allowing us to re-invoke LLM calls. All of this is compatible with the notion of a particular SDK or library as a system, but as an observability product, there's not currently a concrete use case I can see that expanding the definition in this way would facilitate.

python/instrumentation/openinference-instrumentation-litellm/tests/test_batch_embedding.py
axiomofjoy: (This is non-blocking)

python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
codefromthecrypt: How about embedding system? We do need something to correlate with metrics even if there is no embedding playground in the Phoenix ui

codefromthecrypt: progress > perfection, so I will remove the system attribute as we can take it back up later. I'll remove it from openai which is already released so these are doing the same thing. Then, I think for sanity I'll add the docs to this PR since we've been discussing them.

Thanks for your time and consideration! will have this buffed out today eastern US time.

python/instrumentation/openinference-instrumentation-litellm/tests/test_batch_embedding.py
codefromthecrypt: better to use pytest-vcr. yeah I'll port it from the openai tests.

codefromthecrypt:
@axiomofjoy I updated this PR including the specs along with your rationale about why embeddings spans should have no system/provider attributes. If I got anything wrong, lemme know and I'll bump it!

codefromthecrypt:
https://github.com/Arize-ai/openinference/pull/2295 for the unrelated inspector test failures

python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
codefromthecrypt: @axiomofjoy this is all good iiuc. please take a look if you don't mind

python/instrumentation/openinference-instrumentation-litellm/tests/test_responses.py
```diff
@@ -11,11 +11,9 @@
 @pytest.fixture(autouse=True)
 def instrument(
     tracer_provider: TracerProvider,
-    in_memory_span_exporter: InMemorySpanExporter,
 ) -> Iterator[None]:
     LiteLLMInstrumentor().instrument(tracer_provider=tracer_provider)
     yield
-    LiteLLMInstrumentor().uninstrument()
```
axiomofjoy, line 18: Was this pattern causing an issue?

spec/embedding_spans.md
```diff
@@ -1,68 +1,159 @@
...
+The following attributes that are used in LLM spans are **not applicable** to embedding spans:
+
+- `llm.system`: Not used for embedding spans
+- `llm.provider`: Not used for embedding spans
+
+### Rationale
+
+The `llm.system` attribute is defined as "the AI product as identified by the client or server instrumentation." While this definition has been reserved for API providers in LLM spans (e.g., "openai", "anthropic"), it is ambiguous when applied to embedding operations.
+
+In terms of conceptualization, `llm.system` describes the shape of the API, while `llm.provider` describes the owner of the physical hardware that runs those APIs. For observability products like Arize and Phoenix, these conventions are primarily consumed in playground features, allowing re-invocation of LLM calls.
```
axiomofjoy, line 55: Thanks for taking a stab at this. I'm not sure we need to proscribe using `llm.system` and `llm.provider` for embedding spans. Only that we've never used these attributes to describe a client interface like `litellm` before. In the case of the OpenAI SDK, I think it could make sense to set them. It might make less sense to set them if we don't know the specific API being hit, as is the case for LiteLLM.

Let me chat with @RogerHYang to see if he has the same understanding.

codefromthecrypt: > I'm not sure we need to proscribe using llm.system and llm.provider for embedding spans.

ok yeah, this section is describing why to *not* use them, as it is counter intuitive to folks like me as you have no lookup dimension anymore.

python/instrumentation/openinference-instrumentation-litellm/tests/test_responses.py
codefromthecrypt: sorry this is a bug, at some point I had a  `def uninstrument` Once the shape of the PR is ok, I'll go back and fix this prior to merge

axiomofjoy (approved):

PR #2300, opened 2025-10-13
feat(haysack-ai): haystack async pipeline support

### LINKED ISSUE #1317: [feature request] Support for AsyncPipeline in Haystack
There is a new class of Pipeline() in Haystack called AsyncPipeline.
Currently Openinference doesn't track them.

Track them exactly the same as the Pipeline class

/

**Additional context**
Introduced in Haystack 2.10
https://haystack.deepset.ai/release-notes/2.10.0

satyadevai:
Closes #1317 
image
image
image

---

> [!NOTE]
> Adds OpenInference tracing for Haystack `AsyncPipeline` and async component runs, including `run_async` and `run_async_generator`, with new examples, tests, and minor refactors.
> 
> - **Instrumentation (core)**:
>   - Wrap `haystack.AsyncPipeline` methods: `run`, `run_async`, `run_async_generator`, and `_run_component_async`.
>   - Add async wrappers: `_AsyncPipelineWrapper`, `_AsyncPipelineRunAsyncGeneratorWrapper`, `_AsyncPipelineRunComponentWrapper`, `_AsyncComponentRunWrapper`.
>   - Extend component wrapping to cover both `run` and `run_async`; avoid double-wrapping.
>   - Factor request/response attribute collection into helpers: `_set_component_runner_request_attributes`, `_set_component_runner_response_attributes`.
>   - Tweak pipeline span names to use class-specific names (e.g., ``AsyncPipeline.run_async``).
> - **Tests**:
>   - Add async test coverage for chat generator, tool-calling, prompt builder + generator, retriever filtering, and wrapper install/uninstall; include new VCR cassettes.
> - **Examples**:
>   - New async examples: `examples/simple_async_pipeline.py`, `examples/pipeline_run_async_generator.py`.
> - **Config**:
>   - Add `pytest-asyncio` and set `asyncio_mode = "auto"` in `pyproject.toml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dbbeacb8c873de187429a78ade9adb1294c9e4d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
@@ -88,18 +88,35 @@ class HaystackInstrumentor(BaseInstrumentor):  # type: ignore[misc]
         original = Pipeline.__dict__["_run_component"]
         self._original_pipeline_run_component = original.__func__
         self._original_component_run_methods: dict[type[Any], Callable[..., Any]] = {}
+        self._original_component_run_async_methods: dict[type[Any], Callable[..., Any]] = {}
 
         def wrap_component_run_method(
             component_cls: type[Any], run_method: Callable[..., Any]
         ) -> None:
             # To avoid double wrapping, we only wrap the class bound "run" method here.
-            class_method = getattr(component_cls, "run")
-            if component_cls not in self._original_component_run_methods:
+
+            method_name = getattr(run_method, "__name__", None)
+            if method_name is None and hasattr(run_method, "__func__"):
+                method_name = getattr(run_method.__func__, "__name__", None)
+
+            if method_name == "run" and component_cls not in self._original_component_run_methods:
+                class_method = getattr(component_cls, "run")
                 self._original_component_run_methods[component_cls] = class_method
                 wrap_function_wrapper(
                     module=component_cls.__module__,
                     name=f"{component_cls.__name__}.run",
                     wrapper=_ComponentRunWrapper(tracer=self._tracer),
+                )
+            if (
+                method_name == "run_async"
+                and component_cls not in self._original_component_run_async_methods
+            ):
+                class_method = getattr(component_cls, "run_async")
+                self._original_component_run_async_methods[component_cls] = class_method
+                wrap_function_wrapper(
+                    module=component_cls.__module__,
+                    name=f"{component_cls.__name__}.{method_name}",
+                    wrapper=_AsyncComponentRunWrapper(tracer=self._tracer),
                 )
 
         wrap_function_wrapper(
@@ -112,32 +129,12 @@ class HaystackInstrumentor(BaseInstrumentor):  # type: ignore[misc]
 
         async_original = AsyncPipeline.__dict__["_run_component_async"]
         self._original_async_pipeline_run_component_async = async_original.__func__
-        self._original_component_run_async_methods: dict[type[Any], Callable[..., Any]] = {}
-
-        def wrap_component_run_async_method(
-            component_cls: type[Any], run_method: Callable[..., Any]
-        ) -> None:
-            # Component may have both run and run_async methods, so we check for the method name.
-            method_name = getattr(run_method, "__name__", None)
-            if method_name is None and hasattr(run_method, "__func__"):
-                method_name = getattr(run_method.__func__, "__name__", None)
-            if method_name == "run_async":
-                class_method = getattr(component_cls, "run_async")
-                if component_cls not in self._original_component_run_async_methods:
-                    self._original_component_run_async_methods[component_cls] = class_method
-                    wrap_function_wrapper(
-                        module=component_cls.__module__,
-                        name=f"{component_cls.__name__}.{method_name}",
-                        wrapper=_AsyncComponentRunWrapper(tracer=self._tracer),
-                    )
-            elif method_name == "run":
-                wrap_component_run_method(component_cls, run_method)
 
         wrap_function_wrapper(
             AsyncPipeline,
             "_run_component_async",
             _AsyncPipelineRunComponentWrapper(
-                tracer=self._tracer, wrap_component_run_method=wrap_component_run_async_method
+                tracer=self._tracer, wrap_component_run_method=wrap_component_run_method
             ),
         )
 
--- a/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
@@ -132,19 +132,21 @@ class _ComponentRunWrapper:
         component = instance
         component_class_name = _get_component_class_name(component)
         bound_arguments = _get_bound_arguments(wrapped, *args, **kwargs)
-        arguments = bound_arguments.arguments
 
         with self._tracer.start_as_current_span(
             name=_get_component_span_name(component_class_name)
         ) as span:
             component_type = _get_component_type(component)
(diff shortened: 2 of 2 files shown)


python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -747,6 +843,98 @@ def _get_bound_arguments(function: Callable[..., Any], *args: Any, **kwargs: Any
     return sig.bind(*args, **valid_kwargs)
 
 
+def _set_component_runner_request_attributes(
```
axiomofjoy, line 829: Let's make these into pure functions that map into a map of attribute values rather than functions with a side effect.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -82,6 +82,39 @@ def __call__(
...
+class _AsyncPipelineRunComponentWrapper:
+    """
+    Asynchronous wrapper for pipeline component execution.
+    Ensures that both async and sync run methods of a component are wrapped for tracing.
+    """
+
+    def __init__(
+        self,
+        tracer: trace_api.Tracer,
+        wrap_component_run_method: Callable[[type[Any], Callable[..., Any]], None],
```
axiomofjoy, line 94: I know this pattern already existed, but I think we should define the wrapper function in the `__init__` rather than injecting since it is only used once. I think I did it this way originally because the wrapper function needs access to state defined in the scope of the caller, but I think we can accomplish this with a callback function argument.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -104,80 +137,55 @@ def __call__(
...
+
+    async def __call__(
+        self,
+        wrapped: Callable[..., Any],
+        instance: Any,
+        args: Tuple[Any, ...],
+        kwargs: Mapping[str, Any],
+    ) -> Any:
+        if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
+            return await wrapped(*args, **kwargs)
```
axiomofjoy: This is no longer necessary, will be handled by the underlying instrumentation.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -104,80 +137,55 @@ def __call__(
...
+        with self._tracer.start_as_current_span(
+            name=_get_component_span_name(component_class_name)
+        ) as span:
+            component_type = _get_component_type(component)
+            _set_component_runner_request_attributes(span, wrapped, instance, args, kwargs)
+            try:
+                result = await wrapped(*args, **kwargs)
+            except Exception as exception:
+                span.set_status(trace_api.Status(trace_api.StatusCode.ERROR, str(exception)))
+                raise
```
axiomofjoy: Is the try/ except needed? I expected the context manager to handle?

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
```diff
@@ -104,80 +137,55 @@ def __call__(
...
+        component = instance
+        component_class_name = _get_component_class_name(component)
+        bound_arguments = _get_bound_arguments(wrapped, *args, **kwargs)
+        arguments = bound_arguments.arguments
+
+        with self._tracer.start_as_current_span(
+            name=_get_component_span_name(component_class_name)
+        ) as span:
+            component_type = _get_component_type(component)
+            _set_component_runner_request_attributes(span, wrapped, instance, args, kwargs)
```
axiomofjoy: Let's pull this out of the context so we only do the bare minimum inside.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
```diff
@@ -77,16 +110,72 @@ def wrap_component_run_method(
             ),
         )
 
+        async_original = AsyncPipeline.__dict__["_run_component_async"]
+        self._original_async_pipeline_run_component_async = async_original.__func__
```
axiomofjoy, line 131: What is this wrapping? We've had some issues maintaining haystack because I was probably a little over-eager with instrumenting implementation details. This has the look of a private API that is subject to change.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
```diff
@@ -27,6 +31,11 @@ class HaystackInstrumentor(BaseInstrumentor):  # type: ignore[misc]
         "_original_pipeline_run",
         "_original_pipeline_run_component",
         "_original_component_run_methods",
+        "_original_async_pipeline_run",
```
axiomofjoy, line 34: I take it the async pipeline has both sync and async run methods?

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
satyadevai: Made pure functions, it only returns dict object

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
satyadevai: Not needed, Removed

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
satyadevai: Yes, the AsyncPipeline provides both run (sync) and run_async (async) methods. Internally, the run method simply invokes run_async using asyncio to maintain compatibility with synchronous execution.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
satyadevai: I’ve reused the existing implementation of the sync version of _run_component (a private method). The actual component execution happens within the component runner, so wrapping these private methods is necessary for instrumentation. I haven’t introduced any new logic here — just extended the existing sync wrapping pattern to ensure consistent behavior.

python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/_wrappers.py
satyadevai: We’re reusing the same wrapper function for both sync and async components because Haystack components can define only run, or run and run_async both. By using a unified wrapper, we ensure consistent behavior and avoid redundant logic between sync and async paths.

The original component methods are stored in _original_component_run_methods and _original_component_run_async_methods, so each method is wrapped exactly once. This approach prevents double wrapping while maintaining compatibility with components that expose only one of the execution methods.

While we could move the wrapper function definitions to __init__, that would require either moving both run and run_async handling there or duplicating similar logic across _PipelineRunComponentWrapper and _AsyncPipelineRunComponentWrapper. The current implementation avoids this duplication and keeps the wrapping logic centralized and consistent.

PR #2370, opened 2025-10-27
fix: add json type specification for JSON imports in fixtures

mikeldking:


mikeldking:
> I don't have a problem adding this but does it actually do anything in unbundled code?

This is a shift in typescript so that instead of being a config in the tsconfig, you can import JSON like it is being proposed in ECMAScript proposals. For whatever reason this passed CI but it's actually failing eslint on the latest. this is just to make it declarative so as to satisfy eslint. My guess is that this is being changed on the linting so as to expedite the migration and proposal.

PR #2383, opened 2025-10-30
feat(haystack): Added Components instrumentation

### LINKED ISSUE #2328: [feature request] Add Agent.run root span and session hooks to Haystack instrumentation
 When instrumenting Haystack agents with openinference-instrumentation-haystack, only the downstream components (pipelines, tools, generators) produce spans. The Agent.run entry point never gets a root span, so observability backends can’t attach session ids or capture the user prompt/assistant reply. That makes it hard to group tool spans into a single “chat turn” trace.

Have the Haystack instrumentation wrap Agent.run (and related orchestrators) with a top-level span, and expose a hook/API that lets callers update trace metadata (session id, input, output) before and after execution. This would align agent observability with the existing pipeline instrumentation.

 Currently we import Langfuse in application code, start our own span around Agent.run, and manually call update_trace to set the session id and payloads. It works, but duplicates logic across projects and undermines the “automatic instrumentation” goal.

**Additional context**
Adding the root span would help to display full agent replays out of the box.

satyadevai:
In the existing haystack instrumentation, tracing was only applied within the Pipeline execution path. The instrumentation wrapped Pipeline methods, so component methods like run() or run_async() were dynamically wrapped only when they were executed as part of a pipeline. This design ensured that components inside a pipeline were traced properly, but if specific component is invoked directly (outside pipeline) is never traced because their execution did not go through the pipeline wrapping machanism. As a result, invoking Agent.run() directly created no spans at all, since the tracing logic was never attached to the Agent class itself.

Additionally, even when Agent.run() internally invoked Pipeline._run_component, the instrumentation still treated each sub-component call(LLM, Tools, Embeddings or other components) as an independent span. Instead of grouping them under a parent Agent.run span, every internal component run appeared as a separate trace segment. This happened because there was no root span created for Agent.run to serve as the parent for those internal operations—the pipeline instrumentation only captured the individual component spans without establishing hierarchical context.

Haystack will register all the components in component.registry, now we can wrap the all the registered components run and run_async methods at instrumentation time.  This ensures that every component, including Agent gets traced consistently wheather it runs independently or within a pipeline. It also enables proper span hierarchy, allowing Agent.run to created as root span that encapsulates the nested component executions.
 
closes #2328 

image
image
image
image
image
**No Changes detected in Pipeline.run**

image

---

> [!NOTE]
> Wrap all registered Haystack components (including Agent) for tracing, update span names to reflect actual method (run/run_async), and add examples plus tests (with VCR cassettes).
> 
> - **Instrumentation**:
>   - Wrap all registered components from `haystack.core.component.component.registry` to trace `run` and `run_async` outside pipelines (includes `Agent`).
>   - Change component span naming to `ClassName.<method>` using `wrapped.__name__` (e.g., `OpenAIChatGenerator.run_async`, `InMemoryBM25Retriever.run`).
>   - Ensure async/sync component methods are wrapped during pipeline execution as well.
> - **Tests**:
>   - Add `test_agent_run_component_spans` (sync/async) validating parent `Agent.<method>` span and nested LLM/tool spans.
>   - Add tests for individual components outside pipelines (`InMemoryBM25Retriever.run`/`run_async`).
>   - Update expectations where spans now end with `.run_async` for async paths.
>   - Add corresponding VCR cassettes for agent runs.
> - **Examples**:
>   - New examples in `examples/`: `agent_run.py`, `agent_with_pipeline.py`, `retriever_component_run.py` demonstrating tracing setup and component/agent runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1cd56df320f6edd6c8a6a5b923c4b83f03a2bb8a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py
```diff
@@ -137,6 +137,14 @@ def wrap_component_run_method(
...
             ),
         )
+        from haystack.core.component.component import component
+
+        for class_path, cls in component.registry.items():
+            # Ensure the class looks like a Component (has run or run_async)
+            if hasattr(cls, "run"):
+                wrap_component_run_method(cls, cls.run)
+            if hasattr(cls, "run_async"):
+                wrap_component_run_method(cls, cls.run_async)
```
axiomofjoy, line 147: Can we also uninstrument these methods as well?

satyadevai: These methods are already being uninstrumented by the existing code in the uninstrument block. Since we’re not introducing any new logic to capture these original methods and are simply using the existing dictionary, no additional uninstrumentation is required.

https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-haystack/src/openinference/instrumentation/haystack/__init__.py#L165-L178

PR #2390, opened 2025-10-30
ci: add provinence support for npm publish

mikeldking:


.github/workflows/typescript-publish.yaml
```diff
@@ -1,53 +1,58 @@
 name: Typescript Publish
 on:
-    push:
-        branches:
-            - main
+  push:
+    branches:
+      - main
+
```
mikeldking:
```suggestion
```

Issue #2430, opened 2025-11-12
[bug] LangChain.js/LangGraph.js instrumentation not compatible with v1


sinedied:
**Describe the bug**
The LangChain.js Instrumentor is no longer compatible with LangChain.js and LangGraph v1.
API has changed slightly, and peer dependency mismatch event prevents installing the package.

**To Reproduce**
Include langchain>=1.0.0 dependency

Then use `import { LangChainInstrumentation } from '@arizeai/openinference-instrumentation-langchain';`

**Expected behavior**
Install without error.
Trace and spans are emitted.

**Additional context**
Python version had a similar issue that got fixed, might help since both versions are similar: https://github.com/Arize-ai/openinference/issues/2314

sinedied:
@dosu the fix only address the Python compatibility issue, but the JS NPM instrumentation is still broken with v1.

mikeldking:
@sinedied thanks for filing the issue and bringing our attention to this. Let us know if there are any concrete workflows you'd like us to test.

Also PRs welcome if you spot the issue.

mikeldking:
@sinedied I just published a new major version for this. I did some testing and it seemed to work as expected. LMK if it works for you. Happy to work out the kinks and iterate here.

<img width="2056" height="1173" alt="Image" src="https://github.com/user-attachments/assets/2c08ce5e-9807-489d-a2cf-85c407a6c997" />

sinedied:
Thank you!
The package install fines now, but I'm having trouble as `import * as CallbackManagerModule from "@langchain/core/callbacks/manager";` isn't working anymore in LC.js v1.
How did you get it to work?

EDIT: The issue is because of LC v1 when using CommonJS modules, I switched my tsconfig to use Nodenext modules and it fixed the issue.
Now I can confirm that everything works fine! 🙂

PR #2444, opened 2025-11-13
feat!: langchainjs 1.0

mikeldking:
resolves  https://github.com/Arize-ai/openinference/issues/2430
Screenshot 2025-11-13 at 10 48 00 PM

### DIFF
--- a/js/packages/openinference-instrumentation-langchain/package.json
+++ b/js/packages/openinference-instrumentation-langchain/package.json
@@ -40,7 +40,7 @@
     "@opentelemetry/instrumentation": "^0.46.0"
   },
   "peerDependencies": {
-    "@langchain/core": "^1.0.0"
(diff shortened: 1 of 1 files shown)


js/packages/openinference-instrumentation-langchain/tsconfig.esnext.json
```diff
@@ -3,13 +3,10 @@
   "compilerOptions": {
     "outDir": "dist/esnext",
```
cephalization, line 4: I don't actually think that esnext is an accessible dist location. I think we can nuke the build step for esnext. commonjs apps will access the commonjs dist, and esm apps access the esm dist. I don't know which runtimes actually access the esnext dist. We can shrink the package size by quite a bit by dropping esnext builds

PR #2480, opened 2025-11-26
fix: litellm async calls with stream reports token usage and span status

GustavoCaso:
While testing locally Arize, I noticed that when using the `acompletion` function with `stream=True, stream_options={"include_usage": True}` options, the reported traces did not included the count token, and the status was undefined. 

I chekced the code and noticed that the token count issue, was that we pass the `usage_stats` object to the `_set_token_counts_from_usage` directly. When the function checks if the object has the `usage` attribute, it return false and exit. Causing the span to not report token count. 

Screenshot 2025-11-26 at 19 22 51

I modified the code so the object includes the `usage` attribute,  and called the `_set_span_status` function.

After those changes the token count and the status is reported:

Screenshot 2025-11-26 at 19 25 23

---

> [!NOTE]
> Ensure `acompletion`/`completion` streaming reports token counts and sets span status, and add tests covering async streaming and usage reporting.
> 
> - **Instrumentation (LiteLLM)**:
>   - Wrap streaming `usage` with `SimpleNamespace(usage=...)` before calling `_set_token_counts_from_usage` in `_finalize_sync_streaming_span` and `_finalize_streaming_span`.
>   - Set span status via `_set_span_status(span, aggregated_output)` after async streaming completes.
> - **Tests**:
>   - Add async streaming tests for `acompletion` validating output value, status OK, and token counts when `stream_options={"include_usage": True}`.
>   - Cover context attributes in both standard and usage-included async streaming paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 78956e8a973b6f501fdebf2a44ea16f67b646dc5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/tests/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/tests/test_instrumentor.py
@@ -657,6 +657,153 @@ async def test_acompletion(
         )
 
 
+@pytest.mark.parametrize("use_context_attributes", [False, True])
+async def test_acompletion_stream(
+    in_memory_span_exporter: InMemorySpanExporter,
+    setup_litellm_instrumentation: Any,
+    use_context_attributes: bool,
+    session_id: str,
+    user_id: str,
+    metadata: Dict[str, Any],
+    tags: List[str],
+    prompt_template: str,
+    prompt_template_version: str,
+    prompt_template_variables: Dict[str, Any],
+) -> None:
+    in_memory_span_exporter.clear()
+
+    input_messages = [{"content": "What's the capital of China?", "role": "user"}]
+    if use_context_attributes:
+        with using_attributes(
+            session_id=session_id,
+            user_id=user_id,
+            metadata=metadata,
+            tags=tags,
+            prompt_template=prompt_template,
+            prompt_template_version=prompt_template_version,
+            prompt_template_variables=prompt_template_variables,
+        ):
+            response = await litellm.acompletion(
+                model="gpt-3.5-turbo",
+                messages=input_messages,
+                mock_response="Beijing",
+                stream=True,
+            )
+            async for chunk in response:
+                print(chunk)
+    else:
+        response = await litellm.acompletion(
+            model="gpt-3.5-turbo",
+            messages=input_messages,
+            mock_response="Beijing",
+            stream=True,
+        )
+        async for chunk in response:
+            print(chunk)
+
+    spans = in_memory_span_exporter.get_finished_spans()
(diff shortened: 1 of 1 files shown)


GustavoCaso:

GustavoCaso:
@codefromthecrypt  could I get some feedback for theses changes 😄 ?

codefromthecrypt:
I don't have workflow approval, but I think the main step is to add unit test, or another row in the unit test, for this case. 

Main thing is you need to verify the changes are needed, esp the status thing. so you can look at litellm's existing `test_completion_sync_streaming` which needs to be ported for async

there is some code in litellm for async, but basically always the openai code is in the best shape, as that's the most used. you can steal patterns from this python/instrumentation/openinference-instrumentation-openai/tests/openinference/instrumentation/openai/test_instrumentor.py

then you need to run tests the same way they run in CI

codefromthecrypt:
ps one tip is that to run the tests, enter the python directory and do `uvx --with tox-uv tox run -e py313-ci-litellm` this will be the main thing that will ensure CI is happy

GustavoCaso:
Thanks for the feedback @codefromthecrypt.

I added tests 😄 and validated that are passing using `uvx --with tox-uv tox run -e py313-ci-litellm`

GustavoCaso:
@caroger would it be possible to get some feedback on these changes. 

Thanks in advance 😄

hewliyang:
Bump @caroger. This is a pretty straightforward bug affecting streaming sync completions as well.

Here's a monkeypatch for reference.

```py
def _patch_litellm_streaming_token_counts() -> None:
    """
    Patch openinference-instrumentation-litellm to fix streaming token counts.

    The library has a bug where streaming handlers pass `litellm.Usage` directly to
    `_set_token_counts_from_usage`, but that function expects an object WITH a `.usage`
    attribute. This causes token counts to be missing for all streaming responses.

    Fix pending upstream: https://github.com/Arize-ai/openinference/pull/2480
    """
    try:
        import openinference.instrumentation.litellm as litellm_instr
        from litellm.types.utils import Usage

        original_set_token_counts = litellm_instr._set_token_counts_from_usage

        def patched_set_token_counts(span, result):
            # Streaming handlers pass Usage directly, but function expects object with .usage
            if isinstance(result, Usage):
                result = SimpleNamespace(usage=result)
            return original_set_token_counts(span, result)

        litellm_instr._set_token_counts_from_usage = patched_set_token_counts
    except Exception:
        pass
```

PR #2495, opened 2025-12-02
feat(bedrock): Add Routing Classifier Trace event

luke-moehlenbrock:
Adds support for routingClassifierTraces and also splits out the guardrails to pre and post guardrails as separate spans.

---

> [!NOTE]
> Adds `routingClassifierTrace`, separates guardrails into `preGuardrailTrace`/`postGuardrailTrace`, and enhances metadata and input/output extraction across nodes and spans.
> 
> - **Instrumentation (core)**:
>   - `AttributeExtractor`: merge invocation/observation metadata for agent-collaborator I/O; include `routingClassifierTrace` event; improve agent-collaborator I/O attribute construction.
> - **Span assembly**:
>   - `ResponseAccumulator`:
>     - Prioritize node-level chunks; special handling for `agent-collaborator` nodes when extracting parent input/output.
>     - Parse routing-classifier outputs from `rawResponse.content`; fall back sensibly.
>     - Name guardrail node spans as `preGuardrailTrace`/`postGuardrailTrace`; set `Guardrails` span kind/status (ERROR when intervened).
>     - Extract and merge metadata in precedence order; use metadata times to set span start/end.
>   - `TraceCollector`: guardrail-specific unique trace IDs; add chunks to nodes (for metadata propagation); ensure new nodes (incl. guardrail) store chunks/spans correctly.
> - **Tests**:
>   - Add cassette `tests/.../cassettes/test_routing_classifier_trace.yaml` and new `test_routing_classifier_trace`.
>   - Update existing tests to assert orchestration metadata, pre/post guardrail traces, span counts/names, and metadata shapes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 70340c28b780083173c6f0cdde2d5aa5f1606183. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
@@ -168,6 +168,18 @@ def test_tool_calls_with_input_params(
     )
     assert not action_group_span_attributes
 
+    # Check orchestrationTrace span metadata
+    orchestration_span = [span for span in spans if span.name == "orchestrationTrace"][-1]
+    orchestration_attributes = dict(orchestration_span.attributes or {})
+    orchestration_metadata = orchestration_attributes.pop("metadata")
+    assert orchestration_metadata is not None
+    assert isinstance(orchestration_metadata, str)
+    orchestration_metadata_dict = json.loads(orchestration_metadata)
+    assert orchestration_metadata_dict["clientRequestId"] == "edfe2c05-729a-4768-b930-602aece156b0"
+    assert orchestration_metadata_dict["endTime"] == 1.747821025600372e18
+    assert orchestration_metadata_dict["startTime"] == 1.7478210192442092e18
+    assert orchestration_metadata_dict["totalTimeMs"] == 2106
+
 
 @pytest.mark.vcr(
     decode_compressed_response=True,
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
```diff
@@ -168,6 +168,18 @@ def test_tool_calls_with_input_params(
...
     assert not action_group_span_attributes
 
+    # Check orchestrationTrace span metadata
+    orchestration_span = [span for span in spans if span.name == "orchestrationTrace"][-1]
+    orchestration_attributes = dict(orchestration_span.attributes or {})
+    orchestration_metadata = orchestration_attributes.pop("metadata")
+    assert orchestration_metadata is not None
+    assert isinstance(orchestration_metadata, str)
+    orchestration_metadata_dict = json.loads(orchestration_metadata)
+    assert orchestration_metadata_dict["clientRequestId"] == "edfe2c05-729a-4768-b930-602aece156b0"
```
caroger, line 178: I'm wondering if there is a better alternatives to the hard coded expected values used in these test assertions.

In the future, when we need to update/refresh the vcr recording files, these assertions would also be manually updated again, right?

caroger (approved):

python/instrumentation/openinference-instrumentation-bedrock/tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py
luke-moehlenbrock: Yeah I completely agree, it's a huge pain to get the correct hardcoded values here, it might be better just to check if there is a non-null value in these fields. We're not modifying these values after we get them from Bedrock so just checking that they've been added to the span should be good and we don't need to check the exact value.

Issue #1674, opened 2025-05-23
[instrumentation] AWS Strands Agent Instrumentation


mikeldking:
https://aws.amazon.com/blogs/opensource/introducing-strands-agents-an-open-source-ai-agents-sdk/

Note that strands might have native OTEL instrumentation and simply requires normalization via a processor or other.

seanlee10:
Do we have ETA on this?

ryoung562:
Here is a span processor we can use.  Can we create a package for this?

[strands_to_openinference_mapping.py](https://github.com/user-attachments/files/23427935/strands_to_openinference_mapping.py)

so we can do this:

from openinference.instrumentation.strands import StrandsToOpenInferenceProcessor

tracer_provider.add_span_processor(StrandsToOpenInferenceProcessor())

PR #2513, opened 2025-12-11
feat: aws strands instrumentation

Dwij1704:
> [!NOTE]
> Adds a new `openinference-instrumentation-strands` package providing a span processor to transform Strands' GenAI OTEL spans into OpenInference format.
> 
> - Maps span kinds (`invoke_agent`, `execute_event_loop_cycle`, `execute_tool`, `chat`) to `AGENT`/`CHAIN`/`TOOL`/`LLM`
> - Translates attributes (`gen_ai.request.model` → `llm.model_name`, `gen_ai.usage.*` → `llm.token_count.*`, tool fields) and extracts messages to `llm.input_messages`/`llm.output_messages`
> - Handles invocation params, token usage, tags, metadata; mutates spans in-place; strips `gen_ai.*` events post-transform
> - Includes `semantic_conventions` helpers, README with quickstart, CHANGELOG, LICENSE, examples (`basic_agent.py`, `streaming_agent.py`)
> - Adds tests for processor behavior and wiring; project config (`pyproject.toml`, test requirements)
> - Integrates into tox with new `strands`/`strands-latest` environments
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b9106e93e0c1003ace60c01ed9343e0dc69dac7d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
+++ b/python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
@@ -309,7 +309,7 @@ class StrandsToOpenInferenceProcessor(SpanProcessor):
                     if isinstance(item, dict):
                         if "text" in item:
                             text_parts.append(str(item["text"]))
-                        elif "toolUse" in item:
+                        elif "toolUse" in item and isinstance(item["toolUse"], dict):
                             tool_use = item["toolUse"]
                             tool_call = {
                                 "tool_call.id": tool_use.get("toolUseId", ""),
@@ -346,7 +346,7 @@ class StrandsToOpenInferenceProcessor(SpanProcessor):
                 if "text" in content_data:
                     return {"message.role": role, "message.content": str(content_data["text"])}
                 else:
-                    return {"message.role": role, "message.content": str(content_data)}
+                    return {"message.role": role, "message.content": safe_json_dumps(content_data)}
             else:
                 return {"message.role": role, "message.content": str(content_data)}
 
@@ -362,7 +362,7 @@ class StrandsToOpenInferenceProcessor(SpanProcessor):
             if isinstance(item, dict):
                 if "text" in item:
                     text_parts.append(str(item["text"]))
-                elif "toolUse" in item:
+                elif "toolUse" in item and isinstance(item["toolUse"], dict):
                     tool_use = item["toolUse"]
                     tool_call = {
                         "tool_call.id": tool_use.get("toolUseId", ""),


nate-mar:
Hi @Dwij1704 did you look at the span processor approach mentioned in the comment in the related [issue](https://github.com/Arize-ai/openinference/issues/1674). What was the retionale for going with the full instrumentation approach given that it seems like they have a native otel instrumentation?

Dwij1704:
> Hi @Dwij1704 did you look at the span processor approach mentioned in the comment in the related [issue](https://github.com/Arize-ai/openinference/issues/1674). What was the retionale for going with the full instrumentation approach given that it seems like they have a native otel instrumentation?

Yet to check, Will let you know when its ready for review

Dwij1704:
> Hi @Dwij1704 did you look at the span processor approach mentioned in the comment in the related [issue](https://github.com/Arize-ai/openinference/issues/1674). What was the retionale for going with the full instrumentation approach given that it seems like they have a native otel instrumentation?

You're absolutely right - I was unaware about span processor approach, I should have investigated the processor approach first as mentioned in the issue. I've now tested Strands' native OTEL instrumentation, confirmed it's comprehensive, and updated the PR to use a span processor instead of the full instrumentation wrapper. Much cleaner approach!

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        Initialize the processor.
+
+        Args:
+            debug: Whether to log debug information
+        """
+        super().__init__()
+        self.debug = debug
+        self.processed_spans: set[int] = set()
+        self.current_cycle_id: Optional[str] = None
+        self.span_hierarchy: Dict[int, Dict[str, Any]] = {}
```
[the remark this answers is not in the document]
nate-mar: This looks legitimate. Not sure how this would ever be cleared over the lifespan of processor.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+from opentelemetry.sdk.trace import ReadableSpan, SpanProcessor
+from opentelemetry.sdk.trace.export import SpanExportResult
+
+logger = logging.getLogger(__name__)
+
+
+class StrandsToOpenInferenceProcessor(SpanProcessor):
+    """
+    SpanProcessor that converts Strands telemetry attributes to OpenInference format
+    for compatibility with Arize AI, updated for new OpenTelemetry GenAI conventions.
```
nate-mar: this should say compatibility with "open inference" rather than "Arize AI", as technically this is interoperable with any otel compliant backend.

python/instrumentation/openinference-instrumentation-strands/CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.1.0] - 2024-12-12
```
nate-mar: The year is incorrect here

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+    def get_processor_info(self) -> Dict[str, Any]:
+        """
+        Returns information about the processor's capabilities and status.
+
+        Returns:
+            Dict containing processor information.
+        """
+        return {
+            "processor_name": "StrandsToOpenInferenceProcessor",
+            "version": "2.1.0",
```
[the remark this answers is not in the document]
nate-mar: This seems legitimate

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        span_id = span.get_span_context().span_id  # type: ignore[no-untyped-call]
+
+        if span_id in self.span_hierarchy:
+            self.span_hierarchy[span_id]["attributes"] = original_attrs
+
+        try:
+            if "event_loop.cycle_id" in original_attrs:
+                cycle_id_value = original_attrs.get("event_loop.cycle_id")
+                if isinstance(cycle_id_value, str):
+                    self.current_cycle_id = cycle_id_value
```
nate-mar: I'm confused what the purpose of `current_cycle_id`. What is it supposed to do? I don't see it really serving any purpose at the moment.

Doesn't cycle just indicate the type of span kind?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        """Called when a span is started. Track span hierarchy."""
+        span_id = span.get_span_context().span_id  # type: ignore[no-untyped-call]
+        parent_id = None
+
+        if parent_context and hasattr(parent_context, "span_id"):
+            parent_id = parent_context.span_id
+        elif span.parent and hasattr(span.parent, "span_id"):
+            parent_id = span.parent.span_id
+
+        self.span_hierarchy[span_id] = {
```
nate-mar: I know some of this was just copied from a previous proposed implementation, but can't we just get access to the parent id from the ReadableSpan in `on_end`? That way we don't need to worry about tracking any kind of state. Furthermore, start_time should already be included when the span is created.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: This also is odd because we shouldn't be tracking private state on the span processor to track per span information, given that spans can be processed concurrently.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        # Handle tags (both Strands arize.tags and standard tag.tags)
+        self._handle_tags(attrs, result)
+
+        # Handle different span types
+        if span_kind in ["LLM", "AGENT", "CHAIN"]:
+            self._handle_llm_span(attrs, result, input_messages, output_messages)
+        elif span_kind == "TOOL":
+            self._handle_tool_span(attrs, result, events)
+
+        # Handle token usage
```
nate-mar: In general, I think these kinds of comments add fluff since the function names are sufficiently descriptive.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        self._map_token_usage(attrs, result)
+
+        # Important attributes
+        important_attrs = [
```
nate-mar: Same idea here re: the comment. Also, "important" isn't really that meaningful here, in that I'm not sure what that conveys about these attrs vs other attributes.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+                    if isinstance(completion_data, list):
+                        # Handle Strands completion format
+                        message = self._parse_strands_completion(completion_data)
+                        if message:
+                            output_messages.append(message)
+                except json.JSONDecodeError:
+                    # Simple string completion
+                    output_messages.append(
+                        {"message.role": "assistant", "message.content": str(completion)}
+                    )
```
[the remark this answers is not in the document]
nate-mar: Though it is an edge case, this appears to be a legitimate comment.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+                            # Set role to tool for tool results and include tool call ID
+                            message["message.role"] = "tool"
+                            if "toolUseId" in tool_result:
+                                message["message.tool_call_id"] = tool_result["toolUseId"]
+
+                message["message.content"] = " ".join(text_parts) if text_parts else ""
+
+                # Clean up empty tool_calls
+                if not message["message.tool_calls"]:
+                    del message["message.tool_calls"]
```
nate-mar: This  feels a bit unnecessary and should be handled in the flow of the code above rather than a fallback, which makes the update logic flow not super straightforward. That is, why would "message.tool_calls" be empty? if they are empty at any point, shouldn't we just avoid adding it at the appropriate spot?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+                elif "toolUse" in item:
+                    tool_use = item["toolUse"]
+                    tool_call = {
+                        "tool_call.id": tool_use.get("toolUseId", ""),
+                        "tool_call.function.name": tool_use.get("name", ""),
+                        "tool_call.function.arguments": json.dumps(tool_use.get("input", {})),
+                    }
+                    message["message.tool_calls"].append(tool_call)
+
+        message["message.content"] = " ".join(text_parts) if text_parts else ""
```
nate-mar: Shouldn't we just not even add message.content as a key if it's empty?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+
+        # Clean up empty arrays
+        if not message["message.tool_calls"]:
```
nate-mar: Similar comment to the one above

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        attrs: Dict[str, Any],
+        result: Dict[str, Any],
+        input_messages: List[Dict[str, Any]],
+        output_messages: List[Dict[str, Any]],
+    ) -> None:
+        """Handle LLM/Agent span with extracted messages."""
+
+        # Create message arrays
+        if input_messages:
+            result["llm.input_messages"] = json.dumps(input_messages, separators=(",", ":"))
```
nate-mar: Can we do a `safe_json_dumps` here so we don't crash accidentally? You can find the function in the openinference.instrumentation.helpers package

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+
+            # Create output.value
+            if output_messages:
+                last_message = output_messages[-1]
+                content = last_message.get("message.content", "")
+
+                if span_kind == "LLM":
+                    # LLM format
+                    output_structure = {
+                        "choices": [
```
nate-mar, line 461: Just clarifying for my benefit: Is there always only one choice?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        elif span_name.startswith("Tool:"):
+            return "TOOL"
+        elif "Cycle" in span_name:
+            return "CHAIN"
+        elif attrs.get("gen_ai.agent.name") or attrs.get("agent.name"):
+            return "AGENT"
+
+        return "CHAIN"
+
+    def _set_graph_node_attributes(
```
nate-mar, line 517: nit: for readability, I'd generally prefer that we return a dict or similar that contains the list of updated attributes, and then merge it with the result outside of this function so it's easier to reason about what's changing in this function.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+        # Get parent information from span hierarchy
+        span_info = self.span_hierarchy.get(span_id, {})
+        parent_id = span_info.get("parent_id")
+        parent_info = self.span_hierarchy.get(parent_id, {}) if parent_id else {}
+        parent_name = parent_info.get("name", "")
+
+        if span_kind == "AGENT":
+            result["graph.node.id"] = "strands_agent"
+        elif span_kind == "CHAIN":
+            if span_name == "execute_event_loop_cycle":
```
nate-mar, line 536: Can we try to understand and document what execute event loop cycle is and how it's needed in the context of graph nodes? I presume this is why span hierarchy needs to be tracked, but it's a bit of an unfortunate proposition to have to try to track historical spans because we dont' know how long they last before we might have to clean them out. Ideally, there'd be some way that we can extract the necessary information in a stateless way.

nate-mar: It looks like we could potentially omit graph.node.parent_id if it would mean needing to track state about other spans in the processor, which I think we should definitely try to avoid for a number of reasons. It seems like that attribute may be [optional](https://arize.com/docs/ax/observe/tracing/agents) anyway.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Removed: Deleted the variable entirely as it served no practical function

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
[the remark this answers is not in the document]
Dwij1704: Fixed: Removed span_hierarchy dict and on_start tracking. Now using stateless approach.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fixed: Only add message.content and message.tool_calls when non-empty

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fixed: Refactored to conditionally add keys only when values exist

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fixed: Imported and used safe_json_dumps() for all 14 json.dumps() calls

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Strands emits single choice per LLM call; index 0 per OpenTelemetry GenAI semantic conventions

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fixed: _set_graph_node_attributes() now returns Dict[str, Any] instead of mutating in-place

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: The processor is now fully stateless:
Removed all state tracking: span_hierarchy, current_cycle_id, and processed_spans have all been removed

Omitted graph.node.parent_id for LLM/TOOL spans -- Since reliably determining parent info requires state tracking, these are now omitted per your suggestion that the attribute is optional

The processor no longer maintains any cross-span state, eliminating memory leak concerns and concurrency issues. Documented execute_event_loop_cycle: Added comment in _set_graph_node_attributes() explaining it's Strands' agentic loop iteration where the LLM reasons, plans, and optionally selects tools

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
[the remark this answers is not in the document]
Dwij1704: Fixed. Now handles non-list JSON completion data:
Dict with text key: extracts text directly
Dict without text: serializes as JSON content
Other JSON types: converts to string
Previously only lists were processed, silently dropping valid dict completions.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+        """Called when a span is started."""
+        pass
+
+    def on_end(self, span: ReadableSpan) -> None:
+        """
+        Called when a span ends. Transform the span attributes from Strands format
+        to OpenInference format.
+        """
+        if not hasattr(span, "_attributes") or not span._attributes:
+            return
```
nate-mar, line 74: ReadableSpan should have an available `attributes` attribute that you can use, so that we can avoid coupling to private attributes.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+        try:
+            events: List[Any] = []
+            if hasattr(span, "_events"):
+                events = list(span._events)
+            elif hasattr(span, "events"):
+                events = list(span.events)
+
+            transformed_attrs = self._transform_attributes(original_attrs, span, events)
+            span._attributes.clear()  # type: ignore[attr-defined]
+            span._attributes.update(transformed_attrs)  # type: ignore[attr-defined]
```
nate-mar: I think this brings up an interesting discussion as to whether it's more appropriate to use an exporter vs a processor for this kind of transformation/mapping.

Arguably, an exporter is more appropriate because you aren't changing the fundamental data in a span for things like redaction, but rather trying to transform your span data for a specific backend. However, if you look at the Exporter interface in Python, it takes in a sequence of readable spans, so they still are immutable spans, though you could deep copy, then transform and re-set them the Sequence.

I think given that there is ongoing conversations  (see: https://github.com/open-telemetry/opentelemetry-python/issues/4424  and https://github.com/open-telemetry/opentelemetry-specification/issues/1089) about the appropriate way to handle this in the SDK vs a collector, I'm ok with still doing this in a processor for now, though I think long term an exporter probably makes more sense.

However, for this Python implementation, we should at the very least avoid accessing private attributes of the ReadableSpan. I'm pretty sure they expose an `attributes` attribute on the ReadableSpan, and then you can modify the key/values as done in an example like [this](https://github.com/open-telemetry/opentelemetry-js/discussions/2817),

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+                    span.name,
+                    len(original_attrs),
+                    len(transformed_attrs),
+                )
+                logger.info("events=<%d> | processed events", len(events))
+
+        except Exception as e:
+            logger.error(f"Failed to transform span '{span.name}': {e}", exc_info=True)
+            span._attributes.clear()  # type: ignore[attr-defined]
+            span._attributes.update(original_attrs)  # type: ignore[attr-defined]
```
nate-mar: Same thing here, I'd prefer if we not operate on private attributes if we can avoid it

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+
+    def _transform_attributes(
```
nate-mar, line 120: Do we need both attires and span here since attrs comes from the span?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+                )
+            else:
+                input_messages, output_messages = [], []
+
+        model_id = attrs.get("gen_ai.request.model")
+        agent_name = attrs.get("agent.name") or attrs.get("gen_ai.agent.name")
+
+        if model_id:
+            result["llm.model_name"] = model_id
+            result["gen_ai.request.model"] = model_id
```
nate-mar: Can we look into using the genai package here to get the constants, so that we can immediately follow up if there is breakage with their library changes, especially since it's still kinda experimental? See the pydantic instrumentation for an example.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+                else event.get("attributes", {})
+            )
+
+            if event_name == "gen_ai.user.message":
+                content = event_attrs.get("content", "")
+                message = self._parse_message_content(content, "user")
+                if message:
+                    input_messages.append(message)
+
+            elif event_name == "gen_ai.assistant.message":
```
nate-mar: As mentioned in a comment above, it seems like we reference some of these convention constants in the Pedantic instrumentation. I think there is an opportunity here for us to maybe pull out a few of the these constants and put them into a helper package like the base openinference-instrumentation package so that the mapping can be shared/consistent across all implementations..

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Ok, my mistake; I was not looking at a Python example. 

You're right here, but if we're modifying private attributes anyway, can we just set it directly? `span._attributes = transformed_attributes`?

The spec does have info about a on span end hook, but it's unfortunately. not implemented yet, as mentioned in one of the above links.

nate-mar:
Please remember to take a look at the auto generated review comments as well. Thanks.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: See above revised comment.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: See below updated comment.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Acknowledged as valid architectural concern, but appropriate for now given ongoing OTEL discussions

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Yes, we need both. While attrs comes from span, we extract it once at the start and use it throughout transformation to avoid repeated attribute access.

Dwij1704:
> Please remember to take a look at the auto generated review comments as well. Thanks.

Thanks for the review, pushed a bunch of fixes catering your and other reviews! Adding helper for shared semconv is definitely helpful in the longer run

python/openinference-instrumentation/src/openinference/instrumentation/gen_ai_attributes.py
```diff
@@ -0,0 +1,81 @@
+"""
```
nate-mar: Can we get these updated attributes pushed in a separate PR so it's separate from the strands changes? I think there will be some other folks invested in specifically these changes vs the strands processor.

Dwij1704: Makes sense, #2568

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+                            message_data = (
+                                json.loads(message) if isinstance(message, str) else message
+                            )
+                            if isinstance(message_data, list):
+                                text_parts = []
+                                for item in message_data:
+                                    if isinstance(item, dict) and "text" in item:
+                                        text_parts.append(item["text"])
+                                tool_output = (
+                                    " ".join(text_parts) if text_parts else str(message_data)
```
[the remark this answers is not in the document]
Dwij1704: Fixed at line 573. Added str() wrapper to match all other similar code paths (lines 277, 294, 327, 682).

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+                    prompt_data = json.loads(prompt)
+                    if isinstance(prompt_data, list):
+                        for msg in prompt_data:
+                            normalized = self._normalize_message(msg)
+                            if normalized.get("message.role") == "user":
+                                input_messages.append(normalized)
+                    elif isinstance(prompt_data, dict):
+                        # Handle single dict prompt (e.g., {"role": "user", "content": "Hello"})
+                        normalized = self._normalize_message(prompt_data)
+                        input_messages.append(normalized)
```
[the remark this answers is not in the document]
Dwij1704: Design decision. List filtering keeps only user messages; single dicts are treated as complete input. Could unify if needed.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+                            if "toolUseId" in tool_result:
+                                message["message.tool_call_id"] = tool_result["toolUseId"]
+
+                if text_parts:
+                    message["message.content"] = " ".join(text_parts)
+
+                if tool_calls:
+                    message["message.tool_calls"] = tool_calls
+
+                return message
```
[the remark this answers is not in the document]
Dwij1704: Valid inconsistency. _parse_message_content returns a message with just role while _parse_strands_completion returns None. Would need design decision on desired behavior.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+            "temperature": "temperature",
+            "top_p": "top_p",
+        }
+
+        for key, param_key in param_mappings.items():
+            if key in attrs:
+                params[param_key] = attrs[key]
+
+        if params:
+            result["llm.invocation_parameters"] = safe_json_dumps(params)
```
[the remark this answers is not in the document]
Dwij1704: Fixed. Now uses gen_ai.request.max_tokens, gen_ai.request.temperature, gen_ai.request.top_p.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+            if isinstance(content, list):
+                text_parts = []
+                for item in content:
+                    if isinstance(item, dict) and "text" in item:
+                        text_parts.append(str(item["text"]))
+                result["message.content"] = " ".join(text_parts) if text_parts else ""
+            else:
+                result["message.content"] = str(content)
+
+        return result
```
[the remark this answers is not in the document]
Dwij1704: Fixed. Dict messages now default to "user" role via msg.get("role", "user").

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,754 @@
...
+            ("gen_ai.usage.prompt_tokens", "llm.token_count.prompt"),
+            ("gen_ai.usage.input_tokens", "llm.token_count.prompt"),
+            ("gen_ai.usage.completion_tokens", "llm.token_count.completion"),
+            ("gen_ai.usage.output_tokens", "llm.token_count.completion"),
+            ("gen_ai.usage.total_tokens", "llm.token_count.total"),
+        ]
+
+        for strands_key, openinf_key in token_mappings:
+            if value := attrs.get(strands_key):
+                result[openinf_key] = value
```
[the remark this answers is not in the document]
Dwij1704: Not a bug. The code uses if value is not None: which correctly preserves zero values. No walrus operator is used here.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,777 @@
...
+                    if message:
+                        try:
+                            message_data = (
+                                json.loads(message) if isinstance(message, str) else message
+                            )
+                            if isinstance(message_data, list):
+                                text_parts = []
+                                for item in message_data:
+                                    if isinstance(item, dict) and "text" in item:
+                                        text_parts.append(item["text"])
```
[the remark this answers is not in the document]
Dwij1704: Fixed. Added str(item["text"]) to match other code paths.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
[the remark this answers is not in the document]
Dwij1704: _parse_message_content now returns None when list content yields no text or tool_calls, matching _parse_strands_completion behavior.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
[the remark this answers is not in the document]
Dwij1704: Single dict prompts now apply the same role == "user" filter as list prompts for consistent behavior.

python/openinference-instrumentation/src/openinference/instrumentation/gen_ai_attributes.py
Dwij1704: Reverted these changes, considering instability of OTEL semconv

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+
+        if span_kind == "AGENT":
+            graph_attrs["graph.node.id"] = "strands_agent"
+        elif span_kind == "CHAIN":
+            # execute_event_loop_cycle: Strands' agentic loop iteration where the LLM
+            # reasons, plans, and optionally selects tools. Each cycle is a child of
+            # the agent span.
+            if span_name == "execute_event_loop_cycle":
+                cycle_id = attrs.get("event_loop.cycle_id", f"cycle_{span_id}")
+                graph_attrs["graph.node.id"] = f"cycle_{cycle_id}"
```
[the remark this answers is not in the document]
Dwij1704: Valid but edge case only. Applied a fix

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+
+    def _transform_attributes(
+        self, attrs: Dict[str, Any], span: ReadableSpan, events: Optional[List[Any]] = None
+    ) -> Dict[str, Any]:
+        """
+        Transform Strands attributes to OpenInference format, including event processing.
+        """
+        result: Dict[str, Any] = {}
+        span_kind = self._determine_span_kind(span, attrs)
+        result["openinference.span.kind"] = span_kind
```
nate-mar: Can we use the OI span kind constant here?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                    prompt, completion
+                )
+            else:
+                input_messages, output_messages = [], []
+
+        model_id = attrs.get(GEN_AI_REQUEST_MODEL)
+        agent_name = attrs.get("agent.name") or attrs.get(GenAIAttributes.AGENT_NAME)
+
+        if model_id:
+            result["llm.model_name"] = model_id
```
nate-mar: can we use the constants here since we're already importing the semantic conventions?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/__init__.py
```diff
@@ -0,0 +1,43 @@
+"""OpenInference Instrumentation for Strands Agents.
+
+This package provides a span processor that transforms Strands' native OpenTelemetry
+spans (using GenAI semantic conventions) into OpenInference format for compatibility
+with Arize AI.
```
nate-mar: This should state that it has compatibility with openinference supported otel backends.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+            return value
+
+        try:
+            return json.dumps(value, separators=(",", ":"))
+        except (TypeError, OverflowError):
+            return str(value)
+
+    def shutdown(self) -> SpanExportResult:  # type: ignore[override]
+        """Called when the processor is shutdown."""
+        return SpanExportResult.SUCCESS
```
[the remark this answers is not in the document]
nate-mar: Is this not still an issue of the return type mismatching?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                f"Migrate to event-based messaging using {GenAIEventNames.ASSISTANT_MESSAGE} "
+                f"and {GenAIEventNames.CHOICE} events"
+            ),
+            # Span naming changes
+            "Model invoke": "chat",
+            "Cycle [UUID]": "execute_event_loop_cycle",
+            "Tool: [name]": "execute_tool [name]",
+        }
+
+    def get_processor_info(self) -> Dict[str, Any]:
```
nate-mar: What's the purpose of this? Can we not include it unless we know what it's usage will be?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,844 @@
...
+
+            elif event_name == "gen_ai.tool.message":
+                # Tool messages - treat as user messages with tool role
+                content = event_attrs.get("content", "")
+                tool_id = event_attrs.get("id", "")
+                if content:
+                    message = self._parse_message_content(content, "tool")
+                    if message and tool_id:
+                        message["message.tool_call_id"] = tool_id
+                        input_messages.append(message)
```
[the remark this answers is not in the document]
nate-mar: Can you please provide the reasoning for ignoring this feedback even if it is invalid?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+
+        model_id = attrs.get(GEN_AI_REQUEST_MODEL)
+        agent_name = attrs.get("agent.name") or attrs.get(GenAIAttributes.AGENT_NAME)
+
+        if model_id:
+            result["llm.model_name"] = model_id
+            result[GEN_AI_REQUEST_MODEL] = model_id
+
+        if agent_name:
+            result["llm.system"] = "strands-agents"
```
nate-mar: Should this not pull from  `gen_ai.system`?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+            result["llm.provider"] = "strands-agents"
```
nate-mar: Curious, is this always going to be strands-agents? I would assume so, but genai has this `gen_ai.provider.name`. Should we not pull from that if it is there?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+            completion = attrs.get(GenAIAttributes.COMPLETION)
+            if prompt or completion:
+                input_messages, output_messages = self._extract_messages_from_attributes(
+                    prompt, completion
+                )
+            else:
+                input_messages, output_messages = [], []
+
+        model_id = attrs.get(GEN_AI_REQUEST_MODEL)
+        agent_name = attrs.get("agent.name") or attrs.get(GenAIAttributes.AGENT_NAME)
```
nate-mar: Shouldn't we check GenAIAttributes.AGENT_NAME first? looks  like how you've done it over here too https://github.com/Arize-ai/openinference/pull/2513/files#diff-0f2492802a5f3026f0348d09a3d18b48d60d5518e01641840f74ac2815cb4434R479

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+        return result
+
+    def _add_metadata(self, attrs: Dict[str, Any], result: Dict[str, Any]) -> None:
+        metadata = {}
+        skip_keys = {
+            GenAIAttributes.PROMPT,
+            GenAIAttributes.COMPLETION,
+            GenAIAttributes.AGENT_TOOLS,
+            "agent.tools",
+        }
```
nate-mar, line 753: can you provide some context here as to why these are skipped?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                            tool_output = str(message)
+
+            if tool_parameters:
+                result["tool.parameters"] = safe_json_dumps(tool_parameters)
+
+                if tool_name and tool_call_id:
+                    input_messages = [
+                        {
+                            "message.role": "assistant",
+                            "message.content": "",
```
nate-mar, line 631: is this always going to be empty?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                for item in content:
+                    if isinstance(item, dict) and "text" in item:
+                        text_parts.append(str(item["text"]))
+                result["message.content"] = " ".join(text_parts) if text_parts else ""
+            else:
+                result["message.content"] = str(content)
+
+        return result
+
+    def _add_metadata(self, attrs: Dict[str, Any], result: Dict[str, Any]) -> None:
```
nate-mar: This is more a general note based on noticing, but functions like these IMO I think read more easily if we rename these variables these to be a bit more precise about what `attrs` and `result` represents; from what I'm reading, attrs is the original set of attributes, not the "to be returned" attrs, and the final attributes result set you are building up to return. 

Also, I personally think this is easier to read if it's done more functionally, where the return object is the serialized metadata, and it's also less dependent on where this function is run (as written the effect of the function is dependent on the current state of the often changing `result`); essentially that is what it looks like you re doing below but this function kind of obscures that for me so I have to dig in to understand what you are adding.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                result["tag.tags"] = tags
+            elif isinstance(tags, str):
+                result["tag.tags"] = [tags]
+
+    def _determine_span_kind(self, span: ReadableSpan, attrs: Dict[str, Any]) -> str:
+        span_name = span.name
+
+        if span_name == "chat":
+            return "LLM"
+        elif span_name.startswith("execute_tool "):
```
nate-mar, line 500: is the trailing space intentional?

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+                    result["output.mime_type"] = "application/json"
+                else:
+                    result["output.value"] = content
+                    result["output.mime_type"] = "text/plain"
+
+    def _handle_tags(self, attrs: Dict[str, Any], result: Dict[str, Any]) -> None:
+        tags = None
+
+        if "arize.tags" in attrs:
+            tags = attrs["arize.tags"]
```
nate-mar: What is the purpose of the arize specific tags here? not sure they should be needed given this is openinference.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+        original_attrs = dict(span._attributes)
+
+        try:
+            events: List[Any] = []
+            if hasattr(span, "_events"):
+                events = list(span._events)
+            elif hasattr(span, "events"):
+                events = list(span.events)
+
+            transformed_attrs = self._transform_attributes(original_attrs, span, events)
```
luke-moehlenbrock, line 85: Should we also strip events with the gen_ai prefix from the span events after transforming attributes? Arize uses span events strictly for displaying exception messages and stacktraces, and as far as I know Phoenix does as well. So with the gen_ai events, both platforms show that there are events, but they're empty except for the name.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,782 @@
...
+            span._attributes = transformed_attrs
```
luke-moehlenbrock, line 86: I'm also wondering if we should accept a span exporter in `__init__` and basically deepcopy to a new ReadableSpan here and then export that span instead of mutating in-place.

Hypothetically, if a user wanted to export GenAI spans to one provider and OpenInference spans to another, they would have to be careful with the order of their span processors - it would need to look something like:
1. GenAI processor -> GenAI exporter
2. StrandsToOpenInferenceProcessor
3. OpenInference processor -> Openinference exporter

Also if the GenAI exporter is a batch exporter, there's a chance that part of the batch will actually be translated to OpenInference before it's exported due to the asynchronous nature of batch exporting. 

Not saying this would be a common use case or something we should support, it just came up when I was testing this out with Arize and noticed that the spans were being translated even though I accidentally set up the processors in backwards order with this example:
```python
arize_processor = BatchSpanProcessor(space_id=SPACE_ID, api_key=API_KEY)
strands_processor = StrandsToOpenInferenceProcessor()

tracer_provider.add_span_processor(arize_processor)
tracer_provider.add_span_processor(strands_processor)
```
@nate-mar curious if you have any thoughts on this

nate-mar: Good call out

So this is kind of the dilemma with using an exporter vs a processor. With an exporter, I don't know if you have to necessarily worry about this coupling, because you can just send whatever data you want directly to the exporter.

Because this is a processor, I think we could certainly try that but I don't know if there are going to be any other side effects of copying, given that according to the spec, you shouldn't be creating ReadableSpans. I know we're already modifying it, so we've kinda crossed that boundary already, but it's worth a shot if we want to see if it will work, or alternatively, create an exporter version as well using the same building blocks here.

luke-moehlenbrock: We could also keep it all in the same class, just add an optional `span_exporter` arg to the `__init__` function, then do something like
```
if self._span_exporter is None:
  span._attributes = transformed_attrs
else:
  copied_span = copy_readable_span(span, transformed_attrs)
  self._span_exporter.export(copied_span)
```
Although with a separate exporter version I feel like it's a bit more clear that one is mutating spans and the other is copying.

Another thing I hadn't really considered is that adding an exporter essentially makes this a simple span processor, would we need to build a batch version as well? At least with the in-place mutation you can chain it together with whatever processor you want.

Dwij1704:
Pushed fixes for all the review comments:

**OI span kind constant (line 98)** - Done, now using `SpanAttributes.OPENINFERENCE_SPAN_KIND`

**Constants for semantic conventions (line 118)** - Done, switched to `SpanAttributes.LLM_MODEL_NAME`, `SpanAttributes.LLM_SYSTEM`, etc.

**__init__.py compatibility statement** - Updated to "OpenInference-supported OTEL backends (Phoenix, Arize, etc.)"

**shutdown return type** - Fixed to return `None` instead of `SpanExportResult.SUCCESS`

**get_processor_info** - Removed it since there's no concrete use case

**gen_ai.system / gen_ai.provider.name** - Checked the Strands SDK - they hardcode both to "strands-agents" (see tracer.py:770-795), so the value would be the same either way. Happy to add a fallback check if you prefer.

**GenAIAttributes.AGENT_NAME first** - Fixed, now checks `gen_ai.agent.name` before falling back to `agent.name`

**Skipped keys comment** - Added explanation that these are already processed into OI format (prompt/completion -> messages, tools -> llm.tools.{idx}.*)

**Empty message.content for tool spans** - Added comment explaining this is intentional - tool call requests don't have text content, the tool_calls array IS the payload

**_add_metadata variable naming** - Fair point about clarity. Can rename if you want, but the pattern is consistent with other _handle_* methods in the class

**Trailing space** - Couldn't find one at line 467? Let me know if you spot it somewhere specific

**arize.tags** - Removed, now only checks `tag.tags`

**Strip gen_ai events** - Added `_strip_genai_events()` to filter out gen_ai.* events after transformation

**In-place mutation discussion** - Replied in the thread, but tl;dr: keeping current simple approach and documenting processor ordering. The pydantic-ai processor uses the same pattern.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Good discussion. The in-place mutation is definitely a known trade-off.

Looking at the pydantic-ai processor, they take the same approach - mutating `span._attributes` directly in `on_end`. So this pattern is consistent with what we're doing elsewhere in the codebase.

Luke's point about processor ordering is valid. If someone sets up:
```
arize_processor -> BatchExporter
strands_processor (mutates in place)
```
The batch exporter might see partially transformed spans due to async timing. That's a real gotcha.

A few thoughts:

1. **Exporter approach**: We could add an optional `span_exporter` param like Luke suggested. But then we'd need to handle batching ourselves or force users to wrap it with BatchSpanProcessor separately. Adds complexity.

2. **Copy approach**: Making a deep copy of ReadableSpan feels hacky since we're not supposed to construct them directly per the spec. We're already bending the rules by mutating `_attributes`.

3. **Keep it simple for now**: The current approach works fine for the typical use case (single export destination). We should document the processor ordering requirement clearly. Something like:

   > This processor mutates spans in-place. Add it BEFORE any span processors that export spans you want transformed.

I'd lean toward keeping the current simple implementation and adding clear docs about ordering. We can always add an exporter variant later if there's demand for multi-destination setups.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Good call - added _strip_genai_events() that filters out events with gen_ai.* prefix after transformation. The content is already extracted into OI attributes so keeping the original events would just clutter the UI.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Done - now using SpanAttributes.OPENINFERENCE_SPAN_KIND

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Done - switched to SpanAttributes.LLM_MODEL_NAME, SpanAttributes.LLM_SYSTEM, SpanAttributes.LLM_PROVIDER, etc.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/__init__.py
Dwij1704: Updated to "OpenInference-supported OTEL backends (Phoenix, Arize, etc.)"

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Agreed, removed it. No concrete use case for it right now.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Checked the Strands SDK - they hardcode gen_ai.system to "strands-agents" in tracer.py. The value is always the same so pulling from it vs hardcoding makes no difference. Can add the lookup if you prefer though.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Same situation - Strands always sets gen_ai.provider.name to "strands-agents" as well (see tracer.py:786-787). Both attributes are always identical so it doesnt matter which we read from.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fixed - now checks `GenAIAttributes.AGENT_NAME` (gen_ai.agent.name) first before falling back to agent.name.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Added comment explaining why these are skipped. These keys have already been processed into OI format (gen_ai.prompt/completion → llm.input_messages/llm.output_messages, tools → llm.tools.{idx}.*), so including them in metadata would be redundant.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Yes, this is intentional. For tool spans, the assistant message contains only the tool_calls array - there's no text content because the tool call itself IS the message payload. Added a comment to clarify this.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Removed - now only checks `SpanAttributes.TAG_TAGS` (tag.tags). The arize.tags was unnecessary.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: Fair point about clarity. The naming convention (attrs=original, result=transformed) is consistent with other _handle_* methods in this class, but I can see how it's not immediately obvious. Happy to rename if you feel strongly - maybe `original_attrs` and `transformed_attrs`? Though that might get verbose in the method signatures.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
Dwij1704: The trailing space in `"execute_tool "` is intentional. Strands formats tool span names as `execute_tool <tool_name>` (e.g., `execute_tool calculator`), so we need the space to correctly match the prefix pattern. Without it, we might incorrectly match something like `execute_tool_other`.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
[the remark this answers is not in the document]
Dwij1704: I might have resolved this mistakenly, Overwhelming comments from cursor makes it harder to track when I am pushing fixes

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Yeah agreed, let's keep it as is since there is precedent for this + documenting it until there's a more definitive need.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Gotcha, thanks

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Not a huge deal to me but if they do decide to change system, then it'd be nice for this to be updated.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Ok got it. Was just wondering if there would be optional text in addition to the tool call in some cases.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: Sorry my main question is that couldn't you just pull attrs from the beginning of this function so you don't need attrs as a function argument.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
nate-mar: I don't feel necessary strongly about the renaming, if the function signature makes it clearer what the input and output. Fair point about verbosity, but In lieu of that, maybe keep `attrs` as is and then `result` becomes `final_attrs`?

Dwij1704: Changed 'result' parameter to 'final_attrs' in _add_metadata method to better reflect its purpose.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,787 @@
...
+                        if "text" in item:
+                            text_parts.append(str(item["text"]))
+                        elif "toolUse" in item:
+                            tool_use = item["toolUse"]
+                            tool_call = {
+                                "tool_call.id": tool_use.get("toolUseId", ""),
+                                "tool_call.function.name": tool_use.get("name", ""),
+                                "tool_call.function.arguments": safe_json_dumps(
+                                    tool_use.get("input", {})
+                                ),
```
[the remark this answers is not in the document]
Dwij1704: Added type check for toolUse to confirm it is a dictionary before accessing its properties.

python/instrumentation/openinference-instrumentation-strands/src/openinference/instrumentation/strands/processor.py
```diff
@@ -0,0 +1,787 @@
...
+
+                if "message.content" not in message and "message.tool_calls" not in message:
+                    return None
+
+                return message
+            elif isinstance(content_data, dict):
+                if "text" in content_data:
+                    return {"message.role": role, "message.content": str(content_data["text"])}
+                else:
+                    return {"message.role": role, "message.content": str(content_data)}
```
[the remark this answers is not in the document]
Dwij1704: Updated return statement to use safe_json_dumps for content_data to ensure proper serialization.

PR #2515, opened 2025-12-12
feat: google genai instrumentation

### LINKED ISSUE #2321: [feature request] Add in autoinstrumentator for Google GenAI for JS/TS
Users would like to autoinstrument their Google GenAI applications

An autoinstrumentor similar to the python version for JS/TS

Manual OTEL based instrumentation of Google GenAI

Dwij1704:
> [!NOTE]
> Introduces a new OpenInference instrumentation package for Google GenAI (`@arizeai/openinference-instrumentation-google-genai`).
> 
> - Implements `GoogleGenAIInstrumentation` to trace `@google/genai` APIs: `models.generateContent`, `models.generateContentStream`, `models.generateImages`, `chats.create`, `chat.sendMessage`, `chat.sendMessageStream`, and `batches.createEmbeddings`
> - Adds helper `createInstrumentedGoogleGenAI` for easy instance-based instrumentation (ESM-friendly), plus masking via `traceConfig`, context propagation, and tracing suppression
> - Captures OpenInference semantic attributes (model, messages, tools, token usage) and supports multimodal and streaming
> - Adds README, CHANGELOG, examples, tests (Vitest), build configs, and package metadata; updates lockfile
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f072f3c34cc634dafafe6ffed1c84f4e198297f6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,822 @@
...
+
+            span.setStatus({ code: SpanStatusCode.OK });
+            span.end();
+
+            return (async function* () {
+              for (const chunk of chunks) {
+                yield chunk;
+              }
+            })();
+          });
```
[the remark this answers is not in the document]
Dwij1704: Added try/catch around the stream iteration. Errors now properly record exception, set error status, and end the span before re-throwing.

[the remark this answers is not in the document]
Dwij1704: Removed the buffering pattern. The generator now yields chunks directly as they arrive from the upstream API, enabling real-time progressive output.

js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,934 @@
...
+                          [SemanticConventions.OUTPUT_MIME_TYPE]: MimeType.JSON,
+                          ...getOutputMessagesAttributes(lastChunk),
+                          ...getUsageAttributes(lastChunk),
+                        });
+                      }
+                      span.setStatus({ code: SpanStatusCode.OK });
+                    }
+                    span.end();
+                  }
+                })();
```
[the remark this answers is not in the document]
Dwij1704: This is a theoretical edge case where the user requests a stream but never iterates it.

The current implementation correctly handles:

- Full stream consumption
- Partial consumption (break/return in for-await-of)
- Errors during iteration
- Initial API call failures

Addressing the "never iterate" case would require either unreliable GC-based cleanup (FinalizationRegistry) or arbitrary timeouts that could incorrectly close spans during legitimate slow consumption. Other OpenTelemetry instrumentation libraries accept this same limitation for streaming APIs. If a user awaits a stream and never consumes it, the span correctly reflects an incomplete operation.

js/packages/openinference-instrumentation-google-genai/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+ * });
+ *
+ * // Use ai as normal - all calls are automatically traced
+ * const response = await ai.models.generateContent({
+ *   model: "gemini-2.5-flash",
+ *   contents: "Hello!",
+ * });
+ * ```
+ */
+export function createInstrumentedGoogleGenAI(
```
cephalization: Is this necessary? It breaks convention with how we do instrumentation for other packages such as openai or anthropic.

Typically the consumer should be able to import their client, in this case google genai, how they normally would.

Our instrumentation should patch it when the app starts, modifying all future imports of the client so that users can quickly adopt the instrumentation without modifying their source code too much

js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,934 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    models: any,
+    instrumentation: GoogleGenAIInstrumentation,
+  ) {
+    // Wrap generateContent
+    if (models.generateContent) {
+      const originalGenerateContent = models.generateContent.bind(models);
+      models.generateContent = async function patchedGenerateContent(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ...genArgs: any[]
```
cephalization: Is it possible to infer the type of these arguments from generateContent?

cephalization: I am mostly concerned by this since we index into the args and try to pull values out. I am not so worried about cases where we try to just stringify instead.

You could also set this to `unknown[]` so at least you have to defensively explore the arguments with less of a chance of runtime errors later

js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,934 @@
...
+    }
+
+    // Wrap generateContentStream
+    if (models.generateContentStream) {
+      const originalGenerateContentStream =
+        models.generateContentStream.bind(models);
+      models.generateContentStream =
+        async function patchedGenerateContentStream(
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ...genArgs: any[]
```
cephalization: Same here, would prefer unknown to any if we can't infer the real types

js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,934 @@
...
+      traceConfig,
+    });
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  protected init(): InstrumentationModuleDefinition<any> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const module = new InstrumentationNodeModuleDefinition<any>(
+      "@google/genai",
+      ["*"], // Support all versions
```
cephalization: I think it is fair to say ^1.0.0 since that is what we are testing against

js/packages/openinference-instrumentation-google-genai/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+```bash
+pnpm add @arizeai/openinference-instrumentation-google-genai
+```
+
+## Usage
+
+### Basic Usage with Helper Function (Recommended)
+
+Due to ESM module limitations, the recommended approach is to use the `createInstrumentedGoogleGenAI` helper function:
```
cephalization: I'm not sure this is the recommended approach mostly due to the fact that it will require existing user code to change if they want to adopt the instrumentation.

I think this is a fine utility for examples, but it may give the wrong impression that you will have to modify business logic to use the instrumentation

js/packages/openinference-instrumentation-google-genai/README.md
```diff
@@ -0,0 +1,222 @@
...
+  model: "gemini-2.5-flash",
+  contents: "What is the capital of France?",
+});
+
+console.log(response.text);
+```
+
+### Manual Instrumentation
+
+If you need more control, you can manually instrument existing instances:
```
cephalization: This is a combination of more control + being able to import this once before your app starts and then not needing to change any google-genai imports through the rest of your application.

Maybe mention that this could go into a new file called `instrumentation.ts` and then get imported at the entry point of your application before anything else

js/packages/openinference-instrumentation-google-genai/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+```bash
+pip install arize-phoenix
+```
+
+2. Start Phoenix:
+
+```bash
+python -m phoenix.server.main serve
+```
```
cephalization: Maybe a docker command instead so that Typescript developers don't need to mess with python?

```shell
docker run --pull=always -d --name arize-phoenix -p 6006:6006 arizephoenix/phoenix:latest
```

js/packages/openinference-instrumentation-google-genai/src/instrumentation.ts
```diff
@@ -0,0 +1,934 @@
...
+
+  private patchModelsInstance(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    models: any,
+    instrumentation: GoogleGenAIInstrumentation,
+  ) {
+    // Wrap generateContent
+    if (models.generateContent) {
+      const originalGenerateContent = models.generateContent.bind(models);
+      models.generateContent = async function patchedGenerateContent(
```
cephalization: Tool responses, and multi-message inputs, are not instrumented all the way through

Image

Given the following modified tools example:

```ts
/* eslint-disable no-console */
import "./instrumentation";

import { createInstrumentedGoogleGenAI } from "../src";

// Create an instrumented GoogleGenAI instance
const ai = createInstrumentedGoogleGenAI({
  apiKey: process.env.GOOGLE_API_KEY || process.env.GEMINI_API_KEY!,
});

async function main() {
  console.log("Function calling example:\n");

  // Define a tool/function
  const tools = [
    {
      functionDeclarations: [
        {
          name: "get_weather",
          description: "Get the current weather for a location",
          parametersJsonSchema: {
            type: "object",
            properties: {
              location: {
                type: "string",
                description: "The city and state, e.g. San Francisco, CA",
              },
              unit: {
                type: "string",
                enum: ["celsius", "fahrenheit"],
                description: "The temperature unit",
              },
            },
            required: ["location"],
          },
        },
      ],
    },
  ];

  // Make a request with tool definitions
  const response = await ai.models.generateContent({
    model: "gemini-2.5-flash",
    contents: "What's the weather like in San Francisco?",
    config: {
      tools,
    },
  });

  // Check if the model wants to call a function
  if (response.functionCalls && response.functionCalls.length > 0) {
    console.log("Function call requested:");
    console.log("Name:", response.functionCalls[0].name);
    console.log(
      "Arguments:",
      JSON.stringify(response.functionCalls[0].args, null, 2),
    );

    const responseWithResult = await ai.models.generateContent({
      model: "gemini-2.5-flash",
      contents: [
        {
          role: "user",
          parts: [{ text: "What's the weather like in San Francisco?" }],
        },
        {
          role: "model",
          parts: [
            {
              functionCall: {
                name: "get_weather",
                args: response.functionCalls[0].args,
              },
            },
          ],
        },
        {
          role: "model",
          parts: [
            {
              functionResponse: {
                name: response.functionCalls[0].name,
                response: {
                  output: {
                    text: "The weather in San Francisco is sunny.",
                  },
                },
              },
            },
          ],
        },
      ],
    });

    console.log("Response with result:", responseWithResult.text);
  } else {
    console.log("Response:", response.text);
  }
}

main().catch((error) => {
  console.error("Error:", error);
  process.exit(1);
});

```

I would expect to see in Phoenix:

- User message question
- Assistant tool call
- Tool result
- Assistant message response

PR #2555, opened 2025-12-19
fix(langchain): Handle Non-String Dictionary Keys

### LINKED ISSUE #2507: [bug] [langchain] JSON serialization fails with non-string dictionary keys (e.g., Azure PISubType enum)
**Describe the bug**
The LangChain tracer fails to serialize span attributes when dictionary keys are non-string types (such as enum values). The safe_json_dumps fallback only handles non-serializable values via default=str, but does not handle non-string dictionary keys, which causes json.dumps() to fail.

**To Reproduce**
1. Use Azure AI Document Intelligence to process a document with PII detection enabled
2. Pass the resulting document through a LangChain chain with openinference-instrumentation-langchain enabled
3. The tracer fails when attempting to serialize metadata containing PISubType enum keys

**Expected behavior**
The tracer should successfully serialize objects with non-string dictionary keys by converting them to strings, rather than raising a TypeError.

**Screenshots**
```
2025-12-08 14:48:07.927 | ERROR | Failed to get attribute.
Traceback (most recent call last):
  File ".../openinference/instrumentation/langchain/_tracer.py", line 481, in _json_dumps
    return json.dumps(obj, cls=_OpenInferenceJSONEncoder, ensure_ascii=False)
  ...
TypeError: keys must be str, int, float, bool or None, not PISubType

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../openinference/instrumentation/langchain/_tracer.py", line 327, in wrapper
    yield from wrapped(*args, **kwargs)
  File ".../openinference/instrumentation/langchain/_tracer.py", line 421, in _convert_io
    json_value = _json_dumps(obj)
  File ".../openinference/instrumentation/langchain/_tracer.py", line 484, in _json_dumps
    return safe_json_dumps(obj)
  File ".../openinference/instrumentation/helpers.py", line 21, in safe_json_dumps
    return json.dumps(obj, default=str, ensure_ascii=False, **kwargs)
  ...
TypeError: keys must be str, int, float, bool or None, not PISubType
```

- OS: Linux
- Python: 3.11

**Additional context**

ialisaleh:
Closes #2507 

---

> [!NOTE]
> **Fix JSON serialization of dicts with non-string keys**
> 
> - Update `_json_dumps` to coerce dictionary keys to strings before `json.dumps`, retaining fallback to `safe_json_dumps` for edge cases
> - Add `test_serialize_dict_keys.py` validating key conversion for `Enum`, numbers, `bool`, `None`, and that non-dict inputs pass through unchanged
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e101cc5df6bf1ca81ad6726087a56c920707187b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
@@ -48,12 +48,7 @@ from opentelemetry.util.types import AttributeValue
 from typing_extensions import NotRequired, TypeGuard
 from wrapt import ObjectProxy
 
-from openinference.instrumentation import (
-    get_attributes_from_context,
-)
-from openinference.instrumentation import (
-    safe_json_dumps as _original_safe_json_dumps,
-)
+from openinference.instrumentation import get_attributes_from_context, safe_json_dumps
 from openinference.semconv.trace import (
     DocumentAttributes,
     EmbeddingAttributes,
@@ -460,30 +455,6 @@ def _convert_io(obj: Optional[Mapping[str, Any]]) -> Iterator[str]:
     yield OpenInferenceMimeTypeValues.JSON.value  # Always included for structured objects
 
 
-def _serialize_dict_keys(obj: Any) -> Any:
-    """Convert dictionary keys to strings for JSON serialization."""
-    if not isinstance(obj, dict):
-        return obj
-
-    return {str(key): value for key, value in obj.items()}
-
-
-def safe_json_dumps(obj: Any, **kwargs: Any) -> str:
-    """
-    A convenience wrapper around the original `safe_json_dumps` that ensures
-    that any object can be safely normalized first.
-    """
-    try:
-        # Normalize object for JSON safety
-        normalized_obj = _serialize_dict_keys(obj)
-    except Exception:
-        # Fallback to the original object if normalization fails
-        normalized_obj = obj
-
-    # Serialize normalized object using the original safe_json_dumps
-    return _original_safe_json_dumps(normalized_obj, **kwargs)
-
-
 class _OpenInferenceJSONEncoder(json.JSONEncoder):
     """Custom JSON encoder for OpenInference with comprehensive type support."""
 
@@ -537,11 +508,12 @@ def _json_dumps(obj: Any) -> str:
     This approach is much simpler and more robust than manual recursive processing.
     It handles most common types while falling back to safe_json_dumps for edge cases.
     """
+    if isinstance(obj, dict):
+        obj = {str(k): v for k, v in obj.items()}
+
     try:
-        # Normalize object for JSON safety
-        normalized_obj = _serialize_dict_keys(obj)
         # Use standard json.dumps with our custom encoder
-        return json.dumps(normalized_obj, cls=_OpenInferenceJSONEncoder, ensure_ascii=False)
+        return json.dumps(obj, cls=_OpenInferenceJSONEncoder, ensure_ascii=False)
     except (TypeError, ValueError, OverflowError):
         # Fallback to safe_json_dumps for any unsupported types or circular references
         return safe_json_dumps(obj)
--- a/python/instrumentation/openinference-instrumentation-langchain/tests/test_serialize_dict_keys.py
+++ b/python/instrumentation/openinference-instrumentation-langchain/tests/test_serialize_dict_keys.py
@@ -1,7 +1,8 @@
+import json
 from enum import Enum
 from typing import Any
 
-from openinference.instrumentation.langchain._tracer import _serialize_dict_keys
+from openinference.instrumentation.langchain._tracer import _json_dumps
 
 
 class TelemetryAttribute(Enum):
@@ -9,7 +10,7 @@ class TelemetryAttribute(Enum):
     REQUEST_ID = "request.id"
 
 
-def test_serialize_dict_keys_converts_all_keys_to_strings() -> None:
+def test_json_dumps_converts_dict_keys_to_strings() -> None:
     """Converts all dictionary keys to strings to ensure JSON compatibility."""
     input_attributes: dict[Any, Any] = {
         TelemetryAttribute.USER_EMAIL: "alice@example.com",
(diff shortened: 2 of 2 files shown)


python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -423,6 +428,40 @@ def _convert_io(obj: Optional[Mapping[str, Any]]) -> Iterator[str]:
     yield OpenInferenceMimeTypeValues.JSON.value  # Always included for structured objects
 
 
+def _stringify_dict_keys(obj: Any) -> Any:
+    """Recursively convert all dictionary keys to strings to ensure JSON-safe serialization."""
+    if isinstance(obj, dict):
+        # Convert all dict keys to strings
+        return {str(key): _stringify_dict_keys(value) for key, value in obj.items()}
```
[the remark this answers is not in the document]
ialisaleh: Tracing keys are usually strings, so collisions are rare and handling them would slow serialization unnecessarily.

python/instrumentation/openinference-instrumentation-langchain/tests/test_stringify_dict_keys.py
```diff
@@ -0,0 +1,45 @@
...
+    input_obj: dict[Any, Any] = {
+        PersonalInfoSubtype.EMAIL: "alice@example.com",
+        "numbers_set": {1, 2, 3},
+        "nested_items": [{("tuple_key", 1): "value"}, (4, 5)],
+    }
+
+    expected_obj: dict[str, Any] = {
+        "PersonalInfoSubtype.EMAIL": "alice@example.com",
+        "numbers_set": [1, 2, 3],
+        "nested_items": [{"('tuple_key', 1)": "value"}, (4, 5)],
```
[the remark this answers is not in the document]
ialisaleh: This is just a simple test. Exact element order for sets is not guaranteed in Python, so failures in other implementations or versions are expected and harmless.

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -423,6 +428,40 @@ def _convert_io(obj: Optional[Mapping[str, Any]]) -> Iterator[str]:
     yield OpenInferenceMimeTypeValues.JSON.value  # Always included for structured objects
 
 
+def _stringify_dict_keys(obj: Any) -> Any:
```
nate-mar: IMO, I don't think we need to try to handle more complex data types here yet. It seems like the original issue primarily concerns Enums. We could just handle the enum case prior to `safe_json_dumps`:

```
def serialize_dict_with_enums(d):
    return {
        (k.value if isinstance(k, Enum) else k): v 
        for k, v in d.items()
    }
```

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -477,9 +516,11 @@ def _json_dumps(obj: Any) -> str:
     It handles most common types while falling back to safe_json_dumps for edge cases.
     """
     try:
+        # Normalize object for JSON safety
+        normalized_obj = _stringify_dict_keys(obj)
         # Use standard json.dumps with our custom encoder
-        return json.dumps(obj, cls=_OpenInferenceJSONEncoder, ensure_ascii=False)
-    except (TypeError, ValueError, OverflowError):
+        return json.dumps(normalized_obj, cls=_OpenInferenceJSONEncoder, ensure_ascii=False)
+    except (TypeError, ValueError, OverflowError, RecursionError):
```
nate-mar: Won't need RecursionError if we just handle the Enum key types

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
ialisaleh: Hi @nate-mar - I updated the fix to handle only Enum keys using this approach. It is worth noting that it won't recurse into nested dicts, and Enum values can still collide with existing string keys, which could cause silent data loss in rare cases. Please take a look. Thanks!

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
ialisaleh: Hi @nate-mar - I removed the `RecursionError` as we don't need it now. Thanks!

RogerHYang (approved):

python/instrumentation/openinference-instrumentation-langchain/src/openinference/instrumentation/langchain/_tracer.py
```diff
@@ -455,6 +460,30 @@ def _convert_io(obj: Optional[Mapping[str, Any]]) -> Iterator[str]:
     yield OpenInferenceMimeTypeValues.JSON.value  # Always included for structured objects
 
 
+def _serialize_enum_dict_keys(obj: Any) -> Any:
+    """Convert Enum dictionary keys to their values for JSON serialization."""
+    if not isinstance(obj, dict):
+        return obj
+
+    return {(key.value if isinstance(key, Enum) else key): value for key, value in obj.items()}
```
RogerHYang: It's probably simpler to just apply `str()` on the keys.

```suggestion
    return {str(key): value for key, value in obj.items()}
```

RogerHYang: If auto() was used when defining the Enum, then we'll just get integers, which may not be very informative.

Screenshot 2026-01-05 at 8 29 34 PM

ialisaleh: Hi @RogerHYang - I updated the PR to convert dictionary keys to strings for JSON serialization. Please take a look. Thanks!

PR #2568, opened 2025-12-26
feat: Introduce new shared conv

Dwij1704:
> [!NOTE]
> Introduces shared GenAI semantic convention constants and makes them available to consumers.
> 
> - Adds `openinference/instrumentation/gen_ai_attributes.py` to re-export OTEL GenAI attributes and define `GenAIAttributes`/`GenAIEventNames`
> - Updates `__init__.py` to import and export new GenAI constants for public use
> - Adds dependency `opentelemetry-semantic-conventions>=0.60b0` in `pyproject.toml`
> - Bumps version to `0.1.43`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 60d87e429c1b79fccd7237ffc4aec33125c2e788. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


nate-mar:
@Dwij1704 I thought this PR was around introducing some more open-inference semantic conventions or some kind of shared mapping from OI to OTEL genai, but it seems like it's just being used to store constants for OTEL genai semantic conventions and it's being added to the common OI instrumentation package not the OI semantic-conventions package.

Given that's the case, I actually retract what I previously asked about moving this into a separate PR. Can we instead just keep this in the strands instrumentation package for now? The reason is because of the current instability of the OTEL genai conventions, I don't think trying to couple the constants to the common open-inference utils package right is worth it at the moment. I'm a little bit concerned about the song and dance we might have to play based on how each individual 3rd party library updates their usage of the OTEL genai conventions and having to track/work around that in the OI packages.

Dwij1704:
> @Dwij1704 I thought this PR was around introducing some more open-inference semantic conventions or some kind of shared mapping from OI to OTEL genai, but it seems like it's just being used to store constants for OTEL genai semantic conventions and it's being added to the common OI instrumentation package not the OI semantic-conventions package.
> 
> Given that's the case, I actually retract what I previously asked about moving this into a separate PR. Can we instead just keep this in the strands instrumentation package for now? The reason is because of the current instability of the OTEL genai conventions, I don't think trying to couple the constants to the common open-inference utils package right is worth it at the moment. I'm a little bit concerned about the song and dance we might have to play based on how each individual 3rd party library updates their usage of the OTEL genai conventions and having to track/work around that in the OI packages.

You make a good point. After looking at this more closely, I agree the shared module approach isn't worth the coupling risk right now. Looking at the pydantic-ai instrumentation, it already defines its own local GenAIEventNames and related constants in semantic_conventions.py and imports OTEL constants directly from opentelemetry.semconv._incubating. This pattern works well because: Each instrumentation can evolve independently as the 3rd party tools update their OTEL usage at different rates.
I'll move the constants into the strands instrumentation package directly, following the same pattern as pydantic-ai. If we see significant duplication across 3+ instrumentations in the future and the gen_ai conventions stabilize, we can revisit centralizing them.

PR #2598, opened 2026-01-09
feat(google-genai): Handled Image Data & Multiple tool calls in Response

### LINKED ISSUE #2194: [bug] "Other field types of parts are not supported yet" error when attaching images in GoogleGenAIInstrumentor
**Describe the bug**
The `openinference-instrumentation-google-genai` package throws an error "Other field types of parts are not supported yet" when processing image data attached to Google GenAI requests. This error occurs specifically in the `_get_attributes_from_part` function at line 367 of the request attributes extractor, suggesting that the instrumentation library doesn't properly handle `types.Part` objects created from base64 image data.

**To Reproduce**

```python
import base64
import asyncio
from google import genai
from openinference.instrumentation.google_genai import GoogleGenAIInstrumentor

# Enable instrumentation (this is what causes the error)
GoogleGenAIInstrumentor().instrument()

async def reproduce_error():
    client = genai.Client(api_key="your-api-key")
    
    # Create a simple base64 image (1x1 pixel PNG)
    base64_img = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8/5+hHgAHggJ/PchI7wAAAABJRU5ErkJggg=="
    
    # Extract MIME type and decode base64 data
    mime_type = base64_img.split(";")[0].replace("data:", "")
    img_data = base64.b64decode(base64_img.split(",")[1])
    
    # Create Part object from bytes (this is what triggers the error)
    image_part = types.Part.from_bytes(
        data=img_data,
        mime_type=mime_type,
    )
    
    parts = ["Describe this image", image_part]
    
    # This call will trigger the instrumentation error
    response = await client.aio.models.generate_content(
        model="gemini-2.5-flash",
        contents=parts,
        config=types.GenerateContentConfig(temperature=0.0),
    )
    return response

# Run the reproduction
asyncio.run(reproduce_error())
```

**Expected behavior**
The request should complete successfully without instrumentation errors. The OpenInference instrumentation should properly handle `types.Part` objects created from both `from_bytes()` and `from_uri()` methods without throwing unsupported field type errors.

**Error Details**

```
{
  "dt": "2025-09-11T11:52:33.622395+00:00",
  "level": "error", 
  "severity": 4,
  "message": "Other field types of parts are not supported yet",
  "context": {
    "runtime": {
      "function": "_get_attributes_from_part",
      "file": "../lib/python3.13/site-packages/openinference/instrumentation/google_genai/_request_attributes_extractor.py",
      "line": 367,
      "thread_id": 139450478468032,
      "thread_name": "MainThread",
      "logger_name": "openinference.instrumentation.google_genai._request_attributes_extractor"
    }
  }
}
```

**Environment:**

- OS: macOS (Darwin 24.6.0) (but also in prod with debian)
- Python: 3.13
- google-genai: 1.36.0
- openinference-instrumentation-google-genai: 0.1.5

**Root Cause Analysis**
The error suggests that the OpenInference instrumentation library's `_get_attributes_from_part` function doesn't handle all possible `types.Part` variations. Specifically, it appears to not support `Part` objects created via `types.Part.from_bytes()` method, which is commonly used for base64-encoded image data.

**Related Issue**
Previously encountered similar error: "Unexpected input contents type: <class 'PIL.JpegImagePlugin.JpegImageFile'>" when using PIL Image objects directly instead of `types.Part.from_bytes()`. This was resolved by replacing PIL usage with `types.Part.from_bytes()` in commit c476ea9d, but the instrumentation library still doesn't properly handle the resulting Part objects.

**Additional context**

- Error appears as: `ERROR:openinference.instrumentation.google_genai._request_attributes_extractor:Other field types of parts are not supported yet`
- The error is non-blocking but creates noise in logs and monitoring  
- Affects all image processing workflows using base64 image data
- API calls complete successfully despite the instrumentation error

satyadevai:
Closes #2194

Multiple tools Response
image

Stream Image Generator
image

Async Image Artifact
image

---

> [!NOTE]
> **Medium Risk**
> Touches core tracing attribute extraction for requests, responses, and streaming, which can change emitted span schemas and increase attribute size (notably base64 image URLs).
> 
> **Overview**
> Improves `openinference-instrumentation-google-genai` span enrichment for **multimodal and tool-calling** flows.
> 
> Request/response extractors and streaming aggregation now emit structured `message.contents[N]` attributes for multi-part messages, including `inline_data` images as `data:` URLs, and correctly record *multiple* tool calls instead of collapsing/losing them. Logging is also toned down (more `debug`/`warning`, fewer `exception` paths), and tests/cassettes plus new examples were added to cover image inputs, streaming vs non-streaming, and multi-tool-call responses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 032de7143835b4361a6c7f74985a62143113f0f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
```diff
@@ -178,23 +180,31 @@ def get_extra_attributes(self) -> Iterator[Tuple[str, AttributeValue]]:
...
                         yield (
-                            f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.{idx}.{MessageAttributes.MESSAGE_ROLE}",
+                            f"{prefix}.{MessageAttributes.MESSAGE_ROLE}",
                             role,
                         )
                     if parts := content.get("parts"):
-                        text_parts = []
                         tool_call_index = 0
+                        content_index = 0
+                        text_only = len(parts) == 1
```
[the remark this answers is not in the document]
satyadevai: OpenInference MessageContent & MessageContentAttributes is designed to support mixed content (text + image) within the same message object. The spec allows both top-level text content and structured contents entries to coexist without being mutually exclusive.

Sample:
"llm.input_messages.1.message.role": "model",
"llm.input_messages.1.message.content": "Here is a simple 3x3 pixel art image of a happy face!",
"llm.input_messages.1.message.contents.0.message_content.image.image.url": "data:image/png;base64,/9j/4QDeRXhp=",
"llm.input_messages.1.message.contents.0.message_content.type": "image"

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_utils.py
```diff
@@ -141,16 +148,56 @@ def _finish_tracing(
...
+        and "image" in mime_type
+        and (data := get_attribute(inline_data, "data")) is not None
+    ):
+        prefix = f"{MessageAttributes.MESSAGE_CONTENTS}.{tool_call_index}."
+        image_url = f"data:{mime_type};base64,{base64.b64encode(data).decode()}"
+        yield (
+            f"{prefix}{MessageContentAttributes.MESSAGE_CONTENT_IMAGE}.{ImageAttributes.IMAGE_URL}",
+            image_url,
+        )
+        yield f"{prefix}{MessageContentAttributes.MESSAGE_CONTENT_TYPE}", "image"
```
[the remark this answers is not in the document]
satyadevai: This implementation is correct because message.contents is a list of separate content parts, not a single shared object. When both text and inline data are present, each is stored as its own content entry with its own MESSAGE_CONTENT_TYPE. As a result, setting the content type for one part does not overwrite the other, and the message correctly preserves both the text and image content along with their respective types.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -376,33 +379,32 @@ def _get_attributes_from_function_response(
...
+            if text := get_attribute(part, "text"):
+                yield from _get_attributes_from_content_text(text, content_index, is_single_part)
+                increment_content_index = True
+            elif function_call := get_attribute(part, "function_call"):
+                yield from self._get_attributes_from_function_call(function_call, tool_call_index)
+                tool_call_index += 1
+            elif function_response := get_attribute(part, "function_response"):
+                yield from self._get_attributes_from_function_response(function_response)
+            if inline_data := get_attribute(part, "inline_data"):
+                yield from _get_attributes_from_inline_data(inline_data, content_index)
```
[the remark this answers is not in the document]
satyadevai: OpenInference MessageContent & MessageContentAttributes is designed to support mixed content (text + image) within the same message object. The spec allows both top-level text content and structured contents entries to coexist without being mutually exclusive.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -376,33 +379,32 @@ def _get_attributes_from_function_response(
...
-            yield (MessageAttributes.MESSAGE_CONTENT, "\n\n".join(content_values))
+            increment_content_index = False
+            if text := get_attribute(part, "text"):
+                yield from _get_attributes_from_content_text(text, content_index, is_single_part)
+                increment_content_index = True
+            elif function_call := get_attribute(part, "function_call"):
+                yield from self._get_attributes_from_function_call(function_call, tool_call_index)
+                tool_call_index += 1
+            elif function_response := get_attribute(part, "function_response"):
+                yield from self._get_attributes_from_function_response(function_response)
```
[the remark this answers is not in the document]
satyadevai: Function responses are always returned as a single message, so there will only ever be one Message.Content value. Because of this, yielding MESSAGE_CONTENT directly is correct and does not result in data loss. The scenario with multiple function_response parts producing multiple content values does not occur in reality.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -40,7 +43,7 @@ def get_attributes_from_request(
                 _io_value_and_type(request_parameters),
             )
         except Exception:
-            logger.exception(
+            logger.debug(
```
fjcasti1: Why this change? Seems to me it was better before

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -376,33 +379,32 @@ def _get_attributes_from_function_response(
...
+            elif function_call := get_attribute(part, "function_call"):
+                yield from self._get_attributes_from_function_call(function_call, tool_call_index)
+                tool_call_index += 1
+            elif function_response := get_attribute(part, "function_response"):
+                yield from self._get_attributes_from_function_response(function_response)
+            if inline_data := get_attribute(part, "inline_data"):
+                yield from _get_attributes_from_inline_data(inline_data, content_index)
+                increment_content_index = True
+            if increment_content_index:
+                content_index += 1
```
fjcasti1, line 167: Could you explain this change?

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -415,30 +417,21 @@ def _extract_tool_call_index(self, attr: str) -> int:
...
-            )
+            yield from _get_attributes_from_content_text(text, index, is_single_part)
         elif function_call := get_attribute(part, "function_call"):
-            yield from self._get_attributes_from_function_call(function_call, tool_call_index)
+            yield from self._get_attributes_from_function_call(function_call, index)
         elif function_response := get_attribute(part, "function_response"):
             yield from self._get_attributes_from_function_response(function_response)
-        else:
-            logger.exception("Other field types of parts are not supported yet")
+        if inline_data := get_attribute(part, "inline_data"):
```
fjcasti1: You can remove the `elif` for `if` in this case since you're yielding

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
```diff
@@ -120,11 +125,11 @@ def _get_attributes_from_function_call(
                         args_json,
                     )
                 except Exception:
-                    logger.exception(
+                    logger.warning(
```
fjcasti1, line 130: why?

fjcasti1 (commented):
Left some questions, mostly lgtm

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
satyadevai: This will be reported as a failure in the logging systems when logged as an exception in the instrumentation code. This is the issue described in #2194. If a failure occurs within the instrumentation itself, it should not be reported through logging.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
satyadevai: google-genai returns responses as List[Part], where each message may contain a single part or multiple parts. A single part can include both text content and blob (inline) data. We handle this case here.

If the message contains multiple parts, they are processed as MessageContents; otherwise, as MessageContent.

We cannot use elif for inline_data because a single part may contain both text and inline data.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
satyadevai: https://github.com/Arize-ai/openinference/pull/2598#discussion_r2709400675

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
satyadevai: https://github.com/Arize-ai/openinference/pull/2598#discussion_r2709416001

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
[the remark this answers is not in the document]
satyadevai: Implemented as suggested

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_utils.py
```diff
@@ -141,16 +148,56 @@ def _finish_tracing(
...
+            text,
+        )
+        yield (
+            f"{MessageAttributes.MESSAGE_CONTENTS}.{index}.{MessageContentAttributes.MESSAGE_CONTENT_TYPE}",
+            "text",
+        )
+
+
+def _get_attributes_from_inline_data(
+    inline_data: Any, tool_call_index: int
```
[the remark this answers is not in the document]
satyadevai: Implemented as suggested

caroger (commented):
Validated locally against this PR head. The patch that fixed the issues for me did three things: (1) preserve list position as the fallback part index in `_IndexedAccumulator` so streamed multimodal parts do not collapse into slot `0`, (2) add `file_data` image extraction alongside `inline_data` in the request/response/stream paths, and (3) re-enable the streaming multiple-tool-call test with its cassette under `tests/cassettes/test_instrumentation/`.

Verification on the local patch: `ruff check` passed, and `PYTHONPATH=python/instrumentation/openinference-instrumentation-google-genai/src:python/openinference-instrumentation/src:python/openinference-semantic-conventions/src python -m pytest python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py` finished with `23 passed`.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
```diff
@@ -342,7 +375,7 @@ def __iadd__(
             return self
         if isinstance(values, Mapping):
             values = [values]
-        for v in values:
+        for index, v in enumerate(values):
             if v and hasattr(v, "get"):
                 self._indexed[v.get("index") or 0] += v
```
caroger: This still collapses every non-indexed streamed `part` into slot `0`. That preserves text continuation, but it also merges a streamed `[text, image]` response into one synthetic part and can overwrite repeated image parts entirely. I reproduced that locally with `_ResponseAccumulator` using mock chunks. Falling back to the enumerate index when `index` is missing preserves multimodal part boundaries without regressing split tool-call aggregation.

```suggestion
        for index, v in enumerate(values):
            if v and hasattr(v, "get"):
                item_index = v.get("index")
                self._indexed[index if item_index is None else item_index] += v
```

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
```diff
@@ -143,8 +147,18 @@ def _flatten_parts(self, parts: list[Part]) -> Iterator[tuple[str, AttributeValu
                             f"{prefix}.{MessageContentAttributes.MESSAGE_CONTENT_TEXT}",
                             safe_json_dumps(response),
                         )
-                        content_index += 1
+                        increment_content_index = True
                 if id_ := function_response.id:
                     yield (MessageAttributes.MESSAGE_TOOL_CALL_ID, id_)
+            if inline_data := part.inline_data:
```
caroger, line 152: `google.genai.types.Part.from_uri(...)` puts images in `file_data`, not `inline_data`, so URI-based image inputs still emit no `message_content.image.image.url` here. I verified that locally with `Part.from_uri(file_uri=..., mime_type="image/png")`. This needs a `file_data` branch alongside `inline_data` (or a shared helper that both branches can reuse).

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
```diff
@@ -80,21 +82,24 @@ def _get_attributes_from_content_parts(
...
             elif function_call := getattr(part, "function_call", None):
                 # Handle tool/function calls
                 yield from self._get_attributes_from_function_call(function_call, tool_call_index)
                 tool_call_index += 1
-
-        # Always yield message content for consistency, even if empty
-        # This ensures Phoenix can properly display the message structure
-        content = "\n".join(text_content) if text_content else ""
-        yield MessageAttributes.MESSAGE_CONTENT, content
+            if inline_data := getattr(part, "inline_data", None):
```
caroger, line 99: Same gap on the response side: this only checks `inline_data`, so any image parts represented as `file_data` stay invisible in traced output/history even if the request-side extractor is fixed. I’d mirror the same `file_data` handling here and in the streaming extractor so request/response schemas stay aligned.

python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py
```diff
@@ -1172,6 +1260,182 @@ def test_streaming_content_with_tool(
         )
 
 
+@pytest.mark.vcr(
+    decode_compressed_response=True,
+    before_record_request=lambda _: _.headers.clear() or _,
+    before_record_response=lambda _: {**_, "headers": {}},
+)
+@pytest.mark.parametrize("streaming", [False])
```
caroger: This never exercises the streaming branch as written, so the `_stream.py` changes above can still ship without coverage. Turning `[True]` back on caught the regression locally; the existing streaming cassette just needs to live under `tests/cassettes/test_instrumentation/` for `pytest-recording` to find it offline.

```suggestion
@pytest.mark.parametrize("streaming", [False, True])
```

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
satyadevai: Using the index from enumerate(values) might introduce a new issue because self._indexed may not align with the enumerated index. Specifically, some of the values in the current chunk might actually belong to previous index values already processed. I believe it's safer to stick with the previous implementation for the indexing logic. I did find specific issues with how function calls were being handled, which is why I moved that logic outside of the indexed elements to ensure they are captured correctly.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_request_attributes_extractor.py
satyadevai: This is now fixed in a recent commit. URI-based image inputs are correctly handled via a file_data branch alongside inline_data, so the image URL is properly populated.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
satyadevai: This isn’t an issue on the response side. The model always returns generated images as blob content, and we already handle that.

It doesn’t return images as file_data or URLs, so we don’t need to add file_data handling here.

python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py
satyadevai: Initially, I tried to keep this test for both streaming and non-streaming, but streaming wasn’t working correctly since it was only capturing one tool call even when there were two. After moving function calls out of the indexed object, we can now capture both properly.

This was a miss on my side for not setting [True] earlier. I’ve now added it, and the same test runs for both streaming and non-streaming. This works since the underlying functional logic is the same for both, so it now covers both paths properly.

caroger:
Re-verified against the actual `google-genai==1.57.0` source and our real wrapper paths.

My earlier request-side `Part.from_uri(...)->file_data` note no longer applies: that path is addressed in this branch.

The remaining review comments below are the ones I could still reproduce on the actual code paths:
- streamed part indexing still collapses non-indexed `Content.parts`
- split streamed tool-call chunks still become two tool calls instead of one logical call
- response-side `file_data` is still not extracted in the non-stream response extractor

I verified those against the pinned `google-genai` source (`google/genai/models.py` and `google/genai/types.py`) and against our `_SyncGenerateContent` / `_SyncGenerateContentStream` / async wrappers before reposting them.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
caroger, line 99: Verified against the actual library and wrapper path:
- `google/genai/types.py:1541-1543` defines `Part.file_data`.
- `google/genai/types.py:1668-1694` shows `Part.from_uri()` constructs `file_data`.
- Our non-stream wrappers `_SyncGenerateContent` and `_AsyncGenerateContentWrapper` finalize through `_ResponseAttributesExtractor`, so this code runs on the real `generate_content()` path.

Request-side `file_data` is fixed, but response-side `file_data` still falls through here because we only inspect `inline_data`. A response part like `{"file_data": {"file_uri": ..., "mime_type": "image/jpeg"}}` still emits no `message_content.image.image.url`.

Something along these lines should wire it up here as well:

```python
if inline_data := getattr(part, "inline_data", None):
    yield from _get_attributes_from_inline_data(inline_data, content_index)
    increment_content_index = True
if file_data := getattr(part, "file_data", None):
    yield from _get_attributes_from_file_data(file_data, content_index)
    increment_content_index = True
```

with the matching `_get_attributes_from_file_data` import above.

python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py
```diff
@@ -1360,7 +1624,14 @@ def model_dump(self, exclude_unset=True, warnings=False):
                     "index": 0,
                     "content": {
                         "role": "model",
-                        "parts": [{"function_call": {"name": "get_weather"}}],
+                        "parts": [
+                            {
+                                "function_call": {
```
caroger: After re-checking against the actual `google-genai==1.57.0` types, this test no longer models the failure case from the review. `Part.function_call` is one field on each `Part` (`google/genai/types.py:1544-1547`), and the regression happens when one logical tool call is split across streamed chunks. The earlier name-only / args-only pattern caught that. The current version builds two complete function calls, so it still passes when `_stream.py` emits `tool_calls.0.name` and `tool_calls.1.arguments` for one logical call.

I would keep chunk 1 as `{"function_call": {"name": "get_weather"}}`, chunk 2 as `{"function_call": {"args": {"location": "San Francisco", "unit": "fahrenheit"}}}`, and assert that only `tool_calls.0.*` is populated.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
```diff
@@ -167,6 +169,23 @@ def __init__(self) -> None:
...
+            content = candidate.get("content") or {}
+            parts = content.get("parts") or []
+            if isinstance(parts, list):
+                fc_calls = [
+                    p["function_call"]
+                    for p in parts
+                    if isinstance(p, dict) and "function_call" in p
+                ]
+                if fc_calls:
+                    content.setdefault("function_calls", []).extend(fc_calls)
```
caroger: I rechecked this against the same pinned source. `Part.function_call` lives inside `Content.parts` (`google/genai/types.py:1544-1547`), and streamed chunks come through `_Stream` from `generate_content_stream()` (`google/genai/models.py:4031-4105`, `5262-5268`). Appending each chunk fragment into `content["function_calls"]` still turns one logical call into two tool calls when the stream splits name and args across chunks: chunk 1 becomes `tool_calls.0.name`, chunk 2 becomes `tool_calls.1.arguments`.

That is what I still reproduce on this branch. The safer fix is to keep `function_call` inside the indexed `parts` accumulator and merge it by part position once the fallback indexing above is restored.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
caroger, line 1: Verified this against the actual stream path:
- `google/genai/models.py:4031-4105` and `5262-5268` in `google-genai==1.57.0` show that `generate_content_stream()` yields `GenerateContentResponse` chunks.
- `google/genai/types.py:6397-6435` has `Candidate.index`, but `google/genai/types.py:1799-1803` and `1520-1569` show that `Content.parts` is `list[Part]` and `Part` has no `index` field.

On our real stream wrappers (`_SyncGenerateContentStream` / `_AsyncGenerateContentStream`), list position is therefore the only stable fallback for part identity. Collapsing every missing part index to `0` still merges streamed `[text, image]` into one synthetic part and can drop earlier image parts.

I would change the fallback in `_IndexedAccumulator.__iadd__` along these lines:

```python
for index, v in enumerate(values):
    if v and hasattr(v, "get"):
        item_index = v.get("index")
        self._indexed[index if item_index is None else item_index] += v
```

caroger:
Validated a local patch on `fc2d6881` before posting code snippets:

- direct repros passed for the three remaining issues: streamed multimodal part ordering, split streamed tool-call aggregation, and response-side `file_data`
- `ruff check` passed on the touched files
- the full `python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py` file passed in the pinned env (`google-genai==1.57.0`, `wrapt<2`): `25 passed`

I am replying in-thread below with the exact snippets from that verified patch.

python/instrumentation/openinference-instrumentation-google-genai/tests/test_instrumentation.py
caroger: This is the exact regression coverage from the verified local patch.

For the split streamed tool call, I switched the mock chunks back to one logical call split across chunks and asserted that only `tool_calls.0.*` exists:

```diff
-                        "parts": [
-                            {
-                                "function_call": {
-                                    "name": "get_weather",
-                                    "args": {"location": "San Francisco", "unit": "fahrenheit"},
-                                }
-                            }
-                        ],
+                        "parts": [{"function_call": {"name": "get_weather"}}],
...
-                                "function_call": {
-                                    "name": "get_weather",
-                                    "args": {"location": "New York", "unit": "fahrenheit"},
-                                }
+                                "function_call": {
+                                    "args": {"location": "San Francisco", "unit": "fahrenheit"}
+                                }
```

```python
tool_call_name_key = f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.0.{MessageAttributes.MESSAGE_TOOL_CALLS}.0.{ToolCallAttributes.TOOL_CALL_FUNCTION_NAME}"
tool_call_args_key = f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.0.{MessageAttributes.MESSAGE_TOOL_CALLS}.0.{ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}"
assert attributes.get(tool_call_name_key) == "get_weather"
args = json.loads(attributes[tool_call_args_key])
assert args == {"location": "San Francisco", "unit": "fahrenheit"}
assert (
    f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.0.{MessageAttributes.MESSAGE_TOOL_CALLS}.1.{ToolCallAttributes.TOOL_CALL_FUNCTION_NAME}"
    not in attributes
)
assert (
    f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.0.{MessageAttributes.MESSAGE_TOOL_CALLS}.1.{ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}"
    not in attributes
)
```

I also added two direct unit tests that passed locally and catch the two remaining regressions directly:

```python
def test_streaming_multimodal_content_aggregation() -> None:
    ...  # one chunk with [text, inline image, inline image]
    assert message.contents.0 is text
    assert message.contents.1 is the first image
    assert message.contents.2 is the second image


def test_response_attributes_extract_image_from_file_data() -> None:
    ...  # GenerateContentResponse with part.file_data
    assert image_url == file_uri
    assert content_type == "image"
```

Those tests are included in the `25 passed` local run above.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
caroger: This is the exact `_IndexedAccumulator.__iadd__` change from the verified local patch:

```diff
-        for v in values:
-            if v and hasattr(v, "get"):
-                self._indexed[v.get("index") or 0] += v
+        for index, v in enumerate(values):
+            if v and hasattr(v, "get"):
+                item_index = v.get("index")
+                self._indexed[index if item_index is None else item_index] += v
```

That change was part of the patch I validated locally with the direct repros and `25 passed` in the full test file.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_response_attributes_extractor.py
caroger: This is the exact response-side `file_data` patch from the verified local run:

```diff
 from openinference.instrumentation.google_genai._utils import (
     _as_output_attributes,
     _get_attributes_from_content_text,
+    _get_attributes_from_file_data,
     _get_attributes_from_inline_data,
     _get_token_count_attributes_from_usage_metadata,
     _io_value_and_type,
 )
```

```diff
             if inline_data := getattr(part, "inline_data", None):
                 yield from _get_attributes_from_inline_data(inline_data, content_index)
                 increment_content_index = True
+            if file_data := getattr(part, "file_data", None):
+                yield from _get_attributes_from_file_data(file_data, content_index)
+                increment_content_index = True
             if increment_content_index:
                 content_index += 1
```

I also added a direct unit test for a `GenerateContentResponse` containing `{"file_data": {"file_uri": ..., "mime_type": "image/jpeg"}}`, and that passed in the full local run.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_stream.py
caroger: This is the exact `_stream.py` shape that fixed the split-tool-call and streamed multimodal repros locally:

```diff
                         parts=_IndexedAccumulator(
                             lambda: _ValuesAccumulator(
                                 text=_StringAccumulator(),
+                                function_call=_DictReplace(),
                             ),
                         ),
-                        function_calls=[],
                         role=_SimpleStringReplace(),
```

```python
def process_chunk(self, chunk: "GenerateContentResponse") -> None:
    self._is_null = False
    values = chunk.model_dump(exclude_unset=True, warnings=False)
    self._values += values
```

```python
if parts := content.get("parts"):
    content_index = 0
    is_single_part = len(parts) == 1
    tool_call_index = 0
    for part in parts:
        if text := part.get("text"):
            for key, value in _get_attributes_from_content_text(
                text,
                content_index,
                is_single_part,
            ):
                yield f"{prefix}.{key}", value
        elif function_call := part.get("function_call"):
            tc = (
                f"{prefix}.{MessageAttributes.MESSAGE_TOOL_CALLS}."
                f"{tool_call_index}"
            )
            if function_name := function_call.get("name"):
                yield (
                    f"{tc}.{ToolCallAttributes.TOOL_CALL_FUNCTION_NAME}",
                    function_name,
                )
            if function_args := function_call.get("args"):
                yield (
                    f"{tc}.{ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON}",
                    safe_json_dumps(function_args),
                )
            tool_call_index += 1
        if inline_data := part.get("inline_data"):
            for key, value in _get_attributes_from_inline_data(
                inline_data, content_index
            ):
                yield f"{prefix}.{key}", value
        if file_data := part.get("file_data"):
            for key, value in _get_attributes_from_file_data(
                file_data, content_index
            ):
                yield f"{prefix}.{key}", value
        if part.get("text") or part.get("inline_data") or part.get("file_data"):
            content_index += 1
```

The important part is keeping `function_call` inside the indexed `parts` accumulator so a name-only chunk and an args-only chunk still merge into one logical tool call.

caroger (approved):

PR #2632, opened 2026-01-22
fix(openai): add openai sdk as peer dependencies

loopingz:
openai sdk is import for more than just types as it is used in a `instanceof` instruction

Therefore it will fail with strict dependencies management like pnpm or bazel

---

> [!NOTE]
> **Scope**: `@arizeai/openinference-instrumentation-openai`
> 
> - Add `openai` as a `peerDependency` (kept in `devDependencies` for local development) to ensure the SDK is present at runtime for `instanceof` checks under strict managers like pnpm/Bazel.
> - Include a changeset marking a patch release.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aa00c37e2805077aaf806d40d3dafba5b5d0c6e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/js/.changeset/openai-peer-dependency.md
+++ b/js/.changeset/openai-peer-dependency.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/openinference-instrumentation-openai": patch
+---
+
+Add openai as peer dependency to fix strict dependency management issues with pnpm and bazel. The openai SDK is required at runtime for instanceof checks, not just for types.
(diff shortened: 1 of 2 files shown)


loopingz:

fjcasti1 (approved):

mikeldking:
the general rule for auto-instrumentors is to be graceful if the package it's supposed to instrument is not installed. However given that we are for some reason using instanceof (maybe this is the case here), I think it's safe to add this for now until we can maybe use some level of duck typing.

mikeldking:
@claude add a changeset to this PR with the changes so that it can be released.

mikeldking (approved):

PR #2646, opened 2026-01-24
feat(litellm): capture tool calls in streaming responses

hewliyang:
## Summary

Add support for accumulating tool calls from streaming delta chunks in both sync and async streaming handlers for the LiteLLM instrumentation.

## Changes

- Add `_accumulate_tool_calls()` to collect tool call chunks by index
- Add `_build_message_from_accumulated()` to format messages for attribute extraction  
- Update `_finalize_sync_streaming_span()` to accumulate tool calls
- Update `_finalize_streaming_span()` (async) to accumulate tool calls
- Add tests for streaming with single and multiple tool calls

## Context

The existing streaming handlers (`_finalize_sync_streaming_span` and `_finalize_streaming_span`) only captured `delta.content` but missed `delta.tool_calls`. This meant tool calls in streaming responses were not being recorded in spans.

The implementation follows the pattern from OpenAI instrumentation's `_ChatCompletionAccumulator` but in a simpler form since LiteLLM already normalizes the streaming format.

## Testing

- Added some unit tests w/ mock streams (iiuc, this is how its done in the openai instrumentation.) But LMK if we need to get an actual cassette for it

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/examples/litellm_tools_streaming.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/examples/litellm_tools_streaming.py
@@ -1,0 +1,140 @@
+import asyncio
+import json
+from typing import Any, Dict, List
+
+import litellm
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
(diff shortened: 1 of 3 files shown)


mikeldking:
thanks @hewliyang !

mikeldking (approved):

PR #2721, opened 2026-02-10
feat(openinference-vercel): Expand multi-tool results into multiple tool messages

cephalization:
Openinference spec does not specify how multiple tool results in a single message should be handled. This PR handles the case that ai-sdk produces multiple tool results in one message in response to multiple tool calls by expanding the results into distinct tool messages

image

---

> [!NOTE]
> **Medium Risk**
> Changes how prompt message arrays are transformed into OpenInference attributes, which can shift message indexing and tool-result attribution in exported telemetry. Coverage is improved with new tests, but any downstream assumptions about previous indexing/aggregation could be impacted.
> 
> **Overview**
> Updates `getInputMessageAttributes` to **expand a single Vercel `role: "tool"` message containing multiple tool results into multiple OpenInference input messages**, using a separate output index so subsequent messages remain correctly numbered. Tool result extraction is also broadened to handle both AI SDK v6 `output` and legacy `result`, and to populate `message.tool_call_id` and `tool.name` per expanded message.
> 
> Adds comprehensive test coverage for tool-result mapping (object vs string outputs, legacy `result`, multiple tool messages, and multi-result expansion) and bumps `@arizeai/openinference-vercel` with a minor changeset.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 15833d7b31f6a28db4b6894015c738d29d75fea5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/js/packages/openinference-vercel/src/utils.ts
+++ b/js/packages/openinference-vercel/src/utils.ts
@@ -272,7 +272,7 @@ const getInputMessageAttributes = (promptMessages?: AttributeValue) => {
 
   return messages.reduce((acc: Attributes, message) => {
     if (message.role === "tool") {
-      const contentArray = Array.isArray(message.content)
+      const contentArray: unknown[] = Array.isArray(message.content)
         ? message.content
         : message.content
           ? [message.content]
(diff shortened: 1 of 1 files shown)


js/packages/openinference-vercel/src/utils.ts
```diff
@@ -266,53 +266,68 @@ const getInputMessageAttributes = (promptMessages?: AttributeValue) => {
...
+              typeof content?.toolCallId === "string"
+                ? content.toolCallId
+                : undefined,
+            [`${MESSAGE_PREFIX}.${SemanticConventions.TOOL_NAME}`]:
+              typeof content?.toolName === "string"
+                ? content.toolName
+                : undefined,
+          };
+        },
+        {},
```
mikeldking: no need to type the base accumulation value?

cephalization: Good call, the array.isArray check converted the unknown content array to `any[]` and further type checking got disabled. Will fix

cephalization: Fixed. Should upgrade the default tsconfig soon to prevent this in the future

PR #2757, opened 2026-02-19
docs(python spec): add instructions for working on python instrumentation packages

caroger:
- Create `AGENTS.md` at each workspace level (root, `python/`, `js/`, `java/`, `spec/`) and symlink each as `CLAUDE.md`, so agents load both repo-wide and language-specific guidance automatically
- Consolidate duplicated content into root `AGENTS.md`; language files retain only workspace-specific commands, package tokens, and instrumentor patterns
- Add "Preferred Patterns" + "Common Pitfalls" to `python/AGENTS.md` based on audit of popular instrumentors

---

> [!NOTE]
> **Low Risk**
> Docs and `.gitignore` updates only; no runtime behavior changes, with minimal risk aside from potentially ignoring unintended files if paths overlap.
> 
> **Overview**
> Adds a new `python/AGENTS.md` with Python workspace setup, tox/pytest usage, and required instrumentation patterns (suppression, context propagation, `OITracer`/`TraceConfig`).
> 
> Adds `python/CLAUDE.md` pointing to `AGENTS.md`, and expands the instrumentation `.gitignore` to exclude additional instrumentor package directories (e.g. `openlit`, `pipecat`, `strands_agents`) from being tracked in the core module tree.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b9279d28708b1356d397c3e84872b36254a2a978. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/AGENTS.md
+++ b/python/AGENTS.md
@@ -1,338 +1,59 @@
 # Python Workspace Guide
 
-> **Start here — two things matter most:**
-> 1. **VCR/cassette tests** are the workflow for both feature development and debugging —
->    read [Testing Patterns](#testing-patterns) before writing any code or running tests.
-> 2. **Helpers from `openinference.instrumentation`** — check these before writing any
->    attribute-setting or span-creation logic; rolling custom solutions is the #1 review
->    blocker. See [Helper Functions](#helper-functions) for the full reference.
+## Essential Commands
 
-## Setup
+### Setup
 
 ```bash
+# requires Python 3.9+
 cd python
 pip install tox-uv==1.11.2
 pip install -r dev-requirements.txt
-tox run -e add_symlinks              # compose namespace package (required before any imports work)
-pip install -e openinference-instrumentation  # editable install for development
+tox run -e add_symlinks  # required before any imports work
+pip install -e openinference-instrumentation
 ```
 
----
+### Testing and Quality
 
-## Testing Patterns
-
-Every instrumentor uses **pytest-recording** (built on vcrpy) to capture real API
-interactions as YAML cassettes, then replays them in CI. Combined with
-`InMemorySpanExporter`, you can assert on span attributes without a live API key after the
-initial recording.
-
-### conftest.py setup
-
-```python
-import pytest
-from opentelemetry.sdk import trace as trace_sdk
-from opentelemetry.sdk.trace.export import SimpleSpanProcessor
-from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
-from openinference.instrumentation.<name> import <Name>Instrumentor
-
-@pytest.fixture(scope="session")
-def vcr_config():
-    return {
-        "filter_headers": ["authorization", "api-key", "x-api-key"],
-        "decode_compressed_response": True,
-        "record_mode": "once",       # record only if cassette is absent, then replay
-        "match_on": ["method", "scheme", "host", "port", "path", "query"],
-    }
-
-@pytest.fixture(scope="module")
-def in_memory_span_exporter() -> InMemorySpanExporter:
-    return InMemorySpanExporter()
-
-@pytest.fixture(scope="module")
-def tracer_provider(in_memory_span_exporter):
-    provider = trace_sdk.TracerProvider()
-    provider.add_span_processor(SimpleSpanProcessor(in_memory_span_exporter))
-    return provider
-
-@pytest.fixture(autouse=True)
-def instrument(tracer_provider, in_memory_span_exporter):
-    in_memory_span_exporter.clear()          # clean slate before each test
-    <Name>Instrumentor().instrument(tracer_provider=tracer_provider)
-    yield
-    <Name>Instrumentor().uninstrument()
-    in_memory_span_exporter.clear()          # don't leak spans into next test
-```
-
-### Recording cassettes
+Uses **pytest-recording** (vcrpy) — cassettes in `tests/cassettes/` are committed so CI needs no API key.
 
 ```bash
-# Record against the real API (requires API key) — creates cassettes/
-pytest tests/ -k test_my_test --vcr-record=once
-
-# Subsequent runs replay cassettes without an API key
-pytest tests/
+pytest tests/ -k test_name --vcr-record=once  # record against live API
+pytest tests/                                  # replay cassettes
 ```
 
-Cassettes are stored in `tests/cassettes/` and committed to git so CI needs no API key.
+Mark tests with `@pytest.mark.vcr`. Use `InMemorySpanExporter` to assert on spans.
 
-### Writing a test
+Required test categories: suppress tracing, context attribute propagation, trace config masking.
 
-```python
-@pytest.mark.vcr
-def test_basic_call(in_memory_span_exporter):
-    my_framework_call(...)
+### tox Commands
 
-    spans = in_memory_span_exporter.get_finished_spans()
-    assert len(spans) == 1
-    attributes = dict(spans[0].attributes or {})
-
-    # Pop pattern: test fails loudly if attribute is absent
-    assert attributes.pop(OPENINFERENCE_SPAN_KIND) == "LLM"
-    assert attributes.pop(LLM_MODEL_NAME) == "gpt-4o"
-    assert isinstance(attributes.pop(LLM_TOKEN_COUNT_PROMPT), int)
-    assert not attributes  # strict: no unexpected attributes remain
-```
-
-### Required test categories
-
-Test location: `tests/openinference/instrumentation/<name>/`
-
-All three categories must be present:
-
-- **Suppress tracing** — call inside `suppress_tracing()` context; assert zero spans produced.
-- **Context attribute propagation** — call inside `using_session("id")` context; assert `session.id` is set on the span.
-- **Trace config masking** — instrument with `TraceConfig(hide_inputs=True)`; assert `input.value` is absent from span attributes.
-
-Pytest configuration in `pyproject.toml`:
-
-```toml
-[tool.pytest.ini_options]
-asyncio_mode = "auto"
-```
-
-### Debugging with cassettes
-
-To investigate unexpected span output or API response shape:
+`ruff-mypy-test-openai` = hyphen-joined factors; not defined literally in `tox.ini`.
 
 ```bash
-# Re-record a single test against the live API (overwrites existing cassette)
-pytest tests/ -k test_my_test --vcr-record=all
-
-# Record only cassettes that don't exist yet (safe default)
-pytest tests/ -k test_my_test --vcr-record=once
-
-# Run without cassettes (live API, no recording — useful for one-off inspection)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,104 @@
+# OpenInference Repository Guide
+
+OpenInference is an OpenTelemetry-based AI/ML instrumentation library — a multi-language monorepo published to PyPI, `@arizeai` npm, and Maven Central. Owned by Arize AI.
+
+- **Python**: 33 instrumentation packages for OpenAI, LangChain, LlamaIndex, DSPy, and more
+- **JavaScript/TypeScript**: 13 packages (core + instrumentors) with pnpm workspaces
+- **Java**: 4 packages covering LangChain4j and Spring AI
```
cephalization: I would avoid encoding information in the agents file that can become stale; Counts and full path enumerations will probably drift and not get updated.

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+## Repository Structure
+
+```
+openinference/
+├── python/
+│   ├── openinference-instrumentation/        # Core: OITracer, TraceConfig, suppress_tracing
+│   ├── openinference-semantic-conventions/   # Python semconv constants
+│   └── instrumentation/                      # 33 individual instrumentor packages
+├── js/
+│   └── packages/                             # 13 packages (core + 8 instrumentors + specialized)
```
cephalization: same here, counts will get stale

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Development Commands
+
+```bash
+# Python — see python/AGENTS.md for full tox command reference
+pip install tox-uv==1.11.2 && pip install -r python/dev-requirements.txt
+tox run -e add_symlinks && tox run-parallel
+
+# JavaScript — see js/AGENTS.md for full command reference (MUST use pnpm, not npm)
+pnpm install --frozen-lockfile -r && pnpm run -r prebuild && pnpm run -r build
```
cephalization:
```suggestion
pnpm install --frozen-lockfile -r && pnpm run -r build
```

prebuild gets triggered automatically by running build

js/AGENTS.md
```diff
@@ -0,0 +1,228 @@
...
+
+---
+
+## Setup and Build
+
+```bash
+cd js
+nvm use                               # uses .nvmrc for Node v20
+pnpm install --frozen-lockfile -r    # MUST be pnpm, not npm or yarn
+pnpm run -r prebuild                 # REQUIRED: generates version files + cross-package symlinks
```
cephalization: This gets run automatically via `pnpm run -r build`

js/AGENTS.md
```diff
@@ -0,0 +1,228 @@
...
+pnpm run -r build
+```
+
+> **Why prebuild matters**: The prebuild script generates instrumentation version files and creates symlinks for cross-package dependencies. Cross-package imports will fail at test time if prebuild has not been run. Re-run it whenever you change packages in the repo.
```
cephalization: This is true, but it is supposed to run automatically whenever a build command is run

js/AGENTS.md
```diff
@@ -0,0 +1,228 @@
...
+pnpm run -r test
+pnpm run type:check
+pnpm run lint
+pnpm run prettier:check
+pnpm run prettier:write
+```
+
+---
+
+## Package Inventory
```
cephalization: This will get out of date; Maybe the agent md can be updated to tell agents to update the inventory when making new packages?

Either that or we add a CI step to enforce this somehow, like grep for package name inside of agents.md

js/AGENTS.md
```diff
@@ -0,0 +1,228 @@
...
+
+  it("creates a span", async () => {
+    // test body
+  });
+});
+```
+
+---
+
+## Creating a New JS Instrumentor
```
cephalization: Maybe this is a good place to instruct the agent to update AGENTS.md with the new package, update any counts, etc

spec/AGENTS.md
axiomofjoy, line 1: Feels like we're duplicating the spec here to a certain degree. Feels hard to keep in sync.

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+
+```python
+from opentelemetry import context as context_api
+from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY
+
+def patched_function(*args, **kwargs):
+    if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
+        return original_function(*args, **kwargs)  # skip tracing
+    # ... tracing logic ...
+```
```
axiomofjoy: This is implemented in the base instrumentor. Shouldn't be in the concrete implementations

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+
+To permanently disable tracing, implement `_uninstrument()` to reverse all monkey-patching.
+
+### Feature 2 — Context Attribute Propagation
```
axiomofjoy: I think this is also in the base class?

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+   - `src/openinference/instrumentation/<name>/version.py` — `__version__` string
+   - `src/openinference/instrumentation/<name>/package.py` — package name constant
+
+3. **Register in `python/tox.ini`**:
+   - Add `<token>: instrumentation/openinference-instrumentation-<name>/` under `changedir`
+   - Add `<token>: uv pip install ...` lines under `commands_pre`
+
+---
+
+## Testing Patterns
```
axiomofjoy: Testing needs to mention attributes, specifically the `pop` and assert never pattern. That may be something we want to make a skill.

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+    assert spans[0].attributes["session.id"] == "test-session-id"
+
+# 3. Trace configuration masking test
+def test_trace_config_masking(tracer_provider):
+    config = TraceConfig(hide_inputs=True)
+    instrumentor = MyFrameworkInstrumentor()
+    instrumentor.instrument(tracer_provider=tracer_provider, config=config)
+    result = my_framework_call(input="sensitive data")
+    spans = get_spans()
+    assert "input.value" not in spans[0].attributes
```
axiomofjoy: I don't think individual instrumentors need to test these if I recall

python/AGENTS.md
axiomofjoy, line 1: i would really encourage the llms to use the helper functions. single most important thing

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+        try:
+            return obj.model_dump_json(exclude_unset=True)
+        except Exception:
+            pass
+    if hasattr(obj, "model_dump"):
+        return safe_json_dumps(obj.model_dump())
+    if hasattr(obj, "dict"):
+        return safe_json_dumps(obj.dict())
+    return safe_json_dumps(obj)
+```
```
axiomofjoy: I would just pass to safe json dumps, don't think we need this helper all over the place.

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+    span = trace_api.get_current_span()
+    return (
+        span.get_span_context().is_valid
+        and span.is_recording()
+        and isinstance(span, ReadableSpan)
+        and (span.attributes or {}).get(OPENINFERENCE_SPAN_KIND) == "LLM"
+    )
+```
+
+### Pattern 5 — Parametrized tests covering multiple input shapes
```
axiomofjoy: Not sure what this one is covering, maybe nix

python/AGENTS.md
```diff
@@ -0,0 +1,396 @@
...
+
+# Correct — deferred inside _instrument
+def _instrument(self, **kwargs: Any) -> None:
+    try:
+        import my_framework
+    except ImportError as e:
+        raise RuntimeError("my-framework must be installed") from e
+```
+
+`TYPE_CHECKING`-guarded imports (for type annotations only) are fine at module level.
```
axiomofjoy: This needs to say something like don't import the instrumented library at the top of the file, only do nested imports.

python/AGENTS.md
caroger: Does that mean the pattern like following are outdated/unnecessary? 

```python
  # OpenAI _request.py:308
  def __call__(self, wrapped, instance, args, kwargs):
      if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
          return wrapped(*args, **kwargs)   # short-circuit before any tracer call

  # Anthropic _wrappers.py:89
  def __call__(self, wrapped, instance, args, kwargs):

  # Bedrock __init__.py:132 (and lines 185, 239 for other wrappers)
  if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
      return original_invoke_model(*args, **kwargs)

  # LangChain _tracer.py:167
  if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
      return
```

python/AGENTS.md
caroger: The core library tests verify the masking mechanism itself (OITracer/OpenInferenceSpan.set_attribute() +
  TraceConfig.mask()), while some instrumentors (openai, langchain) test that they set the correct attribute keys so masking can trigger. Happy to remove — just confirm and I'll update.

AGENTS.md
caroger: addressed

js/AGENTS.md
caroger: done - instruction in point #4

python/AGENTS.md
caroger: added here for now before turning to skills

python/AGENTS.md
caroger: added as #1 item under Preferred Patterns

python/AGENTS.md
caroger: done

spec/AGENTS.md
caroger: only kept the index

AGENTS.md
caroger: applied the suggestion. Root AGENTS.md now shows pnpm install --frozen-lockfile -r && pnpm run -r build only.

python/AGENTS.md
caroger: the suppress-tracing check in those wrappers is still needed and correct. BaseInstrumentor handles the instrumentation lifecycle (when to call _instrument() / _uninstrument()), but it doesn't inject a suppression check into every wrapper function at call time. Each __call__ / patched_function needs to check _SUPPRESS_INSTRUMENTATION_KEY on every invocation so it can short-circuit before creating a span for that specific request — the base class never sees those calls.

reframing this section to make this lifecycle vs. per-request distinction clearer.

python/AGENTS.md
caroger: reworded

PR #2758, opened 2026-02-19
fix(google-adk): Suppress duplicate generate_content spans from native ADK tracing

sanjeed5:
The instrumentor already suppresses native ADK tracers for `runners` and
`base_agent` via `_PassthroughTracer` to prevent double-instrumentation.
However, it does not suppress the tracer in
`google.adk.telemetry.tracing`, which creates `generate_content` spans
with scope `gcp.vertex.agent`.

This causes every LLM call to produce two spans: a `call_llm` span
(from OpenInference) and a `generate_content` span (from ADK native
tracing). Additionally, `_TraceCallLlm` calls `get_current_span()` after
the wrapped `trace_call_llm` returns, but at that point the
`generate_content` span is still current (it wraps the response generator
yield). So OpenInference attributes like `openinference.span.kind`,
`llm.input_messages`, and token counts land on the wrong span.

This fix extends `_disable_existing_tracers()` to also replace
`google.adk.telemetry.tracing.tracer` with `_PassthroughTracer`, using
the same pattern already applied to the other two ADK tracers. With the
native tracer suppressed, `_use_native_generate_content_span` yields
the current `call_llm` span instead of creating a child, and all
attributes merge onto a single span per LLM call.

# Checklist:

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py
@@ -158,7 +158,7 @@ class GoogleADKInstrumentor(BaseInstrumentor):  # type: ignore
         version = cast(tuple[int, int, int], tuple(int(x) for x in __version__.split(".")[:3]))
 
         if version >= (1, 15, 0):
-            from google.adk.telemetry import (
+            from google.adk.telemetry import (  # type: ignore[attr-defined,import-not-found,unused-ignore]
                 tracing as adk_tracing,  # type: ignore[attr-defined,unused-ignore]
             )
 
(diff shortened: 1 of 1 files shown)


sanjeed5:

sanjeed5:

sanjeed5:
Screenshot 2026-03-02 at 9 42 46 PM

this is the state on latest ADK now

call llm and generate content for a single model event

[Trace on Arize](https://app.arize.com/organizations/QWNjb3VudE9yZ2FuaXphdGlvbjoxNjEzMDpVTzRT/spaces/U3BhY2U6MTY5MTA6dUJjaA==/projects/TW9kZWw6Njc1ODU5NDI3OTovYmpt?queryFilterA=&selectedTab=llmSessions&timeZoneA=Asia%2FCalcutta&selectedTraceId=f27d8df1f3b5cd9e35cea11c99b83afb&modelType=generative_llm&startA=1770834600000&endA=1771486199999&envA=tracing&timeRangeKeyA=custom)

PR #2759, opened 2026-02-19
feat(google-adk): Add fallback output capture and agent description to agent spans

sanjeed5:
## Summary

- Add fallback `output.value` capture for agents that terminate via tool actions (e.g., `escalate=True`) where `is_final_response()` never fires
- Record `gen_ai.agent.description` on agent spans (standard OTel semantic convention)

## Problem

`_BaseAgentRunAsync` only captures `output.value` when `event.is_final_response()` returns `True`. Agents that terminate via tool-based escalation (e.g., a LoopAgent critic calling `approve_response` with `escalate=True`) exit through a tool action, not a text response. Every event in their stream contains function calls or function responses, so `is_final_response()` never fires and the span records no output.

## Fix

**Fallback output:** Track a `has_final_response` flag across the event stream. After the generator is exhausted, if no event triggered `is_final_response()`, write the last event's JSON as `output.value`. For tool-terminating agents this captures the terminal event (typically the `function_response` with the escalation action), showing why the agent stopped.

Uses a flag rather than checking `last_event.is_final_response()` at the end, so the fallback only fires when no event in the entire stream was a final response — avoiding overwrites if a final response event appeared earlier in the stream.

**Agent description:** Record `gen_ai.agent.description` from `BaseAgent.description` when non-empty. This is a [standard OTel GenAI semantic convention](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/) (Development status).

## What this PR intentionally does NOT do

**Agent input capture.** We investigated capturing `input.value` from `InvocationContext.user_content` but decided against it. ADK shares a single `InvocationContext` across the entire invocation tree, so every agent (root and sub-agents) would show the same original user message. For sub-agents, this is misleading — a critic agent appears to receive the user's question when it actually received the drafter's output plus tool results as context. The full agent input (system instruction + conversation history with injected delegation context) is only assembled inside `BaseLlmFlow` at LLM call time, which isn't accessible at the agent span level. Since a partial picture looks like the whole picture and could mislead users, we omit it.

## Test plan

- [x] Verified with google-adk 1.26.0: single-agent, multi-agent delegation, LoopAgent with tool-based escalation
- [x] Agents with text responses: `output.value` captured via existing `is_final_response()` path (no change)
- [x] Agents terminating via tool escalation: `output.value` captured via fallback with `function_response` JSON
- [x] `gen_ai.agent.description` present on agents with descriptions, absent on agents without
- [x] Re-recorded VCR cassettes and updated test assertions — all 7 tests pass


sanjeed5:

sanjeed5:
**Note on CI**: The VCR cassettes for the Google ADK instrumentor tests appear to be stale on `main` — the tests fail even without our changes. Re-recording isn't straightforward because:

1. `conftest.py` has an `autouse` fixture that overrides `GOOGLE_API_KEY` to `"xyz"`, preventing live API calls
2. Tests assert exact token counts (e.g., `== 106`, `== 25`) that change with every re-recording
3. The image artifacts test sends invalid image data that fails against the real API

caroger:
Hi @sanjeed5, we just merged #2789 to fix the issues with the tests set up you described above. 
Please do a merge/rebase with the latest main branch and you should be able to re-record the cassettes now by removing the existing cassette and re-run the test while having a valid GOOGLE_API_KEY or GEMINI_API_KEY set as an environment variable.

For faster feedback loop - you can verify if the CI would pass by running `uvx --with tox-uv tox run -e py314-ci-google_adk-latest` locally

sanjeed5:
@caroger Thanks! Rebased onto latest main (with #2789) and re-recorded all 4 VCR cassettes against the live API. Tests pass locally with `uvx --with tox-uv tox run -e py314-ci-google_adk-latest`. Ready for another look when you get a chance.

caroger (approved):

PR #2767, opened 2026-02-20
fix: update agent span name to include task name

sofia099:
In CrewAI, the agents execute tasks. In the trace structure, there is an agent span and in the attributes we include the task id and name. However, user shared that it would be useful if they could see the task name that was executed in the agent in the **name of the span**


sofia099:
Tested with a simple CrewAI example. The original agent span name:
```
Company Research Agent._execute_core
```
The new agent span name:
```
Company Research Agent.company_research_task._execute_core
```

Screenshot from Arize:
Screenshot 2026-02-24 at 10 56 17 AM

In the screenshot, you can see how useful this change is. There are two Content Personalizer agent calls but they execute different tasks in each run. By including the task name in the span name, its clear to see what the task was

PR #2807, opened 2026-02-26
fix(google-genai): accept dict-shaped content/part payloads

michal-bida:
Handle Mapping-based message params (e.g. pydantic-ai wrappers) without logging unexpected-type exceptions, and add regression tests.

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized change to request attribute extraction plus new regression tests; main risk is misclassifying dict-shaped payloads and extracting incomplete message attributes.
> 
> **Overview**
> **Improves Google GenAI request parsing to handle dict/`Mapping`-shaped message payloads.** `_RequestAttributesExtractor._get_attributes_from_message_param` now accepts TypedDict-style content/part objects (including `UserDict`), routing them through existing content/part extraction instead of logging an unexpected-type exception.
> 
> `get_attribute` is generalized from `dict` to `Mapping` with safer access (`.get`/`__getitem__` fallback). Adds tests covering dict and `UserDict` inputs and asserting message attributes are extracted without noisy exception logs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 286e9552824612ac5961446044087088703716e3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


michal-bida:

michal-bida:
recheck

michal-bida:
I’m integrating OpenInference as a third-party user and ran into an incompatibility between PydanticAI’s model instrumentation and openinference-instrumentation-google-genai.

**What I’m seeing**
`Runtime error: Unexpected input contents type: <class 'dict'>`
Where: in the Google GenAI request attribute extraction path (_request_attributes_extractor).

**Why this happens**
In practice, higher-level frameworks and “provider-agnostic” layers (like PydanticAI) may represent model input as plain Python dicts (or lists containing dicts) to normalize behavior across multiple model providers.
Different providers (and their SDKs) accept/emit “content/message” payloads in slightly different.

When OpenInference’s Google GenAI instrumentation encounters dict-shaped contents, it currently treats that as an unexpected type and raises, which breaks the instrumentation flow.

**The change here**
For better cross-provider compatibility, the Google GenAI request attribute extractor should be lenient when contents (or nested parts) are dict-like:
- don’t raise on dict inputs,
- instead best-effort normalize/serialize/extract what’s useful for span attributes,
- fall back gracefully when the shape is unknown.

**Impact**
This blocks using OpenInference Google GenAI instrumentation in setups where inputs are normalized across providers (e.g., PydanticAI), and it makes integrations brittle when switching model providers that represent messages as dicts.

This could have been addressed on pydantic ai side but I have noticed that other model providers like anthropic/openai do support dict types so felt the change here might be more systematic (and also more concise). 
But feel free to counter/suggest otherwise. Thanks.

RogerHYang:
Thank you for the PR. I appreciate the contribution. Could you share a small, reproducible example that demonstrates the behavior? That would help me step through the relevant code path and review it more effectively.

michal-bida:
> Thank you for the PR. I appreciate the contribution. Could you share a small, reproducible example that demonstrates the behavior? That would help me step through the relevant code path and review it more effectively.

This is basically just a generated rewrite of a test: 
[repro_google_genai_dict_contents_error.py](https://github.com/user-attachments/files/25599880/repro_google_genai_dict_contents_error.py)

What I think you had in mind was a minimal example showing the error when pydantic ai framework and openinference are used at the same time (right?). This is doable just note that for that you actually need to make a real API call to gemini model in order to get the error. I can provide that as well but first maybe it would be nice to address one caveat of this "fix".  That this is only relevant IFF one is using Logfire (via pydantic-ai) and OpenInference (Google GenAI instrumentor) at the same time. I cannot say if this is common or intended. 

This is my callpath analysis from our logs I did to arrive at the change here:
[logfire-trace-019c95298d3384cc137a8ce4fd3c095b-callpath.md](https://github.com/user-attachments/files/25599879/logfire-trace-019c95298d3384cc137a8ce4fd3c095b-callpath.md)

TLDR callstack: 
```
flowchart LR
  mvpService[mvpService] --> pydanticAi[pydanticAi]
  pydanticAi -->|builds_contents_as_dicts| googleModel[GoogleModel]
  googleModel -->|client.aio.models.generate_content(contents=...)| googleGenaiSdk[googleGenaiSdk]
  googleGenaiSdk -->|instrumented_wrapper| oiAsyncWrapper[OpenInferenceAsyncGenerateContentWrapper]
  oiAsyncWrapper -->|get_extra_attributes_from_request| oiRequestExtractor[_RequestAttributesExtractor]
  oiRequestExtractor -->|per_message| oiGetAttributesFromMessage[_get_attributes_from_message_param]
  oiGetAttributesFromMessage -->|logs_on_unhandled_type| unexpectedTypeLog["Unexpected input contents type: <class 'dict'>"]
```

PR #2872, opened 2026-03-10
feat(openai): adds support for python 3.14 to openai autoinstrumentor

### LINKED ISSUE #2647: [feature request] Support for Python 3.14
Can you please make it support for Python 3.14? Because currently all Arch Linux based distro use Python 3.14 mainly.

Currently this project has:
```
requires-python = ">=3.9, <3.14"
```

satyadevai:
Closes #2647


python/instrumentation/openinference-instrumentation-openai/test-requirements.txt
```diff
@@ -1,4 +1,4 @@
-openai==1.99.9
+openai==2.8.0
```
caroger, line 1: is this the minimum openai version that works for py314?

caroger: does it (2.8.0) work for earlier version of python?

python/tox.ini
```diff
@@ -8,7 +8,7 @@ envlist =
   py3{9,14}-ci-instrumentation
   py3{9,14}-ci-{bedrock,bedrock-latest}
   py3{9,14}-ci-{mistralai,mistralai-latest}
-  py3{9,13}-ci-{openai,openai-latest}
+  py3{9,14}-ci-{openai,openai-latest}
```
caroger, line 11: let's see if we can increase the lower bound to py310

python/instrumentation/openinference-instrumentation-openai/test-requirements.txt
satyadevai: @caroger CI is passing for Python 3.9 and 3.14, so the OpenAI version works fine with the lower bound Python version.

python/tox.ini
satyadevai: OpenAI still supports Python 3.9. Do you still want to remove support for 3.9?

PR #2875, opened 2026-03-10
feat(mistralai)!: migrate instrumentor to mistralai>=2.0.0

caroger:
Summary

  - Migrate all mistralai imports to mistralai.client.* to support mistralai>=2.0.0, which restructured its package layout into a namespace package with all classes under mistralai.client
  - Update all 8 wrap_function_wrapper module paths from mistralai.chat/mistralai.agents to mistralai.client.chat/mistralai.client.agents
  - Bump minimum mistralai version from >=1.0.2 to >=2.0.0 and pin test-requirements to ==2.0.0
  - Bump CI lower-bound Python from 3.9 to 3.10 (required by mistralai>=2.0.0)
  - Enable mypy type-checking for mistralai.client imports (ships py.typed) and add targeted # type: ignore comments where needed

  Test plan

  - py314-ci-mistralai (pinned ==2.0.0) — ruff, mypy, 20 tests pass
  - py314-ci-mistralai-latest — ruff, mypy, 20 tests pass
  - Verify no other packages depend on mistralai<2.0.0


mikeldking:
@claude this is a breaking change to the package. Can you add a compatibility matrix to the mistral README.  Support for mistral 2 will be in the next major version of the package

PR #2887, opened 2026-03-14
feat: adding Java annotation instrumentation and programmatic support

duncankmckinnon:
## Summary

- Add `openinference-instrumentation-annotation` module providing annotation-based manual tracing for Java AI applications.
- Typed span classes (`AgentSpan`, `ChainSpan`, `LLMSpan`, `ToolSpan`, `RetrievalSpan`, `EmbeddingSpan`, plus the shared `TracedSpan` base) with try-with-resources
support and automatic OpenInference attribute handling.
- Method annotations (`@Agent`, `@Chain`, `@LLM`, `@Tool`, `@Span`, `@SpanMapping`, `@SpanIgnore`) backed by a ByteBuddy agent for automatic span creation, input capture, and attribute extraction via `SpanHelper`.
- `SpanAttribute` enum + `OpenInferenceAgent` runtime holder, wired into release-please.
- **Restructure:** move the typed span sources/tests (including `OpenInferenceAgent`, `SpanSerializer`, etc.) into `java/openinference-instrumentation/src/{main,test}/java/com/arize/instrumentation/trace`, keeping the shared Trace API inside the base module; update annotation advice/tests accordingly and add `SpanSerializerTest`.
- **Agent packaging & docs:** `java/instrumentation/openinference-instrumentation-annotation/build.gradle` now writes `Premain-Class`/`Agent-Class` (with redefine/retransform flags) so `-javaagent` works; READMEs and the annotation example explain packaging/running the agent and configuring Arize vs Phoenix. 
- Added `TracedSpanAttributeTest` to cover typed-attribute handling, and verified the example with both Phoenix fallback and Arize export.

## Test plan

- [x] `SpanHelperTest` — JSON/primitive serialization, `@SpanIgnore`, empty/all-ignored parameter handling
- [x] `MappingTest` — parameter/return-field extraction (simple, nested, missing, null, empty path cases)
- [x] `TraceAdviceTest` — ByteBuddy advice wiring for annotated methods, tracer registration, suppression
- [x] `AgentSpanTest` — agent span naming, metadata propagation
- [x] `ChainSpanTest` — (part of `IntegrationTest`; spans nested under agent) *
- [x] `LLMSpanTest` — model/system/provider attributes, token counts, costs, invocation params, message capture, hideInput/Output configs
- [x] `ToolSpanTest` — tool name/description, parameter capture, JSON schema, hideToolParameters config
- [x] `RetrievalSpanTest` — retriever span kind, document capture (empty vs populated)
- [x] `EmbeddingSpanTest` — embedding span kind, model name, embedding arrays, hideOutputEmbeddings config
- [x] `SpanSerializerTest` — typed attribute serialization (numbers, booleans, collections, JSON fallback)
- [x] `TracedSpanAttributeTest` — verifies numeric/boolean attribute typing, string arrays, complex-object fallback in exported spans
- [x] `IntegrationTest` — nested agent/chain/LLM/tool flows covering metadata/tags/session IDs
- [x] `JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./gradlew :openinference-instrumentation:test`
- [x] `./gradlew :openinference-instrumentation:spotlessApply :spotlessCheck`
- [x] `./gradlew :examples:annotation-example:run` (validated with Phoenix fallback and with `ARIZE_API_KEY`/`ARIZE_SPACE_ID`)

\*Chain spans are exercised within `IntegrationTest`, which asserts the nested hierarchy (agent → chain → tool → LLM).


java/instrumentation/openinference-instrumentation-annotations/src/main/java/com/arize/instrumentation/annotations/SpanAttribute.java
```diff
@@ -0,0 +1,70 @@
+package com.arize.instrumentation.annotations;
+
+import com.arize.semconv.trace.SemanticConventions;
+import lombok.Getter;
+
+@Getter
+public enum SpanAttribute {
```
nate-mar: Is there a reason this exists as a separate enum? Java annotations accept compile-time constant strings, so `@SpanMapping` could reference `SemanticConventions` constants directly:

```java
// instead of
@SpanMapping(field = "usage.totalTokens", attribute = SpanAttribute.LLM_TOKEN_COUNT_TOTAL)

// this works
@SpanMapping(field = "usage.totalTokens", attribute = SemanticConventions.LLM_TOKEN_COUNT_TOTAL)
```

Maintaining a second copy of these constants means every new attribute added to `SemanticConventions` needs a corresponding enum entry here too.

nate-mar:
Annotation and class naming: these should align more closely with OpenInference span kinds.

The naming feels inconsistent across the module:

| Layer | Pattern | Example |
|-------|---------|---------|
| Annotations | `Trace` + kind | `@TraceChain`, `@TraceLLM` |
| Span classes | `Traced` + kind + `Span` | `TracedChainSpan`, `TracedLLMSpan` |
| Python SDK | just the kind | `@tracer.chain`, `@tracer.llm` |
| Semantic conventions | just the kind | `CHAIN`, `LLM` |

The `Trace`/`Traced` prefixes don't add information since the package already conveys tracing context. The names should map 1:1 to the OpenInference span kinds (`CHAIN`, `LLM`, `TOOL`, `AGENT`, etc.) so the connection to the spec is obvious, and to stay consistent with our Python SDK.

One tension worth noting: in Python, `@tracer.chain` is self-documenting because the tracer instance provides context. In Java, `@Chain` alone on a method is more ambiguous. That said, this is how most Java frameworks handle it. Spring's `@Transactional`, `@Cacheable`, `@Async` don't encode their domain in the name; the import (`import org.springframework.transaction.annotation.Transactional`) provides that context. Same applies here: `import com.arize.instrumentation.annotations.Chain` makes the purpose clear.

There are a few options here. Since OpenInference span kinds have distinct per-kind attributes (e.g., `description` on tool), typed annotations make sense:

    // Typed aliases, 1:1 with OpenInference span kinds
    @Chain(name = "retriever")
    @LLM(name = "generator")
    @Tool(name = "weather", description = "Gets weather")
    @Agent(name = "qa-agent")

    // Span classes, same naming pattern
    try (ChainSpan span = ChainSpan.start(tracer, "retriever")) { ... }
    try (LLMSpan span = LLMSpan.start(tracer, "generator")) { ... }

We could also add a base annotation for less common span kinds (RETRIEVER, EMBEDDING, RERANKER) so we don't need a dedicated annotation every time a new kind is added. Something like `@Span` or `@OISpan` would work better than `@WithSpan`, since ours behaves differently from OTel's `@WithSpan` (auto-captures inputs/outputs, sets OpenInference span kinds, applies `@SpanMapping`, etc.).

nate-mar:
The package name `com.arize.instrumentation.annotations` should probably be singular `com.arize.instrumentation.annotation`.

Java convention is to use singular nouns in package names. The JDK and major frameworks consistently follow this:

- `java.lang.annotation`
- `jakarta.annotation`
- `org.springframework.transaction.annotation`
- `io.opentelemetry.instrumentation.annotation`
- `io.micronaut.tracing.annotation`

All singular, even though they contain multiple annotation classes.

java/instrumentation/openinference-instrumentation-annotations/src/main/java/com/arize/instrumentation/annotations/SpanAttribute.java
duncankmckinnon: Yeah, I wasn't sure if these could be referenced - I will update and see if it works

nate-mar:
The span classes (ChainSpan, LLMSpan, ToolSpan, etc.) should live in the core
`openinference-instrumentation` module, not in the annotation module. They're a
standalone programmatic API that doesn't depend on annotations or ByteBuddy.

Users who only need the programmatic API should be able to depend on
`openinference-instrumentation` alone without pulling in ByteBuddy, which is
only needed for the annotation interception layer.

This is a common pattern in Java libraries:

- OpenTelemetry separates these into two repos entirely:
  `io.opentelemetry.api.trace` (Span, Tracer, SpanBuilder) lives in
  `opentelemetry-java`, while `@WithSpan` lives in
  `opentelemetry-java-instrumentation/instrumentation-annotations/`

- Swagger/OpenAPI splits annotations from model classes across artifacts:
  `swagger-annotations` has `@Operation`, `@Parameter`, `@Schema`;
  `swagger-models` has the corresponding `Operation`, `Parameter`,
  `Schema` classes. Similar naming relationship to our `@Chain` / `ChainSpan`.

- As an example outside observability, Jackson does the same:
  `jackson-core` has the streaming API (JsonParser, JsonGenerator),
  `jackson-annotations` (@JsonProperty, @JsonIgnore) is a separate
  lightweight artifact with minimal dependencies.

Proposed structure:

    com.arize.instrumentation              // core module (already exists)
    ├── OITracer.java
    ├── TraceConfig.java
    └── trace/
        ├── ChainSpan.java
        ├── LLMSpan.java
        ├── ToolSpan.java
        ├── AgentSpan.java
        ├── RetrievalSpan.java
        └── EmbeddingSpan.java

    com.arize.instrumentation.annotation   // annotation module (this PR)
    ├── Chain.java
    ├── LLM.java
    ├── Tool.java
    ├── Agent.java
    ├── Span.java
    └── ...

Gradle-wise, Jackson moves to core (span classes need it for serialization),
and the annotation module drops it (gets it transitively):

    // openinference-instrumentation/build.gradle
    // add:
    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"

    // openinference-instrumentation-annotation/build.gradle
    // remove jackson, keep ByteBuddy
    dependencies {
        api project(':openinference-instrumentation')
        implementation "net.bytebuddy:byte-buddy:${byteBuddyVersion}"
        implementation "net.bytebuddy:byte-buddy-agent:${byteBuddyVersion}"
        implementation "org.slf4j:slf4j-api:${slf4jVersion}"
        testImplementation "io.opentelemetry:opentelemetry-sdk-testing"
    }

nate-mar:
Question on auto-capturing all method parameters as input.value:

In TraceAdvice.onEnter(), all parameters (minus @SpanIgnore) are automatically serialized into the span's input.value:

    Map<String, Object> input = SpanHelper.buildInputMap(method, args);
    if (!input.isEmpty()) {
        span.setInput(input.size() == 1 ? input.values().iterator().next() : input);
    }

What's the specific use case for always setting input.value to the full parameter map? In practice, input.value varies by span kind. For an LLM span it's typically the prompt, for a tool it's the structured arguments, for a chain it might be a question string. Dumping all parameters doesn't seem like the right default.

For example:

    @LLM(name = "generator")
    public String generate(String prompt, String model, int maxTokens) { ... }

This produces input.value = {"prompt": "...", "model": "...", "maxTokens": 100}, when the user probably only wants input.value to be the prompt.

Could we remove the auto-capture and let the user explicitly set input via @SpanMapping or programmatically via Span.current()?

    // Via @SpanMapping
    @LLM(name = "generator", mapping = @SpanMapping(param = "prompt", attribute = "input.value"))
    public String generate(String prompt, String model, int maxTokens) { ... }

    // Or programmatically
    @LLM(name = "generator")
    public String generate(String prompt, String model, int maxTokens) {
        LLMSpan.current().setInput(prompt);
        ...
    }

duncankmckinnon:
@nate-mar Appreciate the feedback - I agree with the restructuring, I had a sense it was limiting the broader functionality but the breakdown really helps. For the second point, I think we need a useful default so they can get something in their traces without too much work. That would at least inform them as to what is truly important if they want to add a mapping.

nate-mar:
Ok, that's fair on the second point. It's my personal opinion that I think it muddies up the input.value and also duplicates the data with whatever other specific attributes are used, but I'm ok with this if it helps with early adoption. I would like to understand how users eventually end up using input.value.

java/openinference-instrumentation/src/main/java/com/arize/instrumentation/trace/LLMSpan.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public void setOutputMessages(List<Map<String, Object>> messages) {
+        if (config.isHideOutputMessages()) return;
+        SpanSerializer.SerializedValue sv = SpanSerializer.serialize(messages);
+        if (sv != null) {
+            span.setAttribute(AttributeKey.stringKey(SemanticConventions.LLM_OUTPUT_MESSAGES), sv.value());
+        }
+    }
+
+    public void setTokenCountPrompt(int count) {
```
nate-mar: `setTokenCountPrompt/Completion/Total` accept `int` but the underlying OTel attribute is `longKey`. Worth fixing before v0.1.0 since changing `int` to `long` after release is source-breaking for every caller.

java/instrumentation/openinference-instrumentation-annotation/src/main/java/com/arize/instrumentation/annotation/TraceAdvice.java
```diff
@@ -0,0 +1,175 @@
...
+        SpanMapping[] mappings = getInputMappings(method);
+        if (mappings == null || mappings.length == 0) return;
+
+        var params = method.getParameters();
+        for (SpanMapping mapping : mappings) {
+            if (mapping.param().isEmpty()) continue;
+            for (int i = 0; i < params.length; i++) {
+                String paramName = params[i].isNamePresent() ? params[i].getName() : "arg" + i;
+                if (paramName.equals(mapping.param())) {
+                    span.setAttribute(mapping.attribute(), args[i]);
```
nate-mar, line 138: `applyInputMappings` and `applyOutputMappings` call `span.setAttribute()` directly, bypassing the `TraceConfig` hiding checks that `setInput()`/`setOutput()` enforce. This means a user who has explicitly configured `hideInputs = true` or `hideOutputs = true` to protect sensitive data will still have that data written to spans if a `@SpanMapping` targets `input.value` or `output.value`. The user has signaled intent to suppress this data, so I think the library should honor that regardless of the code path. `TracedSpan.setAttribute()` should check `config.isHideInputs()` / `config.isHideOutputs()` when the key is `input.value` or `output.value` before writing to the span.

Suggested fix in `TracedSpan.setAttribute()`, after the null checks:

```java
if (SemanticConventions.INPUT_VALUE.equals(key) && config.isHideInputs()) {
    return;
}
if (SemanticConventions.OUTPUT_VALUE.equals(key) && config.isHideOutputs()) {
    return;
}
```

java/instrumentation/openinference-instrumentation-annotation/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+
+    // Test
+    testImplementation "io.opentelemetry:opentelemetry-sdk-testing"
+}
+
+jar {
+    manifest {
+        attributes(
+                'Premain-Class': 'com.arize.instrumentation.annotation.OpenInferenceAgentInstaller',
+                'Agent-Class': 'com.arize.instrumentation.annotation.OpenInferenceAgentInstaller',
```
nate-mar, line 25: The manifest declares `Agent-Class` but there's no `agentmain` method. The major Java instrumentation agents (OTel, Datadog, Elastic) all implement both `premain` and `agentmain`, though none of them officially recommend runtime attach for production. Since this library already has the `install()` method for programmatic self-attach via ByteBuddy, is there a reason to keep the `Agent-Class` manifest entry? If not, I'd suggest removing it. If you do want to support it, it's a one-liner that delegates to `installAgent(inst)`.

java/instrumentation/openinference-instrumentation-annotation/src/main/java/com/arize/instrumentation/annotation/OpenInferenceAgentInstaller.java
```diff
@@ -0,0 +1,67 @@
...
+    public static void premain(String args, Instrumentation inst) {
+        log.info("Installing OpenInference annotation tracing agent (premain)");
+
+        try {
+            String agentJarPath = OpenInferenceAgentInstaller.class
+                    .getProtectionDomain()
+                    .getCodeSource()
+                    .getLocation()
+                    .toURI()
+                    .getPath();
```
nate-mar: `toURI().getPath()` returns `/C:/Users/...` on Windows (see [JDK-8197918](https://bugs.openjdk.org/browse/JDK-8197918)), which is invalid for `new JarFile(...)`. This means the agent JAR won't be added to the bootstrap classloader on Windows, which could cause `ClassNotFoundException` at runtime. Consider using `Paths.get(uri)` or `new File(uri)` instead, which handle Windows paths correctly.

Also, if anyone uses this via `-javaagent` rather than the programmatic `install()` path, the annotation module's agent JAR doesn't bundle its dependencies (ByteBuddy, etc.), so they'll get `NoClassDefFoundError`. Consider adding the Gradle Shadow plugin to the annotation module to produce a self-contained agent JAR, or document that `-javaagent` usage requires ByteBuddy on the classpath separately.

java/examples/annotation-example/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+application {
+    mainClass = 'com.arize.examples.annotation.annotationDemo'
+}
+
+/**
+run {
+    environment 'ARIZE_API_KEY', 'your api key'
+    environment 'ARIZE_SPACE_ID', 'your space id'
+    environment 'PROJECT_NAME', 'annotation-example'
+}
```
nate-mar: I'm a little bit confused about what these examples are intending to target.  This references ARIZE_SPACE_ID, but the README.md makes no mention of it, but instead targets Phoenix and Phoenix cloud. Can you elaborate on what this is intending on covering?

nate-mar:
The name `SpanIgnore` can read like "ignore a span" (suppress the entire span), but it actually excludes a single parameter from being captured. I wonder if something like `@ExcludeFromSpan` would make the target clearer?

The established convention in popular annotation libraries like OTel's `@SpanAttribute` and Spring's `@SpanTag` is opt-in: unannotated parameters are not captured, and you annotate the ones you want included. Since this code captures everything by default which is something you argued for above, it might be worth calling that out explicitly in the docs so developers coming from those libraries aren't surprised by the inverted default.

java/instrumentation/openinference-instrumentation-annotation/README.md
```diff
@@ -0,0 +1,266 @@
...
+public LLMResponse chat(String model, String prompt) { ... }
+```
+
+The `field` parameter uses dot notation to extract nested values from return objects (e.g., `"usage.totalTokens"`).
+
+> **Note:** Parameter mappings rely on Java's `-parameters` compiler flag. If your application
+> isn't compiled with that flag, reference the generated `arg0`, `arg1`, … parameter names in your
+> `@SpanMapping` annotations instead.
+
+## Programmatic API
```
nate-mar, line 181: Can we put this in a separate Readme under the core manual trace library code since it more directly relates to that part of the package after you moved it?

nate-mar:
I noticed the only example right now is annotation-example/, which requires the ByteBuddy agent. The programmatic API (`LLMSpan.start()`, `AgentSpan.start()`, etc. with try-with-resources) is only shown in test files like IntegrationTest.java.

I wonder if a standalone example for the core trace library would be worth adding? The annotations module requires a Java agent (either `OpenInferenceAgentInstaller.install()` or `-javaagent`), which isn't viable in all environments (GraalVM native images, security-restricted deployments). A core lib example would give those users a clear path and also make the two-layer architecture more obvious: openinference-instrumentation as the foundation, annotations as a convenience layer on top.

README.md
```diff
@@ -147,13 +147,15 @@ Normalize and convert data across other instrumentation libraries by adding span
...
+| [`openinference-instrumentation-annotation`](./java/instrumentation/openinference-instrumentation-annotation) | Annotation-based manual tracing with ByteBuddy. | [![Maven Central Version](https://img.shields.io/maven-central/v/com.arize/openinference-instrumentation-annotation)](https://central.sonatype.com/artifact/com.arize/openinference-instrumentation-annotation) |
 
 ### Examples
 
 | Name                                                     | Description                                    | Complexity Level |
 | -------------------------------------------------------- | ---------------------------------------------- | ---------------- |
 | [LangChain4j Example](java/examples/langchain4j-example) | Simple example using LangChain4j with OpenAI   | Beginner         |
 | [Spring AI Example](java/examples/spring-ai-example)     | Spring AI example with OpenAI and tool calling | Beginner         |
+| [Annotation Example](java/examples/annotation-example) | Annotation-based tracing with `@Chain`, `@LLM`, `@Tool`, `@Agent` | Beginner         |
```
nate-mar, line 157: Can we update examples for the programmatic package?

java/instrumentation/openinference-instrumentation-annotation/src/main/java/com/arize/instrumentation/annotation/TraceAdvice.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public static TracedSpan createTypedSpan(
+            OITracer tracer, String name, SemanticConventions.OpenInferenceSpanKind kind) {
+        return switch (kind) {
+            case LLM -> LLMSpan.start(tracer, name);
+            case TOOL -> ToolSpan.start(tracer, name);
+            case AGENT -> AgentSpan.start(tracer, name);
+            case RETRIEVER -> RetrievalSpan.start(tracer, name);
+            case EMBEDDING -> EmbeddingSpan.start(tracer, name);
+            default -> ChainSpan.start(tracer, name);
```
nate-mar: The `default` branch here delegates to `ChainSpan.start()`, which hardcodes the span kind to `CHAIN`. So if someone uses `@Span(kind = RERANKER)` (or GUARDRAIL, EVALUATOR), the kind they specified gets silently overwritten, which could cause unexpected behavior for users who see their spans labeled differently than what they annotated. Is the intent to add dedicated span classes for those later? If not, I wonder if the default branch could just call `startSpan(tracer, name, kind)` directly so the passed-in kind is preserved.

duncankmckinnon: @nate-mar 

## Bug fixes:

- int → long for token count methods in LLMSpan.java
- hideInputs/hideOutputs config now respected in TracedSpan.setAttribute() — closes the @SpanMapping bypass
- Windows path fix using new File(URI) in OpenInferenceAgentInstaller.java
- Span kind preserved in default branch via ChainSpan.start(tracer, name, kind) in TraceAdvice.java

## New features:
- agentmain method added to OpenInferenceAgentInstaller
- Shadow plugin added to build.gradle— produces *-all.jar fat JAR for -javaagent use

## Renames:
- @SpanIgnore → @ExcludeFromSpan (deleted old file, created new, updated all 7 referencing files)

## Documentation:
- Created core library README with programmatic API docs
- Trimmed annotation README to link to core README
- Added Phoenix + Arize examples to annotation example build.gradle
- Added programmatic example row to root README

## New example:
- Created programmatic-example — standalone core trace library example (no annotations/ByteBuddy)

java/openinference-instrumentation/src/main/java/com/arize/instrumentation/trace/TracedSpan.java
```diff
@@ -0,0 +1,183 @@
...
+    public void setAttribute(String key, Object value) {
+        if (key == null || key.isEmpty() || value == null) {
+            return;
+        }
+        if (SemanticConventions.INPUT_VALUE.equals(key) && config.isHideInputs()) {
+            return;
+        }
+        if (SemanticConventions.OUTPUT_VALUE.equals(key) && config.isHideOutputs()) {
+            return;
+        }
```
nate-mar, line 73: The hide checks here only cover `input.value` and `output.value`, but in Python and TypeScript `hideInputs` is the broad parent flag that also suppresses `input.mime_type`, all `llm.input_messages.*`, and `llm.prompts`. Similarly `hideOutputs` suppresses `output.mime_type` and all `llm.output_messages.*`.

The narrower flags (`hideInputMessages`, `hideOutputMessages`, etc.) all exist on `TraceConfig` already but none of them are wired up here or in LangChain4j/SpringAI.

For reference, the Python/TS hierarchy:
- `hideInputs` (broad): suppresses `input.value` + `input.mime_type` + `llm.input_messages.*` + `llm.prompts`
- `hideInputMessages` (narrow): suppresses only `llm.input_messages.*`
- `hideOutputs` (broad): suppresses `output.value` + `output.mime_type` + `llm.output_messages.*`
- `hideOutputMessages` (narrow): suppresses only `llm.output_messages.*`

This is a cross-cutting issue across all three Java instrumentors, so probably best addressed in a follow-up PR. But since `setAttribute` is the centralized write path for both the annotation and programmatic APIs, this would be the natural place to implement it.

nate-mar (approved):
Approving, but let's please address the hideInputs/outputs inconsistency either here are immediately after (and before  a release is pushed).

nate-mar:
Can we also update the title of the PR so it's accurate of what this PR is offering (i.e. both annotation and programmatic support)

PR #2889, opened 2026-03-15
feat(langchain4j): Added Agent and Tool Listeners

### LINKED ISSUE #2741: LangChain4j Autoinstrumentor Records Tool Calls as Attributes Instead of Separate OpenInference Spans
https://github.com/Arize-ai/openinference/blob/7762ebe2c5346d93ae6a442367a561456d567c65/java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jModelListener.java#L196

**Overview**

When using the langchain4j autoinstrumentor from the OpenInference library, tool and agent execution is currently not being represented as separate OpenInference-native spans.

Instead, tool calls are being captured and stored as span attributes on a single, monolithic parent trace span.

This differs from the behavior seen in other OpenInference autoinstrumentors, where tool invocations and agent steps are typically emitted as their own child spans. As a result, traces become difficult to interpret and lose the expected hierarchical structure for tool usage.

**Current Behavior**

Tool calls executed through LangChain4j are not surfaced as individual spans.

Instead:

- The entire agent execution appears as one large span
- 
- Tool call information is embedded inside span attributes
- 
- There is no clear separation between LLM reasoning steps and tool execution
- 
- Trace visualization platforms (ex: Arize) display this as a monolithic trace rather than a structured agent/tool breakdown

This makes it difficult to:

- Analyze tool latency independently
- 
- Debug agent decision-making
- 
- Maintain consistency with OpenInference semantic conventions
- 
- Compare traces across frameworks/instrumentors

_Screenshot 1 attached (Current Autoinstrumentation Output):_
Shows tool calls captured only as attributes within a single span.

**Expected Behavior**

Tool invocations should be represented as first-class OpenInference spans, consistent with other autoinstrumentors and OpenInference tracing conventions.

Specifically:

- Each tool call should generate its own child span
- 
- Tool spans should include relevant metadata (tool name, arguments, outputs)
- 
- Agent execution should appear as a structured span tree rather than flattened into one span

This provides clearer trace structure and improves observability across agent workflows.

_Screenshot 2 attached (Manual Instrumentation Example):_
Shows the desired behavior, where tool calls appear as separate spans in the trace hierarchy.

**Reproduction**

1. Enable OpenInference autoinstrumentation for LangChain4j
2. Run an agent workflow that triggers tool execution
3. Observe that tool calls are stored as span attributes instead of separate spans

**Impact**

This behavior prevents LangChain4j users from getting full OpenInference-native trace structure for agent + tool workflows, and creates inconsistency across supported instrumentors.

It also forces teams to implement manual instrumentation to achieve expected span breakdown, which adds significant overhead.

**Suggested Fix**

Update the LangChain4j autoinstrumentor to emit tool executions as separate OpenInference spans rather than encoding them only as attributes on the parent span.

**Additional Context**

We have validated that manual instrumentation produces the expected span structure, but the goal is for this to work correctly via autoinstrumentation out of the box.

Image

Image

satyadevai:
Closes #2741 

This PR extends the LangChain4j instrumentation to properly capture AiService Invocations Including(LLM, Tool Calls) as first-class OpenInference spans, rather than recording them as flat attributes on the LLM span. It introduces a full set of AI service lifecycle listeners and a tool execution listener that integrate with LangChain4j's observability API.

**New Listeners**

- LangChain4jAiServiceStartedListener — Creates a CHAIN-kind span at the start of an AI service invocation, capturing system and user messages as structured input.value (JSON MIME type).
- LangChain4jServiceCompletedListener — Ends the Chain span on successful completion, recording the final output.
- LangChain4jServiceResponseReceivedListener — Captures intermediate LLM responses during multi-turn/agentic execution.
- LangChain4jServiceRequestIssuedListener — Records outgoing LLM requests issued by the AI service.
- LangChain4jAiServiceErrorListener — Handles error events by recording the exception on the span and setting status to ERROR.
- LangChain4jToolExecutedEventListener — Creates a dedicated span for each tool execution, enabling per-tool visibility in traces.

**Utilities:**
- SpanContext utility — Stores active span + context by invocation ID for correlation across lifecycle events.
- ChatMessageAttributeUtils — Converts ChatMessage objects to structured attribute maps for span serialization.
- LangChain4jInstrumentor — Updated with factory methods (createAiServiceStartedListener(), createToolExecutedListener(), etc.) to expose all new listeners.

**Dependency Update**

Bumped langchain4jVersion from 1.0.0 → 1.11.0 to gain access to the observability API (AiServiceStartedEvent, AiServiceErrorEvent, ToolExecutedEvent, etc.).

**Example**
Added ArizeAXExample.java demonstrating a MathAssistant AI service with 4 tools (add, subtract, multiply, divide), wired up with all new listeners and exporting traces to an OTLP endpoint (e.g., Arize Phoenix).

**Testing**
Manual end-to-end validation using the ArizeAXExample with OpenAI gpt-4.1-nano

**Screenshots**
image
image
image

### DIFF
--- a/java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
+++ b/java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
@@ -77,7 +77,7 @@ public class ChatMessageAttributeUtils {
         } else if (message instanceof ToolExecutionResultMessage toolExecutionResultMessage) {
             return toolExecutionResultMessage.text();
         }
-        return "";
+        return null;
     }
 
     /**
@@ -95,9 +95,13 @@ public class ChatMessageAttributeUtils {
             String role = mapMessageRole(message.type());
             span.setAttribute(AttributeKey.stringKey(prefix + SemanticConventions.MESSAGE_ROLE), role);
 
-            // Set content
-            span.setAttribute(
-                    AttributeKey.stringKey(prefix + SemanticConventions.MESSAGE_CONTENT), getMessageContent(message));
+            String inputMessageContent = getMessageContent(message);
+
+            if (Objects.nonNull(inputMessageContent)) {
+                span.setAttribute(
+                        AttributeKey.stringKey(prefix + SemanticConventions.MESSAGE_CONTENT),
+                        getMessageContent(message));
+            }
 
             // Set ToolCall
             if (message.type().equals(ChatMessageType.AI)) {
@@ -263,8 +267,8 @@ public class ChatMessageAttributeUtils {
         try {
             span.setAttribute(
                     SemanticConventions.LLM_INVOCATION_PARAMETERS, objectMapper.writeValueAsString(invocationParams));
-        } catch (JsonProcessingException e) {
-            logger.log(Level.WARNING, "Failed to serialize invocation parameters", e);
+        } catch (JsonProcessingException jpe) {
+            logger.log(Level.WARNING, "Failed to serialize invocation parameters", jpe);
         }
 
         // Set tool attributes
@@ -284,8 +288,10 @@ public class ChatMessageAttributeUtils {
                 String messagesJson = objectMapper.writeValueAsString(messagesList);
                 span.setAttribute(SemanticConventions.INPUT_VALUE, messagesJson);
                 span.setAttribute(SemanticConventions.INPUT_MIME_TYPE, "application/json");
-            } catch (Exception e) {
+            } catch (JsonProcessingException e) {
                 logger.log(Level.WARNING, "Failed to serialize input messages", e);
+            } catch (Exception e) {
+                logger.log(Level.WARNING, "Failed to process chatRequest", e);
             }
         }
     }
--- a/java/instrumentation/openinference-instrumentation-langchain4j/src/test/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceListenerTest.java
+++ b/java/instrumentation/openinference-instrumentation-langchain4j/src/test/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceListenerTest.java
@@ -277,8 +277,7 @@ class LangChain4jAiServiceListenerTest extends BaseInstrumentationSetup {
 
         assertThat(finalLlmAttrs.remove(AttributeKey.stringKey("llm.input_messages.2.message.role")))
                 .isEqualTo("assistant");
-        assertThat((String) finalLlmAttrs.remove(AttributeKey.stringKey("llm.input_messages.2.message.content")))
-                .contains("");
+
         assertThat(finalLlmAttrs.remove(
                         AttributeKey.stringKey("llm.input_messages.2.message.tool_calls.0.tool_call.id")))
                 .isEqualTo("call_Qm51xgE57uzoj2f6qkZAWwb4");


java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jToolExecutedEventListener.java
```diff
@@ -0,0 +1,75 @@
...
+                SemanticConventions.OPENINFERENCE_SPAN_KIND,
+                SemanticConventions.OpenInferenceSpanKind.TOOL.getValue()
+        );
+
+        span.setAttribute(SemanticConventions.TOOL_CALL_ID, event.request().id());
+        span.setAttribute(SemanticConventions.TOOL_NAME, event.request().name());
+        span.setAttribute(SemanticConventions.TOOL_PARAMETERS, event.request().arguments());
+
+        span.setAttribute(SemanticConventions.OUTPUT_VALUE, event.resultText());
+        span.setStatus(StatusCode.OK);
```
nate-mar, line 75: `LangChain4jToolExecutedEventListener` is missing `INPUT_VALUE`, `INPUT_MIME_TYPE`, and `OUTPUT_MIME_TYPE`. Per OpenInference conventions, every span needs input/output values with MIME types.

There's some ambiguity in the codebase around what `TOOL_PARAMETERS` should contain (actual args vs schema), but the majority of Python instrumentors (autogen-agentchat, agno, google-adk) set both `TOOL_PARAMETERS` and `INPUT_VALUE` to the actual tool arguments, with `INPUT_MIME_TYPE = application/json`. CrewAI is an outlier that puts the schema in `TOOL_PARAMETERS` instead. For output, the consensus is `OUTPUT_MIME_TYPE = text/plain`.

Suggested fix:
```java
span.setAttribute(SemanticConventions.TOOL_PARAMETERS, event.request().arguments());
span.setAttribute(SemanticConventions.INPUT_VALUE, event.request().arguments());
span.setAttribute(SemanticConventions.INPUT_MIME_TYPE, "application/json");

span.setAttribute(SemanticConventions.OUTPUT_VALUE, event.resultText());
span.setAttribute(SemanticConventions.OUTPUT_MIME_TYPE, "text/plain");
```

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
```diff
@@ -0,0 +1,84 @@
...
+     * Sets input attributes on the span based on the event's messages.
+     *
+     * @param span  the span to set attributes on
+     * @param event the AI service started event
+     */
+    public void setInputAttributes(Span span, AiServiceStartedEvent event) {
+        span.setAttribute(
+                SemanticConventions.OPENINFERENCE_SPAN_KIND,
+                SemanticConventions.OpenInferenceSpanKind.CHAIN.getValue()
+        );
```
nate-mar: Consider using `AGENT` instead of `CHAIN` here.

Per the OpenInference spec, `CHAIN` is "a deterministic sequence of operations" while `AGENT` is "a reasoning step in an autonomous agent that may spawn child spans for tool calls, retrievals, or nested LLM calls."

Looking at the LangChain4j 1.11.0 source (`DefaultAiServices.java:323`), the `invoke()` method calls `context.toolService.executeInferenceAndToolsLoop(...)` — a loop where the LLM decides at each iteration whether to call another tool or return a final answer. That's non-deterministic branching decided at runtime by the model, which maps to AGENT.

For reference, other instrumentors use AGENT for equivalent constructs: crewai uses it for `Crew.kickoff()` / `Agent.kickoff()`, autogen-agentchat for `agent.on_messages_stream`, and openai-agents for `AgentSpanData`.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
```diff
@@ -0,0 +1,84 @@
...
+        List<ChatMessage> messages = new ArrayList<>();
+        if (event.systemMessage().isPresent()) {
+            messages.add(event.systemMessage().get());
+        }
+        messages.add(event.userMessage());
+        List<Map<String, Object>> inputList = ChatMessageAttributeUtils.convertMessages(messages);
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            span.setAttribute(SemanticConventions.INPUT_VALUE, objectMapper.writeValueAsString(inputList));
```
nate-mar: `setInputAttributes()` unconditionally sets `INPUT_VALUE` and `INPUT_MIME_TYPE` without checking `TraceConfig`. The existing `LangChain4jModelListener` and `SpringAIInstrumentor` both gate these behind config checks:

```java
// LangChain4jModelListener.java:96
if (!tracer.getConfig().isHideInputMessages()) {
    setInputMessageAttributes(span, request.messages());
    span.setAttribute(SemanticConventions.INPUT_VALUE, messagesJson);
    span.setAttribute(SemanticConventions.INPUT_MIME_TYPE, "application/json");
}
```

`TraceConfig` exists so users can suppress sensitive data (PII, medical info, etc.) from traces while keeping the structural telemetry (latency, token counts, tool calls). If a user sets `hideInputs = true`, this listener would still leak the input data — breaking the privacy guarantee.

The same applies to `ServiceCompletedListener` which sets `OUTPUT_VALUE` without checking `isHideOutputs()`.

Suggested fix:
```java
if (!tracer.getConfig().isHideInputs()) {
    span.setAttribute(SemanticConventions.INPUT_VALUE, objectMapper.writeValueAsString(inputList));
    span.setAttribute(SemanticConventions.INPUT_MIME_TYPE, SemanticConventions.MimeType.JSON.getValue());
}
```

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jServiceCompletedListener.java
```diff
@@ -0,0 +1,39 @@
...
+    public LangChain4jServiceCompletedListener(OITracer tracer, Map<String, SpanContext> activeSpans) {
+        this.tracer = tracer;
+        this.activeSpans = activeSpans;
+    }
+
+    @Override
+    public void onEvent(AiServiceCompletedEvent event) {
+        SpanContext spanContext = activeSpans.remove(event.invocationContext().invocationId().toString());
+        if (spanContext != null) {
+            Span span = spanContext.span();
```
nate-mar, line 28: Missing `OUTPUT_MIME_TYPE`. Per OpenInference conventions, when `OUTPUT_VALUE` is set, `OUTPUT_MIME_TYPE` must also be set so consumers know how to interpret the value.

```java
span.setAttribute(SemanticConventions.OUTPUT_VALUE, event.result().get().toString());
span.setAttribute(SemanticConventions.OUTPUT_MIME_TYPE, "text/plain");
```

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceErrorListener.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @param event the AI service error event
+     */
+    @Override
+    public void onEvent(AiServiceErrorEvent event) {
+        SpanContext spanContext = activeSpans.remove(event.invocationContext().invocationId().toString());
+        if (spanContext != null) {
+            Span span = spanContext.span();
+
+            try (Scope scope = spanContext.context().makeCurrent()) {
```
nate-mar, line 58: `span.setStatus(StatusCode.ERROR)` is missing the error message. The other error handlers in this codebase include it:

```java
// LangChain4jModelListener.onError():175
span.setStatus(StatusCode.ERROR, error.getMessage());
```

Without the message, the span shows an error status but gives no indication of what went wrong. Should be:
```java
span.setStatus(StatusCode.ERROR, event.error().getMessage());
```

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
```diff
@@ -0,0 +1,84 @@
...
+                SemanticConventions.OpenInferenceSpanKind.CHAIN.getValue()
+        );
+        List<ChatMessage> messages = new ArrayList<>();
+        if (event.systemMessage().isPresent()) {
+            messages.add(event.systemMessage().get());
+        }
+        messages.add(event.userMessage());
+        List<Map<String, Object>> inputList = ChatMessageAttributeUtils.convertMessages(messages);
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
```
nate-mar: `new ObjectMapper()` is created on every event. `ObjectMapper` is expensive to construct (reflection metadata, serializer caches, etc.) and thread-safe once built. The rest of the codebase uses static instances:

```java
// ChatMessageAttributeUtils.java
private static final ObjectMapper objectMapper = new ObjectMapper();
```

Should be a `private static final` field, or reuse the one already in `ChatMessageAttributeUtils`.

nate-mar:
There are no tests for the new listeners. I've added test infrastructure and initial LangChain4j WireMock tests on main in #2897 — please rebase/merge from main to pick those up, then add tests for the new AiService and tool listener functionality on top.

At minimum, tests should cover:
- AiService started → completed lifecycle (verify span attributes, span kind, input/output values)
- AiService started → error lifecycle (verify error status + recorded exception)
- Tool execution events (verify TOOL span kind, tool name/parameters, input/output with MIME types)
- TraceConfig masking (verify `hideInputs`/`hideOutputs` actually suppress attributes)
- Exhaustive attribute assertions using the remove-and-verify pattern (pop all attributes, assert map is empty at the end)

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
```diff
@@ -0,0 +1,310 @@
+package com.arize.instrumentation.langchain4j.utils;
+
+import com.arize.instrumentation.OITracer;
+import com.arize.instrumentation.langchain4j.LangChain4jModelListener;
+import com.arize.semconv.trace.SemanticConventions;
```
nate-mar, line 4: Logger references `LangChain4jModelListener.class` instead of `ChatMessageAttributeUtils.class` — likely a copy-paste artifact from the extraction. Log messages from this class will show up under the wrong logger name.

```java
private static final Logger logger = Logger.getLogger(ChatMessageAttributeUtils.class.getName());
```

satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jToolExecutedEventListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jServiceCompletedListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceErrorListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceStartedListener.java
satyadevai: I’ve implemented the changes as recommended.

satyadevai:
@nate-mar added suggested test cases. Please do one round of review

nate-mar:
@claude review

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jToolExecutedEventListener.java
```diff
@@ -0,0 +1,77 @@
...
+
+        span.setAttribute(SemanticConventions.INPUT_VALUE, event.request().arguments());
+        span.setAttribute(SemanticConventions.INPUT_MIME_TYPE, SemanticConventions.MimeType.JSON.getValue());
+
+        span.setAttribute(SemanticConventions.OUTPUT_VALUE, event.resultText());
+        span.setAttribute(SemanticConventions.OUTPUT_MIME_TYPE, SemanticConventions.MimeType.TEXT.getValue());
+
+        span.setAttribute(SemanticConventions.TOOL_CALL_ID, event.request().id());
+        span.setAttribute(SemanticConventions.TOOL_NAME, event.request().name());
+        span.setAttribute(SemanticConventions.TOOL_PARAMETERS, event.request().arguments());
```
nate-mar: The other listeners guard input/output attributes behind `tracer.getConfig().isHideInputMessages()` / `isHideOutputMessages()` (e.g. `LangChain4jAiServiceStartedListener.setInputAttributes` and `LangChain4jServiceCompletedListener.onEvent`), but this listener writes `INPUT_VALUE`, `OUTPUT_VALUE`, and `TOOL_PARAMETERS` unconditionally. This means tool call arguments and results will leak into traces even when the user has explicitly configured hiding. The `traceConfigInputOutputMessagesDisabled` test also doesn't assert on the tool span, which masks this gap. This should have the same config checks as the other listeners.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceErrorListener.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Handles the AI service error event by recording the exception and marking the span as errored.
+     *
+     * @param event the AI service error event
+     */
+    @Override
+    public void onEvent(AiServiceErrorEvent event) {
+        SpanContext spanContext =
+                activeSpans.remove(event.invocationContext().invocationId().toString());
```
nate-mar: This only removes the agent span (keyed by raw `invocationId`), but `LangChain4jServiceRequestIssuedListener` stores the child LLM span at `"chat_" + invocationId`. If an error fires after the LLM request is issued, the child span is never ended or removed from the map.

The root cause is that all 6 listeners share a single map and use a `"chat_"` string prefix to distinguish span types. Every listener that touches LLM child spans needs to know about this prefix, and the error handler doesn't. Consider splitting into two maps:

```java
// In LangChain4jInstrumentor:
private final Map<String, SpanContext> agentSpans = new ConcurrentHashMap<>();
private final Map<String, SpanContext> llmSpans = new ConcurrentHashMap<>();
```

Then every listener uses raw `invocationId` as the key into the appropriate map, and the error handler cleans up both:

```java
SpanContext agentCtx = agentSpans.remove(invocationId);
SpanContext llmCtx = llmSpans.remove(invocationId);
// end both spans
```

No prefix convention to maintain across 6 files, no orphaned entries.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        // Set token usage if available
+        if (response.tokenUsage() != null) {
+            span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_PROMPT, (long)
+                    response.tokenUsage().inputTokenCount());
+            span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_COMPLETION, (long)
+                    response.tokenUsage().outputTokenCount());
+            span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_TOTAL, (long)
+                    response.tokenUsage().totalTokenCount());
```
nate-mar: `tokenUsage()` is null-checked, but `inputTokenCount()`, `outputTokenCount()`, and `totalTokenCount()` each return boxed `Integer` and can be null independently in the LangChain4j API. A provider that reports completion tokens but not prompt tokens would leave `inputTokenCount()` null, and the `(long)` widening on line 294 triggers auto-unboxing on the null `Integer`, which NPEs.

The existing `handleChatRequest` in this same file already does per-field null checks for similar optional values:

```java
// ChatMessageAttributeUtils.java:232
if (request.temperature() != null) {
    invocationParams.put("temperature", request.temperature());
}
```

Same pattern should work here:

```java
if (response.tokenUsage() != null) {
    if (response.tokenUsage().inputTokenCount() != null) {
        span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_PROMPT, (long)
                response.tokenUsage().inputTokenCount());
    }
    if (response.tokenUsage().outputTokenCount() != null) {
        span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_COMPLETION, (long)
                response.tokenUsage().outputTokenCount());
    }
    if (response.tokenUsage().totalTokenCount() != null) {
        span.setAttribute(SemanticConventions.LLM_TOKEN_COUNT_TOTAL, (long)
                response.tokenUsage().totalTokenCount());
    }
}
```

java/instrumentation/openinference-instrumentation-langchain4j/src/test/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceListenerTest.java
```diff
@@ -0,0 +1,643 @@
...
+        try {
+            MathAssistant assistant = createMathAssistant(testInfo, instrumentor);
+            String answer = assistant.chat("What is 45 plus 79?");
+            assertThat(answer).containsIgnoringCase("45 plus 79 is 124.");
+        } finally {
+            instrumentor.uninstrument();
+        }
+
+        List<SpanData> spans = spanExporter.getFinishedSpanItems();
+        assertThat(spans).hasSize(4);
```
nate-mar, line 121: The tests thoroughly verify attributes on each span, but there are no assertions on `getParentSpanId()`. The core value of these listeners is producing a nested trace tree (agent -> LLM, agent -> tool); without parent checks, all the tests would still pass if every span were emitted as a flat sibling with no nesting.

```java
SpanData agentSpan = spans.get(3);
assertThat(llmSpan.getParentSpanId()).isEqualTo(agentSpan.getSpanId());
assertThat(toolSpan.getParentSpanId()).isEqualTo(agentSpan.getSpanId());
assertThat(finalLlmSpan.getParentSpanId()).isEqualTo(agentSpan.getSpanId());
```

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceErrorListener.java
nate-mar: Just FYI, `LangChain4jServiceRequestIssuedListener` stores the child LLM span at `"chat_" + invocationId` (line 47-49):

```java
// LangChain4jServiceRequestIssuedListener.java:47-49
activeSpans.put(
        "chat_" + event.invocationContext().invocationId().toString(),
        new SpanContext(span, Context.current().with(span)));
```

But the error handler only removes the raw `invocationId` key (line 35-36):

```java
// LangChain4jAiServiceErrorListener.java:35-36
SpanContext spanContext =
        activeSpans.remove(event.invocationContext().invocationId().toString());
```

There is no `activeSpans.remove("chat_" + ...)` here. So if `AiServiceRequestIssuedEvent` fires (creating the child LLM span), and then an error fires instead of a response, the `"chat_"` entry is never removed and that span is never ended.

satyadevai:
@claude review

java/instrumentation/openinference-instrumentation-langchain4j/src/test/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceListenerTest.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jAiServiceErrorListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/LangChain4jToolExecutedEventListener.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        // Set input attributes
+        if (!tracer.getConfig().isHideInputMessages()) {
+            try {
+                // Set input messages with proper structure
+                setInputMessageAttributes(span, request.messages());
+
+                // Also set input.value and input.mime_type for compatibility
+                List<Map<String, Object>> messagesList = convertMessages(request.messages());
```
nate-mar, line 287: `convertMessages` uses `Map.of()` which throws `NullPointerException` on null values. If a provider returns unexpected nulls in a tool call, this fails the entire message serialization and the span loses all input data. Would it be worth skipping tool calls with null fields so one bad entry doesn't take out the rest? The outer catch could also be widened from `JsonProcessingException` to `Exception` as a safety net.

Separately, `userMessage.singleText()` at L179 throws on multi-modal messages. That predates this PR, but might be worth a follow-up to guard with `hasSingleText()`.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
```diff
@@ -0,0 +1,309 @@
...
+        for (int i = 0; i < messages.size(); i++) {
+            ChatMessage message = messages.get(i);
+            String prefix = String.format("%s.%d.", SemanticConventions.LLM_INPUT_MESSAGES, i);
+
+            // Set role
+            String role = mapMessageRole(message.type());
+            span.setAttribute(AttributeKey.stringKey(prefix + SemanticConventions.MESSAGE_ROLE), role);
+
+            // Set content
+            span.setAttribute(AttributeKey.stringKey(prefix + SemanticConventions.MESSAGE_CONTENT), message.toString());
```
nate-mar: `message.toString()` here produces Java object representations like `UserMessage { name = null, contents = [TextContent { ... }] }` in the `llm.input_messages.N.message.content` span attributes. Per the OpenInference spec, `message.content` is a plain text string (e.g., `"What's the weather today?"`), as shown in the [semantic conventions](https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md), [traces spec](https://github.com/Arize-ai/openinference/blob/main/spec/traces.md), and [tool calling spec](https://github.com/Arize-ai/openinference/blob/main/spec/tool_calling.md). This was visible in a live run against Phoenix.

This predates the PR (it was in the original `LangChain4jModelListener`), but since the code was moved into this shared utility, would it be worth a follow-up to cast by message type and call the appropriate text accessor? The Spring AI instrumentor already does this correctly (e.g., `message.getText()`).

nate-mar (approved):

satyadevai:
@claude review

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
satyadevai: I’ve implemented the changes as recommended.

java/instrumentation/openinference-instrumentation-langchain4j/src/main/java/com/arize/instrumentation/langchain4j/utils/ChatMessageAttributeUtils.java
satyadevai: I’ve implemented the changes as recommended.

PR #2895, opened 2026-03-17
feat(crewai): instrument event-listener

caroger:
### Summary

This PR adds event-listener-based tracing for CrewAI by implementing an OpenInference listener on top of CrewAI's official event system.

The new mode is intended for CrewAI executions that are surfaced through the event bus, especially AMP / low-code / event-bus-native usage, where wrapper-based monkey-patching may not observe the underlying execution path directly.

Wrapper-based instrumentation remains the default and recommended path for standard Python CrewAI applications.

### Problem

The existing CrewAI instrumentor primarily relies on wrapper-based instrumentation. That works well for direct Python method calls, but it is not a good fit for event-bus-driven CrewAI execution.

CrewAI already exposes a first-class event-listener API. To support that model cleanly, OpenInference needs to translate official CrewAI events into spans while preserving trace hierarchy across:

- crews
- flows
- agents
- standalone `Agent.kickoff()`
- tools
- optional LLM calls

The main implementation challenges are:

- CrewAI dispatches event handlers concurrently
- start/end handlers can race
- some tool execution paths cross worker-thread boundaries
- some child events may arrive without a usable `parent_event_id`

### Goals

- Add official CrewAI event-listener support to `openinference-instrumentation-crewai`
- Preserve correct trace hierarchy for crew / task / agent / tool / flow / LLM events
- Support standalone `Agent.kickoff()` via LiteAgent events
- Tolerate out-of-order event delivery
- Tolerate worker-thread execution paths that would otherwise lose ancestry
- Keep wrapper mode unchanged as the default integration path

### Non-Goals

- Replace wrapper-based instrumentation for normal Python CrewAI apps
- Instrument every CrewAI event type exposed by the framework
- Perfectly reconstruct ancestry when upstream emits insufficient identity data
- Add streaming chunk span support in this PR

### Public API

This PR adds event-listener mode behind the existing instrumentor:

```python
CrewAIInstrumentor().instrument(
    tracer_provider=tracer_provider,
    use_event_listener=True,
)
```

Optional configuration:

```python
CrewAIInstrumentor().instrument(
    tracer_provider=tracer_provider,
    use_event_listener=True,
    create_llm_spans=False,
)
```

`create_llm_spans=False` disables synthetic LLM spans from `LLMCall*` events when the underlying LLM client is already instrumented separately, or when tests only care about crew / agent / tool structure.

#### 1. `OpenInferenceEventListener`

A new `OpenInferenceEventListener` implements CrewAI's `BaseEventListener` and registers handlers for official CrewAI events:

- `CrewKickoffStarted/Completed/Failed`
- `TaskStarted/Completed/Failed`
- `AgentExecutionStarted/Completed/Error`
- `LiteAgentExecutionStarted/Completed/Error`
- `ToolUsageStarted/Finished/Error`
- `LLMCallStarted/Completed/Failed`
- `FlowStarted/Finished`
- `MethodExecutionStarted/Finished/Failed`

The listener itself stays thin: it converts events into span start/end/scope specs and delegates span lifecycle handling to a shared assembler.

#### 2. `CrewAIEventAssembler`

A new `CrewAIEventAssembler` centralizes span assembly logic:

- start span
- end span
- open transparent scope
- close transparent scope
- resolve parents
- recover from out-of-order start/end delivery
- keep short-lived context caches for late child events

This separates event normalization from span lifecycle management and avoids duplicating race-handling logic in individual event handlers.

#### 3. Hierarchy model

Span kinds are mapped as follows:

- crew kickoff -> `CHAIN`
- flow kickoff / flow methods -> `CHAIN`
- agent execution / LiteAgent kickoff -> `AGENT`
- tool usage -> `TOOL`
- LLM calls -> `LLM`

Task events are treated as transparent scopes rather than emitted as spans. Their purpose is to preserve ancestry between crew and agent/tool events without adding extra task spans to the trace.

#### 4. Out-of-order event handling

CrewAI may dispatch handlers concurrently, so completion handlers can run before start handlers.

To handle this safely, the assembler supports:

- deferred ends: store end-state when the matching start span does not exist yet
- pending starts: queue child starts until the parent context exists
- finished-context cache: allow late-arriving children to attach to a recently finished parent context

This allows the listener to preserve hierarchy even under real event-bus concurrency.

#### 5. Worker-thread context propagation

Some CrewAI execution paths, especially native tool execution, cross thread boundaries.

To preserve `contextvars` across those boundaries, event-listener mode patches:

- `Agent._execute_without_timeout`
- `CrewAgentExecutor._execute_single_native_tool_call`

using the same context-preserving descriptor pattern already used elsewhere in the package.

This is needed so worker-thread event emission retains the correct CrewAI event context.

#### 6. Rootless child-event fallback

Some tool/LLM events may still arrive without a usable `parent_event_id`.

For those cases, the assembler performs a best-effort fallback lookup using active execution hints:

- prefer active agent context by `task_id`
- then active task scope by `task_id`
- only then fall back to active agent identity hints (`agent_id`, `agent_key`, `agent_role`)

These hints are scoped to active spans only and are removed when the owning agent span ends, which avoids stale cross-run attachment when the same agent is reused across multiple crews.

#### 7. Standalone `Agent.kickoff()` support

CrewAI standalone agent execution emits LiteAgent events instead of `AgentExecution*` events.

This PR adds explicit support for:

- `LiteAgentExecutionStartedEvent`
- `LiteAgentExecutionCompletedEvent`
- `LiteAgentExecutionErrorEvent`

so `use_event_listener=True` traces standalone `Agent.kickoff()` correctly.

#### 8. Input / output serialization

The listener serializes inputs according to payload shape:

- `str` -> `text/plain`
- `dict` / `list` / `tuple` -> `application/json`

This improves span readability in Phoenix and avoids mislabeling plain-text prompts as JSON.

### Docs / Example

This PR adds:

- README guidance for event-listener mode
- an `examples/event_listener_crew.py` example
- clarification that event-listener mode is for event-bus-native / AMP / low-code CrewAI usage, not a blanket replacement for wrapper-based instrumentation

### Testing

Test coverage includes:

- crew / agent / tool / flow / LLM span creation
- task transparent-scope hierarchy
- standalone LiteAgent tracing
- `create_llm_spans=False`

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
@@ -369,8 +369,8 @@ class CrewAIEventAssembler:
                 logger.debug("Failed to end stale open span for %s", oldest_event_id, exc_info=True)
 
         while self._transparent_contexts:
-            oldest_event_id, entry = next(iter(self._transparent_contexts.items()))
-            if entry.opened_at_monotonic > cutoff:
+            oldest_event_id, scope_entry = next(iter(self._transparent_contexts.items()))
+            if scope_entry.opened_at_monotonic > cutoff:
                 break
             self._transparent_contexts.popitem(last=False)
             if (
@@ -381,11 +381,11 @@ class CrewAIEventAssembler:
                     task_key,
                     oldest_event_id,
                 )
-            self._remember_finished_context_locked(oldest_event_id, entry.context)
+            self._remember_finished_context_locked(oldest_event_id, scope_entry.context)
             logger.warning(
                 "Evicting stale transparent scope for %s after %.1f seconds without a close event",
                 oldest_event_id,
-                self._now_monotonic() - entry.opened_at_monotonic,
+                self._now_monotonic() - scope_entry.opened_at_monotonic,
             )
 
     @staticmethod
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import threading
 import weakref
+from collections import OrderedDict
 from collections.abc import Mapping, Sequence
 from typing import Any, Optional, cast
 
@@ -60,6 +61,7 @@ from openinference.instrumentation import (
     safe_json_dumps,
 )
 from openinference.instrumentation.crewai._event_assembler import (
+    _MAX_BUFFERED_EVENT_ENTRIES,
     CrewAIEventAssembler,
     _SpanEndSpec,
     _SpanStartSpec,
@@ -527,7 +529,7 @@ class OpenInferenceEventListener(BaseEventListener):
         )
         self._assembler = CrewAIEventAssembler(tracer=self._tracer)
         self._create_llm_spans = create_llm_spans
-        self._llm_usage_by_call_id: dict[str, dict[str, Any]] = {}
+        self._llm_usage_by_call_id: OrderedDict[str, dict[str, Any]] = OrderedDict()
         self._llm_usage_lock = threading.RLock()
         if self._create_llm_spans:
             self._patch_llm_token_usage_tracking()
@@ -603,11 +605,25 @@ class OpenInferenceEventListener(BaseEventListener):
 
                 def patched(instance: Any, usage_data: dict[str, Any]) -> None:
                     original(instance, usage_data)
-                    call_id = base_llm_module.get_current_call_id()
+                    try:
+                        call_id = base_llm_module.get_current_call_id()
+                    except Exception:
+                        logger.debug(
+                            "Failed to resolve CrewAI LLM call id while tracking token usage",
+                            exc_info=True,
+                        )
+                        return
                     with cls._llm_patch_lock:
                         listeners = list(cls._llm_patch_listeners)
                     for listener in listeners:
-                        listener._record_llm_token_usage(call_id, usage_data)
+                        try:
+                            listener._record_llm_token_usage(call_id, usage_data)
+                        except Exception:
+                            logger.debug(
+                                "Failed to record CrewAI LLM token usage for %s",
+                                call_id,
+                                exc_info=True,
+                            )
 
                 setattr(base_llm_module.BaseLLM, "_track_token_usage_internal", patched)
 
@@ -627,17 +643,19 @@ class OpenInferenceEventListener(BaseEventListener):
             self._llm_usage_by_call_id.clear()
 
     def _record_llm_token_usage(self, call_id: str, usage_data: Mapping[str, Any]) -> None:
+        if not call_id:
+            return
         normalized_usage = _normalize_token_usage(usage_data)
         with self._llm_usage_lock:
-            aggregate = self._llm_usage_by_call_id.setdefault(
-                call_id,
-                {
+            aggregate = self._llm_usage_by_call_id.pop(call_id, None)
+            if aggregate is None:
+                aggregate = {
                     "prompt": 0,
                     "completion": 0,
                     "total": 0,
                     "prompt_details": {"cache_read": 0},
-                },
-            )
+                }
+            self._llm_usage_by_call_id[call_id] = aggregate
             aggregate["prompt"] += normalized_usage["prompt"]
             aggregate["completion"] += normalized_usage["completion"]
             aggregate["total"] += normalized_usage["total"]
@@ -645,6 +663,14 @@ class OpenInferenceEventListener(BaseEventListener):
             if isinstance(prompt_details, Mapping):
                 aggregate["prompt_details"]["cache_read"] += int(
                     prompt_details.get("cache_read", 0)
+                )
+            while len(self._llm_usage_by_call_id) > _MAX_BUFFERED_EVENT_ENTRIES:
+                evicted_call_id, _ = self._llm_usage_by_call_id.popitem(last=False)
+                logger.warning(
+                    "Evicting oldest LLM token usage entry for %s "
+                    "after reaching %d buffered entries",
+                    evicted_call_id,
+                    _MAX_BUFFERED_EVENT_ENTRIES,
                 )
 
     def _consume_llm_token_usage_attributes(self, call_id: Optional[str]) -> dict[str, Any]:
--- a/python/instrumentation/openinference-instrumentation-crewai/tests/test_event_assembler.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/tests/test_event_assembler.py
@@ -94,6 +94,35 @@ def test_pending_end_preserves_completion_attributes(
     assert not attributes
 
 
+def test_error_end_sets_error_status_without_output_attributes(
+    assembler: CrewAIEventAssembler,
+    in_memory_span_exporter: InMemorySpanExporter,
+) -> None:
+    start_event = _event("chain-start")
+    end_event = _end_event("chain-start")
+
+    assembler.start_span(
+        start_event,
+        _SpanStartSpec(name="crew.kickoff", span_kind=OpenInferenceSpanKindValues.CHAIN),
+    )
+    assembler.end_span(
+        end_event,
+        _SpanEndSpec(output="should not be recorded", error="crew boom"),
+    )
+
+    spans = _finished_spans(in_memory_span_exporter)
+    assert len(spans) == 1
+
+    span = spans[0]
+    attributes: dict[str, Any] = dict(span.attributes or {})
+    assert span.status.status_code == StatusCode.ERROR
+    assert span.status.description == "crew boom"
+    assert attributes.pop(OPENINFERENCE_SPAN_KIND) == OpenInferenceSpanKindValues.CHAIN.value
+    assert attributes.pop(OUTPUT_MIME_TYPE, None) is None
+    assert attributes.pop(OUTPUT_VALUE, None) is None
+    assert not attributes
+
+
 def test_pending_child_start_attaches_to_late_parent(
     assembler: CrewAIEventAssembler,
     in_memory_span_exporter: InMemorySpanExporter,
--- a/python/instrumentation/openinference-instrumentation-crewai/tests/test_event_listener.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/tests/test_event_listener.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import json
 import uuid
-from collections.abc import Generator
+from collections.abc import Generator, Mapping
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
+from types import SimpleNamespace
 from typing import Any
 
 import pytest
@@ -11,6 +13,7 @@ from crewai.events.event_bus import crewai_event_bus
 from opentelemetry import trace as trace_api
 from opentelemetry.sdk.trace import ReadableSpan
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+from opentelemetry.trace import StatusCode
 
 from openinference.instrumentation import (
     REDACTED_VALUE,
@@ -19,6 +22,7 @@ from openinference.instrumentation import (
     using_attributes,
 )
 from openinference.instrumentation.crewai import CrewAIInstrumentor
+from openinference.instrumentation.crewai import _event_listener as event_listener_module
 from openinference.instrumentation.crewai._event_listener import OpenInferenceEventListener
 from openinference.semconv.trace import OpenInferenceSpanKindValues, SpanAttributes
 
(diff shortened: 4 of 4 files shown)


ryoung562:
i dog fooded using this script.   
[test_instrumentor.py](https://github.com/user-attachments/files/26045153/test_instrumentor.py)

attached are csv dump of the spans from arize.
[tracing_export (21).csv](https://github.com/user-attachments/files/26045217/tracing_export.21.csv)


i noticed a few things off:

- LLM span - input / output attributes not set 
- LLM span - token attributes only show on the last LLM span in multiple LLM spans under parent agent
- agent graph is missing

image
image
image


loom video: https://www.loom.com/share/5ced5387b5b44eeab521dbf7c7ae00df

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
```diff
@@ -0,0 +1,513 @@
...
+from openinference.instrumentation.crewai._wrappers import _flatten
+from openinference.semconv.trace import (
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+)
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+_FINISHED_CONTEXT_CACHE_SIZE = 1024
```
nate-mar: I wonder if the single global cap of 1024 on the bounded caches could become a problem for long-running processes. Here's the scenario I'm thinking about:

Multiple concurrent crews (or a flow kicking off crews sequentially) share the same assembler instance and the same 1024 slots across `_finished_contexts`, `_deferred_ends`, `_pending_starts`, etc. Once they fill up, earlier crews' contexts get evicted via `popitem(last=False)`. A late-arriving child event, say a tool callback returning from a thread pool, silently loses its parent or is never created. No log, no error, just missing trace data.

Separately, `_spans` and `_transparent_contexts` have no cap at all, so if end events are dropped they grow without bound.

A couple of questions:

- Is there a reason 1024 was chosen over making this configurable? Exposing the cap as a parameter alongside `create_llm_spans` would let operators tune for their workload without changing the default behavior.
- For the unbounded `_spans` dict, would it make sense to evict with `span.end(StatusCode.ERROR)` so the consumer at least sees something rather than a silently leaked span?
- `_FINISHED_CONTEXT_CACHE_SIZE` is used as the cap for all six bounded caches, not just `_finished_contexts`. Something like `_MAX_CACHE_ENTRIES` would better reflect its actual role.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
```diff
@@ -0,0 +1,513 @@
...
+                )
+                return
+            if entry.agent_hint_keys is not None:
+                self._forget_agent_context_locked(event_id, entry.agent_hint_keys)
+            self._remember_finished_context_locked(event_id, entry.context)
+
+        self._finalize_span(entry, spec, end_time_ns)
+
+    def open_scope(self, event: Any) -> None:
+        def open_scope(parent_context: Context) -> None:
```
nate-mar, line 200: I'm curious about the decision to make tasks transparent scopes instead of real spans. The transparent scope machinery adds a fair amount of complexity: `open_scope`, `close_scope`, `_transparent_contexts`, `_closed_transparent_scopes`, `_task_contexts_by_task_id`, `_task_scope_keys_by_event_id`, plus the out-of-order handling when close arrives before open.

Was there a specific reason not to just make tasks regular spans via `start_span`/`end_span`? The extra nesting level (crew → task → agent) would be one more span in the trace, but it would eliminate a whole category of state and edge cases from the assembler. The task metadata (description, expected output) is already on the agent span, so the consumer wouldn't lose information either way.

Not necessarily saying it's the wrong call, just wondering if the trace-output cleanliness is worth the implementation cost here.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
```diff
@@ -0,0 +1,513 @@
...
+        self._tracer = tracer
+        self._spans: dict[str, _SpanEntry] = {}
+        self._transparent_contexts: dict[str, Context] = {}
+        self._finished_contexts: "OrderedDict[str, Context]" = OrderedDict()
+        self._task_contexts_by_task_id: "OrderedDict[str, _ContextHint]" = OrderedDict()
+        self._task_scope_keys_by_event_id: dict[str, str] = {}
+        self._agent_contexts_by_task_id: "OrderedDict[str, _ContextHint]" = OrderedDict()
+        self._agent_contexts_by_agent_id: "OrderedDict[str, _ContextHint]" = OrderedDict()
+        self._agent_contexts_by_agent_key: "OrderedDict[str, _ContextHint]" = OrderedDict()
+        self._agent_contexts_by_agent_role: "OrderedDict[str, _ContextHint]" = OrderedDict()
```
nate-mar, line 97: Small naming thought: `_deferred_ends` took me a second to parse since the end event already arrived, it's not really "deferred." The assembler is stashing it until the matching start shows up. `_pending_ends` would mirror `_pending_starts` and make the symmetry clearer: both are waiting for their counterpart. Thoughts?

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
caroger: Good call. One clarification: the 1024 limit is per cache, not one shared pool across all of them. That said, the constant name is misleading, and the live-state maps (_spans, _transparent_contexts) are still unbounded if end events never arrive. 

I will rename the constant, add eviction logging for the bounded orphan caches, and add age-based stale-entry cleanup for open spans/scopes rather than hard count-based eviction so we don’t terminate valid long-running crews.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
caroger: The main reason tasks are transparent scopes is trace-shape parity with the existing wrapper path. Wrapper mode today emits `crew -> agent -> tool/llm`, not `crew -> task -> agent`, and I wanted `use_event_listener=True` to preserve that contract. 

The design choice is to make the span hierarchy look "familiar" in these two paths (monkey patched wrapper vs. event_listener).

open to revisit this approach if users tell us differently later

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
caroger: Agreed. `_pending_ends` is clearer and more symmetric with `_pending_starts`. will rename

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
caroger: renamed the shared limit to _MAX_BUFFERED_EVENT_ENTRIES to reflect that it applies across the buffered caches, added warning logs when buffered entries are evicted, and added age-based cleanup for stale live state so _spans and _transparent_contexts don’t grow unbounded if completion events are dropped. 

I kept the limits constants internal in this PR rather than expanding the public event-listener API, but the failure mode is no longer silent or unbounded.

Open to make it configurable based on user feedbacks

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
nate-mar: Gotcha, thanks for the context. I'm okay with the current approach then if that's what we already do with the wrapper and it works for users.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
```diff
@@ -0,0 +1,837 @@
...
+
+        from openinference.instrumentation.crewai.version import __version__
+
+        self._tracer = OITracer(
+            trace_api.get_tracer(__name__, __version__, tracer_provider),
+            config=config,
+        )
+        self._assembler = CrewAIEventAssembler(tracer=self._tracer)
+        self._create_llm_spans = create_llm_spans
+        self._llm_usage_by_call_id: dict[str, dict[str, Any]] = {}
```
nate-mar: I noticed `_llm_usage_by_call_id` is a plain `dict`, while the assembler's caches (`_finished_contexts`, `_deferred_ends`, `_agent_contexts_by_*`, etc.) all use bounded `OrderedDict` with eviction at 1024. Was this intentional? The practical impact is low since each entry is just a few ints, but if a completion event is ever lost (like because of a timeout or something) the entry stays forever. Wondering if there's a reason it doesn't need the same treatment.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
```diff
@@ -0,0 +1,837 @@
...
+                    return
+                cls._llm_patch_original = original
+
+                def patched(instance: Any, usage_data: dict[str, Any]) -> None:
+                    original(instance, usage_data)
+                    call_id = base_llm_module.get_current_call_id()
+                    with cls._llm_patch_lock:
+                        listeners = list(cls._llm_patch_listeners)
+                    for listener in listeners:
+                        listener._record_llm_token_usage(call_id, usage_data)
```
nate-mar: The listener notification loop here has no try/except, which is consistent with the existing wrappers in `_wrappers.py`. I wonder if it's worth a follow-up to add defensive try/except across all the instrumentation patches (both the wrappers and this one). Right now if any instrumentation code throws, it propagates into the user's call. Ideally instrumentation errors shouldn't affect the main application code. This spot is a bit more exposed than the others since it runs inside CrewAI's internal LLM path rather than at a user-facing method boundary, so the resulting traceback would be especially confusing to debug.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
```diff
@@ -0,0 +1,594 @@
...
+class _PendingEnd:
+    output: Any
+    error: Optional[str]
+    attributes: dict[str, Any]
+    end_time_ns: Optional[int]
+
+
+@dataclass
+class _TransparentScopeEntry:
+    context: Context
```
nate-mar, line 80: If a completion or close event is never delivered (e.g. agent timeout), wouldn't entries in `_spans`, `_transparent_contexts`, and `_task_scope_keys_by_event_id` accumulate indefinitely? I know it's an edge case for sure, but `_spans` is probably the most worth watching since each entry holds a live `Span` object, a `Context`, and a dict of attributes. Might be worth the same bounded treatment as a defensive measure for long-running processes.

python/instrumentation/openinference-instrumentation-crewai/tests/test_event_assembler.py
```diff
@@ -0,0 +1,285 @@
...
+) -> None:
+    current_time = [0.0]
+    monkeypatch.setattr(assembler_module, "_MAX_OPEN_ENTRY_AGE_SECONDS", 1.0)
+    monkeypatch.setattr(assembler, "_now_monotonic", lambda: current_time[0])
+
+    with caplog.at_level("WARNING"):
+        assembler.start_span(
+            _event("stale-span"),
+            _SpanStartSpec(name="stale-span", span_kind=OpenInferenceSpanKindValues.CHAIN),
+        )
```
nate-mar, line 279: The assembler tests cover some great edge cases (deferred ends, pending starts, orphan eviction), but I don't see any test that exercises the error path in `_finalize_span`. Every existing test ends spans with `_SpanEndSpec(output=...)`, so the `if spec.error:` branch never runs. Would be nice to have at least one test that ends a span with `_SpanEndSpec(error="something went wrong")` and asserts the span gets `StatusCode.ERROR` with the error description. Same goes for the event listener level, where none of the failure handlers (`_on_crew_failed`, `_on_agent_error`, `_on_tool_error`, etc.) have coverage.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
caroger: Good catch. I bounded _llm_usage_by_call_id, evict oldest entries with a warning, and ignore empty call IDs so a missing completion no longer leaves unbounded token-usage state behind.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_listener.py
caroger: Agreed. The patched token tracker now treats instrumentation failures as non-fatal: each listener callback is wrapped in try/except with debug logging so CrewAI’s LLM path is unaffected.

python/instrumentation/openinference-instrumentation-crewai/tests/test_event_assembler.py
caroger: Added explicit coverage for assembler error finalization plus listener failure handlers for crew, agent, tool, and LLM events.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_event_assembler.py
caroger: addressed.

_spans and _transparent_contexts now get age-based stale-entry eviction via _evict_stale_live_state_locked(...), and stale transparent-scope eviction also removes the corresponding _task_scope_keys_by_event_id entry. I avoided count-bounding live spans/scopes because that could truncate a valid long-running crew; buffered orphan state is count-bounded, while live state is age-bounded and logged on eviction.

PR #2908, opened 2026-03-23
fix(llama-index): Suppress unhandled event warnings for `SparseEmbeddingEndEvent` and `Workflow*OutputEvent`

enrico-stauss:
# Fix LlamaIndex and LlamaIndex Workflows Unhandled Event Logs
This PR registers singledispatch handlers for four previously unhandled event types that were producing spurious WARNING logs:
- WARNING - Unhandled event of type SparseEmbeddingEndEvent
- WARNING - Unhandled event of type WorkflowStepOutputEvent
- WARNING - Unhandled event of type WorkflowRunOutputEvent
- WARNING - Unhandled event of type SpanCancelledEvent  (pre-emptive)

The fix will take effect once the companion fixes in the llama_index and workflows-py repositories are also merged.

## llama-index-core -> `SparseEmbeddingEndEvent`
Find the [PR here](https://github.com/run-llama/llama_index/pull/21119).
Sparse embedding vectors are Dict[int, float] (token-id → weight), incompatible with the EMBEDDING_VECTOR attribute (List[float]).
The handler records the text chunks and skips the vectors, keeping the span populated without violating the semantic convention.
Note: SparseEmbeddingStartEvent no longer needs a handler here because a companion fix in llama-index-core makes it inherit from EmbeddingStartEvent, so singledispatch routes it via MRO.

## llama-index-workflows (optional dependency) -> `WorkflowStepOutputEvent`/`WorkflowRunOutputEvent`/`SpanCancelledEvent`
Find the [PR here](https://github.com/run-llama/workflows-py/pull/445).
These events are emitted by the `workflows` package (llama-index-workflows). Handlers are registered inside a try/except ImportError block so the instrumentor continues to work when that package is not installed.
WorkflowStepOutputEvent and WorkflowRunOutputEvent carry an `output` string that is recorded as OUTPUT_VALUE on the span. SpanCancelledEvent is silenced with a no-op.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
+++ b/python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
@@ -98,6 +98,7 @@ from llama_index.core.instrumentation.events.embedding import (
     EmbeddingEndEvent,
     EmbeddingStartEvent,
     SparseEmbeddingEndEvent,
+    SparseEmbeddingStartEvent,
 )
 from llama_index.core.instrumentation.events.llm import (
     LLMChatEndEvent,
@@ -523,6 +524,11 @@ class _Span(BaseSpan):
             self._span_kind = EMBEDDING
 
     @_process_event.register
+    def _(self, event: SparseEmbeddingStartEvent) -> None:
+        if not self._span_kind:
+            self._span_kind = EMBEDDING
+
+    @_process_event.register
     def _(self, event: EmbeddingEndEvent) -> None:
         i = self._list_attr_len[EMBEDDING_EMBEDDINGS]
         for text, vector in zip(event.chunks, event.embeddings):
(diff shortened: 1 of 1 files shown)


enrico-stauss:
I did the follow-up after realizing that `self[OUTPUT_VALUE] = event.output` serializes pydantic models to just the first characters produced by the `__str__` method which is not very informative. 

I don't really know how LlamaIndex coordinates this type of change with you guys but I would assume that this requires acceptance from maintainers of the LlamaIndex and LlamaIndex Workflows repos?

Anyway I tested this in my local setup by referencing all 3 PRs. The warnings are gone and the output looks good to me (the only change is in the SparseEmbeddingOutputEvent attributes). I hope this gets merged since the sheer amount of logged warnings was honestly quite annoying.

Cheers
Enrico

adrianlyjak:
Chiming in from the llamaindex side. Having this change spread across 3 repos seems pretty hard to get right and make properly compatible.

Perhaps this default should just be changed to a debug, or have something that more reasonably dumps the event contents as (maybe truncated) json

```py
    @singledispatchmethod
    def _process_event(self, event: BaseEvent) -> None:
        logger.warning(f"Unhandled event of type {event.__class__.__qualname__}")
```

ialisaleh:
Hi @enrico-stauss - the following PR is already in progress / under review to handle sparse embeddings & new workflow events, so closing this PR. Thanks!

PR: https://github.com/Arize-ai/openinference/pull/2928

PR #2916, opened 2026-03-25
fix(crewai): serialize agent task input as structured JSON

caroger:
resolves: https://github.com/Arize-ai/arize/issues/65134

before:
Screenshot 2026-03-24 at 11 25 31 PM

after:
Screenshot 2026-03-24 at 11 25 02 PM

## Summary
- serialize the bound CrewAI `agent` task input as structured JSON before span input serialization
- avoid cyclic `agent.crew` and other runtime-only fields by emitting a curated agent payload
- add regression coverage for direct `_get_input_value` serialization and wrapper-based AGENT span inputs

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -72,17 +72,24 @@ class SafeJSONEncoder(json.JSONEncoder):
             return repr(o)
 
 
+# Prefer CrewAI's own Pydantic serialization so span input tracks the agent schema
+# that callers see, but trim fields that are cyclic, runtime-only, or not JSON-safe.
 _AGENT_MODEL_DUMP_EXCLUDE: Dict[str, Any] = {
+    # Cyclic graph back to the crew/task graph.
     "crew": True,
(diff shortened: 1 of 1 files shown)


python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
```diff
@@ -72,6 +72,50 @@ def default(self, o: Any) -> Any:
             return repr(o)
 
 
+def _get_tool_name(tool: Any) -> str:
+    tool_name = getattr(tool, "name", None)
+    if tool_name:
+        return str(tool_name)
+    return str(tool)
+
+
```
nate-mar: Could we use `agent.model_dump(exclude={"crew", "llm", "agent_executor"})` instead of listing fields manually? The allowlist approach silently drops any new fields CrewAI adds to Agent in the future.

nate-mar (commented):
Inline comment on serialization approach.

python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
caroger: See LN126

PR #2925, opened 2026-03-27
fix: update how hideinputs/outputs is handled

duncankmckinnon:
## Summary

Fixes the hide inputs/outputs inconsistency between the annotation and LangChain4j instrumentations by checking `TraceConfig` at the call site in `TraceAdvice`, matching the pattern used in `LangChain4jModelListener`.

Previously, the annotation instrumentation relied solely on internal guards in `TracedSpan.setInput()`/`setOutput()`, which meant `@SpanMapping` attributes were not suppressed when `hideInputs`/`hideOutputs` was configured. Now both `setInput`/`setOutput` and `applyInputMappings`/`applyOutputMappings` are gated by the config check at the call site.

## Changes

- **TraceAdvice.java**: Wrap input capture + input mappings in `!config.isHideInputs()` guard in `onEnter`. Wrap output capture + output mappings in `!config.isHideOutputs()` guard in `onExit`.
- **TraceAdviceTest.java**: Add tests for `hideInputs`, `hideOutputs`, and both combined.
- **MappingTest.java**: Add tests verifying `@SpanMapping` input and output mappings are suppressed when hide config is set.

### DIFF
--- a/java/instrumentation/openinference-instrumentation-annotation/src/main/java/com/arize/instrumentation/annotation/TraceAdvice.java
+++ b/java/instrumentation/openinference-instrumentation-annotation/src/main/java/com/arize/instrumentation/annotation/TraceAdvice.java
@@ -4,7 +4,6 @@ import com.arize.instrumentation.ContextAttributes;
 import com.arize.instrumentation.OITracer;
 import com.arize.instrumentation.OpenInferenceAgent;
 import com.arize.instrumentation.SuppressTracing;
-import com.arize.instrumentation.TraceConfig;
 import com.arize.instrumentation.trace.*;
 import com.arize.semconv.trace.SemanticConventions;
 import java.lang.reflect.Method;
@@ -28,7 +27,6 @@ public class TraceAdvice {
         }
 
         TracedSpan span = null;
-        TraceConfig config = tracer.getConfig();
         try {
             String name = resolveSpanName(method);
             SemanticConventions.OpenInferenceSpanKind kind = resolveSpanKind(method);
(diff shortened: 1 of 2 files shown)


nate-mar:
Since `shouldHide()` now handles all the sensitive key checks in `TracedSpan.setAttribute()`, the outer `hideInputs`/`hideOutputs` guards in `TraceAdvice` can be removed.

`setInput()`/`setOutput()` already check the config internally, and `applyInputMappings`/`applyOutputMappings` go through `setAttribute()` which calls `shouldHide()`. So the call-site guards are redundant.

Removing them would also fix an over-suppression issue: right now `hideInputs=true` blocks all `@SpanMapping` input mappings, including non-sensitive ones like `llm.model_name`. With the guards removed, `shouldHide()` would still block sensitive keys but let things like model name through, which aligns with how LangChain4j behaves.

duncankmckinnon:
> Since `shouldHide()` now handles all the sensitive key checks in `TracedSpan.setAttribute()`, the outer `hideInputs`/`hideOutputs` guards in `TraceAdvice` can be removed.
> 
> `setInput()`/`setOutput()` already check the config internally, and `applyInputMappings`/`applyOutputMappings` go through `setAttribute()` which calls `shouldHide()`. So the call-site guards are redundant.
> 
> Removing them would also fix an over-suppression issue: right now `hideInputs=true` blocks all `@SpanMapping` input mappings, including non-sensitive ones like `llm.model_name`. With the guards removed, `shouldHide()` would still block sensitive keys but let things like model name through, which aligns with how LangChain4j behaves.

@nate-mar - should be fixed with most recent change!

PR #2928, opened 2026-03-30
feat(llama_index): Instrument Sparse Embedding & Workflow Events

### LINKED ISSUE #2918: [bug] openinference-instrumentation-llama-index: Unhandled event of type WorkflowStepOutputEvent
LlamaIndex Workflows introduces several new Events:

```
WARNING  Unhandled event of type WorkflowStepOutputEvent   _handler.py:481
```

https://github.com/run-llama/workflows-py/blob/main/packages/llama-index-workflows/src/workflows/runtime/types/step_function.py

ialisaleh:
Closes #2918 

**Sparse Embedding**

llama-index-sparse-embedding

**Workflow Events (WorkflowStepOutputEvent + WorkflowRunOutputEvent)**

llama-index-workflows-events-1

**Workflow Events (SpanCancelledEvent)**

llama-index-workflows-events-2


python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -1218,6 +1257,31 @@ def _get_token_counts_impl(
             pass
 
 
+if WorkflowStepOutputEvent is not None:
+
+    @_Span._process_event.register(WorkflowStepOutputEvent)  # type: ignore[attr-defined,misc]
+    def _workflow_step_output_event(self: _Span, event: WorkflowStepOutputEvent) -> None:  # type: ignore[misc]
+        # Pre-summarised step output produced by workflows.runtime
+        self[OUTPUT_VALUE] = event.output
```
caroger, line 1265: The new workflow output handlers now populate `output.value` before `process_output()` runs, so real `Workflow.run()` spans regress from structured JSON plus `output.mime_type=application/json` to the `summarize_event(...)` string with no MIME type. I reproduced that locally on this branch with a minimal workflow, while the merge base still emits JSON. Since these events are summary payloads, could we avoid writing them into `OUTPUT_VALUE` and let `process_output()` serialize the real returned event instead?

python/instrumentation/openinference-instrumentation-llama-index/tests/openinference/instrumentation/llama_index/test_multiple_workflow_events.py
```diff
@@ -0,0 +1,76 @@
...
+
+from openinference.instrumentation.llama_index import LlamaIndexInstrumentor
+from openinference.semconv.trace import SpanAttributes
+
+dispatcher = get_dispatcher(__name__)
+
+
+@dispatcher.span  # type: ignore[misc,unused-ignore]
+def step_span() -> None:
+    dispatcher.event(WorkflowStepOutputEvent(output="StopEvent(result='hello')"))
```
caroger, line 21: These tests synthesize `WorkflowStepOutputEvent` and `WorkflowRunOutputEvent` directly, then assert the summary string, so they lock in the behavioral regression above instead of exercising the real workflow runtime. A minimal `Workflow` integration test catches the difference immediately and also lets us assert `output.mime_type` and JSON output with pop-style checks. Could we drive this through `Workflow.run()` and use `handler.cancel_run()` for the cancellation case instead of dispatching the events by hand?

ialisaleh: Addressed.

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
ialisaleh: Addressed.

RogerHYang:
If we are changing the minimum version for the tests, then we should change the minimum version that's supported by the instrumentor and bump the major version of the instrumentor.

Screenshot 2026-04-21 at 2 33 04 PM

python/instrumentation/openinference-instrumentation-llama-index/src/openinference/instrumentation/llama_index/_handler.py
```diff
@@ -153,6 +155,20 @@
...
+        WorkflowRunOutputEvent,
+        WorkflowStepOutputEvent,
+    )
+
+    _HAS_WORKFLOW_EVENTS = True
+except ImportError:
+    SpanCancelledEvent = None  # type: ignore[misc,assignment]
+    WorkflowStepOutputEvent = None  # type: ignore[misc,assignment]
+    WorkflowRunOutputEvent = None  # type: ignore[misc,assignment]
+    _HAS_WORKFLOW_EVENTS = False
```
RogerHYang: do we still need these after bumping the version lower bound for llama index?

ialisaleh: Addressed.

caroger (approved):

PR #2932, opened 2026-03-30
ci: add Claude Code auto-fix step to Python canary cron

nate-mar:
## Summary

- Adds an `auto-fix` job to `python-cron.yaml` that runs when canary jobs fail, invoking Claude Code to investigate the upstream dependency change, draft a fix, and open a PR automatically
- Extracts failure collection into a shared `collect-failures` job so both `slack` and `auto-fix` can consume the failure list
- Adds a `python-canary-fix` skill (`.claude/skills/python-canary-fix/`) that guides Claude Code through the investigation and fix workflow

## How it works

1. Canary jobs run as before, uploading artifacts on failure
2. New `collect-failures` job aggregates failures into an output
3. If failures exist, `auto-fix` checks out the repo, installs Claude Code, and runs it with a prompt that includes the failure list
4. Claude Code investigates logs, identifies the upstream change, fixes the code, and opens a PR

## Test plan

- [ ] Trigger via `workflow_dispatch` to verify the auto-fix job runs correctly
- [ ] Verify Slack notification still fires (refactored to use `collect-failures` output)
- [ ] Verify `ANTHROPIC_API_KEY` secret is accessible to the workflow

### DIFF
--- a/.github/workflows/python-cron.yaml
+++ b/.github/workflows/python-cron.yaml
@@ -109,8 +109,6 @@ jobs:
               "type": "mrkdwn",
               "text": ${{ toJSON(steps.message.outputs.text) }}
             }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
 
   auto-fix:
     name: Propose fix for failures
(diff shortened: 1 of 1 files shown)


caroger:
consider creating the skill under .`agents/skills` and then symlink it under .claude and .codex

nate-mar:
Will do that in a follow up thanks @caroger

PR #2935, opened 2026-03-31
fix(agno): Agno Team Span Tree fix, Context issues in streaming and async streaming

### LINKED ISSUE #2640: [bug][agno] Remote team spans are no longer being nested properly
**Describe the bug**
I think version 0.1.25 accidentally broke remote agent nesting.  With all of our remote agents on version 0.1.23 they nest properly in arize with the teamleader as a root span, but if a remote agent moves to 0.1.25 or 0.1.26 then it shows as an independent span, no longer linked to the original.
**To Reproduce**
Override the arun method to extract the context , attach it to headers and make a call to a remote agentos endpoint wita an agui interface. I am paraphrasing as our implementation is a lot more complex with multiple hops until it reaches the agentos server but I can guarantee the agentos server is reached with the header from upstream and it worked with version 0.1.23.

**Expected behavior**
Remote runs of remote agents should show nested under the local team span as they did before.
For POST http://localhost:8081/agui with body:
```
{
  "threadId": "d635539-8f7d-494c-a8d0-f97ecc38e3dd",
  "runId": "example-test",
  "state": {},
  "messages": [
    {
      "id": "example-message",
      "role": "user",
      "content": "ask all subordinates to provide a poem from their own subordinates or from themselves if they don't have one"
    }
  ],
  "tools": [],
  "context": [],
  "forwardedProps": {
    "user_id": "andy"
  }
}
```

Image

Same setup with version 0.1.26:

Now there are 2 roots:

Image

Image

 - OS: fedora, debian,windows
 - Version 0.1.26

**Additional context**
As you can see the issue only affects remote teams, remote agents work as before. We have a custom remote agents implementation from before agno added the feature and the way we link the instrumentation is through headers.
```    def _extract_and_set_trace_context(self, headers: Dict[str, str]):
        """Extract OTEL trace context from request headers and set as current context."""
        if is_telemetry_disabled():
            return None
        try:
            if headers:
                ctx = extract(headers)
                token = otel_context.attach(ctx)
                logger.debug(
                    f"Extracted and attached trace context from headers: {list(headers.keys())}"
                )
                return token
            else:
                logger.debug(
                    "No headers found in request context for trace propagation"
                )
        except Exception as e:
            logger.warning(f"Failed to extract trace context: {str(e)}")
        return None
```

satyadevai:
Closes #2640 
image

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
@@ -251,10 +251,10 @@ def detach_context_tokens(
 ) -> None:
     """Helper function to detach context token with error handling."""
     if initial_thread is current_thread:
-        if ctx_token is not None:
-            context_api.detach(ctx_token)
         if team_token:
             context_api.detach(team_token)
+        if ctx_token is not None:
+            context_api.detach(ctx_token)
 
 
 class _RunWrapper:
--- a/python/instrumentation/openinference-instrumentation-agno/tests/test_team_context_isolation.py
+++ b/python/instrumentation/openinference-instrumentation-agno/tests/test_team_context_isolation.py
@@ -267,3 +267,97 @@ async def test_mixed_sync_async_team_runs_create_separate_traces(
         assert parent_span_context is None or not parent_span_context.is_valid, (
             "Mixed sync/async runs are nesting"
         )
+
+
+def test_streamed_team_run_outer_span_remains_current(
+    instrumented_team: Tuple[Team, InMemorySpanExporter],
+) -> None:
+    """Regression: outer workflow span must remain current after a streamed Team run.
+
+    Verifies that team_token is detached before ctx_token so the OTel context
+    stack is unwound in LIFO order.
+    """
+    from opentelemetry import trace as trace_api
+
+    team, exporter = instrumented_team
+    tracer = trace_api.get_tracer("test")
+
+    outer_span = tracer.start_span("outer_workflow")
+    outer_ctx = trace_api.use_span(outer_span, end_on_exit=False)
+    outer_ctx.__enter__()
+
+    try:
+        try:
+            for _ in team.run("Stream question", stream=True, yield_run_output=True):
+                pass
+        except Exception:
+            pass
+
+        current_span = trace_api.get_current_span()
+        assert current_span.get_span_context().span_id == outer_span.get_span_context().span_id, (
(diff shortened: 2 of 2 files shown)


python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
```diff
@@ -241,6 +246,17 @@ def _get_team_span_context(agent_or_team: Optional[Union[Agent, Team]]) -> Optio
...
 
+def detach_context_tokens(
+    initial_thread: Any, current_thread: Any, team_token: Any, ctx_token: Any
+) -> None:
+    """Helper function to detach context token with error handling."""
+    if initial_thread is current_thread:
+        if ctx_token is not None:
+            context_api.detach(ctx_token)
+        if team_token:
+            context_api.detach(team_token)
```
caroger, line 255: The streamed cleanup helper is unwinding the OpenTelemetry context stack in the wrong order. `run_stream` / `arun_stream` attach the span context first and then attach the Team context key on top of it. Detaching `ctx_token` before `team_token` reactivates the ended Team span, so the next span after a streamed Team run becomes a child of the Team span instead of the outer workflow span.

```suggestion
    if initial_thread is current_thread:
        if team_token:
            context_api.detach(team_token)
        if ctx_token is not None:
            context_api.detach(ctx_token)
```

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
```diff
@@ -385,47 +401,44 @@ def run_stream(
...
-            with trace_api.use_span(span, end_on_exit=False):
-                team_token, team_ctx = _setup_team_context(agent_or_team, node_id)
-                for response in wrapped(*args, **kwargs):
-                    if hasattr(response, "run_id"):
-                        current_run_id = response.run_id
-                        if current_run_id:
-                            span.set_attribute("agno.run.id", current_run_id)
+            # Manually attach/detach instead of use_span() context manager to context errors
+            # when yielding results across threads.
+            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
```
caroger, line 415: These changes only affect streamed execution, but the current Agno tests do not cover the `Team.run(..., stream=True, yield_run_output=True)` / `Team.arun(..., stream=True, yield_run_output=True)` path that Agno workflows use for team steps. Please add a regression test that keeps an outer workflow span current during a streamed Team run and asserts the next span is parented back to the workflow span rather than the ended Team span. Sync and async variants would both be useful.

PR #2944, opened 2026-04-01
fix(pydantic-ai): set output.value for plain text responses

ritiztambi:
Fixes #2943

## Summary
- When a PydanticAI agent uses the default `output_type=str`, the LLM responds with plain text (no `final_result` tool call). `output.value` was never set in this case, causing downstream consumers (e.g. Langfuse) to show null output.
- Captures the first assistant text part as `output_value` in `_extract_from_gen_ai_messages()`, mirroring how `input.value` is derived from the last user text part
- If a `final_result` tool call is also present, it still overwrites `output_value`, preserving existing behavior for structured output agents

## Change
One-line addition in `semantic_conventions.py` — after yielding `llm.output_messages.*.message.content` for a text part, also set `output_value` if not already set.


ritiztambi:

ritiztambi:
recheck

ialisaleh:
Hi Ritiz - This issue is addressed in a different PR, so closing this one. Thanks!

PR #2974, opened 2026-04-10
fix: withSpan sync errors, tracer resolution, and receiver forwarding

### LINKED ISSUE #2975: withSpan: sync errors aren't recorded/ended; withSpan also loses this-binding when wrapping unbound methods
## Summary

Two related bugs in `@arizeai/openinference-core`'s `withSpan` helper (`js/packages/openinference-core/src/helpers/withSpan.ts`):

1. **Synchronous errors are not recorded on the span, and the span is never ended.** The async path has a `.catch(...).finally(span.end)` chain, but the sync path calls `fn(...args)` with no try/catch, so any synchronously-thrown error escapes past `span.recordException` and `span.end()`. The span leaks and stays open with no error status.

2. **`this`-binding is lost when wrapping a method directly.** The wrapped function calls `fn(...args)` (bare) rather than `fn.apply(this, args)`, so `withSpan(obj.method)` loses its receiver. The `observe` decorator works around this internally via `originalMethod.bind(this)`, but users calling `withSpan` directly on a method hit the bug.

## Repro (bug 1 — sync errors)

```ts
import { withSpan } from "@arizeai/openinference-core";

const traced = withSpan(() => {
  throw new Error("boom");
}, { name: "sync-failure" });

try {
  traced();
} catch (_) {}
// Expected: span ended with ERROR status, exception recorded.
// Actual: span is never ended; no exception recorded on the span.
```

## Repro (bug 2 — this-binding)

```ts
import { withSpan } from "@arizeai/openinference-core";

class Service {
  prefix = "svc";
  run() { return \`\${this.prefix}:done\`; }
}

const svc = new Service();
const traced = withSpan(svc.run);
traced(); // TypeError: Cannot read properties of undefined (reading 'prefix')
```

## Proposed fix

In `withSpan.ts`, wrap the sync path in try/catch that mirrors the async `.catch` handler, and invoke `fn.apply(this, args)` from inside a `function (...args)` wrapper so `this` is preserved:

```ts
const wrappedFn = function (this: ThisParameterType<Fn>, ...args: Parameters<Fn>) {
  return tracer.startActiveSpan(spanName, { /* ... */ }, (span) => {
    try {
      const result = fn.apply(this, args);
      if (isPromise(result)) {
        return result
          .then((value) => {
            span.setAttributes(processOutput(value));
            span.setStatus({ code: SpanStatusCode.OK });
            return value;
          })
          .catch((err) => {
            span.recordException(err);
            span.setStatus({
              code: SpanStatusCode.ERROR,
              message: String(err?.message ?? err),
            });
            throw err;
          })
          .finally(() => span.end());
      }
      span.setAttributes(processOutput(result));
      span.setStatus({ code: SpanStatusCode.OK });
      span.end();
      return result;
    } catch (err) {
      span.recordException(err);
      span.setStatus({
        code: SpanStatusCode.ERROR,
        message: String(err?.message ?? err),
      });
      span.end();
      throw err;
    }
  });
};
```

## Context

Found while collapsing a local reimplementation of these helpers in `@arizeai/phoenix-otel` back into pure re-exports from `@arizeai/openinference-core`. Phoenix's fork carries exactly these two fixes; once the fix lands upstream, the fork can go away entirely.

mikeldking:
## Summary

resolves #2975

This fixes three `@arizeai/openinference-core` issues in `withSpan`.

- wrap the synchronous execution path in `try/catch` so sync throws record an exception, set span status to `ERROR`, end the span, and rethrow
- invoke the wrapped function with `fn.apply(this, args)` so the traced wrapper forwards the call-site receiver
- defer default tracer resolution until invocation so wrappers created before a global provider change use the current global tracer provider
- add regression coverage for sync throws, method-style invocation, and default-vs-explicit tracer behavior, and update the hand-written tracing docs

## Root Cause
`withSpan` only handled promise rejections in its error path. If the wrapped function threw before returning a promise, the exception bypassed `recordException` and `span.end()`. Separately, the wrapper invoked `fn(...args)` as a bare function call, which discarded the wrapper's `this` when the traced function was invoked as a method.

The default tracer path was also resolved when `withSpan(...)` was called, so wrappers without an explicit tracer stayed bound to the global tracer provider that was active at declaration time rather than the provider active when the wrapper ran.

## Impact
Synchronous failures now close spans correctly with error status and exception events instead of leaking open spans. Method calls through the traced wrapper now preserve the receiver from the call site. Wrappers that rely on the global tracer provider now pick up provider changes on later invocations, while explicit tracers remain pinned.

Note: detached method references still require explicit `.bind(instance)` before wrapping or calling without a receiver, because JavaScript does not retain the original object once a method is extracted.

## Validation
- `pnpm --filter @arizeai/openinference-core test`
- `pnpm --filter @arizeai/openinference-core build`

### DIFF
--- a/js/packages/openinference-core/test/helpers/withSpan.test.ts
+++ b/js/packages/openinference-core/test/helpers/withSpan.test.ts
@@ -29,6 +29,7 @@ describe("withSpan", () => {
     // Clean up after each test
     spanExporter.reset();
     tracerProvider.shutdown();
+    trace.disable();
   });
 
   it("should wrap synchronous functions and create spans", () => {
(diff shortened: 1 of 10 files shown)


js/packages/openinference-core/test/helpers/withSpan.test.ts
```diff
@@ -194,6 +272,31 @@ describe("withSpan", () => {
     const span = spans[0];
     expect(span.attributes["test-attribute"]).toBe("test-value");
   });
+
+  it("should preserve this when the traced wrapper is invoked as a method", () => {
```
mikeldking, line 276: this is to work with dynamic tracers so we can change the tracing context dynamically during testing and experimentation.

PR #2977, opened 2026-04-10
fix(agno): finalize arun_stream output on early stream close

### LINKED ISSUE #2976: [bug] openinference-instrumentation-agno: arun_stream drops output.value when stream closes after final RunOutput
**Describe the bug**
When `openinference-instrumentation-agno` instruments `arun_stream()`, the agent span can finish without `output.value` if the consumer closes the async generator immediately after receiving the final `RunOutput`.

**To Reproduce**
1. Install `openinference-instrumentation-agno` from `Arize-ai/openinference` `main`.
2. Run the repro below.
3. Observe that the finished span has `output.value=None` even though the final `RunOutput` contains structured content.

```python
import asyncio

from agno.agent.agent import Agent
from agno.run.agent import RunOutput
from openinference.instrumentation import OITracer, TraceConfig
from openinference.instrumentation.agno._runs_wrapper import _RunWrapper


class FinalAnswer(BaseModel):
    answer: str


async def fake_arun_stream(*_args, **_kwargs):
    yield RunOutput(run_id="run-123", content=FinalAnswer(answer="done"))


def build_wrapper():
    provider = TracerProvider()
    tracer = OITracer(
        trace_api.get_tracer("openinference-main-repro", tracer_provider=provider),
        config=TraceConfig(),
    )
    return _RunWrapper(tracer=tracer), exporter


async def main():
    wrapper, exporter = build_wrapper()
    agent = Agent(name="test-agent")
    stream = wrapper.arun_stream(
        fake_arun_stream,
        None,
        (agent,),
        {"yield_run_output": True},
    )
    event = await anext(stream)
    assert isinstance(event, RunOutput)
    await stream.aclose()
    await asyncio.sleep(0)

    spans = exporter.get_finished_spans()
    print(spans[0].attributes.get("output.value"))


asyncio.run(main())
```

Observed output:

```text
None
```

Relevant control flow in `python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py`:

- the wrapper saves the final `RunOutput` into `run_response`
- but it only writes `output.value` after the `async for` completes normally
- if the consumer closes the generator at `yield response`, the post-loop finalization block is skipped

**Expected behavior**
If `arun_stream()` has already received the final `RunOutput`, the span should still record `output.value` even when the consumer closes the generator immediately afterwards via `aclose()` or `async for ... break`.

**Screenshots**
Not applicable.

 - OS: Linux 6.8.0-100-generic
 - Version: `openinference` main at `d563e428983808b9151d2b7c868d3778c33441a8`

**Additional context**
- Tested package versions in the repro environment:
  - `agno==2.5.15`
  - `openinference-instrumentation==0.1.46`
  - `opentelemetry-api==1.41.0`
  - `opentelemetry-sdk==1.41.0`
- I did not find an existing upstream issue/PR for this exact remaining bug via repository search.
- A minimal fix is to bind the wrapped async generator to `iterator`, catch `GeneratorExit` around `yield response`, `await iterator.aclose()`, and then let the existing post-loop finalization write `output.value` from the cached `run_response`.

amarobus:
fixes #2976

Preserve `output.value` on Agno `arun_stream` spans when the consumer closes the
stream immediately after receiving the final `RunOutput`.

## Summary

- bind the wrapped async generator to `iterator` so the wrapper can explicitly
  close it during `GeneratorExit`
- catch `GeneratorExit` around `yield response` and close the inner iterator
  before breaking out of the loop
- add a focused regression test covering both `aclose()` and `async for ...
  break` after the final `RunOutput`

## Why

`arun_stream()` already caches the final `RunOutput` in `run_response`, but it
only writes `output.value` after the `async for` completes normally. If the
consumer closes the generator at the outer `yield response`, the loop-finalizing
code is skipped and the span ends without `output.value`.

This change keeps the existing context-management fix from `#2935` intact while
ensuring span output is finalized for early-close consumers too.

- Added a focused regression test for `aclose()` and `async for ... break` after the final `RunOutput`
- Validated locally in an isolated environment with:
  - `PYTHONPATH=... python -m pytest python/instrumentation/openinference-instrumentation-agno/tests/test_async_stream_output_finalization.py -q`

## Notes

- This PR is intentionally small and limited to the remaining `output.value`
  finalization bug in `openinference-instrumentation-agno`.
- I did not include screenshots because the bug only changes a missing span
  attribute and the regression test captures the behavior directly.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
@@ -3,6 +3,7 @@ import json
 import threading
 from typing import (
     Any,
+    AsyncIterator,
     Awaitable,
     Callable,
     Iterator,
@@ -581,7 +582,7 @@ class _RunWrapper:
                 yield_run_output_set = True
                 kwargs["yield_run_output"] = True  # type: ignore
             run_response = None
-            iterator = wrapped(*args, **kwargs)  # type: ignore
+            iterator = cast(AsyncIterator[Any], wrapped(*args, **kwargs))
             ctx_token = context_api.attach(trace_api.set_span_in_context(span))
             team_token, team_ctx = _setup_team_context(agent_or_team, node_id)
             async for response in iterator:
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
@@ -581,9 +581,10 @@ class _RunWrapper:
                 yield_run_output_set = True
                 kwargs["yield_run_output"] = True  # type: ignore
             run_response = None
+            iterator = wrapped(*args, **kwargs)  # type: ignore
             ctx_token = context_api.attach(trace_api.set_span_in_context(span))
             team_token, team_ctx = _setup_team_context(agent_or_team, node_id)
-            async for response in wrapped(*args, **kwargs):  # type: ignore
+            async for response in iterator:
                 if hasattr(response, "run_id"):
                     current_run_id = response.run_id
                     if current_run_id:
@@ -594,7 +595,12 @@ class _RunWrapper:
                     if yield_run_output_set:
                         continue
 
-                yield response
+                try:
+                    yield response
+                except GeneratorExit:
+                    if hasattr(iterator, "aclose"):
+                        await iterator.aclose()
+                    break
 
             if run_response is not None:
                 output = _extract_run_response_output(run_response)
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
@@ -21,8 +21,10 @@ from opentelemetry.util.types import AttributeValue
 
 from agno.agent import Agent
 from agno.models.message import Message
+from agno.run.agent import RunCompletedEvent as AgentRunCompletedEvent
 from agno.run.agent import RunOutput
 from agno.run.messages import RunMessages
+from agno.run.team import RunCompletedEvent as TeamRunCompletedEvent
 from agno.run.team import TeamRunOutput
 from agno.team import Team
 from agno.tools.function import Function
@@ -94,6 +96,18 @@ def _extract_run_response_output(run_response: Union[RunOutput, TeamRunOutput]) 
         else:
             return str(run_response.content.model_dump_json())
     return ""
+
+
+def _extract_completed_event_output(
+    completed_event: Union[AgentRunCompletedEvent, TeamRunCompletedEvent],
+) -> str:
+    if completed_event.content is None:
+        return ""
+    if isinstance(completed_event.content, str):
+        return completed_event.content
+    if hasattr(completed_event.content, "model_dump_json"):
+        return str(completed_event.content.model_dump_json())
+    return str(completed_event.content)
 
 
 def _strip_method_args(arguments: Mapping[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


amarobus:

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
```diff
@@ -594,7 +595,12 @@ async def arun_stream(
                     if yield_run_output_set:
                         continue
 
-                yield response
+                try:
+                    yield response
+                except GeneratorExit:
```
caroger, line 619: This fixes #2976's exact `aclose()` / `break`-after-`RunOutput` repro, but I think `stream_events=True` is still broken.

This branch only finalizes from `RunOutput` / `TeamRunOutput`. If a consumer stops on `RunCompletedEvent` before the terminal `RunOutput`, `output.value` is still dropped.

Exact repro:

```python
import asyncio

from agno.agent import Agent
from agno.run.agent import RunCompletedEvent, RunOutput

from openinference.instrumentation import OITracer, TraceConfig
from openinference.instrumentation.agno._runs_wrapper import _RunWrapper


class FinalAnswer(BaseModel):
    answer: str


async def fake_arun_stream(*_args, **_kwargs):
    yield RunCompletedEvent(content=FinalAnswer(answer="done"))
    yield RunOutput(run_id="run-123", content=FinalAnswer(answer="done"))


async def main():
    provider = TracerProvider()
    tracer = OITracer(trace_api.get_tracer("repro", tracer_provider=provider), config=TraceConfig())

    wrapper = _RunWrapper(tracer=tracer)
    agent = Agent(name="test-agent")

    async for event in wrapper.arun_stream(
        fake_arun_stream,
        None,
        (agent,),
        {"yield_run_output": True, "stream_events": True},
    ):
        if isinstance(event, RunCompletedEvent):
            break

    await asyncio.sleep(0.05)
    spans = exporter.get_finished_spans()
    print(spans[0].attributes.get("output.value"))  # None


asyncio.run(main())
```

Can we also finalize from `RunCompletedEvent` / `TeamRunCompletedEvent`, or add a regression test for this path and adjust the wrapper until it passes?

python/instrumentation/openinference-instrumentation-agno/tests/test_async_stream_output_finalization.py
```diff
@@ -0,0 +1,70 @@
...
+    tracer = OITracer(
+        trace_api.get_tracer("test-agno-arun-stream", tracer_provider=tracer_provider),
+        config=TraceConfig(),
+    )
+    return _RunWrapper(tracer=tracer), exporter
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("close_mode", ["aclose", "break"])
+async def test_arun_stream_records_output_value_when_closed_after_final_run_output(
```
caroger, line 53: Please add the `stream_events=True` variant here.

This test only covers closing after the final `RunOutput`. It does not cover the case where the consumer breaks on `RunCompletedEvent`, which is the remaining early-close path that can still finish the span without `output.value`.

caroger (changes requested):
Thanks a lot for jumping in with this fix, and welcome to OpenInference @amarobus! Really appreciate the first contribution here.

This PR fixes the original #2976 repro cleanly. I left one inline blocking concern: I think there is still a remaining early-close gap when `stream_events=True`. If a consumer stops on `RunCompletedEvent` before the terminal `RunOutput`, `output.value` can still be dropped. I included a small repro in the inline comment and suggested adding a regression test for that path.

For fast local iteration, it may help to run the package CI target directly with `cd python && uvx --with tox-uv tox run -e py314-ci-agno`.

Once that `stream_events=True` case is covered, this should be in good shape. Thanks again for the patch.

python/instrumentation/openinference-instrumentation-agno/tests/test_async_stream_output_finalization.py
amarobus: Thanks, good catch. I reproduced the `stream_events=True` / `RunCompletedEvent` path, added a regression test for it, and updated the wrapper to preserve `output.value` from completion events as well.

caroger (approved):

PR #3003, opened 2026-04-16
fix: ensure pipecat metadata is JSON

duncankmckinnon:
### Problem

When using pipecat with Google Gemini (or other providers with Pydantic model settings), the `metadata` span attribute was being set to a Python repr string like:

```
GoogleLLMSettings(model='gemini-3-flash-preview', extra={}, system_instruction=None, ...)
```

This caused export failures with `StatusCode.UNKNOWN` because the Arize backend requires `metadata` to be valid JSON.

### Fix

In `_attributes.py`, the `LLMServiceAttributeExtractor.extract_from_service` method now converts non-dict settings objects to dicts before JSON serialization, using `model_dump()` (Pydantic v2), `dict()` (Pydantic v1), or `__dict__` as fallbacks.

### Testing

All 66 existing tests pass.


python/instrumentation/openinference-instrumentation-pipecat/src/openinference/instrumentation/pipecat/_attributes.py
```diff
@@ -818,10 +818,16 @@ def extract_from_service(self, service: FrameProcessor) -> Dict[str, Any]:
...
         # Extract LLM settings/configuration as metadata
         if hasattr(service, "_settings"):
-            if isinstance(service._settings, dict):
-                results[SpanAttributes.METADATA] = safe_json_dumps(service._settings)
-            else:
-                results[SpanAttributes.METADATA] = str(service._settings)
+            settings = service._settings
+            if not isinstance(settings, dict):
+                # Convert Pydantic models, dataclasses, or other objects to dicts
+                if hasattr(settings, "model_dump"):
```
nate-mar: You'll probably want to make sure that model_dump and dict() are callable as well. You can find examples of this on other instruments as well where we ensure that it both has an attr and is callable.

nate-mar (approved):

PR #3016, opened 2026-04-21
fix(google-genai): don't double-count cached tokens in prompt

### LINKED ISSUE #3013: google-genai: cached_content_token_count added twice into llm.token_count.prompt
## Summary

In `openinference-instrumentation-google-genai`, the Gemini `usage_metadata` extraction adds `cached_content_token_count` into the emitted `llm.token_count.prompt` attribute. Because Gemini's API already includes cached tokens inside `prompt_token_count`, this double-counts the cached portion, inflating the prompt attribute by exactly `cached_content_token_count`.

The same count is then also emitted separately under `llm.token_count.prompt_details.cache_read`, so downstream consumers that bill both keys effectively count the cached tokens three times (once in the full input, once at the discounted cache rate, and a latent third time if they also forward `total`).

## Reproduction

Package version: `openinference-instrumentation-google-genai==0.1.14` (also present on `main` and in `v0.1.15`).

`python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_utils.py`
Function: `_get_token_count_attributes_from_usage_metadata`

```python
prompt_token_count = 0
if usage_metadata.prompt_token_count:
    prompt_token_count += usage_metadata.prompt_token_count
if usage_metadata.tool_use_prompt_token_count:
    prompt_token_count += usage_metadata.tool_use_prompt_token_count
if usage_metadata.cached_content_token_count:
    yield (
        SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ,
        usage_metadata.cached_content_token_count,
    )
    prompt_token_count += usage_metadata.cached_content_token_count   # <- double-count
if prompt_token_count:
    yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT, prompt_token_count
```

## Why this is incorrect

Per Google's official Gemini REST API reference (https://ai.google.dev/api/generate-content) for `UsageMetadata`:

> `promptTokenCount`: Number of tokens in the prompt. When `cachedContent` is set, this is still the total effective prompt size meaning this includes the number of tokens in the cached content.

The `python-genai` SDK (`google/genai/types.py`, `GenerateContentResponseUsageMetadata`) says the same thing. So `cached_content_token_count` is a subset of `prompt_token_count`, not an additional quantity. Adding it produces `prompt = P + C` where `P` was already the total.

Empirical check with a real response: a call with `prompt_token_count = 2560`, `cached_content_token_count = 2295` is only coherent if cached is a subset of prompt — otherwise 89% of the prompt would have to be duplicated content.

## Downstream impact

Observability platforms that map `llm.token_count.prompt` to an input-token usage type and price it at the uncached input rate will overstate Gemini cost. The magnitude scales with cache hit rate `X`:

- X = 33%: ~1.4x overstatement
- X = 75%: ~2.8x
- X = 90%: ~5.3x

When the observability platform additionally prices `prompt_details.cache_read` at the cache rate (which some do), cached tokens are effectively billed twice — once at full input rate inside the inflated prompt, once at the cache rate as a separate line.

## Context

PR #2779 introduced this extraction path. The PR description states the intent is to *subtract* cached tokens from prompt to represent active tokens, but the merged implementation adds. The `total_token_count` computation in the same function uses the inflated `prompt_token_count`, so totals are also affected.

## Suggested fix

Stop adding `cached_content_token_count` into `prompt_token_count`. The emitted values should reflect Gemini's API semantic directly:

- `llm.token_count.prompt = prompt_token_count + tool_use_prompt_token_count` (no cached addend; cached is already included in `prompt_token_count`)
- `llm.token_count.prompt_details.cache_read = cached_content_token_count` (unchanged)

Updating the test `test_get_attributes_from_generate_content_usage_cached` in `tests/test_response_attributes_extractor.py` to assert `prompt: 30` rather than `prompt: 50` given `prompt_token_count=30, cached=20` would codify the corrected behaviour.

Happy to open a PR if it's useful.

rivernate:
Fixes #3013.

Gemini's `prompt_token_count` already includes `cached_content_token_count` (per the `google-genai` SDK docstring and the Gemini REST API reference). The instrumentation was adding the cached count again, inflating `llm.token_count.prompt` and `llm.token_count.total` by the cache-read amount — causing downstream platforms to overstate Gemini cost, and to double-bill cached tokens when they also price `prompt_details.cache_read`.

## Change

Drop the `prompt_token_count += cached_content_token_count` line in `_get_token_count_attributes_from_usage_metadata`. `cached_content_token_count` is still emitted separately as `llm.token_count.prompt_details.cache_read` (a subset of `prompt`).

## Test plan

- Added regression test `test_cached_content_tokens_not_double_counted_in_prompt` in `tests/test_utils.py`.
- Updated the existing `test_get_attributes_from_generate_content_usage_cached` in `tests/test_response_attributes_extractor.py` whose expectations previously codified the bug (`prompt: 50 → 30`, `total: 150 → 130`).
- Full package test suite passes (62 tests).

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_utils.py
+++ b/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_utils.py
@@ -53,7 +53,7 @@ def _get_token_count_attributes_from_usage_metadata(
         if prompt_details_audio:
             yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_AUDIO, prompt_details_audio
 
-    # Calculate total prompt tokens (base + tool use + cached)
+    # Calculate total prompt tokens (base + tool use; cached is already in `prompt_token_count`)
     prompt_token_count = 0
     if usage_metadata.prompt_token_count:
         prompt_token_count += usage_metadata.prompt_token_count
@@ -64,7 +64,6 @@ def _get_token_count_attributes_from_usage_metadata(
             SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ,
             usage_metadata.cached_content_token_count,
         )
-        prompt_token_count += usage_metadata.cached_content_token_count
     if prompt_token_count:
         yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT, prompt_token_count
 
--- a/python/instrumentation/openinference-instrumentation-google-genai/tests/test_response_attributes_extractor.py
+++ b/python/instrumentation/openinference-instrumentation-google-genai/tests/test_response_attributes_extractor.py
@@ -72,8 +72,8 @@ def test_get_attributes_from_generate_content_usage(
                 thoughts_token_count=20,
             ),
             {
-                "llm.token_count.total": 150,
-                "llm.token_count.prompt": 50,
+                "llm.token_count.total": 130,
+                "llm.token_count.prompt": 30,
                 "llm.token_count.completion": 100,
                 "llm.token_count.completion_details.reasoning": 20,
                 "llm.token_count.prompt_details.cache_read": 20,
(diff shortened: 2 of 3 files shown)


rivernate:

rivernate:
recheck

RogerHYang (approved):

PR #3039, opened 2026-04-28
feat(bedrock): add Amazon Nova model support for invoke_model

### LINKED ISSUE #3006: [feature request]
## Is your feature request related to a problem? Please describe.

The Python Bedrock auto-instrumentation (`openinference-instrumentation-bedrock`) does not support Amazon Nova Sonic 1 or Nova Sonic 2 models. When using these models, traces may not be captured properly or at all. Additionally, some Bedrock calls go through AWS Smithy rather than `boto3`, which completely bypasses the current instrumentation since it only patches `botocore.client.ClientCreator` and `aiobotocore.client.AioClientCreator`.

The [[supported models documentation](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-bedrock/README.md#supported-models)](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-bedrock/README.md#supported-models) only lists Anthropic Claude, Meta Llama, and Mistral models—no Amazon Nova models are mentioned. The JavaScript/TypeScript instrumentation already has Nova support (tested with Nova Lite and Nova Micro), but Python currently has no Nova coverage or tests.

---

## Describe the solution you'd like

1. Add official support for Amazon Nova models, including:

   * Nova Sonic 1 (`amazon.nova-sonic-v1:0`)
   * Nova Sonic 2 (`amazon.nova-sonic-v2:0`)

2. Add test coverage for Nova models using VCR cassettes (similar to existing tests for other models)

3. Update documentation to include Nova models in the supported models table

4. Investigate Smithy-based instrumentation to handle Bedrock calls that bypass `boto3` / `aioboto3`

The solution should align with the JavaScript implementation patterns, which already include:

* Nova-specific helper functions (`isNovaRequest`, `isNovaResponse`, `processNovaStreamChunk`)
* Streaming support for Nova models
* Test coverage for Nova Lite and Nova Micro

---

## Describe alternatives you've considered

1. **Relying on vendor-based detection alone**
   The existing code uses `model_id.split(".")` to extract the vendor, so Nova models might technically work since they use the `"amazon"` prefix. However, without explicit request/response handling and tests, this is unreliable.

2. **Manual instrumentation**
   Users could manually create spans, but this defeats the purpose of auto-instrumentation and adds maintenance burden.

3. **Using only boto3 paths**
   Users could ensure all calls go through `boto3`, but this doesn't solve the problem for systems already using Smithy-based calls.

---

### Reproduction example

```python
import boto3
from openinference.instrumentation.bedrock import BedrockInstrumentor

BedrockInstrumentor().instrument()

session = boto3.session.Session()
client = session.client("bedrock-runtime")

# This may or may not be traced properly
response = client.converse(
    modelId="amazon.nova-sonic-v1:0",
    messages=[{
        "role": "user",
        "content": [{"text": "Hello!"}]
    }]
)
```

### References

* JavaScript reference implementation: `js/packages/openinference-instrumentation-bedrock/` (has working Nova support)
* Relevant Python files:

  * `python/instrumentation/openinference-instrumentation-bedrock/README.md`
  * `python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/__init__.py`
  * `python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py`

### Notes

AWS recently released Nova Sonic models with improved performance, and users need reliable tracing support for these models.

satyadevai:
Closes #3006 

Adds Amazon Nova model support for the Bedrock invoke_model and invoke_model_with_response_stream APIs, including multi-turn tool-use messages.

Changes
Provider attribute: Sets llm.provider = aws on Nova spans.
invoke_model (non-streaming):
Detects Nova requests via vendor + _is_nova_request and extracts structured input messages (system + user/assistant turns) into llm.input_messages.*.
Builds structured llm.output_messages.* from Nova response bodies, including assistant text and toolUse blocks.
Captures token counts (inputTokens, outputTokens, totalTokens) and Nova prompt-cache metrics from usage.
invoke_model_with_response_stream (streaming):
New _NovaStreamCallback wraps the Bedrock EventStream, parses contentBlockDelta/messageStop/metadata chunks, accumulates output text, and records token usage on stream completion.
Multi-turn tool messages: Input message construction now handles tool-result content blocks across turns for Nova invoke_model.
Examples: examples/bedrock_nova.py and examples/bedrock_nova_tool_use.py.
Tests: New cassette-based tests covering Nova Micro/Lite text, system prompts, tool use, streaming, and a Converse-path Nova test. Updates to test_instrumentor.py.

image
image
image
image


python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_wrappers.py
```diff
@@ -119,6 +182,15 @@ def handle_response(response: Any, kwargs: Mapping[str, Any], span: Span) -> Any
...
                 )
                 return response
+            model_id = str(kwargs.get("modelId", ""))
+            if model_id.startswith("amazon.nova"):
+                parsed_request = {**kwargs, "body": body}
+                response["body"] = _EventStream(
+                    response["body"],
+                    _NovaStreamCallback(span, parsed_request),
+                    _use_span(span),
+                )
```
mikeldking, line 282: **Nova streaming detection misses cross-region inference profile IDs**

Nova model detection uses `model_id.startswith("amazon.nova")`, which misses AWS-supported cross-region inference profile IDs (e.g. `us.amazon.nova-pro-v1:0`, `eu.amazon.nova-lite-v1:0`, `apac.amazon.nova-micro-v1:0`). For these IDs the streaming branch is skipped and execution falls through to the non-Nova branch which calls `span.set_attribute(INPUT_VALUE, body)` and `(LLM_INVOCATION_PARAMETERS, body)` where `body` is a dict (from `json.loads`) — not a valid OpenTelemetry attribute value, so streaming attributes are lost and OTel emits warnings.

The same vendor-prefix flaw exists in `_extract_invoke_model_attributes.set_input_attributes` where `model_id.split(".")[0]` yields `"us"` instead of `"amazon"` for CRI IDs, causing the Nova branch to be missed and `INPUT_VALUE` to be set to `None`.

Consider using a substring check like `"amazon.nova" in model_id` or a body-content check (e.g. `schemaVersion == "messages-v1"`) similar to how Anthropic detection uses `"anthropic_version" in body`.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_wrappers.py
```diff
@@ -39,6 +46,62 @@
...
+                            span.set_attribute(LLM_TOKEN_COUNT_TOTAL, v)
+                except Exception:
+                    pass
+        elif isinstance(obj, (StopIteration, StopAsyncIteration)):
+            if self._text_chunks:
+                span.set_attribute(OUTPUT_VALUE, "".join(self._text_chunks))
+            _finish(span, None, {})
+        elif isinstance(obj, BaseException):
+            _finish(span, obj, {})
+        return obj
```
mikeldking, line 108: **`_NovaStreamCallback` drops tool-use blocks and never sets `totalTokens`**

Two parity gaps with the non-streaming Nova path:

1. It only handles `contentBlockDelta.delta.text`. Nova streaming tool-use events (`contentBlockStart` with `start.toolUse` and `contentBlockDelta` with `delta.toolUse.input`) are silently dropped, so streaming spans for Nova tool-use prompts are missing all `tool_calls` attributes that the non-streaming `_build_nova_output_messages` emits.
2. Nova streaming `metadata.usage` emits `inputTokens`/`outputTokens` but no `totalTokens` (verified in the new streaming cassette and acknowledged by `test_invoke_model_nova.py:323`). The callback only sets `LLM_TOKEN_COUNT_TOTAL` when `usage.totalTokens` is present, so streaming spans always lack it.

Fix: accumulate `toolUse` start/input-delta events and emit tool_call attributes on stop (same shape as `_ConverseStreamCallback._handle_content_block`); compute `total = input + output` as a fallback like `_set_token_count_attributes` does for the non-streaming path.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
```diff
@@ -127,10 +152,30 @@ def set_input_attributes(span: Span, request_body: Dict[str, Any]) -> None:
...
+        input_messages = _build_nova_input_messages(request_body)
+        if input_messages:
+            span.set_attributes(get_llm_attributes(input_messages=input_messages))
+    else:
+        # All other models (anthropic completion style, cohere, meta, ai21):
+        # input is the prompt field, remaining body fields are invocation params
+        input_value = request_body.pop("prompt", None)
+        invocation_parameters = safe_json_dumps(request_body)
+
+    _set_span_attribute(span, SpanAttributes.INPUT_VALUE, input_value)
```
mikeldking, line 182: **Nova invoke_model drops `toolConfig` and never sets `LLM_TOOLS`**

`set_input_attributes` for Nova invoke_model serializes only `request_body['inferenceConfig']` into `LLM_INVOCATION_PARAMETERS`, dropping `toolConfig`, `schemaVersion`, and `system`; the Nova path also never populates the `LLM_TOOLS` span attribute.

As a result, traces of Nova tool-use invoke_model calls record the assistant's tool *call* on the output side but completely lose the tool catalog on the input side, breaking observability parity with the Converse path.

Fix by including the tool schemas via `get_llm_attributes(tools=...)` (mirroring `_converse_attributes.py:159-167`) and considering whether non-`inferenceConfig` fields belong in `invocation_parameters`.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
```diff
@@ -189,3 +236,136 @@ def is_claude_message_api(model_id: str) -> bool:
...
+
+
+def _set_nova_body_token_attributes(span: Span, response_body: Dict[str, Any]) -> None:
+    usage = response_body.get("usage", {})
+    if not isinstance(usage, dict):
+        return
+    if (v := usage.get("cacheReadInputTokens")) is not None:
+        _set_span_attribute(span, SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ, int(v))
+    if (v := usage.get("cacheWriteInputTokens")) is not None:
+        _set_span_attribute(span, SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, int(v))
```
mikeldking, line 366: **Nova invoke_model cache token field names use wrong suffix**

`_set_nova_body_token_attributes` reads `usage.get("cacheReadInputTokens")` / `usage.get("cacheWriteInputTokens")` (no `Count` suffix), but every Nova `invoke_model` response in this PR's cassettes uses the `Count`-suffixed names (`cacheReadInputTokenCount` / `cacheWriteInputTokenCount`).

The helper's only purpose for `invoke_model` is silently a no-op: `LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ` / `CACHE_WRITE` will never be set.

Fix: change the two key names to `cacheReadInputTokenCount` / `cacheWriteInputTokenCount`. (The no-`Count` names are correct for the Converse API only — see `_converse_attributes.py:199-201`.)

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
```diff
@@ -189,3 +236,136 @@ def is_claude_message_api(model_id: str) -> bool:
...
+
+
+def _extract_nova_output(response_body: Dict[str, Any]) -> str:
+    """Extract assistant text from a Nova response body (text blocks only)."""
+    content_blocks = response_body.get("output", {}).get("message", {}).get("content", [])
+    return "\n".join(
+        block.get("text", "")
+        for block in content_blocks
+        if isinstance(block, dict) and block.get("text")
+    )
```
mikeldking: **Nit: dead code — `_extract_nova_input` and `_extract_nova_output` have no callers**

`_extract_nova_input` (lines 260-268) and `_extract_nova_output` (lines 271-278) are defined here but never called anywhere in the codebase (verified via grep — only the two definitions appear in `src/` and `tests/`).

The actual Nova input/output extraction is performed by `_build_nova_input_messages`, `_build_nova_output_messages`, `get_output_attributes(message)`, and `safe_json_dumps(messages)` in `set_input_attributes` and `_set_model_name_attributes`. These two helpers appear to be leftovers from an earlier implementation and should be deleted before merging.

mikeldking (commented):
Automated review findings on the Nova invoke_model support. 4 normal-severity issues and 1 nit. Inline comments below.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_wrappers.py
satyadevai: Fixed. Changed model_id.startswith("amazon.nova") to "amazon.nova" in model_id in all places

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_wrappers.py
satyadevai: Fixed and verified both issues

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
satyadevai: Fixed and verified.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
satyadevai: Removed Dead code.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/utils/_extract_invoke_model_attributes.py
satyadevai: Fixed and verifeid

caroger (changes requested):
Please consider the following:

`_NovaStreamCallback._finalize_output` only emits structured `llm.output_messages.*` when tool calls exist, so text-only Nova streams currently get `output.value` but no `output.mime_type` and no structured assistant output message. The pure tool-call case has the inverse problem: it emits the tool call under `llm.output_messages.*`, but never sets `output.value`/`output.mime_type` because there is no text chunk.

That diverges from the non-streaming Nova path and means downstream chat/tool renderers lose either the structured assistant text or the top-level output, depending on the stream shape. Please make the streaming callback mirror the non-streaming response extraction for both text-only and tool-call outputs: always set `output.value`/`output.mime_type`, and emit `llm.output_messages.*` for text as well as tool calls.

Verification run locally:

`uvx tox run -e py314-ci-bedrock -- tests/test_instrumentor.py::test_invoke_model_nova tests/openinference/instrumentation/bedrock/test_invoke_model_nova.py::TestNovaStreamCallbackUnit::test_streaming_tool_use_assembles_call_and_derives_total`

I also ran a synthetic callback probe for text-only and pure tool-use streams, which confirmed the missing attributes above.

python/instrumentation/openinference-instrumentation-bedrock/src/openinference/instrumentation/bedrock/_wrappers.py
```diff
@@ -39,6 +51,144 @@
...
+                    tool_input = raw_input
+            content_blocks.append(
+                {
+                    "toolUse": {
+                        "toolUseId": entry.get("id", ""),
+                        "name": entry.get("name", ""),
+                        "input": tool_input,
+                    }
+                }
+            )
```
caroger: Mixed Nova streams can interleave text and tool-use blocks by `contentBlockIndex`. Aggregating all text in `_text_chunks` and then appending all tool calls moves any text that arrives after a tool call ahead of that tool call in both `output.value` and `llm.output_messages`. That breaks downstream chat/tool rendering for streams shaped like `text -> toolUse -> text`.

This keeps the same emitted attributes but stores the streamed state by content block and materializes the final content in provider order:

```suggestion
        self._content_blocks: Dict[int, Dict[str, Any]] = {}
        body = request.get("body", {})
        span.set_attribute(OPENINFERENCE_SPAN_KIND, LLM)
        span.set_attribute(LLM_PROVIDER, AWS)
        if model_id := request.get("modelId"):
            span.set_attribute(LLM_MODEL_NAME, str(model_id))
        if messages := body.get("messages"):
            span.set_attribute(INPUT_VALUE, safe_json_dumps(messages))
            span.set_attribute(INPUT_MIME_TYPE, JSON)
        if inference_config := body.get("inferenceConfig"):
            span.set_attribute(LLM_INVOCATION_PARAMETERS, safe_json_dumps(inference_config))
        input_messages = _build_nova_input_messages(body)
        if input_messages:
            span.set_attributes(get_llm_attributes(input_messages=input_messages))
        tools = _build_nova_tools(body)
        if tools:
            span.set_attributes(get_llm_tool_attributes(tools))

    def __call__(self, obj: Any) -> Any:
        span = self._span
        if isinstance(obj, dict):
            if "chunk" in obj and "bytes" in obj["chunk"]:
                try:
                    payload = json.loads(obj["chunk"]["bytes"])
                    self._process_payload(payload)
                except Exception:
                    pass
        elif isinstance(obj, (StopIteration, StopAsyncIteration)):
            self._finalize_output()
            _finish(span, None, {})
        elif isinstance(obj, BaseException):
            _finish(span, obj, {})
        return obj

    def _content_block(self, index: int) -> Dict[str, Any]:
        return self._content_blocks.setdefault(index, {})

    def _process_payload(self, payload: Mapping[str, Any]) -> None:
        span = self._span
        if (start_event := payload.get("contentBlockStart")) and isinstance(start_event, dict):
            index = start_event.get("contentBlockIndex")
            tool_use = (start_event.get("start") or {}).get("toolUse")
            if isinstance(index, int) and isinstance(tool_use, dict):
                self._content_block(index)["toolUse"] = {
                    "toolUseId": tool_use.get("toolUseId", ""),
                    "name": tool_use.get("name", ""),
                    "input": "",
                }
        if (delta_event := payload.get("contentBlockDelta")) and isinstance(delta_event, dict):
            index = delta_event.get("contentBlockIndex")
            delta = delta_event.get("delta") or {}
            if isinstance(index, int):
                block = self._content_block(index)
                if text := delta.get("text"):
                    block["text"] = f"{block.get('text', '')}{text}"
                elif isinstance(tool_use_delta := delta.get("toolUse"), dict):
                    tool_use = block.setdefault(
                        "toolUse", {"toolUseId": "", "name": "", "input": ""}
                    )
                    if (fragment := tool_use_delta.get("input")) is not None:
                        tool_use["input"] += fragment if isinstance(fragment, str) else str(fragment)
        if metadata := payload.get("metadata"):
            usage = metadata.get("usage", {}) if isinstance(metadata, dict) else {}
            input_tokens = usage.get("inputTokens") if isinstance(usage, dict) else None
            output_tokens = usage.get("outputTokens") if isinstance(usage, dict) else None
            if input_tokens is not None:
                span.set_attribute(LLM_TOKEN_COUNT_PROMPT, input_tokens)
            if output_tokens is not None:
                span.set_attribute(LLM_TOKEN_COUNT_COMPLETION, output_tokens)
            # Nova streaming metadata omits ``totalTokens``; derive it when possible
            # so consumers don't have to reconstruct the sum themselves.
            total = usage.get("totalTokens") if isinstance(usage, dict) else None
            if total is None and input_tokens is not None and output_tokens is not None:
                total = input_tokens + output_tokens
            if total is not None:
                span.set_attribute(LLM_TOKEN_COUNT_TOTAL, total)
            if isinstance(usage, dict):
                if (v := usage.get("cacheReadInputTokenCount")) is not None:
                    span.set_attribute(LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ, int(v))
                if (v := usage.get("cacheWriteInputTokenCount")) is not None:
                    span.set_attribute(LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, int(v))

    def _finalize_output(self) -> None:
        span = self._span
        content_blocks: list[dict[str, Any]] = []
        for idx in sorted(self._content_blocks):
            block = self._content_blocks[idx]
            if text_output := block.get("text"):
                content_blocks.append({"text": text_output})
            if tool_use := block.get("toolUse"):
                raw_input = tool_use.get("input", "")
                tool_input: Any = raw_input
                if isinstance(raw_input, str) and raw_input:
                    try:
                        tool_input = json.loads(raw_input)
                    except Exception:
                        tool_input = raw_input
                content_blocks.append(
                    {
                        "toolUse": {
                            "toolUseId": tool_use.get("toolUseId", ""),
                            "name": tool_use.get("name", ""),
                            "input": tool_input,
                        }
                    }
                )
```

caroger (approved):
Follow-up review looks good to me.

The prior Nova streaming ordering issue is addressed at the current head. I verified the mixed text/tool/text streaming path now preserves `contentBlockIndex` order in `output.value`, and the targeted Bedrock checks pass locally.

PR #3060, opened 2026-05-06
fix(litellm): preserve CustomStreamWrapper type in async wrapper

aaoobd3:
# fix(litellm): preserve `CustomStreamWrapper` type in async wrapper

## Context

`_acompletion_wrapper` in
`src/openinference/instrumentation/litellm/__init__.py` wraps a streaming
result through `_finalize_streaming_span`, an `async def ... yield` helper.
Calling that helper returns a plain Python `async_generator` — not the
`litellm.CustomStreamWrapper` that the underlying call produced.

That breaks `isinstance(response, CustomStreamWrapper)` checks downstream.
The most immediate impact is in litellm's Responses-API → Chat-Completions
bridge:

```python
# litellm/responses/litellm_completion_transformation/handler.py
litellm_completion_response = await litellm.acompletion(**acompletion_args)
if isinstance(litellm_completion_response, ModelResponse):
    ...
elif isinstance(litellm_completion_response, litellm.CustomStreamWrapper):
    ...
raise ValueError(f"Unexpected response type: {type(...)}")
```

When `litellm.aresponses(stream=True)` is called against a non-native
provider (Gemini, Anthropic, Bedrock, Vertex), the bridge calls
`await litellm.acompletion(stream=True)`. With this instrumentor active, the
wrapped `acompletion` returns an `async_generator`, the bridge's `isinstance`
check fails, and we get:

```
ValueError: Unexpected response type: <class 'async_generator'>
```

Reproduction:

```python
import asyncio, litellm
from openinference.instrumentation.litellm import LiteLLMInstrumentor

LiteLLMInstrumentor().instrument()

async def main():
    await litellm.aresponses(
        model="gemini/gemini-2.5-flash",
        input="hello",
        stream=True,
        api_key="...",
    )
    # raises ValueError: Unexpected response type: async_generator

asyncio.run(main())
```

The sync sibling (`_completion_wrapper`) is unaffected because it routes
through `_finalize_sync_streaming_span` — same shape but called via the sync
streaming path that doesn't go through the same bridge.

## Fix

Introduce `_AsyncInstrumentedCustomStreamWrapper`, a subclass of
`litellm.CustomStreamWrapper` that:

- delegates `__anext__` to the wrapped stream
- spies on each yielded chunk to populate OpenInference output-message and
  token-count attributes (same logic the previous `_finalize_streaming_span`
  performed)
- finalizes the span on `StopAsyncIteration` or any exception
- **preserves `isinstance(stream, CustomStreamWrapper)` for downstream
  consumers** — this is the load-bearing change

The subclass is built lazily on first use to avoid importing
`litellm.litellm_core_utils.streaming_handler` at module load.

## Tests

Adds `test_acompletion_stream_preserves_custom_stream_wrapper_type` in
`tests/test_instrumentor.py`. The test asserts:

1. `await litellm.acompletion(..., stream=True)` returns an
   `isinstance(., CustomStreamWrapper)` while the instrumentor is active.
2. The stream is iterable and produces chunks.
3. Span finalization still happens on iteration completion (`acompletion`
   span is recorded).

Local verification — direct probe with the fix applied:

```
type returned: _AsyncInstrumentedCustomStreamWrapper
isinstance(CustomStreamWrapper): True
chunks: 3
```

## Behavior change

None for code that just iterates the stream — same chunks, same span.
Behavior changes only for code that does an `isinstance` check on the
return value of `await litellm.acompletion(stream=True)` while the
instrumentor is active: it now correctly identifies the wrapper as a
`CustomStreamWrapper`, where before it incorrectly identified it as an
`async_generator`.

## Files

- `src/openinference/instrumentation/litellm/__init__.py` — fix +
  `_AsyncInstrumentedCustomStreamWrapper`
- `tests/test_instrumentor.py` — regression test

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
@@ -583,6 +583,131 @@ async def _finalize_streaming_span(span: trace_api.Span, stream: Any) -> Any:
         span.end()
 
 
+def _make_async_instrumented_stream_wrapper_class() -> Any:
+    """Lazily build a `CustomStreamWrapper` subclass used to wrap the result of
+    `await litellm.acompletion(stream=True)` while preserving its public type.
+
+    Why a subclass: the previous implementation passed the stream through
+    `_finalize_streaming_span` (an `async def ... yield` function), which
+    returned a bare `async_generator`. That broke litellm's
+    Responses->Chat-Completions bridge handler at
+    `litellm/responses/litellm_completion_transformation/handler.py`, whose
+    `isinstance(result, CustomStreamWrapper)` check then fell through to
+    `raise ValueError("Unexpected response type: ...")`.
+
+    By subclassing `CustomStreamWrapper`, the returned object is still a
+    `CustomStreamWrapper` for downstream consumers, while iteration is
+    intercepted to populate OpenInference span attributes.
+    """
+    from litellm.litellm_core_utils.streaming_handler import CustomStreamWrapper
+
+    class _AsyncInstrumentedCustomStreamWrapper(CustomStreamWrapper):  # type: ignore[misc]
+        # Don't call CustomStreamWrapper.__init__ — `wrapped` is fully
+        # initialized; we only need to expose its state on this instance.
+        def __init__(
+            self, span: trace_api.Span, wrapped: "CustomStreamWrapper"
+        ) -> None:
+            self.__dict__.update(wrapped.__dict__)
+            self._oi_span = span
+            self._oi_wrapped = wrapped
+            self._oi_output_messages: Dict[int, Dict[str, Any]] = {}
+            self._oi_usage_stats: Any = None
+            self._oi_finalized = False
+
+        def __aiter__(self) -> "_AsyncInstrumentedCustomStreamWrapper":
+            return self
+
+        async def __anext__(self) -> Any:
+            try:
+                token = await self._oi_wrapped.__anext__()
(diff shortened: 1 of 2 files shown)


aaoobd3:
CLA is signed but it shows as failed 
image

aaoobd3:

aaoobd3:
Follow-up commit e366c4f addresses two regressions found during a self-review pass against the previous behavior of the async def ... yield finalizer:

CancelledError handling — __anext__ previously caught Exception, but asyncio.CancelledError is BaseException in Python 3.8+. Now caught, span ends correctly on cancellation. The old async-generator finalizer handled this implicitly via its finally: block.
aclose() override — explicit consumer-side close (FastAPI streaming-response disconnect, anyio task-group cancellation) had no path to finalize the span. Now overridden with a try/finally.
Two new regression tests cover these:

test_acompletion_stream_finalizes_span_on_cancellation_in_anext
test_acompletion_stream_finalizes_span_on_aclose
Both pass alongside the original test_acompletion_stream_preserves_custom_stream_wrapper_type.

While reviewing I also confirmed the same root-cause bug (returning a generator/async-generator instead of preserving the upstream iterator type) exists in three more places in __init__.py: _completion_wrapper (sync, line 851), _aresponses_wrapper (line 825), and _responses_wrapper (sync, line 803). I left those out to keep this PR scoped to async completion as titled, but happy to extend the PR or split into follow-ups depending on what you'd prefer.

aaoobd3 (commented):
Looks good

caroger:
I'm closing this PR now because it doesn't have a linked issue. 

please see our contribution guide here - https://github.com/Arize-ai/openinference/blob/main/CONTRIBUTING#L21

PR #3081, opened 2026-05-11
feat: Introduce attributes for multi-modal content

nosnaws:
## Summary

Semantic conventions for multimodal content (image, audio, document) using a generic `message_content.file.*` namespace, plus a small set of supporting attributes for streaming and voice-conversation agents.

## What's added

- `message_content.type` — closed-enum discriminator: `text` | `image` | `audio` | `document`
- `message_content.file.url` / `mime_type` / `name` / `file_id` — shared across all non-text content
- `message_content.file.transcript` — applies when `type = "audio"`
- `llm.voice_name` — top-level invocation parameter for the requested TTS voice preset
- `openinference.end_reason` — top-level span-lifecycle outcome (`completed`, `interrupted_by_user`, `cancelled`, `session_closed`). Distinct from `llm.finish_reason`.
- `llm.first_token` span event — streaming TTFT marker (replaces an earlier attribute approach)
- `OPENINFERENCE_HIDE_INPUT_AUDIO` / `OPENINFERENCE_HIDE_INPUT_DOCUMENTS` env vars
- 12-month consumer support window for deprecated shapes

## What's deprecated

- `message_content.image.image.url` (double-nested) → `message_content.file.url` with `type = "image"`
- Span-level `audio.url` / `audio.mime_type` / `audio.transcript` → `message_content.file.*` with `type = "audio"`

## Provider payload references

- OpenAI Responses (image, audio): https://developers.openai.com/api/reference/resources/responses/methods/create
- OpenAI Realtime Events: https://developers.openai.com/api/reference/resources/realtime/server-events
- Anthropic content blocks: https://docs.anthropic.com/en/api/messages
- Gemini parts (inlineData / fileData): https://ai.google.dev/api/generate-content

## Follow-ups (not in this PR)

- Rename llama-index's informal `"First Token Stream Event"` to `llm.first_token` for spec compliance.


spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
 ```
 llm.input_messages.0.message.contents.1.message_content.type = "image"
-llm.input_messages.0.message.contents.1.message_content.image.image.url = "data:image/png;base64,iVBORw0KGgo..."
+llm.input_messages.0.message.contents.1.message_content.image.url = "https://example.com/image.jpg"
+llm.input_messages.0.message.contents.1.message_content.image.mime_type = "image/jpeg"
 ```
 
+The previous double-nested form `message_content.image.image.url` is **deprecated** but still accepted. New instrumentors should emit the single-nested shape.
+
 ### Audio Content
```
nosnaws, line 64: The following section describes how to handle audio content, but doesn't go deep on audio content for audio-to-audio applications. Is it worth adding another section that clearly goes into that topic?

spec/semantic_conventions.md
```diff
@@ -42,12 +42,14 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
 | `llm.invocation_parameters`                    | JSON string                 | `"{model_name: 'gpt-3', temperature: 0.7}"`                                                                                   | Parameters used during the invocation of an LLM or API                                                                                                                              |
 | `llm.finish_reason`                            | String                      | `"stop"`, `"length"`                                                                                                          | The reason the model stopped generating tokens.                                                                                                                                     |
+| `llm.end_reason`                               | String                      | `"completed"` or `"interrupted_by_user"`                                                                                      | The reason an LLM span ended. Applies to any LLM call, including voice-conversation rounds and agent-harness invocations. See [LLM End Reason Values](#llm-end-reason-values) for the enum. |
```
nosnaws: This has a distinction from `llm.finish_reason`, which covers why the model stopped generating tokens, `llm.end_reason` is from the observer's perspective so it's able to capture things like `interrupted_by_user`. 

Since it is from the observer's perspective, it may belong at the top level of span instead of nested under `llm`, which could also help clear up confusion between the two values.

spec/semantic_conventions.md
```diff
@@ -256,8 +280,17 @@ These attributes use a nested indexed format with discriminated union structure,
 For messages containing multiple content items (text, images, audio):
 
 - `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.text` - Text content item
```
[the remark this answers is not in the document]
nosnaws: Good catch, but terrible change suggestion. Will fix manually.

spec/multimodal_attributes.md
```diff
@@ -19,8 +19,9 @@ Where:
 
 Each content item has a `type` attribute that identifies its kind:
 - `"text"` - Text content
-- `"image"` - Image content (URL or base64)
-- `"audio"` - Audio content (URL or base64)
```
duncankmckinnon, line 23: Is it still URL or base64 for now?

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
+llm.output_messages.0.message.contents.0.message_content.type = "audio"
+llm.output_messages.0.message.contents.0.message_content.audio.url = "gs://voice-bucket/turn-1-out.mp3"
+llm.output_messages.0.message.contents.0.message_content.audio.mime_type = "audio/mpeg"
+llm.output_messages.0.message.contents.0.message_content.audio.transcript = "Hello! What's your confirmation number?"
+llm.output_messages.0.message.contents.0.message_content.audio.voice_id = "marin"
 ```
 
+The previous double-nested form `message_content.audio.audio.url` is **deprecated** but still accepted. New instrumentors should emit the single-nested shape. The span-level `audio.url` / `audio.mime_type` / `audio.transcript` attributes are likewise deprecated in favor of the content-block-level attributes shown above.
+
+### Document Content
```
duncankmckinnon: Besides PDFs, will we need to add handling for any type of document they may want to use separately in the backend (xlsx, doc, ppt etc.)? Just wondering about the scope here

spec/semantic_conventions.md
```diff
@@ -83,9 +85,18 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
-| `message_content.image`                        | Image Object                | `{"image.url": "https://sample-link-to-image.jpg"}`                                                                           | The image content of the message, if the type is "image".                                                                                                                           |
+| `message_content.image`                        | Image Object                | `{"image.url": "https://sample-link-to-image.jpg"}`                                                                           | **Deprecated** — use the single-nested `message_content.image.url` / `message_content.image.mime_type` attributes instead. The double-nested `message_content.image.image.url` shape is still accepted but will be phased out. |
```
duncankmckinnon: These attributes are 1:1 mapping from message_content.type, right - so naming conventions for type attributes needs to be consistent and validated.

spec/semantic_conventions.md
```diff
@@ -253,11 +277,20 @@ These attributes use a nested indexed format with discriminated union structure,
...
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.type` - Content type ("text", "image", "audio", "document")
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.image.url` - Image URL
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.image.mime_type` - Image MIME type
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.image.image.url` - **Deprecated** double-nested image URL; still accepted, prefer `message_content.image.url`
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.audio.url` - Audio URL
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.audio.mime_type` - Audio MIME type
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.audio.transcript` - Audio transcript
+- `llm.input_messages.<messageIndex>.message.contents.<contentIndex>.message_content.audio.voice_id` - Provider-defined voice identifier (assistant audio outputs)
```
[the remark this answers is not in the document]
duncankmckinnon: I think it's correct to have the option of voice_id on input as well as output - we want to remain flexible to different patterns of invocation, so that either the LLM or the person can send the request.

duncankmckinnon (commented):
Just a few qqs on these patterns but looks good!

spec/multimodal_attributes.md
nosnaws: Generally, we want URLs, but backends such as Arize AX can support data URLs.

spec/multimodal_attributes.md
nosnaws: I used the word `document` to leave that open for the future, but at the moment it would just be PDFs. Worth adding into the spec?

spec/semantic_conventions.md
nosnaws: You mean that for a message_content to have an `audio` attribute, the `message_content.type` must be `audio`? Yes that is correct.
We could add a callout to each field saying it requires a particular `type`, would that work?

mikeldking (commented):
Spec-only change — adds image/audio/document content-block attributes, deprecates the double-nested `message_content.image.image.url` and span-level `audio.*` forms, and introduces `llm.end_reason` (+ enum) and `latency.streaming.time_to_first_response_ms`. A few inquisitive questions / concerns before this lands:

**General**
- Large semantic surface for an "XS" PR — consider splitting into (1) image/audio/document content blocks + deprecation, (2) `llm.end_reason`, (3) `latency.streaming.*`. Each has independent review concerns.
- No links to provider docs. For each new shape, please link how it's actually obtained from foundation-model providers (OpenAI chat image parts, Anthropic content blocks, Gemini `inlineData`/`fileData`, OpenAI Realtime audio, etc.) — reviewers can't validate the modeling without seeing the source payloads.

**1. Deprecation clarity — `message_content.image.image.url`**
- "Deprecated but still accepted" — accepted by whom? The Phoenix UI reader, instrumentor authors, or both? Spell out the consumer contract.
- Is there a migration timeline / target removal version? "Will be phased out" with no horizon tends to mean never.
- Why was it double-nested originally, and is the single-nested form confirmed compatible with existing Phoenix rendering?

**2. URLs assume public accessibility**
- The examples (`https://example.com/image.jpg`, `gs://...`) assume a fetchable URL, which is the rare case — most provider payloads carry base64 inline data or opaque file IDs.
- What's the canonical representation for inline base64? The old text said "URL or base64" and showed `data:image/png;base64,...`; the new text dropped that. Is a `data:` URI still the answer for `image.url`/`audio.url`/`document.url`? If so, say so explicitly with an example.

**3. Purpose of `mime_type` on these blocks**
- For a URL, MIME type is generally only inferable from the extension anyway — what does carrying it explicitly buy us? Is the intent that it's only meaningful when the payload is inline bytes (base64), where there's no extension? If so, document that.
- Is `mime_type` REQUIRED or OPTIONAL on each block? The table and prose disagree (image prose implies required, audio table implies optional).

**4. Redaction / privacy**
- This adds `audio.url`, `audio.transcript`, `document.url`, `document.name` etc. but doesn't extend the Privacy Considerations section. There's an existing "Hiding Images" mechanism — what's the equivalent for audio (the transcript IS the message content), documents, and the new URL fields? Needs an `OPENINFERENCE_HIDE_*` story or an explicit statement that existing flags cover it.
- `document.name` can itself be sensitive — should it be redactable?

**5. `voice_id`**
- The name isn't self-describing. Is it the TTS voice preset (`alloy`, `marin`)? If so, clarify in the description that it's a *provider preset name*, not a speaker/diarization id or user identity — and consider `audio.voice` / `audio.voice_name`.
- Why nested under `message_content.audio` rather than an invocation parameter (it's a request-side knob)? On *input* audio blocks it makes no sense, yet the doc lists `message_content.audio.voice_id` under `llm.input_messages.*`. Should it be output-only?

**6. `document` content type — why separate semantics?**
- A document handed to an LLM is conceptually just an attachment, same as an image or audio asset. The fields are `url` + `mime_type` + `name` — identical to what image/audio need. Why three near-identical parallel namespaces rather than one generic "file/attachment" block discriminated by `mime_type`? At minimum the PR should justify the duplication.
- If we keep `document`, why does only `document` get a `name` field and not `image`/`audio`? Providers attach filenames to all of them.

**7. Provider File IDs section**
- "Instrumentors MUST drop these file IDs" is a strong normative requirement in a prose paragraph with zero examples. Please show: (a) the provider payload containing a file ID, (b) the resulting OI attributes (block with `mime_type`, no `url`).
- "MUST drop" loses information users want (which file did the model see?). Is dropping really better than a clearly-namespaced `message_content.<type>.file_id`? The section says "a future revision may introduce" it — why not now, since you're already touching this surface? Drop-then-add-later is itself a breaking change for consumers.
- How does an instrumentor robustly distinguish "opaque provider file ID" from a URL? `gs://` / `s3://` are also not Phoenix-fetchable but aren't file IDs. The MUST needs a definition of what counts.

**8. `latency.streaming.time_to_first_response_ms`**
- This feels like it belongs on a span event (a `first_token`/`first_byte` event with a timestamp), not a scalar attribute — events naturally carry timestamps. Why an attribute?
- Capturing it is hard — the instrumentor must record request-start and observe the first streamed chunk. Please add a strawman showing *how* an instrumentor would populate this (wrap the stream iterator? hook the transport?).
- "for non-streaming responses this is equivalent to the span duration" — then why emit it at all in that case? That just duplicates span duration and invites double-counting. Recommend streaming-only, absent otherwise.
- Naming: is `latency.streaming.*` a new namespace with a plan for more keys, or a one-off? If one-off, `llm.time_to_first_token_ms` (a well-known term) may be clearer. Also "first byte ... reaching the client" is ambiguous for SSE/token streams — TTFT is the usual concept.

**9. `llm.end_reason` vs `llm.finish_reason`**
- Easy to confuse; the doc doesn't contrast them. Add an explicit note: `finish_reason` = why the model stopped generating tokens (provider-reported: `stop`/`length`/`tool_calls`); `end_reason` = why the OI span ended (lifecycle: completed/interrupted/cancelled/error/timeout/session_closed).
- Overlap with OTel span status: `end_reason=error`/`timeout` seems redundant with span `status=ERROR`. Consider dropping those two and keeping `end_reason` for things span status can't express (`interrupted_by_user`, `cancelled`, `session_closed`).
- The enum description references "the native path" / "the cascaded path" / "round CHAIN span" — undefined anywhere in this PR or (apparently) the spec. Either define voice-conversation span topology first, or drop those parentheticals.
- It's added to the "reserved and MUST be supported" table — does every LLM span now have to emit it? Please clarify it's optional-but-reserved.

**Nits**
- `audio.voice_id` example `"marin"` — if illustrative, maybe use a documented voice (`alloy`, `verse`) so readers don't read it as normative.
- Good catch on the trailing-newline fix at EOF of `multimodal_attributes.md`.

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
-llm.input_messages.0.message.contents.1.message_content.image.image.url = "https://example.com/image.jpg"
-```
+Image content blocks use the single-nested `message_content.image.url` shape along with a MIME type:
 
-For base64-encoded images:
 ```
 llm.input_messages.0.message.contents.1.message_content.type = "image"
-llm.input_messages.0.message.contents.1.message_content.image.image.url = "data:image/png;base64,iVBORw0KGgo..."
+llm.input_messages.0.message.contents.1.message_content.image.url = "https://example.com/image.jpg"
+llm.input_messages.0.message.contents.1.message_content.image.mime_type = "image/jpeg"
```
mikeldking: The example assumes a publicly fetchable URL, which in practice will be the *rare* case — most provider payloads are base64 inline data or opaque file IDs. The old `data:image/png;base64,...` example was removed here; is a `data:` URI still the canonical representation for inline bytes in `image.url`? Please keep an explicit example.

Also: what is `mime_type` for? For a URL it's only inferable from the extension anyway, so it seems only meaningful when the payload is inline bytes (no extension). Is that the intent? And is `mime_type` REQUIRED or OPTIONAL on the block? The prose here implies required; the attribute table implies optional.

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
 ```
 
+The previous double-nested form `message_content.image.image.url` is **deprecated** but still accepted. New instrumentors should emit the single-nested shape.
```
mikeldking: "Deprecated but still accepted" — accepted by whom? The Phoenix UI reader, instrumentor authors, or both? Spell out the consumer contract, and give a migration timeline / target removal version ("will be phased out" with no horizon tends to mean never). Also worth a sentence on why it was double-nested originally and confirmation the Phoenix renderer already handles the single-nested shape.

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
+llm.input_messages.0.message.contents.2.message_content.audio.transcript = "Hello, how are you?"
+```
+
+For assistant audio outputs, include `voice_id` when known:
+```
+llm.output_messages.0.message.contents.0.message_content.type = "audio"
+llm.output_messages.0.message.contents.0.message_content.audio.url = "gs://voice-bucket/turn-1-out.mp3"
+llm.output_messages.0.message.contents.0.message_content.audio.mime_type = "audio/mpeg"
+llm.output_messages.0.message.contents.0.message_content.audio.transcript = "Hello! What's your confirmation number?"
+llm.output_messages.0.message.contents.0.message_content.audio.voice_id = "marin"
```
mikeldking: `voice_id` isn't self-describing. Is this the TTS voice preset (`alloy`, `marin`)? If so, `audio.voice` / `audio.voice_name` reads better, and the description should make clear it's a *provider preset name* — not a speaker/diarization id, not a user identity.

Also: why is it nested under `message_content.audio` rather than being an invocation parameter (it's a request-side knob for outputs)? On *input* audio blocks `voice_id` makes no sense at all, yet `message_content.audio.voice_id` is listed under `llm.input_messages.*` elsewhere in this PR. Should it be output-only?

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
 ```
 
+The previous double-nested form `message_content.audio.audio.url` is **deprecated** but still accepted. New instrumentors should emit the single-nested shape. The span-level `audio.url` / `audio.mime_type` / `audio.transcript` attributes are likewise deprecated in favor of the content-block-level attributes shown above.
+
+### Document Content
+
+Document content blocks (PDFs and similar) use `message_content.document.url` along with a MIME type and an optional filename:
```
mikeldking: Why does `document` need its own semantics? A document handed to an LLM is conceptually just an attachment — same as an image or audio asset — and the fields here (`url` + `mime_type` + `name`) are identical to what image/audio need. Why three near-identical parallel namespaces rather than one generic "file/attachment" content block discriminated by `mime_type`? At minimum the PR should justify the duplication. And if we keep `document`: why does only `document` get a `name` field and not `image`/`audio`? Providers attach filenames to all of them.

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
+```
+llm.input_messages.0.message.contents.1.message_content.type = "document"
+llm.input_messages.0.message.contents.1.message_content.document.url = "gs://docs-bucket/policies/travel-policy.pdf"
+llm.input_messages.0.message.contents.1.message_content.document.mime_type = "application/pdf"
+llm.input_messages.0.message.contents.1.message_content.document.name = "travel-policy.pdf"
+```
+
+## Provider File IDs
+
+Some providers (OpenAI Files API, Anthropic Files API, Gemini File API, and similar) expose opaque file IDs that reference assets stored on the provider's side. Instrumentors MUST drop these file IDs rather than emit them — they should not be placed in the `url` field (which would mis-type the value) nor under a vendor namespace. When a provider call is made with a file ID and no fetchable URL is available, emit the content block with `mime_type` (and `transcript`, where applicable) and omit `url`. A future revision of this spec may introduce dedicated `file_id` fields; until then, dropping is the expected behavior.
```
mikeldking: This is a strong normative requirement ("Instrumentors MUST drop these file IDs") buried in a prose paragraph with zero examples. Please show: (a) the provider payload that contains a file ID, (b) the resulting OI attributes (block with `mime_type`, no `url`).

Also — "MUST drop" discards information users genuinely want (which file did the model see?). Is dropping really better than a clearly-namespaced `message_content.<type>.file_id`? The section says a future revision "may introduce" it — why not now, since you're already reworking this surface? Drop-then-add-later is itself a breaking change for consumers. And: how does an instrumentor robustly distinguish an opaque provider file ID from a URL? `gs://` / `s3://` URIs are also not Phoenix-fetchable but aren't file IDs — the MUST needs a definition of what counts.

spec/multimodal_attributes.md
```diff
@@ -31,24 +32,53 @@ llm.input_messages.0.message.contents.0.message_content.text = "What is in this
...
+
 ## Privacy Considerations
 
 ### Hiding Images
```
mikeldking, line 118: Privacy Considerations isn't updated for the new fields. This PR adds `audio.url`, `audio.transcript`, `document.url`, `document.name` — there's an existing "Hiding Images" mechanism, but no equivalent for audio (the transcript literally *is* the message content), documents, or the new URL fields. Needs an `OPENINFERENCE_HIDE_*` story or an explicit statement that existing flags already cover them. `document.name` can itself be sensitive (`2025-termination-letter-jane-doe.pdf`) — should it be redactable?

spec/semantic_conventions.md
```diff
@@ -42,12 +42,14 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
 | `input.value`                                  | String                      | `"{'query': 'What is the weather today?'}"`                                                                                   | The input value to an operation                                                                                                                                                     |
+| `latency.streaming.time_to_first_response_ms`  | Number                      | `100`                                                                                                                         | Time, in milliseconds, from the start of the request to the first byte of the model's response reaching the client. For non-streaming responses this is equivalent to the span duration; for streaming responses (including streamed audio in voice conversations) this measures how long the client waited before receiving any output. |
```
mikeldking: A few concerns with `latency.streaming.time_to_first_response_ms`:
- This feels like it belongs on a span *event* (a `first_token`/`first_byte` event with a timestamp), not a scalar span attribute — events naturally carry timestamps and compose with streaming. Why an attribute?
- Capturing it is non-trivial: the instrumentor must record request-start and observe the first streamed chunk. Please add a strawman showing *how* an instrumentor would actually populate this (wrap the stream iterator? hook the transport?).
- "for non-streaming responses this is equivalent to the span duration" — then why emit it at all in that case? It just duplicates span duration and invites double-counting in dashboards. Recommend streaming-only, absent otherwise.
- Naming: is `latency.streaming.*` a new top-level namespace with a roadmap of more keys, or a one-off? If one-off, `llm.time_to_first_token_ms` (a well-known term) is clearer. "first byte ... reaching the client" is also ambiguous for SSE/token streams — TTFT is the usual concept.

spec/semantic_conventions.md
mikeldking: How does `llm.end_reason` differ from `llm.finish_reason`? The two are easy to confuse and the doc doesn't contrast them. Add an explicit note: `finish_reason` = why the *model* stopped generating tokens (provider-reported: `stop`/`length`/`tool_calls`); `end_reason` = why the *OI span* ended (lifecycle). Also: `end_reason=error`/`timeout` overlaps with OTel span `status=ERROR` — consider dropping those two values and keeping `end_reason` for what span status can't express (`interrupted_by_user`, `cancelled`, `session_closed`). And it's been added to the "reserved and MUST be supported" table — does every LLM span now have to emit it? Please clarify it's optional-but-reserved.

spec/semantic_conventions.md
```diff
@@ -83,9 +85,18 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
+| `message_content.image`                        | Image Object                | `{"image.url": "https://sample-link-to-image.jpg"}`                                                                           | **Deprecated** — use the single-nested `message_content.image.url` / `message_content.image.mime_type` attributes instead. The double-nested `message_content.image.image.url` shape is still accepted but will be phased out. |
+| `message_content.image.url`                    | String                      | `"https://sample-link-to-image.jpg"`                                                                                          | The link to the image content. Replaces the double-nested `message_content.image.image.url` shape.                                                                                  |
```
mikeldking: For a URL, the MIME type is generally only inferable from the extension — what does carrying `image.mime_type` explicitly buy us beyond the inline-bytes case? And is it REQUIRED or OPTIONAL? Same question applies to `audio.mime_type` / `document.mime_type` below.

spec/semantic_conventions.md
```diff
@@ -219,6 +230,19 @@ The `llm.cost` prefix is used to group cost-related attributes. When these keys
 }
 ```
 
+### LLM End Reason Values
+
+The `llm.end_reason` attribute identifies how an LLM span ended. It applies to any LLM call — voice-conversation rounds (set on the round `LLM` span on the native path or the round `CHAIN` span on the cascaded path), child STT / thinker / TTS spans within a cascaded voice round, and ordinary single-shot LLM invocations (including agent-harness calls where a user may cancel mid-generation). The following values are well-known; if one of them applies, the respective value MUST be used.
```
mikeldking: This description references "the native path", "the cascaded path", "the round `CHAIN` span", and "child STT / thinker / TTS spans" — none of which are defined in this PR or (as far as I can tell) anywhere in the spec. Either define the voice-conversation span topology somewhere first, or drop these parentheticals; as written they're unintelligible to anyone not on the voice project.

mikeldking (commented):
Inline questions on the multi-modal additions — mostly modeling/clarity concerns. (See line comments.)

spec/semantic_conventions.md
nosnaws: I like `llm.time_to_first_token_ms`, more than willing to get that changed and would prefer to use well known terms. 
For the thought on using a span event, is that something we use for any current open inference attributes? I didn't consider it because I hadn't seen that used at all. If so, can you point me to an example?

spec/semantic_conventions.md
nosnaws: I like reducing what it represents to remove overlap, will get it adjusted. Will clarify that's it's optional-but-reserved.

nosnaws:
> No links to provider docs. For each new shape, please link how it's actually obtained from foundation-model providers (OpenAI chat image parts, Anthropic content blocks, Gemini inlineData/fileData, OpenAI Realtime audio, etc.) — reviewers can't validate the modeling without seeing the source payloads.

@mikeldking We don't link out to provider documentation in the semantic conventions today, do you mean just getting links in the PR description to help with reviews?

spec/multimodal_attributes.md
nosnaws: That was an overzealous change, I'll get that example back to the way it was.

For `mime_type`, the base64 case is still valid, so it helps out in that case. For a regular URL case, I agree it's less useful. However, it's very useful for other formats where the mime type gives you more information, such as audio. We could drop it for images, but I prefer being uniform across types.

spec/multimodal_attributes.md
nosnaws: Agreed on `voice_id` will get that updated and add clearer documentation.

For the nesting, I wanted it as close to the actual artifact that was generated as possible, but I do agree that invocation parameters sounds like the right spot. @duncankmckinnon I know you liked that it was able to exist on input and output, what do you think about moving it to the invocation parameters?

spec/multimodal_attributes.md
nosnaws: Agreed on the filename, we should extend that to all types.

For the different parallel namespaces, I was trying to conform to the existing handling for images, but I'm on board with moving to a generic "file" namespace. My only hesitation is that we then have fields such as "transcript" that are dependent on the `mime_type` vs there just being different fields on different namespaces. @mikeldking If you aren't worried about them being dependent on the `mime_type`, then I'm fully on board with changing to a generic `file` namespace.

spec/semantic_conventions.md
nosnaws: It does look like we use a span event for literally the same information in the Llama Index instrumentor. I'm going to move this to be an optional span event.

spec/multimodal_attributes.md
nosnaws: Got this covered, thanks for the catch!

spec/semantic_conventions.md
```diff
@@ -219,6 +225,35 @@ The `llm.cost` prefix is used to group cost-related attributes. When these keys
...
+
+## Deprecation Policy
+
+Attributes and shapes that are renamed, restructured, or replaced are marked **Deprecated** in this spec, with a pointer to the replacement. Deprecated shapes are subject to the following support window:
+
+- **Instrumentors** — SHOULD migrate to the replacement shape promptly. New instrumentation MUST emit the replacement shape.
+- **Consumers** (backends, UIs, dashboards) — MUST continue to accept the deprecated shape for at least **12 months** after the deprecation marker first appears in a tagged spec release. Removal from a consumer before that window is a breaking change.
+- **Spec** — MAY remove deprecated entries after the 12-month support window has elapsed. Spec changes that drop a deprecated entry SHOULD reference the release in which the deprecation was introduced.
+
+The 12-month window starts at the spec release that introduces the deprecation marker, not the release that introduces the replacement (these may differ when a replacement ships before its predecessor is formally deprecated).
```
nosnaws, line 248: I tried to formalize deprecations here, let me know what y'all think.

spec/multimodal_attributes.md
nosnaws: Added this bit of text back in to be clear.

spec/semantic_conventions.md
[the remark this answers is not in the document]
nosnaws: Moved this to be part of the invocation parameters since it's mainly an argument the user makes to the LLM model.

spec/multimodal_attributes.md
nosnaws: This should be clearly called out now!

spec/semantic_conventions.md
nosnaws: I think I've captured this.

spec/multimodal_attributes.md
nosnaws: I've made some changes to get this whole thing more uniform.

spec/multimodal_attributes.md
nosnaws: I added a larger section on deprecations, but would love some guidance on this and how we've done deprecations in past (if we have).

spec/multimodal_attributes.md
nosnaws: Moved into invocation_parameters.

spec/multimodal_attributes.md
nosnaws: I've moved to a generic `file` namespace. Let me know what you think!

spec/multimodal_attributes.md
nosnaws: I completely agree on the file ids, so I've updated the spec to include them as optional. That way user's won't have to deal with the "add later" case.

spec/semantic_conventions.md
nosnaws: I've reduced the enum options so that it doesn't overlap and I've moved it to the top level of the span, since it's really a span value, not an LLM value.

spec/semantic_conventions.md
nosnaws: While I agree it doesn't add much when talking about images, but for audio it can convey a lot of information such as sampling rate.
I've updated things to know only have a `file` namespace, making the mime_type optional.

spec/semantic_conventions.md
nosnaws: Good call out, I've got the language clean up.

spec/multimodal_attributes.md
```diff
@@ -29,32 +31,108 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
...
-- Image URLs in input messages will be replaced with `"__REDACTED__"`
+- All fields of input content blocks with `type = "image"` will be replaced with `"__REDACTED__"`
+- This only applies when input messages are not already completely hidden
+
+### Hiding Audio
+
+When `OPENINFERENCE_HIDE_INPUT_AUDIO` is set to true:
+- All fields of input content blocks with `type = "audio"` will be replaced with `"__REDACTED__"`
+- The transcript is particularly sensitive because it *is* the rendered message content for voice conversations
+- This only applies when input messages are not already completely hidden
```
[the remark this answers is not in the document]
nosnaws: This is a good comment, I think it's worth providing a output flag for each of the multi-modal types.

spec/semantic_conventions.md
```diff
@@ -83,10 +84,15 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
+| `message_content.type`                         | String                      | `"text"`, `"image"`, `"audio"`, or `"document"`                                                                               | The semantic category of the content. Closed enum of well-known values; new values require a spec update. Discriminates which `file.*` fields apply.                                |
+| `message_content.text`                         | String                      | `"This is a sample text"`                                                                                                     | The text content of the message, when `message_content.type` is `"text"`.                                                                                                           |
+| `message_content.file.url`                     | String                      | `"https://example.com/image.jpg"`                                                                                             | URL of the file asset, or a `data:` URI carrying inline base64 bytes. Applies to any non-text `message_content.type`.                                                               |
```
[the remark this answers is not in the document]
nosnaws: Fixed!

spec/multimodal_attributes.md
```diff
@@ -29,32 +31,108 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
...
+
+### Hiding Documents
+
+When `OPENINFERENCE_HIDE_INPUT_DOCUMENTS` is set to true:
+- All fields of input content blocks with `type = "document"` will be replaced with `"__REDACTED__"`
+- `file.name` may itself be sensitive (e.g., filenames that disclose the subject matter)
 - This only applies when input messages are not already completely hidden
```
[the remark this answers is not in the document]
nosnaws: Done!

spec/multimodal_attributes.md
```diff
@@ -29,32 +31,108 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
...
 
 ### Base64 Image Truncation
```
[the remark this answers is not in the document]
nosnaws: Handled!

spec/configuration.md
```diff
@@ -9,20 +9,25 @@ The possible settings are:
...
 | OPENINFERENCE_HIDE_INPUT_MESSAGES            | Hides all input messages (independent of HIDE_INPUTS)                                                                          | bool | False   |
 | OPENINFERENCE_HIDE_OUTPUT_MESSAGES           | Hides all output messages (independent of HIDE_OUTPUTS)                                                                        | bool | False   |
 | OPENINFERENCE_HIDE_INPUT_IMAGES              | Hides images from input messages (only applies when input messages are not already hidden)                                     | bool | False   |
+| OPENINFERENCE_HIDE_OUTPUT_IMAGES             | Hides images from output messages (only applies when output messages are not already hidden)                                   | bool | False   |
+| OPENINFERENCE_HIDE_INPUT_AUDIO               | Hides audio content from input messages (only applies when input messages are not already hidden)                              | bool | False   |
+| OPENINFERENCE_HIDE_OUTPUT_AUDIO              | Hides audio content from output messages (only applies when output messages are not already hidden)                            | bool | False   |
```
mikeldking, line 19: you probably need to demarcate the difference between auto content and audio transcription. Or are both controlled together?

spec/configuration.md
```diff
@@ -9,20 +9,25 @@ The possible settings are:
...
+| OPENINFERENCE_HIDE_INPUT_DOCUMENTS           | Hides document content from input messages (only applies when input messages are not already hidden)                           | bool | False   |
+| OPENINFERENCE_HIDE_OUTPUT_DOCUMENTS          | Hides document content from output messages (only applies when output messages are not already hidden)                         | bool | False   |
```
mikeldking, line 21: this naming is confusing because we have document semantic conventions and it relates to document retrieval - we have to be extremely precise here as otherwise it's not clear what we are talking about.

I'm not sure if this is explicitly related to responses. if that's the case maybe I misunderstood and we should converge on files: https://developers.openai.com/api/docs/guides/file-inputs

spec/multimodal_attributes.md
```diff
@@ -29,47 +31,162 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
 llm.input_messages.0.message.contents.0.message_content.text = "What is in this image?"
 ```
 
-### Image Content
+### File Content (Image, Audio, Document)
```
mikeldking, line 34: this is confusing to me as phrasing as each one of these are different "parts" of content bodies.

curl "https://api.openai.com/v1/responses" \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $OPENAI_API_KEY" \
    -d '{
        "model": "gpt-5.5",
        "input": [
            {
                "role": "user",
                "content": [
                    {
                        "type": "input_text",
                        "text": "Analyze the letter and provide a summary of the key points."
                    },
                    {
                        "type": "input_file",
                        "file_url": "https://www.berkshirehathaway.com/letters/2024ltr.pdf"
                    }
                ]
            }
        ]
    }'

we should create a discriminated union on the parts and have a section

spec/multimodal_attributes.md
```diff
@@ -29,47 +31,162 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
...
+**Type-conditional fields**:
+
+- `file.transcript` — applies when `type = "audio"`; rendered text of the audio content (input or assistant output)
+
+#### Image
 
 ```
 llm.input_messages.0.message.contents.1.message_content.type = "image"
-llm.input_messages.0.message.contents.1.message_content.image.image.url = "https://example.com/image.jpg"
+llm.input_messages.0.message.contents.1.message_content.file.url = "https://example.com/image.jpg"
```
mikeldking, line 53: file conventions used here - not used in config - which is inconsistent

spec/multimodal_attributes.md
```diff
@@ -29,47 +31,162 @@ llm.input_messages.0.message.contents.0.message_content.type = "text"
...
+- `file.file_id` — optional opaque provider file ID; used when no fetchable URL is available
+
+**Type-conditional fields**:
+
+- `file.transcript` — applies when `type = "audio"`; rendered text of the audio content (input or assistant output)
+
+#### Image
 
 ```
 llm.input_messages.0.message.contents.1.message_content.type = "image"
```
mikeldking, line 52: wait, this is an image? I'm confused...

spec/multimodal_attributes.md
mikeldking, line 53: I wouldn't change how all of our conventions work. They are designed to un-nest cleanly into discriminated unions. Let's not break semantics that already work and adopted by backends.

mikeldking (changes requested):
Given how much this diverges from the files multi-part content of the main 3 providers I'm going to ask this to be broken up into parts:

1. files as multi-part content:
https://developers.openai.com/api/docs/guides/file-inputs
https://ai.google.dev/gemini-api/docs/file-input-methods
https://platform.claude.com/docs/en/build-with-claude/files

2. Adding auto as input output - and in that case what does a span represent 

https://developers.openai.com/api/docs/guides/file-inputs
https://ai.google.dev/gemini-api/docs/file-input-methods
https://platform.claude.com/docs/en/build-with-claude/files

nosnaws:
We've determined that general Multi-modal support needs more work. We are going to continue with an audio-only instrumentor for OpenAI realtime to prove things out before coming back with changes to the semantic conventions. That PR is [here](https://github.com/Arize-ai/openinference/pull/3173)

PR #3085, opened 2026-05-11
fix(agno): finalize workflow stream contexts without detach errors

### LINKED ISSUE #3078: [bug][agno] Same async-generator context leak (#2640 fix) still present in _workflow_wrapper.py (Workflow / Step / Parallel) on 0.1.32
**Describe the bug**

The async-generator context leak fixed for `_runs_wrapper.py` in PR #2681 / v0.1.32 (closing #2640 and #2917) is **still present** in `_workflow_wrapper.py`, affecting `_WorkflowWrapper`, `_StepWrapper`, and `_ParallelWrapper` when running with `stream=True`.

The exact same anti-pattern — `with trace_api.use_span(span, end_on_exit=False):` wrapping a body that contains `yield` — was rewritten in `_runs_wrapper.py` (Agent runs) but left untouched in `_workflow_wrapper.py` (Workflow / Step / Parallel runs).

When an Agno Workflow runs with `stream=True` (or `Step.execute_stream` / `Step.aexecute_stream`, or any `Parallel` streaming variant), every streamed response triggers:

```
ERROR opentelemetry.context - Failed to detach context
ValueError: <Token var=<ContextVar name='current_context' ...> at 0x...> was created in a different Context
```

This is the same root cause described in #2917 / #2640 / #2698: `use_span()` attaches a `ContextVar` token at entry; the `yield` cedes control to consumers (httpx, fastapi, anthropic, etc.) that perform their own attach/detach on the same ContextVar in non-LIFO order; when the generator finishes (or is closed via `aclose()` / `GeneratorExit`), `use_span.__exit__` calls `context.detach(token)` whose token no longer refers to the current head of the ContextVar stack.

**Affected file** (v0.1.32): `python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py`

**Affected locations (6):**

| Class | Method | Kind | Line |
|---|---|---|---|
| `_WorkflowWrapper` | `_run_stream_continue` | sync generator | 277 |
| `_WorkflowWrapper` | `_arun_stream_continue` | async generator | 424 |
| `_StepWrapper` | `_run_stream_continue` | sync generator | 571 |
| `_StepWrapper` | `_arun_stream_continue` | async generator | 704 |
| `_ParallelWrapper` | `_execute_stream_continue` | sync generator | 854 |
| `_ParallelWrapper` | `_aexecute_stream_continue` | async generator | 986 |

All six share this pattern (line numbers from v0.1.32):

```python
async def _arun_stream_continue(self, async_iter, span, workflow_token, instance=None):
    ...
    try:
        with trace_api.use_span(span, end_on_exit=False):     # ← attaches ContextVar token
            try:
                async for response in async_iter:
                    final_response = response
                    yield response                            # ← cedes control across attach/detach
            finally:
                if workflow_token:
                    try:
                        context_api.detach(workflow_token)
                    except Exception:
                        pass
        ...
```

This is exactly the pattern PR #2681 rewrote in `_runs_wrapper.py` to use manual `attach` + `detach_context_tokens(...)` in the outer `finally`.

**To Reproduce**

```python
import asyncio
from agno.agent import Agent
from agno.models.openai import OpenAIChat
from agno.workflow import Step, Workflow
from openinference.instrumentation.agno import AgnoInstrumentor

tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
AgnoInstrumentor().instrument()

agent = Agent(name="A", model=OpenAIChat(id="gpt-4o-mini"))
workflow = Workflow(name="W", steps=[Step(name="S", agent=agent)])

async def main():
    async for _event in await workflow.arun(input="hi", stream=True, stream_events=True):
        pass

asyncio.run(main())
```

**Observed behavior**

Every streamed event yields a `Failed to detach context — ValueError: <Token ...> was created in a different Context` log. Logs originate from `_workflow_wrapper.py` (frames vary depending on which of the six methods are on the stack).

**Expected behavior**

No `Failed to detach context` errors when consuming a Workflow / Step / Parallel stream — matching the behavior already achieved for Agent runs in v0.1.32.

**Suggested fix**

Apply the same restructure used by PR #2681 in `_runs_wrapper.py` to all six methods in `_workflow_wrapper.py`:

1. Replace the outer `with trace_api.use_span(span, end_on_exit=False):` around the `yield` loop with a manual `ctx_token = context_api.attach(trace_api.set_span_in_context(span))` at the start of the `try` block.
2. Move the `detach` of both `ctx_token` and the wrapper-specific token (`workflow_token` / `step_token` / `parallel_token`) into the outer `finally`, guarded by the same `initial_thread is current_thread` / `initial_task is current_task` check used by `detach_context_tokens(...)` from `_runs_wrapper.py`.

The `detach_context_tokens` helper already lives in `_runs_wrapper.py:264` and is shaped to be reused; promoting it to a shared utility (e.g. `utils.py`) and calling it from both wrappers keeps the fix consistent across the package.

**Environment**

- `agno`: 2.5.17 (also reproducible on 2.6.5 — same Workflow/Step API surface)
- `openinference-instrumentation-agno`: 0.1.32 (verified by unpacking the wheel from PyPI)
- `opentelemetry-api` / `opentelemetry-sdk`: 1.20.0+
- Python: 3.12
- OS: macOS (Darwin 25.4.0); also reproduced under Linux via Docker

**Workaround in use**

Until a fix lands, we maintain a local monkey-patch that re-implements all four `run / arun / run_stream / arun_stream` of `_RunWrapper` with manual attach/detach. We will extend it to cover `_WorkflowWrapper`, `_StepWrapper`, and `_ParallelWrapper` while this issue is open.

**Related**

- PR #2681 — partial fix in `_runs_wrapper.py` (Agent runs only)
- Issue #2640 — root cause originally reported (closed as fixed)
- Issue #2917 — duplicate report including Teams-in-Workflows nesting (closed as duplicate)
- Issue #2698 — same `detach` error class

Genmin:
## Summary

- promotes the Agno stream context detach helper into `utils.py` so run and workflow wrappers share the same guarded detach behavior
- updates Workflow, Step, and Parallel streaming continuations to manually attach the span context around iteration instead of keeping `trace_api.use_span(...)` open across `yield`
- detaches the workflow/step/parallel setup token before returning async stream continuations, so the setup span context unwinds in order
- adds sync and async regression coverage for closing Workflow, Step, and Parallel streams from a different thread/task without OpenTelemetry detach errors

## Validation

- `python -m pytest tests/test_workflow_stream_context_finalization.py -q`
- `python -m pytest tests/test_async_stream_output_finalization.py tests/test_workflow_stream_context_finalization.py -q`
- `python -m pytest tests/test_team_context_isolation.py -k 'streamed_team_run_outer_span_remains_current or streamed_async_team_run_outer_span_remains_current' -q`
- `python -m ruff check src/openinference/instrumentation/agno/_workflow_wrapper.py src/openinference/instrumentation/agno/_runs_wrapper.py src/openinference/instrumentation/agno/utils.py tests/test_workflow_stream_context_finalization.py`
- `python -m compileall -q src/openinference/instrumentation/agno/_workflow_wrapper.py src/openinference/instrumentation/agno/_runs_wrapper.py src/openinference/instrumentation/agno/utils.py`
- `git diff --check`

Closes #3078.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
@@ -1,5 +1,3 @@
-import asyncio
-import threading
 from typing import (
     Any,
     Awaitable,
@@ -9,11 +7,6 @@ from typing import (
     Tuple,
 )
 
-from openinference.semconv.trace import (
-    OpenInferenceMimeTypeValues,
-    OpenInferenceSpanKindValues,
-    SpanAttributes,
-)
 from opentelemetry import context as context_api
 from opentelemetry import trace as trace_api
 from opentelemetry.util.types import AttributeValue
@@ -24,7 +17,11 @@ from openinference.instrumentation.agno.utils import (
     _bind_arguments,
     _flatten,
     _generate_node_id,
-    detach_context_tokens,
+)
+from openinference.semconv.trace import (
+    OpenInferenceMimeTypeValues,
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
 )
 
 
@@ -281,13 +278,23 @@ class _WorkflowWrapper:
     ) -> Any:
         """Continue streaming workflow with existing span"""
         final_response = None
-        ctx_token = None
-        workflow_token = None
-        initial_thread = threading.current_thread()
+        iterator = iter(iterator)
         try:
-            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
-            workflow_token = _setup_workflow_context(node_id)
-            for response in iterator:
+            while True:
+                ctx_token = None
+                workflow_token = None
+                try:
+                    ctx_token = context_api.attach(trace_api.set_span_in_context(span))
+                    workflow_token = _setup_workflow_context(node_id)
+                    response = next(iterator)
+                except StopIteration:
+                    break
+                finally:
+                    if workflow_token is not None:
+                        context_api.detach(workflow_token)
+                    if ctx_token is not None:
+                        context_api.detach(ctx_token)
+
                 final_response = response
                 yield response
 
@@ -313,9 +320,6 @@ class _WorkflowWrapper:
             raise
 
         finally:
-            detach_context_tokens(
-                initial_thread, threading.current_thread(), workflow_token, ctx_token
-            )
             span.end()
 
     def arun(
@@ -430,13 +434,23 @@ class _WorkflowWrapper:
     ) -> Any:
         """Continue streaming async workflow with existing span"""
         final_response = None
-        ctx_token = None
-        workflow_token = None
-        initial_task = asyncio.current_task()
+        iterator = async_iter.__aiter__()
         try:
-            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
-            workflow_token = _setup_workflow_context(node_id)
-            async for response in async_iter:
+            while True:
+                ctx_token = None
+                workflow_token = None
+                try:
+                    ctx_token = context_api.attach(trace_api.set_span_in_context(span))
+                    workflow_token = _setup_workflow_context(node_id)
+                    response = await anext(iterator)
+                except StopAsyncIteration:
+                    break
+                finally:
+                    if workflow_token is not None:
+                        context_api.detach(workflow_token)
+                    if ctx_token is not None:
+                        context_api.detach(ctx_token)
+
                 final_response = response
                 yield response
 
@@ -462,9 +476,6 @@ class _WorkflowWrapper:
             raise
 
         finally:
-            detach_context_tokens(
-                initial_task, asyncio.current_task(), workflow_token, ctx_token
-            )
             span.end()
 
 
@@ -576,13 +587,23 @@ class _StepWrapper:
     ) -> Any:
         """Continue streaming step with existing span"""
         final_response = None
-        ctx_token = None
-        step_token = None
-        initial_thread = threading.current_thread()
+        iterator = iter(iterator)
         try:
-            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
-            step_token = _setup_step_context(node_id)
-            for response in iterator:
+            while True:
+                ctx_token = None
+                step_token = None
+                try:
+                    ctx_token = context_api.attach(trace_api.set_span_in_context(span))
+                    step_token = _setup_step_context(node_id)
+                    response = next(iterator)
+                except StopIteration:
+                    break
+                finally:
+                    if step_token is not None:
+                        context_api.detach(step_token)
+                    if ctx_token is not None:
+                        context_api.detach(ctx_token)
+
                 final_response = response
                 yield response
 
@@ -600,9 +621,6 @@ class _StepWrapper:
             raise
 
         finally:
-            detach_context_tokens(
-                initial_thread, threading.current_thread(), step_token, ctx_token
-            )
             span.end()
 
     def arun(
@@ -711,13 +729,23 @@ class _StepWrapper:
     ) -> Any:
         """Continue streaming async step with existing span"""
         final_response = None
-        ctx_token = None
-        step_token = None
-        initial_task = asyncio.current_task()
+        iterator = async_iter.__aiter__()
         try:
-            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
-            step_token = _setup_step_context(node_id)
-            async for response in async_iter:
+            while True:
+                ctx_token = None
+                step_token = None
+                try:
+                    ctx_token = context_api.attach(trace_api.set_span_in_context(span))
+                    step_token = _setup_step_context(node_id)
+                    response = await anext(iterator)
+                except StopAsyncIteration:
+                    break
+                finally:
+                    if step_token is not None:
+                        context_api.detach(step_token)
+                    if ctx_token is not None:
+                        context_api.detach(ctx_token)
+
                 final_response = response
                 yield response
 
@@ -735,7 +763,6 @@ class _StepWrapper:
             raise
 
         finally:
-            detach_context_tokens(initial_task, asyncio.current_task(), step_token, ctx_token)
             span.end()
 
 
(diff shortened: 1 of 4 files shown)


caroger:
hi @Genmin , thanks for your contribution. I pushed a commits to get your PR over the line.

but please address the CLA bot comment above - it's required for all contributors in their first PR

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
```diff
@@ -415,24 +425,20 @@ async def _arun_stream_continue(
...
-                        final_response = response
-                        yield response
-                finally:
-                    if workflow_token:
-                        try:
-                            context_api.detach(workflow_token)
-                            workflow_token = None
-                        except Exception:
-                            pass
+            ctx_token = context_api.attach(trace_api.set_span_in_context(span))
```
caroger: Concrete suggestion: don’t keep OpenTelemetry context tokens attached across the caller-facing `yield`. In the current async continuations, `ctx_token = context_api.attach(...)` and `_setup_*_context(node_id)` happen before the `async for`, so the consumer gets control while the workflow/step/parallel span is still current. If `aclose()` runs from another task, the guard skips detach and the original consumer task keeps the ended span / `_AGNO_PARENT_NODE_CONTEXT_KEY`.

A safer shape is to attach only while advancing Agno’s underlying iterator, then detach before yielding the event to caller code:

```python
iterator = async_iter.__aiter__()
while True:
    ctx_token = context_api.attach(trace_api.set_span_in_context(span))
    node_token = _setup_workflow_context(node_id)
    try:
        response = await anext(iterator)
    except StopAsyncIteration:
        break
    finally:
        context_api.detach(node_token)
        context_api.detach(ctx_token)

    final_response = response
    yield response
```

Use the corresponding `_setup_step_context` / `_setup_parallel_context` in the other async wrappers, and the same idea for sync streams with `next(iterator)`. Then update the regression to assert that after cross-task `aclose()`, the original consumer task has `trace_api.get_current_span()` restored to its pre-stream value and `context_api.get_value(_AGNO_PARENT_NODE_CONTEXT_KEY) is None.

caroger:
Pushed a follow-up commit (`bd9d2ff2`) to get this over the line.

What changed:
- Updated Workflow, Step, and Parallel stream continuations so OpenTelemetry context is attached only while advancing the underlying iterator, then detached before yielding control back to caller code.
- Added stronger regression assertions that the consumer thread/task has its current span and `_AGNO_PARENT_NODE_CONTEXT_KEY` restored after the first yielded event and after cross-thread/task stream close.
- Applied Ruff import ordering/formatting fixes that were blocking CI.

Validated locally with:

`uvx --with tox-uv tox run -e py310-ci-agno-latest,py314-ci-agno-latest`

Results:
- py310: 46 passed
- py314: 46 passed

The py314 run emitted existing `pytest_asyncio` deprecation warnings, but no failures.

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
ialisaleh: These three functions (`_setup_workflow_context` - `_setup_step_context` - `_setup_parallel_context`) are identical. They can be replaced with `_setup_node_context` to avoid confusion about whether they differ in behaviour.

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
```diff
@@ -268,24 +273,30 @@ def _run_stream_continue(
...
-                    if workflow_token:
-                        try:
-                            context_api.detach(workflow_token)
-                            workflow_token = None
-                        except Exception:
-                            pass
+                    if workflow_token is not None:
+                        context_api.detach(workflow_token)
+                    if ctx_token is not None:
+                        context_api.detach(ctx_token)
```
ialisaleh, line 284: On all the streaming finally blocks, these detach unconditionally. If the generator is abandoned without an explicit `close()`/`aclose()` and GC finalises it on a different thread, the detach fires from the wrong context.

Try capturing `initial_thread = threading.current_thread()` (or `initial_task = asyncio.current_task()`) before the loop and guarding with `detach_context_tokens` from `utils.py` module (which was promoted there precisely for this purpose but is not imported or called here).

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
ialisaleh: On `non_stream_wrapper` in `_WorkflowWrapper.arun`, `_StepWrapper.arun`, and `_ParallelWrapper.aexecute`, `use_span` attaches a context token before the await and holds it across the suspension point. If the awaitable internally does its own `attach`/`detach` (e.g. httpx, anthropic client), the detach ordering on resumption can be non-LIFO, which is the same root cause as the original bug, just narrower. Try manual `attach`/`detach` bracketing the await instead.

ialisaleh, line 466: Add `except (StopIteration, StopAsyncIteration): raise` above this line in all six `*_stream_continue` methods to avoid recording a false ERROR span.

python/instrumentation/openinference-instrumentation-agno/tests/test_workflow_stream_context_finalization.py
```diff
@@ -0,0 +1,178 @@
...
+
+            await asyncio.create_task(close_stream())
+            restored_context.append(
+                (
+                    trace_api.get_current_span() is initial_span,
+                    context_api.get_value(_AGNO_PARENT_NODE_CONTEXT_KEY) == initial_parent_node,
+                )
+            )
+
+        await asyncio.create_task(consume_and_close())
```
ialisaleh, line 205: These tests cover explicit `close()`/`aclose()` from another thread/task, which is the primary scenario. But there's no test for an iterator that yields zero items.

PR #3098, opened 2026-05-13
feat(google-genai)!: require google-genai sdk 2

### LINKED ISSUE #3086: Google GenAI SDK breaking changes
**Describe the bug**
google-genai published 2.0, with breaking [changes](https://ai.google.dev/gemini-api/docs/interactions-breaking-changes-may-2026) in the interaction api
**To Reproduce**
View Python Canary Cron Logs

Summary: https://github.com/Arize-ai/openinference/actions/runs/25690024816

caroger:
## Summary
- require google-genai SDK 2.x and remove legacy 1.x Interactions fallback handling
- update Interactions span extraction for SDK 2 response/stream event shapes
- update Interactions examples from outputs to steps
- refresh cassettes generated with SDK 2.x where possible; keep original file URI image cassettes because live re-recording is quota-blocked

Closes #3086


python/instrumentation/openinference-instrumentation-google-genai/tests/conftest.py
```diff
@@ -64,10 +64,8 @@ def pytest_recording_configure(config: Any, vcr: Any) -> None:
 
 @pytest.fixture(scope="session", autouse=True)
 def _bridge_google_api_key() -> None:
-    if not os.environ.get("GEMINI_API_KEY"):
-        google_api_key = os.environ.get("GOOGLE_API_KEY")
-        if google_api_key:
-            os.environ["GEMINI_API_KEY"] = google_api_key
+    if google_api_key := os.environ.get("GOOGLE_API_KEY"):
```
caroger, line 67: Many tests still construct genai.Client(api_key=os.environ["GEMINI_API_KEY"]). Prefer GOOGLE_API_KEY here so CI/local re-recording can use that key while preserving the existing explicit-client pattern.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_interactions_stream.py
```diff
@@ -98,106 +99,123 @@ class _InteractionAccumulator:
...
-                        self._outputs[idx].data = delta.data or ""
-                    else:
-                        self._outputs[idx].data += delta.data or ""
-                    if delta.mime_type is not None:
-                        self._outputs[idx].mime_type = delta.mime_type
-                    if delta.resolution is not None:
-                        self._outputs[idx].resolution = delta.resolution
-        elif event_type == "interaction.complete":
-            # Update final metadata
+        elif event_type == "step.start":
```
caroger, line 122: SDK 2 Interactions streams emit step.start and step.delta events. This accumulator keeps the stream path shaped like the non-streaming response so the shared response extractor can process steps consistently.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/interactions_attributes.py
```diff
@@ -171,9 +230,10 @@ def get_attributes_from_response(
             **get_output_attributes(safe_json_dumps(response)),
         }
 
+    steps = get_attribute(response, "steps")
```
caroger, line 294: SDK 2 Interactions responses expose steps rather than the old outputs field. Deriving both output.value and LLM output messages from steps removes the 1.x fallback path.

mikeldking:
Screenshot 2026-05-13 at 8 41 23 PM
## Tested examples against local Phoenix

Ran the 5 changed examples end-to-end. All produced spans (status=OK), but found three instrumentation issues worth flagging:

### 1. Streaming drops prompt token count
Trace from `interactions_multi_model.py` (streaming + image input): `llm.token_count.prompt = 0`, `completion = 769`, `total = 1853` — 1084 tokens unaccounted for. `get_token_object_from_response` in `interactions_attributes.py:161-170` only sums `total_input_tokens + total_thought_tokens + total_output_tokens` and ignores `input_tokens_by_modality` / `total_cached_tokens` / `total_tool_use_tokens` that the streaming response carries (visible in cassettes). Result: `total ≠ prompt + completion` for any multimodal or cached request.

### 2. Tool-call-only responses report `completion = 0`
In `interactions_with_tools.py`, the follow-up `create()` whose model output is purely a `tool_calls` message (no text) records `completion=0, prompt=85, total=85` — tool-call output tokens aren't being counted as completion. Siblings that return text do count correctly (`completion=12`).

### 3. `background=True` agent interactions emit a near-empty span with no children
Trace from `agent_interactions.py` (`agent="deep-research-pro-preview-12-2025"`, `background=True`): the span closes in 640ms with `status='in_progress'`, `steps=None`, `usage=None`. No descendants because:
- `background=True` returns immediately; the actual agent work runs server-side and is invisible to the client.
- `InteractionsResource.get` (the 22 polling calls) is **not** wrapped by the instrumentor — only `create` is. So the polls emit zero spans.
- Even if `.get()` were wrapped, those would be sibling root spans, not children, since they aren't inside any `with span` context.

Suggested fixes:
- **Example**: wrap the polling loop in a manual span, or drop `background=True` so `.create()` blocks until the agent finishes and captures `steps` + `usage`.
- **Instrumentor**: wrap `InteractionsResource.get` so polls are at least observable; consider linking spans that share an `interaction.id`.

Otherwise looks healthy — LLM input/output messages, tool-call IDs, arguments, and roles all populate correctly on the model-based traces.

mikeldking (approved):

PR #3104, opened 2026-05-14
feat(go): add openinference-semantic-conventions + openinference-instrumentation

gabe0912:
PR1 of 3 in the Go OpenInference rollout — splits the original combined PR (#3102) per the maintainer's Slack feedback. This PR adds the two foundation packages; PR2 (Anthropic instrumentor) and PR3 (OpenAI instrumentor) follow once this is merged + tagged.

## What ships

| Path | Module | Role |
|---|---|---|
| `go/openinference-semantic-conventions/` | `github.com/Arize-ai/openinference/go/openinference-semantic-conventions` | Attribute keys + value constants. 122 wire-format keys mirroring `python/openinference-semantic-conventions/`. |
| `go/openinference-instrumentation/` | `github.com/Arize-ai/openinference/go/openinference-instrumentation` | Customer-facing config: `WithSuppression`, `WithSession`/`WithUser`/`WithMetadata`/`WithTags`, and `TraceConfig` masking driven by `OPENINFERENCE_HIDE_*` env vars. |

Each module ships its own `go.mod` and gets its own release-please entry (slash-separated tags: `go/openinference-semantic-conventions/v0.1.0` and `go/openinference-instrumentation/v0.1.0`). `go.work` at `go/` makes in-repo builds resolve cross-module deps from the working tree.

### semconv
Plain string constants — pass directly to `attribute.String(...)` without casts. Indexer helpers (`LLMInputMessageRoleKey(i)`, etc.) build array-shaped keys. Drift script `scripts/check_go_semconv.py` parses the Python source and asserts every wire-format key value is present in Go; CI fails if Go falls behind.

### openinference-instrumentation
- **Suppression** — `WithSuppression(ctx)` makes child LLM calls bypass instrumentation entirely (evaluator/grader use case).
- **Context attrs** — `WithSession` / `WithUser` / `WithMetadata` / `WithTags` propagate via OTel baggage. Provider instrumentors (in PR2/PR3) auto-apply them to every LLM span descended from the context. `WithTags` is variadic `...string` and emits `tag.tags` as `attribute.StringSlice`, matching the OpenInference spec.
- **Masking** — `TraceConfig` honors the canonical `OPENINFERENCE_HIDE_*` env vars. Behavior matches Python `TraceConfig.mask` exactly: `HideInputs` drops the entire `llm.input_messages.*` family AND drops `llm.tools.*` AND replaces `input.value` with `__REDACTED__`; `HideOutputs` mirror; `HideInputText` / `HideOutputText` keep structure and redact only `.content`. Token counts, model name, and `llm.finish_reason` are never affected. `WithTraceConfig(...)` lets callers override env at construction (PR2/PR3 will expose this on their instrumentors).

## Releases

`release-please-config.json` adds two entries with `release-as: "0.1.0"` so the inaugural tags are `go/openinference-semantic-conventions/v0.1.0` and `go/openinference-instrumentation/v0.1.0`. Verified via `npx release-please manifest-pr --dry-run`. After v0.1.0 ships, the `release-as` overrides should be removed (follow-up).

The `openinference-instrumentation` go.mod has a relative `replace` for semconv so raw `go build` works inside the module dir; the replace is ignored by external consumers (Go only applies replaces in the *main* module's go.mod).

## Tests

`.github/workflows/go-CI.yaml` runs on paths under either module, the workspace file, the drift script, or the Python semconv source. Steps: `gofmt -l .`, `go vet`, `go test -race -count=1`, `python3 scripts/check_go_semconv.py`.

Coverage:
- semconv: 4 tests pinning wire-format values + Python parity
- openinference-instrumentation: 12 tests covering `TraceConfigFromEnv` env-var parsing, mask helper truth tables, suppression on/off/no-mutate, baggage propagation (including partial baggage and tags with embedded commas), and the empty-value no-op for the With* helpers.

All pass with `-race`.

## Follow-on PRs (blocked on this one)

- **PR2**: `go/openinference-instrumentation-anthropic/` — `option.Middleware` for `anthropics/anthropic-sdk-go` + `examples/` (Arize AX + Phoenix wire-up, manual + auto patterns)
- **PR3**: `go/openinference-instrumentation-openai/` — `http.RoundTripper` for `sashabaranov/go-openai` + same `examples/` surface

Once this PR is merged and release-please cuts the two `v0.1.0` tags, PR2 and PR3 can resolve their `require` lines normally.

## Supersedes

Closes the spirit of #3102 (which bundled all four packages into one PR). #3102 will be closed once this PR merges; the other instrumentor code is staged locally and ready for PR2/PR3.


README.md
```diff
@@ -142,6 +142,19 @@ Normalize and convert data across other instrumentation libraries by adding span
...
+## Go
+
+### Libraries
+
+| Package | Description | Version |
+| --- | --- | --- |
+| [`openinference-semantic-conventions`](./go/openinference-semantic-conventions) | Semantic conventions for tracing of LLM Apps. | `go get github.com/Arize-ai/openinference/go/openinference-semantic-conventions` |
+| [`openinference-instrumentation`](./go/openinference-instrumentation) | Suppression, context attribute propagation, and `TraceConfig` masking. Honors `OPENINFERENCE_HIDE_*` env vars. | `go get github.com/Arize-ai/openinference/go/openinference-instrumentation` |
+| [`openinference-instrumentation-anthropic`](./go/openinference-instrumentation-anthropic) | OpenInference Instrumentation for the Anthropic Go SDK (`anthropics/anthropic-sdk-go`). | `go get github.com/Arize-ai/openinference/go/openinference-instrumentation-anthropic` |
+| [`openinference-instrumentation-openai`](./go/openinference-instrumentation-openai) | OpenInference Instrumentation for the OpenAI Go SDK (`sashabaranov/go-openai`). | `go get github.com/Arize-ai/openinference/go/openinference-instrumentation-openai` |
```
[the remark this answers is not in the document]
gabe0912: Addressed at 2f047db5. README now lists only the two packages this PR introduces (semconv + openinference-instrumentation); the anthropic and openai rows have moved to their respective follow-up PRs (#3105, #3106) where the linked directories actually exist.

go/openinference-instrumentation/instrumentation.go
```diff
@@ -0,0 +1,177 @@
...
+		return ""
+	}
+	return m.Value()
+}
+
+// setBaggage is the internal builder used by the With* helpers.
+// Returns ctx unchanged on a zero-length value or on a baggage-API
+// rejection (e.g., a value containing reserved characters); we never
+// want a tracing helper to error a customer's request path.
+func setBaggage(ctx context.Context, key, value string) context.Context {
```
caroger: I’m concerned about using OTel baggage for OpenInference context attributes here.

`WithSession`, `WithUser`, `WithMetadata`, and `WithTags` are meant to make these values available to OpenInference spans, but baggage is also propagation state. If an application has the baggage propagator configured, these values can be serialized into outbound `baggage` headers on downstream requests, including LLM provider calls.

That is especially risky for `metadata`, since callers may put tenant IDs, workflow labels, internal routing details, or other customer/user data there. The intended behavior is “copy these values onto spans,” not “propagate these values across service boundaries.”

I think these should use private in-process `context.Context` keys instead of `go.opentelemetry.io/otel/baggage`. Provider instrumentors can still call `ApplyContextAttributes(ctx, span)` and set `session.id`, `user.id`, `metadata`, and `tag.tags` on the span, but generic baggage propagators would no longer be able to emit them as HTTP headers.

This also matches the safer pattern used by the other SDKs: Java uses OTel `ContextKey`s, JS uses `createContextKey`, and Python uses OTel context values rather than baggage for these OpenInference attributes.

.github/workflows/go-CI.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: filter
+        with:
+          filters: |
+            diff:
+              - ".github/workflows/go-CI.yaml"
+              - "go/go.work"
```
caroger, line 33: Small Go CI nit: this PR adds `go/go.work.sum`, but the path filter only watches `go/go.work`. If `go.work.sum` changes independently, Go CI may not run.

The setup-go cache dependency list also names `go/openinference-semantic-conventions/go.sum`, but that file does not exist today. I’d suggest adding `go/go.work.sum` to the filter and using a cache dependency list like:

```yaml
cache-dependency-path: |
  go/go.work.sum
  go/**/go.sum
```

I compared this workflow to the existing Python/Java/TypeScript CI setup; the overall shape is consistent with those workflows, so this is just a Go-specific tightening.

release-please-config.json
```diff
@@ -196,6 +196,22 @@
...
                 }
             ]
+        },
+        "go/openinference-semantic-conventions": {
+            "package-name": "openinference-semantic-conventions-go",
+            "component": "go/openinference-semantic-conventions",
+            "release-type": "go",
+            "tag-separator": "/",
+            "include-component-in-tag": true,
+            "release-as": "0.1.0"
```
caroger, line 206: Non-blocking publishing/release note: using `release-as: "0.1.0"` looks fine for bootstrapping the inaugural Go module tags, since Go publication is tag-based and does not need a `publish.yaml` upload job.

Please make sure this override is removed after the first Go release is cut; otherwise future release-please runs can keep forcing the version instead of computing it from commits.

go/openinference-instrumentation/instrumentation.go
gabe0912: Fixed in 800629a. WithSession / WithUser / WithMetadata / WithTags now use unexported `context.Context` keys (sessionKey{}, userKey{}, metadataKey{}, tagsKey{}) — the same in-process pattern as Java/JS/Python. Values flow through the customer's call graph but can't leak via a baggage propagator into outbound HTTP headers, which matters most for metadata. Package doc and README updated to explain the choice. Thanks for catching this.

.github/workflows/go-CI.yaml
gabe0912: Fixed in e7a5eb9. Added `go/go.work.sum` to the paths-filter, and switched the cache list to the glob you suggested (`go/go.work.sum` + `go/**/go.sum`) so the missing semconv go.sum no longer matters and future modules are picked up automatically.

release-please-config.json
gabe0912: Acknowledged — there's a follow-up task to remove the `release-as: "0.1.0"` overrides once the inaugural Go tags are cut, so subsequent release-please runs compute the version from commits.

caroger (approved):

PR #3105, opened 2026-05-14
feat(go): add openinference-instrumentation-anthropic with examples

gabe0912:
PR2 of 3 in the Go OpenInference rollout. Adds the Anthropic instrumentor for `anthropics/anthropic-sdk-go` plus runnable AX + Phoenix examples. Builds on top of #3104 (now merged on `main`, `go/openinference-semantic-conventions/v0.1.0` and `go/openinference-instrumentation/v0.1.0` tagged); this module's `require` lines resolve via the Go module proxy.

## Package

`go/openinference-instrumentation-anthropic/` — `option.Middleware` for `anthropic-sdk-go` v1.43+. Every `/v1/messages` call emits an OpenInference LLM-kind span with model, invocation params, input/output messages, finish reason, and prompt/completion/total token counts (+ cache details when present).

Honors `openinference-instrumentation`'s `WithSuppression`, the in-process context-attribute helpers (`WithSession` / `WithUser` / `WithMetadata` / `WithTags` — backed by unexported `context.Context` keys, not OTel baggage, so they never leak as `baggage` HTTP headers on downstream calls), and `TraceConfig` masking. Streaming responses pass through unchanged with span end tied to body Close/EOF. Error responses and JSON parse failures record exception events.

## Examples

`go/openinference-instrumentation-anthropic/examples/`
- `auto-instrumentation/` — drop-in middleware with `-backend=ax|phoenix`
- `manual-instrumentation/` — hand-built spans using semconv constants

Each example is its own Go module with local `replace` directives so it builds in-tree before tags exist.

## Repo wiring

- Adds release-please entry for the new package.
- `go.work` registers the new module + its two example modules.
- CI workflow (added in #3104) covers `go/openinference-instrumentation-anthropic/**` via paths-filter.


go/openinference-instrumentation-anthropic-sdk-go/anthropic.go
```diff
@@ -0,0 +1,379 @@
...
+			}
+		}
+	}
+
+	if r.Usage != nil {
+		span.SetAttributes(
+			attribute.Int64(semconv.LLMTokenCountPrompt, r.Usage.InputTokens),
+			attribute.Int64(semconv.LLMTokenCountCompletion, r.Usage.OutputTokens),
+			attribute.Int64(semconv.LLMTokenCountTotal, r.Usage.InputTokens+r.Usage.OutputTokens),
+		)
```
caroger, line 392: Cached Anthropic tokens are part of prompt usage. Python’s Anthropic instrumentor folds `cache_creation_input_tokens` and `cache_read_input_tokens` into `llm.token_count.prompt`; this code reports only `input_tokens`, so cached calls undercount prompt tokens. Since this package also emits `llm.token_count.total`, that total should use the same cache-inclusive prompt count.

```suggestion
			promptTokens := r.Usage.InputTokens + r.Usage.CacheReadInputTokens + r.Usage.CacheCreationInputTokens
			span.SetAttributes(
				attribute.Int64(semconv.LLMTokenCountPrompt, promptTokens),
				attribute.Int64(semconv.LLMTokenCountCompletion, r.Usage.OutputTokens),
				attribute.Int64(semconv.LLMTokenCountTotal, promptTokens+r.Usage.OutputTokens),
			)
```

Please add a regression test with nonzero `cache_creation_input_tokens` / `cache_read_input_tokens`.

go/openinference-instrumentation-anthropic-sdk-go/anthropic.go
```diff
@@ -0,0 +1,379 @@
...
+		return
+	}
+	if p.Model != "" {
+		span.SetAttributes(attribute.String(semconv.LLMModelName, p.Model))
+	}
+	if !m.config.HideLLMInvocationParameters {
+		if invocation := invocationParamsJSON(p); invocation != "" {
+			span.SetAttributes(attribute.String(semconv.LLMInvocationParameters, invocation))
+		}
+	}
```
caroger, line 180: `Tools` is decoded into the request payload, but it never becomes span attributes. That means tool-calling requests lose the advertised schemas, and `OPENINFERENCE_HIDE_LLM_TOOLS` has no observable effect. The Python Anthropic instrumentor emits `llm.tools.{i}.tool.json_schema` for this field.

```suggestion
	if !m.config.HideLLMInvocationParameters {
		if invocation := invocationParamsJSON(p); invocation != "" {
			span.SetAttributes(attribute.String(semconv.LLMInvocationParameters, invocation))
		}
	}
	if !m.config.HideInputs && !m.config.HideLLMTools {
		for i, tool := range p.Tools {
			toolJSON, err := json.Marshal(tool)
			if err != nil {
				continue
			}
			span.SetAttributes(attribute.String(semconv.LLMToolKey(i), string(toolJSON)))
		}
	}

```

I would also add tests for normal tool emission and for the hide-tools config path.

caroger (commented):
Inline review findings from local pass.

caroger (approved):

go/go.work
```diff
@@ -8,4 +8,7 @@ go 1.25.0
 use (
 	./openinference-semantic-conventions
 	./openinference-instrumentation
+	./openinference-instrumentation-anthropic
```
caroger: rename the auto instrumentor to `openinference-instrumentation-anthropic-sdk-go`

the convention is to have it end with the actual package/module name that we are instrumenting - in this case `anthropic-sdk-go`

- change folder name
- imports
- readme

release-please-config.json
```diff
@@ -212,6 +212,14 @@
             "tag-separator": "/",
             "include-component-in-tag": true,
             "release-as": "0.1.0"
+        },
+        "go/openinference-instrumentation-anthropic": {
+            "package-name": "openinference-instrumentation-anthropic-go",
```
caroger: openinference-instrumentation-anthropic-sdk-go

caroger (changes requested):

go/openinference-instrumentation-anthropic-sdk-go/anthropic.go
gabe0912: Fixed at a18a63e5. `setResponseAttrs` now computes `promptTokens = InputTokens + CacheReadInputTokens + CacheCreationInputTokens` and threads it into both `llm.token_count.prompt` and `llm.token_count.total`. Added `TestMiddleware_CachedTokensFoldIntoPromptCount` with nonzero cache_read=3 and cache_creation=7 asserting prompt=20 and total=25.

go/openinference-instrumentation-anthropic-sdk-go/anthropic.go
gabe0912: Fixed at a18a63e5. `setRequestAttrs` now iterates request `Tools` and emits `llm.tools.{i}.tool.json_schema` per entry, gated behind `!HideInputs && !HideLLMTools`. Added `TestMiddleware_AdvertisedToolsEmittedAsAttributes` (normal emission with a `get_weather` tool) and `TestMiddleware_HideLLMToolsOmitsAdvertisedTools` (HideLLMTools drops the attribute while leaving input_messages intact).

go/go.work
gabe0912: Renamed at a18a63e5. Module is now `github.com/Arize-ai/openinference/go/openinference-instrumentation-anthropic-sdk-go` — directory, module path in go.mod, internal/httputil import, examples (both `go.mod` files and `main.go` imports), package `README.md` install/import snippets, root README.md row, go.work `use` paths, release-please-config + manifest entries. Build + tests pass; cross-language sibling READMEs updated to reference the new name.

release-please-config.json
gabe0912: Done at a18a63e5. `release-please-config.json` entry is now keyed by `go/openinference-instrumentation-anthropic-sdk-go` with `package-name: openinference-instrumentation-anthropic-sdk-go` and matching `component`. Manifest entry updated too.

.release-please-manifest.json
```diff
@@ -15,7 +15,7 @@
     "python/instrumentation/openinference-instrumentation-litellm": "0.1.33",
     "python/instrumentation/openinference-instrumentation-groq": "0.1.15",
     "python/instrumentation/openinference-instrumentation-instructor": "0.1.16",
-    "python/instrumentation/openinference-instrumentation-anthropic": "1.0.4",
+    "python/instrumentation/openinference-instrumentation-anthropic-sdk-go": "1.0.4",
```
caroger: undo, this is python :)

caroger (changes requested):
Second-round review focused on the new Go Anthropic autoinstrumentor's consistency with the existing Python/JS instrumentations and on whether the previous review comments were addressed.

Requesting changes for the issues below.

1. The Go package rename was over-applied to existing Python/JS Anthropic packages.

   The previous naming comment was about the new Go module, but this PR also changes existing Python/JS release metadata and README entries to `anthropic-sdk-go`. Those paths/packages do not exist and this would break release-please tracking/docs for the existing Anthropic instrumentations:

   - `.release-please-manifest.json:18`
   - `release-please-config.json:81`
   - `README.md:56`
   - `README.md:92`
   - `README.md:118`

   Please restore the existing Python/JS Anthropic entries and keep the `anthropic-sdk-go` name scoped to the new Go module.

2. The new Go module is not covered by Go CI.

   `.github/workflows/go-CI.yaml` still filters and tests only `go/openinference-semantic-conventions/**` and `go/openinference-instrumentation/**`. This PR currently triggers Go CI because it touches `go/go.work`, but future changes isolated to `go/openinference-instrumentation-anthropic-sdk-go/**` can miss CI, and the existing `go vet` / race-test commands do not include the new package.

   Please add the new module path to the workflow path filter and include `./openinference-instrumentation-anthropic-sdk-go/...` in the Go vet/test coverage.

3. The examples create an OTLP tracer provider but use the global no-op tracer.

   Both examples build an exporter-backed `tp`, but then call `otel.Tracer(...)` without calling `otel.SetTracerProvider(tp)` and without using `tp.Tracer(...)`. As written, the examples can complete the Anthropic request while exporting no spans:

   - `go/openinference-instrumentation-anthropic-sdk-go/examples/auto-instrumentation/main.go:42`
   - `go/openinference-instrumentation-anthropic-sdk-go/examples/manual-instrumentation/main.go:39`

   Please either set the global tracer provider before creating the tracer or create the tracer from `tp` directly.

The previous cache-token, tool-schema, Go module naming, and Go release-please naming comments are mostly addressed in the Go package itself. The cache-token regression test and `llm.tools.*.tool.json_schema` coverage are present. The remaining naming problem is the accidental Python/JS metadata and docs regression above.

Checks run:

- `gofmt -l go/openinference-instrumentation-anthropic-sdk-go go/openinference-instrumentation-anthropic-sdk-go/examples/auto-instrumentation go/openinference-instrumentation-anthropic-sdk-go/examples/manual-instrumentation`
- `go test -mod=readonly ./...` in the new module and both examples
- `go vet -mod=readonly ./...` in the new module and both examples
- `go test -race -count=1 -mod=readonly ./...` in the new module
- `go test -race -count=1 ./openinference-semantic-conventions/... ./openinference-instrumentation/...` from `go/`
- `git diff --check main...HEAD`

caroger (approved):

caroger (approved):

PR #3106, opened 2026-05-14
feat(go): add openinference-instrumentation-openai with examples

gabe0912:
PR3 of 3 in the Go OpenInference rollout. Adds the OpenAI instrumentor for `sashabaranov/go-openai` plus runnable AX + Phoenix examples. Builds on top of #3104 (now merged on `main`, `go/openinference-semantic-conventions/v0.1.0` and `go/openinference-instrumentation/v0.1.0` tagged); this module's `require` lines resolve via the Go module proxy.

## Package

`go/openinference-instrumentation-openai/` — `http.RoundTripper` that wraps the OpenAI HTTP client's transport. Every `/v1/chat/completions` call emits an OpenInference LLM-kind span with model, invocation params, input/output messages (incl. nested `tool_calls`), advertised tools, finish reason, prompt/completion/total token counts plus `prompt_tokens_details` (cache, audio) and `completion_tokens_details` (reasoning, audio for o1 / gpt-4o).

Honors `openinference-instrumentation`'s `WithSuppression`, the in-process context-attribute helpers (`WithSession` / `WithUser` / `WithMetadata` / `WithTags` — backed by unexported `context.Context` keys, not OTel baggage, so they never leak as `baggage` HTTP headers on downstream calls), and `TraceConfig` masking. Streaming responses pass through unchanged with span end tied to body Close/EOF. Error responses and JSON parse failures record exception events.

## Examples

`go/openinference-instrumentation-openai/examples/`
- `auto-instrumentation/` — transport wrap with `-backend=ax|phoenix`
- `manual-instrumentation/` — hand-built spans using semconv constants

Each example is its own Go module with local `replace` directives so it builds in-tree before tags exist.

## Repo wiring

- Adds release-please entry for the new package.
- `go.work` registers the new module + its two example modules.
- CI workflow (added in #3104) covers `go/openinference-instrumentation-openai/**` via paths-filter.

## Tracks

Per the maintainer's Slack thread requesting separate PRs by published unit.


README.md
```diff
@@ -150,8 +150,9 @@ Normalize and convert data across other instrumentation libraries by adding span
 | --- | --- | --- |
 | [`openinference-semantic-conventions`](./go/openinference-semantic-conventions) | Semantic conventions for tracing of LLM Apps. | `go get github.com/Arize-ai/openinference/go/openinference-semantic-conventions` |
 | [`openinference-instrumentation`](./go/openinference-instrumentation) | Suppression, context attribute propagation, and `TraceConfig` masking. Honors `OPENINFERENCE_HIDE_*` env vars. | `go get github.com/Arize-ai/openinference/go/openinference-instrumentation` |
+| [`openinference-instrumentation-openai`](./go/openinference-instrumentation-openai) | OpenInference Instrumentation for the OpenAI Go SDK (`sashabaranov/go-openai`). | `go get github.com/Arize-ai/openinference/go/openinference-instrumentation-openai` |
```
caroger: just want to make sure we are intentionally autoinstrument the unofficial https://github.com/sashabaranov/go-openai instead of the official https://github.com/openai/openai-go

`sashabaranov/go-openai` seems to have been stale for at least 7 months

go/openinference-instrumentation-openai-go/openai.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	ctx, span := t.tracer.Start(req.Context(), "openai.chat.completions.create")
+	req = req.WithContext(ctx)
+
+	span.SetAttributes(
+		attribute.String(semconv.OpenInferenceSpanKind, semconv.SpanKindLLM),
+		attribute.String(semconv.LLMSystem, semconv.LLMSystemOpenAI),
+		attribute.String(semconv.LLMProvider, semconv.LLMProviderOpenAI),
+	)
```
caroger, line 103: This always records `llm.provider=openai`, but this transport also wraps Azure OpenAI chat-completions requests. Please derive the provider from `req.URL.Host`/Azure path before setting the attribute, e.g. map `*.openai.azure.com`, `*.services.ai.azure.com`, and `*.cognitiveservices.azure.com` to `semconv.LLMProviderAzure`, and add an Azure URL test.

go/openinference-instrumentation-openai/openai.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	if len(params) == 0 {
+		return ""
+	}
+	out, err := json.Marshal(params)
+	if err != nil {
+		return ""
+	}
+	return string(out)
+}
```
caroger: `llm.invocation_parameters` currently keeps only four fields, so newer/common chat params (`max_completion_tokens`, `reasoning_effort`, `response_format`, `tool_choice`, `stream_options`, penalties, etc.) disappear from traces. 

Can be a follow up if the MVP requirement is narrow. Otherwise please build this from the raw request JSON: unmarshal to `map[string]any`, delete `messages`, `functions`, and separately-emitted `tools`, then marshal the remaining params. Add a test with one omitted field.

go/openinference-instrumentation-openai/examples/auto-instrumentation/main.go
```diff
@@ -0,0 +1,117 @@
...
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		_ = tp.Shutdown(shutdownCtx)
+	}()
+
+	// Wrap the OpenAI HTTP client's transport so every
+	// /v1/chat/completions request emits an LLM-kind span.
+	cfg := openai.DefaultConfig(mustGetenv("OPENAI_API_KEY"))
+	cfg.HTTPClient = &http.Client{
+		Transport: openaiotel.NewTransport(http.DefaultTransport, otel.Tracer("openai-auto-example")),
```
caroger: This example creates an SDK tracer provider but never installs or uses it. `otel.Tracer(...)` will use the default no-op provider, so the example will not export spans. Please call `otel.SetTracerProvider(tp)` after `newTracerProvider(...)` or pass `tp.Tracer(...)` into `NewTransport`; the same fix is needed in the manual example.

caroger (commented):
Inline review findings from local pass.

release-please-config.json
```diff
@@ -212,6 +212,14 @@
             "tag-separator": "/",
             "include-component-in-tag": true,
             "release-as": "0.1.0"
+        },
+        "go/openinference-instrumentation-openai": {
+            "package-name": "openinference-instrumentation-openai-go",
```
caroger, line 225: similar naming convention:
`openinference-instrumentation-go-openai` if we are instrumenting the [3rd party sdk](https://github.com/sashabaranov/go-openai)
`openinference-instrumentation-openai-go` if we are instrumenting the [official sdk](https://github.com/openai/openai-go)

caroger (changes requested):

README.md
gabe0912: Switched to the official SDK at 9b6516e6. The package now depends on `openai/openai-go v1.12.0` and uses its first-class `option.Middleware` surface, dropping the raw `http.RoundTripper` wrap. Module also renamed to `openinference-instrumentation-openai-go` to match the convention. All 15 package tests pass against the official SDK's typed call surface.

go/openinference-instrumentation-openai-go/openai.go
gabe0912: Fixed at 9b6516e6. `llm.provider` is now derived from `req.URL.Host` via a `providerForHost` helper that suffix-matches `*.openai.azure.com`, `*.services.ai.azure.com`, and `*.cognitiveservices.azure.com` (also strips an explicit port so `my-resource.openai.azure.com:443` resolves correctly). `llm.system` stays `openai` regardless of host. Added `TestMiddleware_AzureHostMapsProviderToAzure` with sub-tests for all three Azure suffixes, a port-suffixed host, and a `api.openai.com` negative case.

go/openinference-instrumentation-openai/openai.go
gabe0912: Fixed at 9b6516e6. `invocationParamsJSON` now unmarshals the raw request body to `map[string]any`, deletes `messages`, `functions`, and `tools` (surfaced as separate attributes), and marshals the rest. `max_completion_tokens`, `reasoning_effort`, `response_format`, `tool_choice`, `stream_options`, penalties, `n`, `seed`, etc. all round-trip without us having to enumerate them. Added `TestMiddleware_InvocationParamsCaptureForwardCompatibleFields` that builds a request with several non-content fields and asserts they all land in the attribute (and that the three stripped keys do not).

go/openinference-instrumentation-openai/examples/auto-instrumentation/main.go
gabe0912: Fixed at 9b6516e6 in both examples. Each now passes `tp.Tracer(...)` directly into the middleware (auto) or uses it for manual span construction (manual) rather than `otel.Tracer(...)`. Avoids touching global state and ensures spans actually flow through the locally-built TracerProvider — without this fix the no-op global provider swallowed all spans.

release-please-config.json
gabe0912: Renamed at 9b6516e6 — since the package now instruments the official SDK (`openai/openai-go`), the module / release-please entry is `openinference-instrumentation-openai-go` per your guidance. `package-name` and `component` both updated.

caroger:
Direct maintainer changes pushed in `94114075`:

- Added `go/openinference-instrumentation-openai-go/**` to Go CI path filters and included the OpenAI module plus examples in the CI `go vet` / `go test -race` package lists.
- Fixed the OpenAI Go README usage snippet by importing `github.com/openai/openai-go/shared`.
- Added legacy chat `function_call` request/response parsing so spans emit `message.function_call_name` and `message.function_call_arguments_json`, with test coverage.

Validation:

- `gofmt` clean.
- `go vet` passed for the OpenAI module and the expanded workspace package list.
- `go test -race -count=1` passed for the OpenAI module and the expanded workspace package list.
- GitHub checks after push passed, including `CI Go`, `Go CI Required`, and `claude-review`.

caroger (approved):

caroger (approved):

PR #3107, opened 2026-05-14
ci: add weekly skills update workflow

mikeldking:
## Summary
- Adds `.github/workflows/skills-update.yml`, a scheduled workflow that runs every Monday at 15:00 UTC (9am America/Denver during MDT) and on `workflow_dispatch`.
- Runs `npx --yes skills` to refresh the agent skills under `.agents/skills/` and `skills-lock.json`, then opens a PR via `peter-evans/create-pull-request` when there's a diff.
- Uses the existing `MY_RELEASE_PLEASE` PAT so the resulting PR triggers downstream CI (same pattern as `release.yml` and the `bump-openinference-*` workflows).

## Notes
- Workflow-level permission is `contents: read`; the job opts into `contents: write` + `pull-requests: write` with explanatory comments (mirrors `python-cron.yaml`).
- Adds a `concurrency` group and `timeout-minutes: 10`.
- Default `zizmor` audit reports no findings; `--persona=pedantic` reports only the Low-confidence `superfluous-actions` suggestion to swap `peter-evans/create-pull-request` for `gh pr create`, which I left as-is to match repo convention.

## Test plan
- [ ] Manually trigger via the Actions tab (`workflow_dispatch`) and confirm:
  - [ ] When there are no skill updates, the workflow finishes without creating a PR.
  - [ ] When there are updates, a PR titled `chore: weekly skills update` is opened against `main`, contains only changes under `.agents/skills/` and `skills-lock.json`, and CI runs on it.
- [ ] Wait for the first scheduled Monday run to confirm cron firing.

### DIFF
--- a/.github/workflows/skills-update.yml
+++ b/.github/workflows/skills-update.yml
@@ -32,12 +32,13 @@ jobs:
           node-version: "20"
 
       - name: Run skills update
-        run: npx --yes skills
+        run: npx --yes skills@1.5.6
 
       - name: Check for changes
         id: verify-changed-files
         run: |
-          if git diff --quiet -- .agents/skills skills-lock.json; then
+          git add -- .agents/skills skills-lock.json
+          if git diff --quiet --cached -- .agents/skills skills-lock.json; then
             echo "changed=false" >> $GITHUB_OUTPUT
           else
             echo "changed=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


mikeldking:
Addressed both review comments in e843fc15:

- **Pinned `npx skills`** to `skills@1.5.6` (current latest published by vercel-labs/skills). Bumps will now be deliberate.
- **Fixed `git diff` gating** so untracked new skill directories are detected: `git add -- .agents/skills skills-lock.json` then `git diff --quiet --cached`.

PR #3112, opened 2026-05-15
feat: specify reasoning semantic conventions

mikeldking:
## Summary

- Clarifies the OpenInference reasoning semantic conventions for ordered reasoning content, provider continuity tokens, and tool-call reasoning state.
- Adds and documents `message_content.id` for provider-assigned content identifiers such as OpenAI `ResponseReasoningItem.id`.
- Separates provider continuity fields by source:
  - OpenAI `encrypted_content` -> `message_content.encrypted_content`
  - Anthropic `thinking.signature` -> `message_content.signature`
  - Anthropic `redacted_thinking.data` -> `message_content.data`
  - Gemini `thoughtSignature` on text/non-tool parts -> `message_content.signature`
  - Gemini `thoughtSignature` on `functionCall` parts -> `tool_call.reasoning_signature`
- Documents `message_content.type = "tool_use"` for provider-native ordered tool-use parts, reusing `tool_call.*` suffixes when a tool call must remain ordered relative to reasoning/text parts.
- Updates Python, JavaScript, Java, and Go semantic-convention constants/tests to match the spec.
- Fixes the reasoning round-trip demos so they emit and replay the new keys consistently:
  - OpenAI summary-less reasoning blocks still get `message_content.type = "reasoning"`.
  - OpenAI encrypted payloads use `message_content.encrypted_content`.
  - Anthropic `redacted_thinking.data` uses `message_content.data`.
  - Gemini tool-call signatures use `tool_call.reasoning_signature`.

## Why

The previous draft normalized too much under generic signature fields, which made OpenAI encrypted content and Anthropic redacted data indistinguishable from provider signatures. Gemini `tool_call.signature` was also ambiguous because the provider calls the field `thoughtSignature`; `tool_call.reasoning_signature` makes clear that this is reasoning-continuity state attached to a tool call.

## Test plan

- [x] `python3 -m py_compile internal_docs/specs/reasoning/scripts/common.py internal_docs/specs/reasoning/scripts/openai_roundtrip.py internal_docs/specs/reasoning/scripts/anthropic_roundtrip.py internal_docs/specs/reasoning/scripts/gemini_roundtrip.py`
- [x] `uv run --with ruff ruff check ...`
- [x] `uv run --with ruff ruff format --check ...`
- [x] `PYTHONPATH=python/openinference-semantic-conventions/src uv run --with pytest python -m pytest python/openinference-semantic-conventions/tests/openinference/semconv/test_attributes.py`
- [x] `go test ./...` from `go/openinference-semantic-conventions`
- [x] `pnpm --dir js --filter @arizeai/openinference-semantic-conventions type:check`
- [x] `pnpm --dir js exec oxfmt --check packages/openinference-semantic-conventions/src/trace/SemanticConventions.ts`
- [x] `git diff --check`

Java Gradle verification could not complete locally because this machine only has Java 8 installed, while the repo's Spring Boot Gradle plugin resolution requires Java 17.


spec/semantic_conventions.md
```diff
@@ -83,9 +83,11 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
-| `message_content.type`                         | String                      | `"text"` or `"image"`                                                                                                         | The type of the content, such as "text" or "image".                                                                                                                                 |
-| `message_content.text`                         | String                      | `"This is a sample text"`                                                                                                     | The text content of the message, if the type is "text".                                                                                                                             |
+| `message_content.type`                         | String                      | `"text"`, `"image"`, `"audio"`, or `"reasoning"`                                                                              | The type of the content (e.g. `"text"`, `"image"`, `"audio"`, `"reasoning"`).                                                                                                       |
```
mikeldking: picked reasoning as it's the most common name (reasoning models)

spec/semantic_conventions.md
```diff
@@ -83,9 +83,11 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
 | `message_content.image`                        | Image Object                | `{"image.url": "https://sample-link-to-image.jpg"}`                                                                           | The image content of the message, if the type is "image".                                                                                                                           |
+| `message_content.signature`                    | String                      | `"EuYBCkQYAiJA..."`                                                                                                           | Opaque vendor-issued echo token captured verbatim. Maps to OpenAI `encrypted_content`, Anthropic `signature` (or `data` for redacted blocks), or Gemini `thoughtSignature`. Required on stateless multi-turn echo. |
```
mikeldking: this would be any sort of continuation token - note the name is different across vendors but they roughly all mean the signature of the data and signature is the simplest

internal_docs/specs/reasoning/vendor_comparison.md
```diff
@@ -0,0 +1,342 @@
...
+  "model": "gpt-5",
+  "store": false,
+  "include": ["reasoning.encrypted_content"],
+  "reasoning": { "effort": "medium" },
+  "input": [
+    { "role": "user", "content": "..." },
+    {
+      "type": "reasoning",
+      "id": "rs_1",
+      "summary": [{ "type": "summary_text", "text": "..." }],
```
mikeldking, line 239: We would just put this as content. It's not entirely right but seems odd to special case. We could as summary...

cephalization (approved):

spec/semantic_conventions.md
nate-mar: I'm not totally following why encrypted content and thoughtSignature or data would fall under signature, because it's technically carrying the whole value as opposed to being some hashed signature for verification (like is used in the Anthropic case for normal thinking)

spec/semantic_conventions.md
```diff
@@ -83,9 +83,11 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
-| `message_content.text`                         | String                      | `"This is a sample text"`                                                                                                     | The text content of the message, if the type is "text".                                                                                                                             |
+| `message_content.type`                         | String                      | `"text"`, `"image"`, `"audio"`, or `"reasoning"`                                                                              | The type of the content (e.g. `"text"`, `"image"`, `"audio"`, `"reasoning"`).                                                                                                       |
+| `message_content.id`                           | String                      | `"rs_abc123"` or `"msg_def456"`                                                                                               | Vendor-assigned identifier for the content block, when one exists. Required on stateless echo for OpenAI Responses (`rs_*` for reasoning items, `msg_*` for message items). Unset for Anthropic/Gemini. |
```
nate-mar: Is there anything we can meet here to ensure that this stays up to day (vendor support)?

spec/semantic_conventions.md
nate-mar: Also, the descriptions of message_content.signature and tool_call.signature together make it sound like Gemini only puts thoughtSignature on tool calls. In practice, Gemini attaches it to whichever data-bearing part the reasoning landed on: functionCall when tools are involved, text when they're not. [Docs do say ](https://ai.google.dev/gemini-api/docs/thought-signatures)that it's recommended, but it's not clearly covered by either field here and has no example in llm_spans.md.

internal_docs/specs/reasoning/vendor_comparison.md
```diff
@@ -0,0 +1,342 @@
...
+## 4. Quick reference: capturing for telemetry
+
+| Aspect | OpenAI | Anthropic | Gemini |
+|---|---|---|---|
+| Intent (request) | `reasoning.effort`, `reasoning.summary`, `include[]` contains `reasoning.encrypted_content` | `thinking.type`, `thinking.budget_tokens`, `thinking.display` | `thinkingConfig.thinkingBudget` / `thinkingLevel`, `includeThoughts` |
+| Arrival (response) | count of `output[].type == "reasoning"`; presence of `summary[]` vs `content[]` | count of `content[].type == "thinking"` and `"redacted_thinking"`; presence of `signature` | count of `parts[].thought == true`; count of `parts[].thoughtSignature` (and which kinds of parts carry it) |
+| Continuity token | presence + length of `encrypted_content` | presence + length of `signature` / `data` | presence + length of `thoughtSignature` per part |
+| Tokens | `output_tokens_details.reasoning_tokens` | (no separate field; included in `output_tokens`) | `usageMetadata.thoughtsTokenCount` |
+
+Never log the continuity token's payload bytes — they're large, opaque, base64-encoded, and not human-readable. Log presence, length, and the discriminator that produced them.
```
nate-mar, line 353: I assume given the context, "log" is referring to on the spans, and not log telemetry as separate from spans? If so, aren't we logging the token per the spec?

spec/semantic_conventions.md
mikeldking: Will remove - I think as we noticed. It's missing from Gemini / OpenAI and it doesn't  serve anything right now.

internal_docs/specs/reasoning/vendor_comparison.md
mikeldking: This seems wrong - will re-work. Thanks for flagging.

spec/semantic_conventions.md
mikeldking: will split into .signature and .encrypted_content - good callout

PR #3113, opened 2026-05-15
fix(smolagents): Handle Crash On Dropped OTEL Span

### LINKED ISSUE #2747: [bug] openinference-instrumentation-smolagents crash for NonRecordingSpan
**Describe the bug**
If opentelemetry is not well configured or the span is dropped, the instrumentation for smolagents crashes with the following error:
```
File "/home/xxxx/.cache/pypoetry/virtualenvs/agentdeltasud-Sy8Wvlmq-py3.12/lib/python3.12/site-packages/openinference/instrumentation/smolagents/_wrappers.py", line 241, in _finalize_step_span
if span.status.status_code != trace_api.StatusCode.ERROR: # type: ignore[attr-defined]

AttributeError: 'NonRecordingSpan' object has no attribute 'status'
```
**To Reproduce**
Steps to reproduce the behavior

Run any smolagents after calling `SmolagentsInstrumentor().instrument()` and without providing any OTEL configuration

**Expected behavior**
The span should be silently dropped

 - OS: linux
 - Version: 1.22

**Additional context**

ialisaleh:
Closes #2747

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/_wrappers.py
@@ -243,6 +243,9 @@ def _finalize_step_span(
     - Sets status to OK if no error is present.
     - Captures & logs any errors that occur.
     """
+    if not span.is_recording():
+        return
+
     observations = getattr(step_log, "observations", None)
     if observations is not None:
         span.set_attribute(OUTPUT_VALUE, str(observations))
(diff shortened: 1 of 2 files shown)


ialisaleh:
@claude review

caroger (approved):
Thanks for the fix. The `is_recording()` guard prevents `_finalize_step_span` from dereferencing proxied span state on a dropped/non-recording OTEL span, which addresses the crash reported in #2747.

I also checked the added coverage for dropped spans with observations, errors, expected tool errors, no observations, and no side effects. Local focused tox passed for `TestFinalizeStepSpanWithDroppedSpan`.

PR #3116, opened 2026-05-17
fix(python): preserve falsy GenAI tool results

pragnyanramtha:
No linked issue; this is a small bug fix for GenAI semantic convention conversion.

## Summary
- Preserve converted GenAI tool call arguments and results when the converted value is a valid falsy string.
- Add regression coverage for empty string, `0`, and `false` tool outputs.

## Why
Tool spans with `output.value` set to an empty text result or another falsy result were dropping `gen_ai.tool.call.result` because the conversion used a truthiness check. When the OpenInference output attribute is present, the GenAI tool call result should be emitted even if its serialized value is falsy.

## Validation
- `uvx --with tox-uv tox run -e py314-test-instrumentation -- tests/test_genai.py` - 23 passed.
- `uvx --with tox-uv tox run -e py314-ruff-mypy-test-instrumentation -- tests/test_genai.py` - ruff format/check passed, mypy passed, 23 tests passed.


pragnyanramtha:

pragnyanramtha:
recheck

mikeldking:
thanks for this @pragnyanramtha

PR #3124, opened 2026-05-19
fix(google_adk): Handle Session ID For Sub-Agent Spans

### LINKED ISSUE #3117: [bug] google-adk: sub-agent spans get wrong session.id when using AgentTool
### **Describe the bug**
When using `openinference-instrumentation-google-adk`, spans generated during sub-agent execution (invoked via `AgentTool`) are stamped with an incorrect `session.id`. Instead of inheriting the `session_id` from the top-level `Runner`, sub-agent spans are stamped with an ADK-internal sub-session UUID.

This causes traces to be split across different sessions in backends like Langfuse or Arize Phoenix.

### **To Reproduce**
I have created a minimal reproduction repository: https://github.com/MrOrz/openinference-google-adk-session-id-bug

1. Clone the repo.
2. Install dependencies: `uv sync`.
3. Set `GOOGLE_API_KEY` in `.env`.
4. Run the PoC: `uv run python poc.py`.
5. Observe the output table where `agent_run [sub_agent]` and its child `call_llm` spans have a different `session.id` than the root span.

### **Expected behavior**
All spans within the same trace, including those from sub-agents invoked via `AgentTool`, should share the same `session.id` provided to the initial `runner.run_async()` call.

### **Actual behavior**
```text
──────────────────────────────────────────────────────────────────────────────────────────────────────────────
SPAN NAME                           PARENT                    session.id
──────────────────────────────────────────────────────────────────────────────────────────────────────────────
call_llm                            agent_run [sub_agent]     c74cba2c-5bc6-4f1a-9e7b-f43b2b9cc251  ✗ BUG
agent_run [sub_agent]               invocation [poc-app]      c74cba2c-5bc6-4f1a-9e7b-f43b2b9cc251  ✗ BUG
invocation [poc-app]                execute_tool sub_agent    known-session-id-abc123
execute_tool sub_agent              call_llm                  known-session-id-abc123
call_llm                            agent_run [root_agent]    known-session-id-abc123
call_llm                            agent_run [root_agent]    known-session-id-abc123
agent_run [root_agent]              invocation [poc-app]      known-session-id-abc123
invocation [poc-app]                (root)                    known-session-id-abc123
──────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

### **Desktop (please complete the following information):**
 - OS: macOS (darwin)
 - Version: openinference-instrumentation-google-adk==0.1.13, google-adk==1.31.1

### **Additional context**

**Root Cause Analysis:**
The investigation using a reproduction trace reveals that the session ID fragmentation occurs because `_RunnerRunAsync` unconditionally prioritizes the `session_id` argument from `kwargs` over the ambient OpenTelemetry context:

1.  **Sub-agent Runner (`AgentTool`)**: ADK's `AgentTool` creates a new sub-runner and generates an internal sub-session UUID, passing it as the `session_id` argument to `run_async()`.
2.  **Instrumentation Overwrite**: In `_RunnerRunAsync.__call__`, the instrumentation extracts this internal UUID and stamps it onto the `invocation` span attributes, even if a valid `session.id` is already in the context.
3.  **Context Pollution**: In `__aiter__`, it calls `stack.enter_context(using_session(session_id))`. This switches the ADK telemetry context to the internal UUID, causing all subsequent child spans (like `agent_run` and `call_llm`) to inherit the wrong ID from the polluted context.

**Proposed Fix:**
The instrumentation should prioritize the ambient `session.id` already present in the OpenTelemetry context. It should only fallback to the `session_id` argument if the context is empty.

**In `_RunnerRunAsync.__call__`:**
```python
ambient_attributes = get_attributes_from_context()
if SpanAttributes.SESSION_ID in ambient_attributes:
    # Prioritize parent session from OTel context
    attributes[SpanAttributes.SESSION_ID] = ambient_attributes[SpanAttributes.SESSION_ID]
elif (session_id := kwargs.get("session_id")) is not None:
    # Fallback to kwargs for top-level runners
    attributes[SpanAttributes.SESSION_ID] = session_id
```

**In `_RunnerRunAsync.__aiter__`:**
```python
# Only switch ADK context if we are not already in a session
if session_id is not None and SpanAttributes.SESSION_ID not in ambient_attributes:
    stack.enter_context(using_session(session_id))
```

ialisaleh:
Closes #3117

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
@@ -24,7 +24,6 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
     instrument: Any,
     in_memory_span_exporter: InMemorySpanExporter,
 ) -> None:
-    # Build a sub-agent exposed as an AgentTool to the root agent
     sub_agent_name = "Sub_Agent"
     sub_agent = Agent(
         name=sub_agent_name,
@@ -42,7 +41,6 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
         tools=[AgentTool(agent=sub_agent)],
     )
 
-    # Wire up runner with a known session ID
     app_name = token_hex(4)
     user_id = token_hex(4)
     session_id = token_hex(4)
@@ -65,7 +63,6 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
     ):
         ...
 
-    # Collect and index spans
     spans = sorted(
         in_memory_span_exporter.get_finished_spans(),
         key=lambda s: s.start_time or 0,
@@ -74,10 +71,10 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
     for span in spans:
         spans_by_name[span.name].append(span)
 
-    # Only one top-level invocation span should exist.
     invocation_spans = spans_by_name[f"invocation [{app_name}]"]
     assert len(invocation_spans) >= 1
 
+    # Root invocation has no parent and carries the correct session/user IDs.
     root_invocation = next(s for s in invocation_spans if not s.parent)
     assert root_invocation.status.is_ok
     root_attrs = dict(root_invocation.attributes or {})
@@ -85,7 +82,7 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
     assert root_attrs.get(SpanAttributes.USER_ID) == user_id
     assert root_attrs.get("openinference.span.kind") == "CHAIN"
 
-    # Root agent_run span is a child of the root invocation
+    # Root agent_run is a direct child of the root invocation.
     root_agent_run_spans = spans_by_name[f"agent_run [{root_agent_name}]"]
     assert len(root_agent_run_spans) == 1
     root_agent_run = root_agent_run_spans[0]
@@ -97,7 +94,7 @@ async def test_sub_agent_session_id_not_overwritten_by_adk_internal_uuid(
     assert root_agent_run_attrs.get("openinference.span.kind") == "AGENT"
     assert root_agent_run_attrs.get("agent.name") == root_agent_name
 
-    # Check execute_tool span for the AgentTool call
+    # execute_tool span is produced when the root agent invokes the AgentTool.
     tool_spans = spans_by_name[f"execute_tool {sub_agent_name}"]
     assert len(tool_spans) == 1
     tool_span = tool_spans[0]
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
```diff
@@ -7,15 +7,138 @@
...
 from google.genai import types
 from opentelemetry.sdk.trace import ReadableSpan
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 
+from openinference.semconv.trace import SpanAttributes
+
 _VERSION = cast(tuple[int, int, int], tuple(int(x) for x in __version__.split(".")[:3]))
 
 
+@pytest.mark.vcr
```
caroger, line 30: This regression test currently runs under the minimum-version Google ADK CI jobs, but that path installs `google-adk==1.2.1` and the test fails before verifying the session-id behavior because `sub_invocation_spans` is empty.

I checked upstream `google/adk-python`: before `v1.17.0`, `AgentTool` creates the child runner/session with the wrapped agent name as the app name. Starting in `v1.17.0`, it derives the child app from the parent invocation context, which is the invocation shape this test is asserting.

Please gate this test on the ADK version, for example:

```python
@pytest.mark.skipif(
    _VERSION < (1, 17, 0),
    reason=(
        "AgentTool uses the wrapped agent name as the child runner app before "
        "google-adk 1.17; this regression asserts the newer parent-app "
        "sub-agent invocation shape."
    ),
)
@pytest.mark.vcr
```

That should keep coverage in the `google_adk-latest` jobs while restoring the supported minimum-version matrix.

caroger (changes requested):
Thanks for taking this on. The wrapper change is headed in the right direction for #3117: it preserves the ambient `session.id` and avoids pushing the ADK-internal sub-session when a parent session is already active.

I am requesting changes because the new regression test currently breaks the supported non-latest Google ADK matrix. I reproduced the same failure locally with `py314-ci-google_adk`: under `google-adk==1.2.1`, the test never observes a nested `invocation [{app_name}]` span and fails before it reaches the session-id assertions.

Please version-gate the new AgentTool regression test so it only runs on Google ADK versions with the parent-app sub-agent invocation shape. I checked upstream and `v1.17.0` is the first tag where `AgentTool` derives the child runner app name from the parent invocation context; older versions use the wrapped agent name, so this exact assertion shape is not valid there.

python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
ialisaleh: Addressed.

ialisaleh:
@claude review

caroger (approved):

PR #3125, opened 2026-05-19
fix(bedrock): set clean text from routing classifier output and set action group tool description

### LINKED ISSUE #2352: [feature request] Ensure Parity between JS and Python Bedrock Agent instrumentation
Comparing traces between the JS and Python Bedrock instrumentors, there are some slight differences that would be good to rectify between the two instrumentation packages. These differences include:

- The JS instrumentor parses just the output.message.context.text for the routing classifier LLM span, the Python instrumentor captures the entire response which includes a lot of noise:

JS:

Image

Python:

Image

- JS instrumentor also has the correct tool attributes on the action_group tool span, in the Python instrumentor it's empty for the tool attributes.

- For the JS instrumentor, the Guardrails traces have a nested metadata key for each guardrail where the clientRequestId, startTime, etc. are all nested, in the Python instrumentor each guardrail metadata is just at the top level of the intervening/non_intervening guardrails list

Update the Python instrumentor to ensure parity between the two instrumentation packages

satyadevai:
closes #2352


caroger (approved):
Reviewed the routing-classifier reasoning output and action-group tool description changes. The output extraction now handles reasoning blocks with null text before the actual routing decision, and the new cassette covers the intended shape.

Verification run locally:

`uvx tox run -e py314-ci-bedrock -- tests/openinference/instrumentation/bedrock/test_invoke_agent_instrumentor.py::test_routing_classifier_with_reasoning`

`git diff --check` also passed.

PR #3140, opened 2026-05-20
fix(claude_agent_sdk): pick max-output-tokens model from modelUsage

### LINKED ISSUE #3136: [bug][claude-agent-sdk] _extract_model_name_from_usage returns first dict key, mis-attributes multi-model runs
`_extract_model_name_from_usage` (in `_wrappers.py:121-141`) extracts the
model name from a `model_usage` mapping by returning
`str(next(iter(model_usage.keys())))`:

```python
def _extract_model_name_from_usage(model_usage: Any) -> str | None:
    if isinstance(model_usage, MappingABC) and model_usage:
        return str(next(iter(model_usage.keys())))   # <-- first dict key wins
    ...
```

This value is later written to the AGENT span via
`_maybe_set_model(span, msg)` (`_wrappers.py:167-175`) on every streamed
message. The `ResultMessage` final flush overwrites whatever the per-message
`AssistantMessage` handlers had stamped, so the **last** call to
`_extract_model_name_from_usage` is what the span ends up carrying.

### Reproduction

Run `claude_agent_sdk.query(...)` against any configuration where the
SDK ends up reporting usage for two different models in the same run
(for example: the main model plus a separate small/fast / router
model). Inspect the AGENT span's `llm.model_name` attribute - it
carries whichever model key happens to be iterated first out of
`model_usage`, which is not necessarily the model that produced the
majority of the run's output.

### Expected behavior

The model attribute should reflect the model that did the bulk of the
work, not whichever key happens to land first in dict iteration. Dict
ordering is brittle even within a single-model run.

### Code reference

`python/instrumentation/openinference-instrumentation-claude-agent-sdk/openinference/instrumentation/claude_agent_sdk/_wrappers.py:121-141`

Kymi808:
## Summary

Closes #3136.

`_extract_model_name_from_usage` (in `_wrappers.py:121-141`) returned `next(iter(model_usage.keys()))` when given a dict-shaped `modelUsage` — i.e. whichever key landed first in iteration order. This value is set on the AGENT span via `_maybe_set_model` on every streamed message, so multi-model runs (main model + a separate router/fast model) ended up with the span's `llm.model_name` carrying an essentially-random model name rather than the one that actually did the bulk of the generation.

## Fix

The claude-agent-sdk emits `modelUsage` as:

```python
{
    "claude-sonnet-4-6": {"outputTokens": 350, "inputTokens": 8, "costUSD": 0.041, ...},
    "claude-haiku-4-5":  {"outputTokens": 5,   "inputTokens": 200, ...},
}
```

so the dict branch now picks the model that did the bulk of the generation work by weighting on `outputTokens` (with `output_tokens` snake-case accepted as a fallback). Non-mapping shapes — list/tuple of entries, attribute-bearing objects, `None` — keep their existing semantics. **This PR only changes the dict branch.**

Defensive: max-key extraction handles missing `outputTokens`, non-int values, and non-mapping entry values without raising.

## Tests

New `tests/test_wrappers_unit.py` (15 tests, all passing). The tests target `_extract_model_name_from_usage` directly so they run without `claude-agent-sdk` integration setup:

- Single-model dict returns that model
- Multi-model dict picks max-output-tokens irrespective of insertion order
- `output_tokens` snake-case accepted alongside `outputTokens`
- Missing / non-int `outputTokens` does not crash
- Non-mapping entry value (string instead of dict) does not crash
- Empty dict returns `None`
- List-of-entries branch preserves original "first named entry" semantics (unchanged)
- Object-with-`.model` branch preserves original semantics (unchanged)
- `None` and other falsy inputs return `None`

**Regression coverage verified**: `git stash` of the fix → 4 of 15 tests fail against the prior code (the multi-model-picks-max ones), confirming the tests would have caught the original bug.

`ruff format` and `ruff check` clean on the touched files.

## Notes

- No semantic-conventions changes — span attribute keys and values are unchanged; only the *value chosen* for `llm.model_name` is now correct on multi-model runs.
- I'd love to add a cassette-driven integration test for the multi-model case, but I couldn't find an existing cassette with a multi-model `modelUsage` — happy to add one if a maintainer can point me at a recording with multiple models present.

---

*AI-assisted contribution — diff was reviewed line-by-line; the unit tests were run locally against both the original and patched implementation to verify the regression coverage.*

### DIFF
--- a/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/__init__.py
@@ -8,7 +8,7 @@ from opentelemetry import trace as trace_api
 from opentelemetry.instrumentation.instrumentor import (  # type: ignore[attr-defined]
     BaseInstrumentor,
 )
-from wrapt.patches import resolve_path, wrap_function_wrapper
+from wrapt import resolve_path, wrap_function_wrapper
 
 from openinference.instrumentation import OITracer, TraceConfig
 from openinference.instrumentation.claude_agent_sdk._wrappers import (
@@ -76,8 +76,8 @@ class ClaudeAgentSDKInstrumentor(BaseInstrumentor):  # type: ignore[misc]
         self._originals: List[Tuple[Any, Any, Any]] = []
         for qualified, wrapper in method_wrappers.items():
             module, name = qualified.split(":", 1)
-            self._originals.append(resolve_path(module, name))  # type: ignore[no-untyped-call]
-            wrap_function_wrapper(module, name, wrapper)  # type: ignore[no-untyped-call]
+            self._originals.append(resolve_path(module, name))
+            wrap_function_wrapper(module, name, wrapper)
 
         # Sync package export so "from claude_agent_sdk import query" resolves to the wrapper
         import claude_agent_sdk
--- a/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
@@ -8,14 +8,6 @@ from collections.abc import Mapping as MappingABC
 from typing import TYPE_CHECKING, Any, Callable, Mapping, Tuple
 
 import opentelemetry.context as context_api
-from openinference.semconv.trace import (
-    MessageAttributes,
-    OpenInferenceLLMSystemValues,
-    OpenInferenceMimeTypeValues,
-    OpenInferenceSpanKindValues,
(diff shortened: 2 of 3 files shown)


Kymi808:

ialisaleh:
Hi Kyle - This issue is addressed in a different PR, so closing this one. Thanks!

caroger (commented):
Thanks @Kymi808 for the contribution. This is the right direction for #3136: the fix handles the Claude Agent SDK's real `modelUsage` shape with camelCase `outputTokens`, while preserving the existing fallback behavior for other usage shapes.

I pushed a small maintainer follow-up commit to get the PR over the line:

- cleaned up the import ordering in `_wrappers.py` for ruff
- moved the Claude Agent SDK instrumentor to the public `wrapt` import path and removed now-unneeded type ignores so mypy passes with the current test environment
- updated the PR title to conventional-commit format

Local verification after the maintainer commit:

- `tox run -e py310-ci-claude_agent_sdk -- tests/test_wrappers_unit.py`
- direct camelCase `modelUsage` probe selected the high-output model as expected

caroger (approved):

PR #3146, opened 2026-05-21
feat(semconv): Add audio I/O and USER/AUDIO span kinds

nosnaws:
# Summary
1. Adds `AUDIO` span kind. This will be used to represent calls to models that operate on audio such as `gpt-audio`, TTS, STT, or realtime APIs.
2. Adds `USER` span kind. This will be used to represent user input in streaming applications, intended for use with realtime streaming audio APIs, but can be used for any streaming system.
3. Adds an `audio` namespace to `input` and `output`, allowing us to represent input or output audio.

### DIFF
--- a/go/openinference-semantic-conventions/attributes.go
+++ b/go/openinference-semantic-conventions/attributes.go
@@ -16,6 +16,14 @@ const (
 	OutputValue    = "output.value"
 	OutputMimeType = "output.mime_type"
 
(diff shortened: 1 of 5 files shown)


nosnaws:
To properly prove out the audio tracing, I'm going to close this PR and open another one that locally defines these constants in the instrumentor, that way the blast radius is contained.

PR #3178, opened 2026-05-29
fix(google_adk): Handle Thinking Token Inclusion Correctly

### LINKED ISSUE #3177: fix(google-adk): thoughts_token_count double-counted in completion tokens
## Describe the bug

When using Gemini models with thinking enabled, `llm.token_count.completion` is inflated because `thoughts_token_count` is added on top of `candidates_token_count`, even though `thoughts_token_count` is already a subset of `candidates_token_count` per the Gemini API spec.

---

## To Reproduce

Use `GoogleADKInstrumentor` with a Gemini thinking model (e.g. `gemini-2.5-flash`) and inspect the trace in Phoenix.

Compare the raw API `usage_metadata`:

```json
{
  "candidates_token_count": 3362,
  "thoughts_token_count": 3347,
  "total_token_count": 3703
}
```

Phoenix reports `llm.token_count.completion` as **6709** (3362 + 3347) instead of the correct **3362**.

---

## Expected behavior

`llm.token_count.completion` should equal `candidates_token_count` (**3362**), not `candidates_token_count + thoughts_token_count` (**6709**).

---

## Root cause

In `_wrappers.py`, `_get_attributes_from_usage_metadata` does:

```python
completion = 0
if candidates := obj.candidates_token_count:
    completion += candidates          # 3362 (already includes thoughts)
if thoughts := obj.thoughts_token_count:
    yield LLM_TOKEN_COUNT_COMPLETION_DETAILS_REASONING, thoughts
    completion += thoughts            # adds 3347 again → double count
```

---

## Additional context

- The identical pattern was fixed for `google-genai` in #3013. The fix is the same: emit `thoughts_token_count` as `llm.token_count.completion_details.reasoning` only, without adding it again to the completion total.
- For non-Gemini models (e.g. OpenAI, Anthropic via LiteLLM), `thoughts_token_count` is `None` or `0` so this branch never fires — no impact on other models.
- Worth noting: if a future Google model reports `thoughts_token_count` as a value **not** already included in `candidates_token_count`, this fix would under-count. Flagging for maintainer awareness.

Fung1117:
Closes #3177 

This PR fixes the token-count discrepancy where `llm.token_count.completion` was getting double-counted or under-counted depending on the backend, where the Gemini Developer API already includes thinking tokens in `candidates_token_count`, while Vertex AI (and ADK's LiteLLM wrapper) reports them separately. It now checks which case we're in, using the same `prompt + candidates == total` heuristic LiteLLM uses internally, before deciding whether to add `thoughts_token_count`, and falls back to adding them when there isn't enough info to tell.

google-adk-tokens


Fung1117:

caroger:
Maintainer note: I pushed follow-up commits to tighten Google ADK token accounting, ground the tests in a recorded provider response, and resolve the branch conflicts with current `main`.

Changes pushed:
- `bae24dd4`: include `tool_use_prompt_token_count` in Google ADK prompt-token accounting, and use the full prompt-side total when deciding whether `candidates_token_count` already includes `thoughts_token_count`
- `95e82428`: add a VCR cassette-backed ADK code-execution test that records real Gemini 2.5 Flash usage metadata with `promptTokenCount=75`, `toolUsePromptTokenCount=573`, `candidatesTokenCount=414`, `thoughtsTokenCount=117`, and `totalTokenCount=1179`
- `046714fd`: merge current `main` into the PR branch and resolve the only conflict in `tests/test_instrumentor.py` by keeping both main's new ADK event-output tests and the cassette-backed token-count test
- revise the synthetic usage-metadata tests to avoid provider-specific claims we could not verify from cassettes, while keeping a small schema guard for the inclusive-total edge case

Validation:
- `uvx tox run -e py314-ci-google_adk -- tests/test_get_attributes_from_usage_metadata.py tests/test_instrumentor.py::test_google_adk_instrumentor_code_execution_token_counts`
- after conflict resolution: `uvx tox run -e py314-ci-google_adk -- tests/test_get_attributes_from_usage_metadata.py tests/test_instrumentor.py::test_runner_run_async_keeps_content_output_after_state_delta_final_event tests/test_instrumentor.py::test_base_agent_run_async_keeps_content_output_after_state_delta_final_event tests/test_instrumentor.py::test_google_adk_instrumentor_code_execution_token_counts`

Note: the recorded Google responses we found all use additive accounting for `thoughtsTokenCount`; I did not find a real cassette proving the inclusive-total provider behavior, so the test coverage now separates the hard provider evidence from the defensive schema edge.

caroger (approved):

PR #3195, opened 2026-06-02
fix(agno): Stream Context Leak In Workflow, Step, & Parallel Wrappers

### LINKED ISSUE #3078: [bug][agno] Same async-generator context leak (#2640 fix) still present in _workflow_wrapper.py (Workflow / Step / Parallel) on 0.1.32

ialisaleh:
Closes #3078


ialisaleh:
@claude review

python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py
```diff
@@ -152,18 +152,6 @@ def _workflow_run_arguments(arguments: Mapping[str, Any]) -> Iterator[Tuple[str,
         yield USER_ID, user_id
 
 
-def _setup_workflow_context(node_id: str) -> Any:
```
ialisaleh, line 155: These three functions (`_setup_workflow_context` - `_setup_step_context` - `_setup_parallel_context`) are identical. They are replaced with `_setup_node_context` to avoid confusion about whether they differ in behaviour.

caroger (changes requested):
The stream context lifetime fix looks directionally right, and the new finalization tests are good coverage. I found one regression in the refactor:

Workflow spans no longer capture `agno.run.id`. Before this PR, the Workflow completion paths copied `result.run_id` / `final_response.run_id` onto the span. The new sync, async, sync-stream, and async-stream Workflow finalization blocks still set output, workflow id, and user id, but they dropped the `span.set_attribute("agno.run.id", run_id)` logic. That removes a useful workflow-run identifier from telemetry even though the response still carries it.

Please restore the `agno.run.id` extraction in all Workflow completion paths and add a focused test so this does not regress again.

ialisaleh:
Hi @caroger - I restored the `agno.run.id` extraction in all workflow completion paths. This PR (https://github.com/Arize-ai/openinference/pull/3221) added this attribute with testcases 3 days ago. This working branch was not up-to-date, so I just merged the latest changes from the main branch. Please take a look when you have a chance. Thanks!

caroger (approved):
Thanks for the update.

PR #3199, opened 2026-06-03
fix(claude-agent-sdk): skip session.id write when already propagated via OTel baggage

### LINKED ISSUE #3198: [bug] claude-agent-sdk instrumentor overwrites application session.id with internal Claude CLI session UUID
**Describe the bug**
We use Langfuse v4 (OTel-native) for tracing. Langfuse's propagate_attributes(session_id=...) sets session.id on spans via its SpanProcessor.on_start hook, reading the session from OTel baggage. The ClaudeAgentSDKInstrumentor is registered alongside Langfuse

Langfuse reads session.id from exported spans to assign trace.sessionId, but because the Claude instrumentor overwrites the propagated application session with an internal CLI UUID, traces are incorrectly split into separate auto-generated sessions, breaking correlation with the parent application session even when the session was correctly propagated via OTel baggage.

**To Reproduce**
Set session.id on spans via any standard OpenTelemetry mechanism before the Claude SDK runs:

```
  with propagate_attributes(session_id="dedf7759-99ee-46ad-a5fc-3837892a0d78", as_baggage=True):
      with tracer.start_as_current_span("my-root-span"):
          async with ClaudeSDKClient(options=options) as client:
              await client.query(prompt)
              async for msg in client.receive_response():
                  pass
```


**Expected behavior**
  laudeAgentSDK.ClaudeSDKClient.receive_response retains session.id = "dedf7759-..." through export.


**Actual behavior** 
The OTel span processor confirms session.id was correctly set at span creation:

  DEBUG langfuse: Propagated 3 attributes to span '273fd5b0ecca359d':
    {'session.id': 'dedf7759-99ee-46ad-a5fc-3837892a0d78', ...}

But when the same span is exported after ResultMessage processing completes, the value has been overwritten:

  DEBUG langfuse: Trace: Processing span 'ClaudeAgentSDK.ClaudeSDKClient.receive_response':
    "session.id": "4e00c355-0cb1-4a44-a7ec-50739f9aabcd"

 The ClaudeAgentSDK span is affected, because _extract_init_attributes and _extract_usage_and_cost_attributes call span.set_attribute(SESSION_ID, message.session_id) during message processing.


**Screenshots**
N/A

openinference-instrumentation-claude-agent-sdk  version = "0.1.5"

**Additional context**
 Workaround

  Patch both functions at startup to remove SESSION_ID from their return values before the instrumentor writes to the span:
```
  import openinference.instrumentation.claude_agent_sdk._wrappers as _w
  from openinference.semconv.trace import SpanAttributes

  _sid = SpanAttributes.SESSION_ID
  _orig_init = _w._extract_init_attributes
  _orig_usage = _w._extract_usage_and_cost_attributes

  def _patched_init(msg):
      attrs = _orig_init(msg)
      attrs.pop(_sid, None)
      return attrs

  def _patched_usage(msg):                                                                                                                                                                                                                                      
      attrs = _orig_usage(msg)
      attrs.pop(_sid, None)
      return attrs

  _w._extract_init_attributes = _patched_init
  _w._extract_usage_and_cost_attributes = _patched_usage
```

[demo_openinference_session_overwrite.py](https://github.com/user-attachments/files/28521839/demo_openinference_session_overwrite.py)

BandaruDheeraj:
## Summary

Fixes #3198.

When a caller propagates `session.id` via OTel baggage (e.g. Langfuse's
`propagate_attributes(session_id=..., as_baggage=True)`), a `SpanProcessor`
can write that value on `span.on_start`. However `_extract_init_attributes`
and `_extract_usage_and_cost_attributes` both unconditionally set `SESSION_ID`
from the Claude CLI's internal session UUID mid-stream, overwriting the
application-provided value and breaking trace correlation in Langfuse.

**Root cause:** two unconditional `attributes[SESSION_ID] = session_id` writes
in `_wrappers.py` — one in `_extract_init_attributes` (system init message)
and one in `_extract_usage_and_cost_attributes` (result message).

**Fix:** add a `_baggage_has_session_id()` guard that checks current OTel
baggage before each write. If `session.id` is already in baggage the
instrumentation skips its write, deferring to whatever the SpanProcessor set
on `on_start`. For callers not using baggage the behaviour is unchanged — the
Claude internal UUID is still recorded.

## Test plan

- New regression test `test_session_id_not_overwritten_by_claude_internal_uuid`
  in `test_instrumentor.py` that:
  - Sets `session.id` in OTel baggage (simulating `propagate_attributes`)
  - Wires a Langfuse-style `SpanProcessor` that writes baggage → span on start
  - Replays an existing cassette (no API key needed)
  - Asserts the exported span retains the application session ID, not the Claude UUID
- Test **fails** before this fix and **passes** after (verified in CI)

## Alternatives considered

- **Rename** Claude's session to `claude.session_id` — breaking change, rejected
- **Check span attribute before writing** — no reliable `span.get_attribute()` in public OTel API
- **Only write on init message** — still races with `on_start` processors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
@@ -58,6 +58,26 @@ TOOL_CALL_FUNCTION_NAME = ToolCallAttributes.TOOL_CALL_FUNCTION_NAME
 TOOL_CALL_FUNCTION_ARGUMENTS_JSON = ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON
 
 
+def _baggage_has_session_id() -> bool:
+    """Return True when session.id is already propagated via OTel baggage.
+
+    When a caller sets session.id through OTel baggage (e.g. Langfuse's
+    ``propagate_attributes(session_id=..., as_baggage=True)``), a SpanProcessor
+    may write that value to the span on ``on_start``.  The Claude CLI also
+    emits an internal session UUID on every message, and without this guard
+    ``_extract_init_attributes`` / ``_extract_usage_and_cost_attributes`` would
+    unconditionally overwrite the application session with the internal UUID,
+    breaking trace correlation in Langfuse and similar backends.
+
+    By checking baggage first we only skip the write when the application has
+    explicitly propagated its own session; in all other cases the Claude UUID
+    is recorded as before.
+    """
+    from opentelemetry import baggage
+
+    return bool(baggage.get_baggage(SESSION_ID))
+
+
 def _get_field(obj: Any, key: str, default: Any = None) -> Any:
     if isinstance(obj, MappingABC):
         return obj.get(key, default)
@@ -199,7 +219,7 @@ def _extract_init_attributes(msg: Any) -> dict[str, Any]:
         session_id = _get_field(_get_field(msg, "data", {}), "session_id")
     model = _extract_model_name(msg)
     attributes: dict[str, Any] = {}
-    if session_id:
+    if session_id and not _baggage_has_session_id():
         attributes[SESSION_ID] = session_id
     if model:
         attributes[LLM_MODEL_NAME] = model
@@ -258,7 +278,7 @@ def _extract_usage_and_cost_attributes(msg: Any) -> dict[str, Any]:
         attributes[LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE] = cache_write_tokens
     if (cost := _safe_float(_get_field(msg, "total_cost_usd"))) is not None:
         attributes[LLM_COST_TOTAL] = cost
-    if session_id := _get_field(msg, "session_id"):
+    if (session_id := _get_field(msg, "session_id")) and not _baggage_has_session_id():
         attributes[SESSION_ID] = session_id
     return attributes
 
--- a/python/instrumentation/openinference-instrumentation-claude-agent-sdk/tests/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-claude-agent-sdk/tests/test_instrumentor.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from typing import Any, Callable
 
 import pytest
+from opentelemetry.sdk.trace import SpanProcessor as _BaseSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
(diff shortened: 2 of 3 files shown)


python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -58,6 +58,26 @@
 TOOL_CALL_FUNCTION_ARGUMENTS_JSON = ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON
 
 
+def _baggage_has_session_id() -> bool:
```
ialisaleh, line 61: This helper answers "was session.id put in baggage?" not "does this span already have session.id set?" so any SpanProcessor that writes session.id from a source other than baggage (a custom propagator, W3C TraceContext extensions, a direct `span.set_attribute` call) would still be silently overwritten.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -58,6 +58,26 @@
...
+    unconditionally overwrite the application session with the internal UUID,
+    breaking trace correlation in Langfuse and similar backends.
+
+    By checking baggage first we only skip the write when the application has
+    explicitly propagated its own session; in all other cases the Claude UUID
+    is recorded as before.
+    """
+    from opentelemetry import baggage
+
+    return bool(baggage.get_baggage(SESSION_ID))
```
ialisaleh, line 78: There's also a timing hazard because baggage lives in the OTel context, which in async Python can be detached by the time a result message arrives. If the frame that held the baggage context has already exited, `get_baggage` returns `None` and the overwrite happens anyway.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/tests/test_instrumentor.py
```diff
@@ -778,3 +779,91 @@ async def test_client_real_agent_span(
...
+
+    def shutdown(self) -> None:
+        pass
+
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
+        return True
+
+
+@pytest.mark.asyncio
+async def test_session_id_not_overwritten_by_claude_internal_uuid(
```
ialisaleh, line 817: The test passes because it keeps the baggage context alive for the full duration of the `async with` block, which mirrors the Langfuse happy path but doesn't exercise the async context drift scenario. Checking `span.attributes` directly is the authoritative fix for this issue.

PR #3201, opened 2026-06-04
fix(claude_agent_sdk): Capture Thinking Blocks as Reasoning Message Content

### LINKED ISSUE #3137: [bug][claude-agent-sdk] _get_output_message_attributes ignores ThinkingBlock content blocks
`_get_output_message_attributes` (the helper that turns assistant
messages into `llm.output_messages.{N}.*` flat attributes) only branches
on `_is_tool_use_block` and `_is_text_block`. The Anthropic API's
`ThinkingBlock` (extended thinking) falls into neither, so its `.thinking`
content is dropped and never appears on the AGENT span.

### Reproduction

Run any `claude_agent_sdk.query()` against a model with extended thinking
enabled (e.g. `claude-sonnet-4-6`) and inspect the AGENT span's
`llm.output_messages.{N}.*` attributes - the thinking transcript is
absent even when the assistant message clearly contained one.

### Expected behavior

Extended-thinking content is a first-class part of the assistant
message and should be exported on the span alongside text and
tool_use blocks. There is no OpenInference semconv key for thinking
blocks yet, so the spec discussion is part of the fix.

### Code reference

`python/instrumentation/openinference-instrumentation-claude-agent-sdk/openinference/instrumentation/claude_agent_sdk/_wrappers.py` - the
`_get_output_message_attributes` function.

ialisaleh:
Closes #3137


ialisaleh:
@claude review

caroger (changes requested):
Thanks for adding reasoning block capture. I think the extraction needs one adjustment before merge:

The new reasoning blocks are emitted under `message.contents`, but text blocks still go to legacy `message.content` and tool calls still go to `message.tool_calls`, with the reasoning index counted separately. That loses the provider's original content-block order for mixed outputs such as reasoning -> tool call -> text. The reasoning content convention says to use `message.contents` when ordered provider parts need to be replayed; here, once a Claude assistant message includes thinking/redacted thinking, the adjacent text/tool-use parts should be emitted into the same ordered `message.contents` sequence as well.

The current test also locks in the lossy mixed representation by asserting the text block remains at `message.content.0`; it would be better to update the test to cover a mixed sequence and assert a single content index over the original provider blocks.

ialisaleh:
Hi @caroger - I addressed the content indexing. When thinking blocks are present, all block types now share a single index under `message.contents` so the original provider order is preserved for replay. Legacy paths stay untouched. Tests updated accordingly. Please take a look when you have a chance. Thanks!

caroger (approved):

PR #3202, opened 2026-06-04
fix(pipecat): instrument PipelineWorker for pipecat 1.x

vipyne:
The wrapper targeted PipelineTask, which pipecat 1.x deprecated in favor of
  PipelineWorker. Bots using the recommended class got no observer injection
  and no spans. Switch the wrap target to PipelineWorker.__init__ — the
  deprecated PipelineTask alias is still covered via inheritance (its
  __init__ delegates to super), so the wrap fires exactly once for either
  class.

Fixes silent zero-spans regression introduced in Pipecat 1.0.0.


RogerHYang (approved):

PR #3219, opened 2026-06-07
fix(google_adk): Preserve Final Response Content Output

### LINKED ISSUE #3200: [bug] google-adk: final response output is overwritten and cleared by subsequent state_delta events
**Describe the bug**
In `openinference-instrumentation-google-adk` (observed in `v0.1.15`), the `_RunnerRunAsync` and `_BaseAgentRunAsync` wrappers intercept events and set the span's `OUTPUT_VALUE` when `event.is_final_response()` is `True`.

However, the Google ADK can yield multiple events where `is_final_response()` is `True` during a single invocation (for instance, the actual model response containing content, followed by a state synchronization event like `actions.state_delta` which has no content).

Because the wrapper unconditionally overwrites `OUTPUT_VALUE` on every final response event, the last event (which lacks content) wins. This effectively erases the LLM/Agent's actual text response from the trace output in downstream visualization tools like Langfuse or Arize Phoenix.

**To Reproduce**

1. Run a Google ADK agent that emits state updates or navigation deltas after generating its text response.
2. Observe the invocation span in Phoenix/Langfuse.
3. See error: The output field is empty or only contains the state delta JSON, missing the core `content` object.

**Expected behavior**
The instrumentation should preserve the actual generated content. It should only overwrite the `OUTPUT_VALUE` if the new final event actually contains content, or alternatively, retain the last final response that possessed a non-empty content payload.

**Screenshots**
*N/A*

**Desktop:**

* OS: Linux
* Version: `openinference-instrumentation-google-adk==0.1.15`

**Additional context**

### Proposed Fix

In `openinference/instrumentation/google_adk/_wrappers.py`, track whether an event with content has already been captured during the stream evaluation.

Inside `_RunnerRunAsync` and `_BaseAgentRunAsync`:

```python
has_set_content = False
async for event in self.__wrapped__:
    if event.is_final_response():
        event_has_content = getattr(event, "content", None) is not None
        if event_has_content or not has_set_content:
            if event_has_content:
                has_set_content = True
            try:
                span.set_attribute(
                    SpanAttributes.OUTPUT_VALUE,
                    event.model_dump_json(exclude_none=True),
                )
                # ... rest of the span tracking logic

```

liuzemei:
Closes #3200

It keeps the first content-bearing final Google ADK event as the span output when a later final `state_delta` event has no content, and applies the same guard to `invocation` and `agent_run` spans.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
@@ -26,17 +26,6 @@ from google.adk.models.llm_response import LlmResponse
 from google.adk.tools.base_tool import BaseTool
 from google.genai import types
 from google.genai.types import Blob
-from openinference.semconv.trace import (
-    ImageAttributes,
-    MessageAttributes,
-    MessageContentAttributes,
-    OpenInferenceLLMProviderValues,
-    OpenInferenceMimeTypeValues,
-    OpenInferenceSpanKindValues,
-    SpanAttributes,
-    ToolAttributes,
-    ToolCallAttributes,
-)
 from opentelemetry import context as context_api
 from opentelemetry import trace as trace_api
 from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY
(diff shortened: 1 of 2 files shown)


ialisaleh:
Hi @liuzemei - Please sign the CLA (per the bot comment above) before this can be merged.

This PR resolves the issue. The fix is correctly applied to both `_RunnerRunAsync` and `_BaseAgentRunAsync`, along with `_event_has_content` checking non-empty `parts`, which is a nice improvement over the original proposal. The new tests directly reproduce the reported scenario, verifying that the span output retains the correct content. Thanks!

PR #3227, opened 2026-06-08
fix(litellm): record completion text_tokens as a token count, not a cost (USD) attribute

aaravanmay:
In `_set_token_counts_from_usage` (openinference-instrumentation-litellm), the completion-side `text_tokens` (a token **count**) is written to `SpanAttributes.LLM_COST_COMPLETION_DETAILS_OUTPUT` — i.e. `llm.cost.completion_details.output`, a **USD cost** field (the semconv docstring says values should be in USD). Every sibling in that function writes counts to `llm.token_count.*` (including the parallel prompt-side `text_tokens` and the completion `reasoning_tokens` / `audio_tokens`); only this one branch lands in the cost namespace, which looks like a copy-paste slip.

Effect: any litellm call against a model that returns `completion_tokens_details.text_tokens` (some multimodal / audio models populate it) emits a span where a token count sits under a cost attribute — so a cost dashboard reads e.g. `$500` for 500 text tokens. No exception is raised; it's the only write to that attribute, so nothing masks it.

This PR:
1. adds the missing symmetric semconv constant `LLM_TOKEN_COUNT_COMPLETION_DETAILS_TEXT = "llm.token_count.completion_details.text"` (mirroring the existing `_AUDIO` / `_REASONING` constants),
2. routes the completion `text_tokens` count to that token-count attribute instead of the cost one,
3. adds a regression test (fake span + stub usage, no live LLM) asserting the value lands under the token-count attribute and not under the cost attribute. The existing semconv `test_nesting` stays green.

Happy to rename the constant if you'd prefer a different convention. Found while building a fault-injection tester for agent tools.


aaravanmay:

aaravanmay:
Related context for this PR: I found it with faultline, an open-source deterministic fault-injection tester for agents (it corrupts tool outputs and flags when an agent — or instrumentation — silently records wrong values, like this tokens-as-USD mixup). It feels complementary to observability: you trace what agents did; faultline stress-tests what they do under bad data. If a small regression suite for the litellm instrumentation would be useful, I'm glad to contribute one — happy to follow your contribution process.

mikeldking:
I'm fairly confused about this pr. It's caused the fair amount of thrash on whether or not this is meant to be a breakdown of the token usage given its created some thrash I'm going to close this PR until there is an issue reporting incorrect token calculation. Thanks for the understanding.

PR #3266, opened 2026-06-22
fix(claude_agent_sdk): Invalid Parent for Subagent Tool in PreToolUse Hook

### LINKED ISSUE #3255: claude-agent-sdk: subagent tool calls mis-parented to root agent (PreToolUse hook delivers parent_tool_use_id=None)
## Summary

In `openinference-instrumentation-claude-agent-sdk`, tool calls made **inside a subagent** (spawned via the `Task`/`Agent` tool) are mis-parented to the **top-level agent span** instead of nesting under the subagent's `ClaudeAgentSDK.Agent` span. Result: a multi-subagent run renders as a mostly-flat list of tool spans, defeating subagent grouping in Phoenix.

## Versions

- `openinference-instrumentation-claude-agent-sdk==0.1.6` (latest)
- `claude-agent-sdk==0.2.104`
- `opentelemetry-sdk==1.42.1`, Python 3.12.8

## Root cause

The instrumentor creates tool spans from the injected **`PreToolUse` hook** and parents them via `parent_tool_use_id` read from the hook payload (`_wrappers.py`, `pre_tool_use` -> `_get_field(input_data, "parent_tool_use_id")` -> `start_tool_span(...)`).

But the `PreToolUse` hook payload delivers **`parent_tool_use_id=None`** for subagent tool calls — even though the streamed `AssistantMessage`s for those same tool calls **do** carry the correct `parent_tool_use_id`. So the hook-created span has no subagent parent and falls back to the root agent.

## Evidence

A task that spawns one general-purpose subagent which runs 3 `Bash` calls + 1 `Write`. Same run, comparing the message stream vs. the hook payload vs. the resulting span parent:

| tool | `AssistantMessage.parent_tool_use_id` | `PreToolUse` hook `parent_tool_use_id` | span parent |
|------|------|------|------|
| `Agent` (Task) | `None` | `None` | root agent (correct) |
| `Bash` ×3 (subagent) | `toolu_01NSj9…` (the Agent id) | **`None`** | **root agent (wrong)** |
| `Write` (subagent) | `toolu_01NSj9…` | **`None`** | subagent (raced through the message path) |

The `parent_tool_use_id` the SDK exposes on the message stream is correct and sufficient to nest these tools under the subagent; only the hook payload is missing it.

## Minimal repro

```python
import asyncio
from claude_agent_sdk import (AssistantMessage, ClaudeAgentOptions, ClaudeSDKClient, HookMatcher)
from openinference.instrumentation.claude_agent_sdk import ClaudeAgentSDKInstrumentor

exporter = InMemorySpanExporter(); provider = TracerProvider()
ClaudeAgentSDKInstrumentor().instrument(tracer_provider=provider)

HOOK_LOG = []
async def pre(input_data, tool_use_id=None, context=None):
    g = (lambda d,k: d.get(k) if isinstance(d,dict) else getattr(d,k,None))
    HOOK_LOG.append((g(input_data,"tool_name"), g(input_data,"parent_tool_use_id")))
    return {}

PROMPT = ("Use the Task tool to spawn ONE general-purpose subagent. Instruct it to run three "
          "Bash commands (echo alpha/beta/gamma) then Write subout.txt containing DONE. Then reply 'finished'.")

async def main():
    opts = ClaudeAgentOptions(allowed_tools=["Bash","Write","Task"], max_turns=40,
        system_prompt={"type":"preset","preset":"claude_code"},
        hooks={"PreToolUse":[HookMatcher(hooks=[pre])]})
    c = ClaudeSDKClient(options=opts); await c.connect(PROMPT)
    async for m in c.receive_response():
        if isinstance(m, AssistantMessage):
            for b in m.content:
                if getattr(b,"name",None):
                    print("MSG", b.name, "parent=", getattr(m,"parent_tool_use_id",None))
    await c.disconnect()

asyncio.run(main()); provider.force_flush()
print("HOOK parent_tool_use_id values:", HOOK_LOG)
nm = {s.context.span_id: s.name for s in exporter.get_finished_spans()}
for s in exporter.get_finished_spans():
    print(s.name, "->", nm.get(s.parent.span_id) if s.parent else "ROOT")
```

## Suggested fix

Prefer the **message-stream** `parent_tool_use_id` for tool-span parenting (it is reliably populated in the `AssistantMessage` that introduces each `tool_use` block), rather than relying on the `PreToolUse` hook payload's value — or fall back to the message value when the hook delivers `None`. (Separately, the hook payload arguably *should* include `parent_tool_use_id`, which may warrant an upstream fix in `claude-agent-sdk` itself.)

ialisaleh:
Closes #3255

### DIFF
--- a/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
@@ -386,9 +386,10 @@ def _create_tool_hook_matchers(
             tool_input = _get_field(input_data, "tool_input")
             resolved_tool_use_id = tool_use_id or _get_field(input_data, "tool_use_id")
             parent_tool_use_id = _get_field(input_data, "parent_tool_use_id")
-            # The SDK currently omits parent_tool_use_id from PreToolUse for some
-            # subagent tool calls. Defer those starts to the message stream, where
-            # AssistantMessage.parent_tool_use_id is authoritative.
+            # A missing hook parent is ambiguous: it can be a root-level tool or
+            # a subagent tool whose parent was omitted by the SDK. Starting now
+            # would permanently root the span, so defer no-parent tool starts to
+            # the message stream, where parent_tool_use_id is authoritative.
             if not _has_parent_tool_use_id(parent_tool_use_id):
                 return {}
             tool_tracker.start_tool_span(
@@ -628,14 +629,9 @@ class _ToolSpanTracker(_ToolSpanTrackerBase):
             **get_input_attributes(safe_json_dumps(tool_input), mime_type=JSON),
         }
 
-        resolved_parent_tool_use_id = parent_tool_use_id
         parent_span = self._parent_span
-        if (
-            resolved_parent_tool_use_id is not None
-            and str(resolved_parent_tool_use_id) != ""
-            and self._parent_span_resolver is not None
-        ):
-            resolved = self._parent_span_resolver(resolved_parent_tool_use_id)
+        if _has_parent_tool_use_id(parent_tool_use_id) and self._parent_span_resolver is not None:
+            resolved = self._parent_span_resolver(parent_tool_use_id)
             if resolved is not None:
                 parent_span = resolved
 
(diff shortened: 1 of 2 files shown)


caroger:
Maintainer note on the follow-up changes I pushed to this PR:

I pushed two commits directly to the PR branch to address the remaining parentage issue and the follow-up cleanup comments:

- `f489edae fix(claude_agent_sdk): defer missing-parent hook spans`
  - Removed the stack-based fallback that guessed the parent from the most recent in-flight `Task`/`Agent` tool.
  - Changed `PreToolUse` handling so hook payloads without `parent_tool_use_id` do not start a span immediately. This avoids permanently rooting a subagent tool span when the SDK omits the parent in the hook payload.
  - Lets the existing message-stream path create the tool span instead, where `parent_tool_use_id` is authoritative.
  - Added regressions for concurrent/root-level agent tools and root-level sibling tools so we do not mis-parent spans under the wrong subagent.

- `5b0f7a69 test(claude_agent_sdk): cover deferred hook completion`
  - Cleaned up the helper usage in `start_tool_span` by reusing `_has_parent_tool_use_id()` and removing the dead alias.
  - Expanded the comment around no-parent hook deferral so it is clear this path is intentionally deferred until the SDK provides a parent id.
  - Added coverage for the hook/message interleaving where `PostToolUse` arrives before the deferred message-created span; the later `tool_result` still closes the span and records output.

Validation:

- Local: `uvx --with tox-uv tox -c python/tox.ini r -e py310-ci-claude_agent_sdk` passed with 40 tests.
- GitHub: checks are green at the current head, 18 passing and 0 pending/failing.

Net effect: this keeps the fix aligned with #3255 by avoiding irreversible span creation until the parent relationship is known, while preserving root-level tool sibling behavior and message-stream completion.

caroger (approved):

PR #3270, opened 2026-06-22
fix(agno): guard model_dump_json on non-pydantic run response content

FredAtNeo:
## Problem

`_extract_run_response_output` in `_runs_wrapper.py` calls
`run_response.content.model_dump_json()` unconditionally for any non-`str`
content:

```python
def _extract_run_response_output(run_response: Union[RunOutput, TeamRunOutput]) -> str:
    if run_response and run_response.content:
        if isinstance(run_response.content, str):
            return run_response.content
        else:
            return str(run_response.content.model_dump_json())  # <-- unguarded
    return ""
```

A standalone `agent.run` can return a plain `dict` as `content` (e.g. when the
agent is configured with `output_schema` / JSON mode). In that case the call
raises:

```
'dict' object has no attribute 'model_dump_json'
```

This breaks output capture on the span and, depending on the caller's retry
logic, can fail the run entirely. The sibling helper
`_extract_completed_event_output` (same file) already guards this with a
`hasattr` check — so team-member paths that route through the completed-event
helper are unaffected, but the standalone `agent.run` path crashes.

## Fix

Mirror the existing guard in `_extract_completed_event_output`: serialize via
`model_dump_json()` only when the content exposes it, otherwise fall back to
`str(content)`.

## Tests

Added unit tests for `_extract_run_response_output` covering `str`,
pydantic-like (`model_dump_json`), and `dict` content. The `dict` case
reproduces the crash on the previous code and passes with the fix.

Verified locally:

```
tests/test_instrumentor.py::test_extract_run_response_output_str_content PASSED
tests/test_instrumentor.py::test_extract_run_response_output_pydantic_content PASSED
tests/test_instrumentor.py::test_extract_run_response_output_dict_content PASSED
```

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_runs_wrapper.py
@@ -93,8 +93,9 @@ def _extract_run_response_output(run_response: Union[RunOutput, TeamRunOutput]) 
     if run_response and run_response.content:
         if isinstance(run_response.content, str):
             return run_response.content
-        else:
+        if hasattr(run_response.content, "model_dump_json"):
             return str(run_response.content.model_dump_json())
+        return str(run_response.content)
     return ""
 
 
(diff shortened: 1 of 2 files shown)


FredAtNeo:

mikeldking (approved):
Thank you

PR #3301, opened 2026-06-26
fix(agno)!: Handle Background Workflow Runs

### LINKED ISSUE #3234: [feature request][agno] Workflow span ends prematurely for background runs

When an agno workflow is run with `background=True` (which is what AgentOS uses for POST `/workflows/{id}/runs`), the workflow span that is exported is basically broken:

  - It lasts ~40ms even when the real workflow runs for minutes.
  - `output.value` is the literal string "None".
  - Step, agent, and LLM spans inside the workflow end up as orphans under an already-closed parent.

Currently `Workflow.arun` is wrapped. For background runs, agno's `_arun_background` returns a placeholder `WorkflowRunOutput` (with content=None, status=pending) the moment it kicks off `asyncio.create_task(execute_workflow_background())`. The wrapper treats that placeholder as the
  result, closes the span, and stringifies None into "None". The real work happens later in the background task, after the span is already gone.

The same thing happens for background streaming runs, which go through `_aexecute_stream` instead.

Wrap `Workflow._aexecute` and `Workflow._aexecute_stream`. They're the coroutines that actually run the steps and write the real output onto `workflow_run_response.content`, so a span around them has the right duration, the right output, and the right children.

To avoid double-spanning the normal path (where `arun` already wraps things and then calls `_aexecute` inline), the `arun` span can be skipped when `background=True`, and new wrapper can bail if it's already inside an `arun` span (via `_AGNO_PARENT_NODE_CONTEXT_KEY`).

Instead of avoiding double-spanning, the duplicate child span can just be left on the non-background path.

**Additional context**
N/A

ialisaleh:
Closes #3234

This PR fixes workflow instrumentation for background runs (`background=True`), where the previous `arun` span captured only the short placeholder return instead of the real execution, producing a `"None"` output and orphaned child spans. The fix skips span creation in `arun` when `background=True` and instead wraps `Workflow._aexecute` and `Workflow._aexecute_stream` directly, where the real work and output live. On the foreground path, a task-scoped OpenTelemetry context marker prevents `_aexecute` from double-spanning alongside `arun` while allowing detached background tasks to create their own execution span.

**Breaking change:** raises the minimum supported Agno version from 2.5.0 to 2.7.2.

Agno 2.5.0 uses an older background execution path: async background tasks are unnamed, and `background=True, stream=True` is not available without a WebSocket. Those differences prevent the instrumentor from guaranteeing the foreground/background and streaming/non-streaming context-capture behavior introduced here. Agno 2.7.2 is the earliest version verified end to end across those paths, so retaining `agno>=2.5.0` would advertise support for behavior this release cannot consistently provide. CI now tests both the exact 2.7.2 floor and the latest compatible Agno release on Python 3.10 and 3.14.

**Background Run Trace**
agno-background-run

**Foreground Run Trace**
agno-foreground-run

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/examples/agno_workflow_background_run.py
+++ b/python/instrumentation/openinference-instrumentation-agno/examples/agno_workflow_background_run.py
@@ -1,0 +1,122 @@
+"""
+This example shows how to instrument your agno agent with OpenInference
+and send traces to Arize Phoenix.
+
+Install dependencies:
+pip install openai opentelemetry-sdk opentelemetry-exporter-otlp
+pip install openinference-instrumentation-agno
+"""
+
+import asyncio
+
+from agno.agent.agent import Agent
+from agno.models.openai import OpenAIChat
+from agno.workflow.step import Step
+from agno.workflow.workflow import Workflow
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk import trace as trace_sdk
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+
+from openinference.instrumentation.agno import AgnoInstrumentor
+
+endpoint = "http://127.0.0.1:6006/v1/traces"
+tracer_provider = trace_sdk.TracerProvider()
+tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter(endpoint)))
+
+AgnoInstrumentor().instrument(tracer_provider=tracer_provider)
+
+
+summariser = Agent(
+    name="Summariser",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    role="Summarise the user's input in one sentence.",
+)
+step = Step(name="summarise", agent=summariser)
+workflow = Workflow(
+    name="Background Workflow",
+    description="Demonstrates correct span lifetime for background=True runs.",
+    steps=[step],
+)
+
+
+def _print_section(title: str) -> None:
+    width = 60
+    print(f"\n{'=' * width}")
+    print(f"  {title}")
+    print(f"{'=' * width}")
+
+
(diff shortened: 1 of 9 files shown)


caroger (changes requested):
Thanks for working on the background workflow span lifetime here. I found one correctness issue that I think should be fixed before merging.

The `_oi_arun_spanned` marker is stored on the reusable `Workflow` instance, so one foreground run can suppress span creation for a concurrent background run on the same object. In this PR, `arun` sets the marker on the instance before awaiting the foreground execution (`python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_workflow_wrapper.py:394`), and `_WorkflowExecuteWrapper.aexecute` skips whenever that instance marker is present (`.../_workflow_wrapper.py:580`). I reproduced the overlap case by starting a foreground `workflow.arun(...)` task, then starting `workflow.arun(..., background=True)` on the same `Workflow` before the foreground completed. The background run emitted step spans, but no `Workflow._aexecute` root span; only the foreground `Workflow.arun` root span was present.

Concretely, I think this suppression needs to be scoped to the active foreground coroutine context, not the workflow instance. One fix shape would be:

- replace `_ARUN_SPANNED_ATTR` with an OpenTelemetry context key such as `agno_workflow_arun_spanned`;
- attach that context value only while awaiting the foreground `arun` result, and while advancing the foreground async stream;
- make `_WorkflowExecuteWrapper._already_spanned()` read that context key instead of `getattr(instance, ...)`;
- create/detach the workflow node context inside the returned coroutine/iterator task, so tokens are detached in the same async context where they were attached.

Please also add a regression test that overlaps a slow foreground `arun` with a `background=True` run on the same workflow instance and asserts the background run still gets its `_aexecute` span while the foreground path still avoids double-spanning.

caroger (approved):

PR #3347, opened 2026-07-03
fix(strands_agents): Prevent Non-Strands Span Attribute Overwrite

### LINKED ISSUE #3144: StrandsAgentsToOpenInferenceProcessor destroys attributes on non-Strands spans
## Summary

`StrandsAgentsToOpenInferenceProcessor.on_end()` runs on **every span** in the tracer provider, not just spans originating from the Strands SDK. It replaces `span._attributes` entirely with a new dict, which destroys original attributes on non-Strands spans (botocore, httpx, etc.).

## Problem

The processor's `on_end()` only checks:

```python
if not hasattr(span, "_attributes") or not span._attributes:
    return
```

There is no check for Strands-specific attributes (e.g., `gen_ai.system = "strands-agents"` or `gen_ai.provider.name = "strands-agents"`). Any span with attributes gets processed.

Because the processor does a full replacement (`span._attributes = transformed_attrs`) rather than an additive merge, non-Strands spans lose their original top-level attributes (`rpc.system`, `rpc.service`, `http.status_code`, etc.) — those get packed into a serialized `metadata` JSON blob instead.

## Impact

In our deployment this caused downstream OTel collector filter rules (which match on `rpc.system`, `rpc.service`, etc.) to stop working, since those attributes no longer existed at the top level.

## Suggested fix

Either:

1. **Add a guard** — check for `gen_ai.system == "strands-agents"` or `gen_ai.provider.name == "strands-agents"` at the top of `on_end()` and return early for non-Strands spans.
2. **Use additive merge** — instead of `span._attributes = transformed_attrs`, use `span._attributes = {**span._attributes, **transformed_attrs}` (this is the approach used by `OpenInferenceSpanProcessor` for pydantic-ai).

Option 1 is safer since it avoids any side effects on spans that weren't produced by the Strands instrumentation.

## Workaround

We wrapped the processor in a guard that checks for Strands-specific attributes before delegating `on_end()`:

```python
class StrandsProcessorGuard(SpanProcessor):
    def __init__(self, inner: SpanProcessor) -> None:
        self._inner = inner

    def on_end(self, span: ReadableSpan) -> None:
        attrs = span.attributes
        if attrs and (
            attrs.get("gen_ai.system") == "strands-agents"
            or attrs.get("gen_ai.provider.name") == "strands-agents"
        ):
            self._inner.on_end(span)
```

## Environment

- `openinference-instrumentation-strands-agents` (latest as of 2026-05)
- OpenTelemetry SDK with botocore + httpx auto-instrumentation

ialisaleh:
Closes #3144 

This PR restricts the processor to transform only Strands Agents SDK spans by prioritizing Strands-specific identifiers with limited fallbacks, preventing non-Strands spans from being modified.

non-strands-span

### DIFF
--- a/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
+++ b/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
@@ -147,7 +147,6 @@ class StrandsAgentsToOpenInferenceProcessor(SpanProcessor):
     def _is_strands_span(self, span: ReadableSpan) -> bool:
         """Return True if the span was emitted by Strands Agents SDK."""
         attrs = getattr(span, "_attributes", None) or {}
-
         if (
             attrs.get(GEN_AI_SYSTEM) == _STRANDS_SDK_NAME
             or attrs.get(GEN_AI_PROVIDER_NAME) == _STRANDS_SDK_NAME
@@ -155,13 +154,10 @@ class StrandsAgentsToOpenInferenceProcessor(SpanProcessor):
             return True
 
         name = span.name
-
         if name in _STRANDS_EXACT_SPAN_NAMES:
             return True
-
         if any(name.startswith(prefix) for prefix in _STRANDS_SPAN_PREFIXES):
             return True
-
         if any(substring in name for substring in _STRANDS_SPAN_SUBSTRINGS):
             return True
 
--- a/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
+++ b/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
@@ -109,7 +109,7 @@ class StrandsAgentsToOpenInferenceProcessor(SpanProcessor):
         if not self._is_strands_span(span):
             return
 
-        original_attrs = dict(span._attributes)
+        original_attrs = dict(span._attributes)  # type: ignore[arg-type]
 
         try:
             events: List[Any] = []
--- a/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
+++ b/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
@@ -13,14 +13,7 @@ The processor transforms:
 
 import json
 import logging
-from typing import (
-    Any,
-    Dict,
-    FrozenSet,
-    List,
-    Optional,
-    Tuple,
-)
+from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py
```diff
@@ -38,6 +49,29 @@
 
 logger = logging.getLogger(__name__)
 
+# Rules for identifying spans emitted by the Strands Agents SDK.
```
ialisaleh, line 45: The processor now transforms only Strands Agents SDK spans by first checking Strands-specific `gen_ai.system` & `gen_ai.provider.name` attributes, then falling back to known Strands span names & attributes for spans that do not yet emit those identifiers (for example, `execute_event_loop_cycle`). This prevents non-Strands spans from being modified while preserving existing Strands behavior.

caroger (changes requested):
Thanks for tightening this up. The additive merge is a good improvement, but I think the span classifier still needs one more change before this can safely close #3144.

Right now `_is_strands_span()` checks `gen_ai.system` / `gen_ai.provider.name` first, but if those attributes are present and explicitly identify another GenAI SDK, it still falls through to name-only matches like `chat`. I verified that a span with `name="chat"` and `gen_ai.system="openai"` is still transformed: it gets `openinference.span.kind=LLM`, Strands graph metadata, status `OK`, and its `gen_ai.*` events are stripped. That is still cross-instrumentation mutation in mixed traces.

I also checked the supported `strands-agents==1.19.0` emitter paths. Strands `chat`, `execute_tool ...`, and `invoke_agent ...` spans already carry either `gen_ai.system="strands-agents"` or `gen_ai.provider.name="strands-agents"`. The one current Strands span shape that lacks those identifiers is `execute_event_loop_cycle`, and it carries `event_loop.cycle_id`.

Please make the provider/system attributes authoritative and avoid broad name-only fallbacks. The shape I would expect is roughly:

```python
system = attrs.get(GEN_AI_SYSTEM)
provider = attrs.get(GEN_AI_PROVIDER_NAME)

if system is not None or provider is not None:
    return system == _STRANDS_SDK_NAME or provider == _STRANDS_SDK_NAME

return span.name == "execute_event_loop_cycle" and "event_loop.cycle_id" in attrs
```

Then please add/adjust tests for:

- `name="chat", gen_ai.system="openai"` remains unchanged and retains `gen_ai.*` events/status.
- `name="execute_event_loop_cycle"` without `event_loop.cycle_id` remains unchanged.
- `execute_event_loop_cycle` with `event_loop.cycle_id` still transforms.
- Strands `chat` / tool / agent tests include `gen_ai.system="strands-agents"` or `gen_ai.provider.name="strands-agents"`.

With that change, the processor should preserve the current Strands behavior without reopening the mixed-trace pollution bug.

caroger (approved):

PR #3382, opened 2026-07-13
feat(openai): capture cache write and read token usage

mikeldking:
OpenAI reports billable cache writes separately from cache reads for GPT-5.6 and later models, including GPT-6 Astra. These instrumentors captured `cached_tokens` but omitted `cache_write_tokens`, leaving downstream consumers unable to distinguish ordinary input from cache writes.

Capture `cache_write_tokens` as `llm.token_count.prompt_details.cache_write` alongside existing cache reads in Python and JavaScript OpenAI and OpenAI Agents instrumentation. Preserve explicit zeros, omit unavailable counts, and leave provider-reported prompt/completion/total counts unchanged. Cache reads and writes are subsets of prompt tokens; they must not be added to the prompt total or inferred from cache misses.


js/packages/openinference-instrumentation-openai-agents/src/processor.ts
```diff
@@ -505,6 +507,24 @@ function accumulateChatCompletionUsage({
   } else if (promptTokens !== undefined || completionTokens !== undefined) {
     totals.totalTokens = (totals.totalTokens ?? 0) + (promptTokens ?? 0) + (completionTokens ?? 0);
   }
+  accumulatePromptCacheUsage({ totals, usage });
+  // o-series completion_tokens_details.reasoning_tokens
```
mikeldking, line 532: should this be called accumulate reasoning tokens?

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -797,6 +797,12 @@ function getUsageAttributes(completion: ChatCompletion | Completion): Attributes
       [SemanticConventions.LLM_TOKEN_COUNT_TOTAL]: completion.usage.total_tokens,
       [SemanticConventions.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ]:
         completion.usage.prompt_tokens_details?.cached_tokens,
+      [SemanticConventions.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE]:
```
caroger: ## Capture cache-write usage in Chat Completions streams

The new cache-write extraction only runs for non-streaming Chat Completions. With `stream: true` and `stream_options: { include_usage: true }`, `consumeChatCompletionStreamChunks` skips the final usage-only chunk because `choices` is empty. The guide documents streaming support, but this path still drops the reported counts.

This is the documented Chat Completions streaming contract. With `include_usage: true`, OpenAI sends an additional chunk before `data: [DONE]` containing usage for the entire request and an empty `choices` array. Other chunks carry `usage: null`. See the [official OpenAI API reference](https://developers.openai.com/api/reference/resources/chat). The installed OpenAI JavaScript SDK `6.7.0` documents this same shape, and its stream parser yields the parsed JSON payload directly.

I reproduced this at `5f35c733` using the real OpenAI JavaScript SDK `6.7.0` and instrumentor with a synthetic HTTP/SSE response supplied through a custom `fetch`. The payload and token numbers were constructed for the test, not captured from a live API call. The same usage values, `cached_tokens: 6` and `cache_write_tokens: 4`, reach the caller in both modes. The non-streaming span records both counts, while the streaming span has no token-count attributes. The reproduction follows the documented final-chunk structure; it does not establish which models report `cache_write_tokens`. This is an existing streaming omission that leaves the new feature incomplete.

Suggested fix - Let the existing helper accept stream chunks:

```diff
-function getUsageAttributes(completion: ChatCompletion | Completion): Attributes {
+function getUsageAttributes(
+  completion: ChatCompletion | Completion | ChatCompletionChunk,
+): Attributes {
```

Then process usage before the empty-choices guard in `consumeChatCompletionStreamChunks`:

```diff
   for await (const chunk of stream) {
+    if (chunk.usage) {
+      span.setAttributes(getUsageAttributes(chunk));
+    }
     if (chunk.choices.length <= 0) {
       continue;
     }
```

This reuses the existing mapping for cache reads, cache writes, and token totals. It assigns provider-reported values directly without summing chunks or adding cache counts to prompt totals.

Add a streaming regression test whose final chunk contains:

```ts
{
  choices: [],
  usage: {
    prompt_tokens: 10,
    completion_tokens: 5,
    total_tokens: 15,
    prompt_tokens_details: {
      cached_tokens: 6,
      cache_write_tokens: 4,
    },
  },
}
```

Request `stream_options: { include_usage: true }`, fully consume the stream, and assert all five exported counts.

caroger (approved):

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
mikeldking: Fixed in 92f4c748. `getUsageAttributes` now accepts a `ChatCompletionChunk`, and `consumeChatCompletionStreamChunks` records usage from any chunk that carries it before the empty-`choices` guard, so the final `stream_options.include_usage` chunk lands on the span with prompt, completion, total, cache read, and cache write taken directly from the provider. Added a streaming regression test in `openai.cacheTokens.test.ts` with the final chunk you described (`cached_tokens: 6`, `cache_write_tokens: 4`) asserting all five counts. Thanks for the clear reproduction.

PR #3385, opened 2026-07-13
fix(smolagents): trace streaming model calls

### LINKED ISSUE #1636: [bug] SmolagentsInstrumentor does not create model spans when using CodeAgent

Sanjays2402:
Fixes #1636.

`CodeAgent(stream_outputs=True)` calls `generate_stream()` directly, while the smolagents instrumentor only wrapped `generate()`, so those model calls emitted no LLM span. This wraps concrete streaming implementations, keeps the span open while deltas are consumed, and aggregates them before applying the same output and token attributes as non-streaming calls.

The regression test fails on `main` with zero finished spans and passes with one `OpenAIModel.generate_stream` span; the package suite passes all 49 tests, and the focused test also passes against smolagents 1.26.0.

Streaming CodeAgent trace with the generated LLM span

### DIFF
--- a/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-smolagents/src/openinference/instrumentation/smolagents/__init__.py
@@ -12,6 +12,7 @@ from openinference.instrumentation import (
     TraceConfig,
 )
 from openinference.instrumentation.smolagents._wrappers import (
+    _ModelStreamWrapper,
     _ModelWrapper,
     _RunWrapper,
     _StepWrapper,
@@ -28,6 +29,7 @@ class SmolagentsInstrumentor(BaseInstrumentor):  # type: ignore
         "_original_step_stream_methods",
         "_original_tool_call_method",
         "_original_model_generate_methods",
+        "_original_model_generate_stream_methods",
         "_original_executor",
         "_tracer",
     )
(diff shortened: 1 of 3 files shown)


Sanjays2402:

Sanjays2402:
Recheck

caroger (commented):
Maintainer note: I pushed `0468ba53` to address two streaming lifecycle edge cases found during review:

- keep the LLM span current only while advancing the underlying iterator, so caller work between chunks is not incorrectly parented to that span and subsequent model calls are not suppressed;
- finalize accumulated output attributes when a consumer closes a partially consumed stream.

Both behaviors now have regression coverage that failed against the prior head and passes with the fix. I also ran the full Smolagents package checks against the pinned 1.24.0 and latest 1.26.0 dependencies: 50 tests passed in each environment, with Ruff and mypy clean.

The completed-stream telemetry schema remains unchanged. Thanks @Sanjays2402 for the original fix—this should be ready once the refreshed CI checks are green.

caroger (approved):

PR #3388, opened 2026-07-13
feat(litellm): instrument Anthropic messages create and acreate

### LINKED ISSUE #3378: [feature request] Support LiteLLM's Anthropic methods acreate / create for tracing
I'm using LiteLLM's Anthropic-specific methods (create() / acreate()) and these calls are not captured by the current OpenInference LiteLLM instrumentation. This means I have no observability into these calls when using Phoenix tracing. The current instrumentation only covers completion(), acompletion(), completion_with_retries(), embedding(), aembedding(), image_generation(), and aimage_generation().

Add instrumentation support for LiteLLM's Anthropic-specific methods — create() and acreate() — so that they are automatically traced and appear in Phoenix/OpenInference spans just like the other supported methods.

ardahk:
### TL;DR

LiteLLM’s Anthropic Messages API (`litellm.anthropic.create` / `acreate` and `messages.create` / `acreate`) produced no OpenInference spans — only `completion`, embeddings, and image generation were wrapped. Instruments those entry points (sync/async, streaming, tools, cache token counts, masking) so they show up like other LiteLLM LLM spans. Fixes #3378.

**Files to review (~1.5k lines):**

| File | Why |
|---|---|
| `__init__.py` *(start here)* | Wrappers on `litellm.anthropic_interface`, request attrs, stream finalize, uninstrument. |
| `_anthropic_messages_attributes.py` *(new)* | Input/output content-block attrs + `AnthropicMessagesStreamAccumulator`. |
| `test_anthropic_messages.py` *(new)* | Coverage for create/acreate, streaming, tools, suppress, masking, uninstrument. |
| `README.md` | Lists the new supported methods. |

## Why

Users calling LiteLLM’s Anthropic-native surface get no Phoenix/OI traces today. #3378 asks for parity with the already-instrumented LiteLLM methods.

## How

1. Patch `litellm.anthropic_interface.messages.create` / `acreate` (and the module-level aliases).
2. Record Anthropic-shaped input messages, system, tools, and invocation params on the span.
3. Finalize non-stream responses from dict / `model_dump` payloads; accumulate SSE/stream chunks before setting output and token counts (including Anthropic cache token re-derivation, matching the Anthropic instrumentor).

## Reviewer notes

- **No live Anthropic calls in tests.** Fixtures monkeypatch the underlying create/acreate — no VCR cassettes.
- **Span name is `messages.create`** for both sync and async, matching the Anthropic Messages surface rather than LiteLLM’s function name.
- **Focus area:** stream finalize paths (`__next__` / `__anext__`, awaitable-from-create, split SSE bytes) — easiest place for a missed `span.end()` or dropped usage attrs.

## Tests

```bash
tox run -e test-litellm
# or: pytest python/instrumentation/openinference-instrumentation-litellm/tests/test_anthropic_messages.py
```

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/README.md
+++ b/python/instrumentation/openinference-instrumentation-litellm/README.md
@@ -10,6 +10,8 @@ This package implements OpenInference tracing for the following LiteLLM function
 - aembedding()
 - image_generation()
 - aimage_generation()
+- anthropic.messages.create()
+- anthropic.messages.acreate()
 
 These traces are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize Phoenix](https://github.com/Arize-ai/phoenix) or [Arize AX](https://arize.com/docs/ax).
 
--- a/python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
   "opentelemetry-api",
   "opentelemetry-sdk",
   "opentelemetry-instrumentation",
-  "openinference-instrumentation>=0.1.51",
+  "openinference-instrumentation>=0.1.54",
   "openinference-semantic-conventions>=0.1.30",
   "wrapt",
   "setuptools",
--- a/python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-litellm/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
   "opentelemetry-sdk",
   "opentelemetry-instrumentation",
   "openinference-instrumentation>=0.1.51",
-  "openinference-semantic-conventions>=0.1.25",
+  "openinference-semantic-conventions>=0.1.30",
   "wrapt",
   "setuptools",
 ]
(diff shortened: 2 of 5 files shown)


ardahk:

caroger (commented):
Maintainer note: I pushed two small follow-up commits directly to this PR:

- preserve Anthropic thinking signatures as `message_content.signature`
- record `redacted_thinking.data` as opaque `message_content.data` instead of reasoning text
- accumulate streaming `signature_delta` events
- align the OpenInference dependency minimums with the reasoning semantic conventions and current `main`
- add input and streaming regression coverage

To be explicit about `redacted_thinking`: the block is still represented with
`message_content.type = "reasoning"`, but its opaque provider payload is stored verbatim in
`message_content.data`; `message_content.text` is not emitted for that block. “Opaque” means the
value is preserved for ordered round-tripping rather than interpreted or displayed as readable
reasoning. Normal Anthropic `thinking` blocks continue to use `message_content.text` plus
`message_content.signature`. This matches both the OpenInference reasoning semantic convention and
the direct Anthropic auto-instrumentor's input, output, and streaming mappings.

Validation is green: `ruff`, `mypy`, and all 86 LiteLLM package tests passed locally; GitHub reports 17 passing checks, including Python 3.10 and 3.14, with no failures. The branch is mergeable again.

Commits: `b3bbd8e6`, `052a1d52`.

caroger (approved):

PR #3389, opened 2026-07-14
fix(openai-agents): add input/output and tool metadata to agent, chain, and handoff spans

### LINKED ISSUE #3337: OpenAI Agents parent spans are missing input/output
## Problem identified
OpenAI Agents parent AGENT/CHAIN spans are missing useful input and output content. In a multi-span agent trace, the user prompt and final answer are only visible on the leaf LLM span. Adjacent workflow spans can also share the same `Agent workflow` name, making the trace hard to scan.

## Reproduce
1. Instrument an OpenAI Agents workflow with `OpenAIAgentsInstrumentor`.
2. Run an agent that produces a parent AGENT span, one or more CHAIN spans, and a final LLM span.
3. Inspect the generated trace.
4. Open the parent AGENT/CHAIN spans and compare their `input.value` / `output.value` attributes with the leaf LLM span.

## Expected behavior
Parent AGENT/CHAIN spans should include meaningful input/output summaries when tracing is not configured to hide them. Adjacent workflow spans should be distinguishable by name and/or displayed span kind without requiring users to open each span.

jimbobbennett:
fixes #3337

## Problem

The Agents SDK's own tracing spans carry no input/output on the agent, task, turn, and handoff spans — including the **root span the trace list displays** — so those spans render with an empty Input/Output panel. The instrumentor also omits the tool schema from function spans, leaves handoff spans with no name or input/output, and stamps `llm.system` on every span kind.

A minimal reproduction (one agent, one tool, one question) is here: https://github.com/jimbobbennett/openai-agents-openinference-span-gaps

## Change

`OpenInferenceTracingProcessor` now propagates a child LLM span's input/output up to its ancestor agent/task/turn spans and the trace root (first input wins, latest output wins). The propagated output is the assistant's text, or — for a turn that only made a tool call — the tool call the model requested, rather than the full serialized `Response` object. Alongside that:

- **Function spans** get `tool.description` / `tool.parameters` from the tool schema captured on the response span (the SDK gives function spans only a name, input, and output).
- **Handoff spans** get `input.value` (from-agent), `output.value` (to-agent), and a `tool.name` derived from the SDK's own `transform_string_function_style("transfer_to_<agent>")`, so it matches the tool call the model actually made.
- **`llm.system`** is set only on LLM spans instead of every span.
- **`agent.name`** is added to agent spans (consistent with the Google ADK instrumentor).

Both the Responses API (`ResponseSpanData`) and Chat Completions (`GenerationSpanData`) paths are handled.

## Trace output

Root `AGENT` span before — empty Input/Output:

before

Root `AGENT` span after — user question as input, final answer as output:

after

## Tests

Adds `tests/test_processor_span_attributes.py` (11 tests) covering both API paths: parent/root input/output, the tool-call-turn output fallback, function-span tool schema (both seeded and captured from `response.tools`), handoff-span population, `llm.system` scoping, and `agent.name`.

## Notes

- The handoff `tool.name` reuses `agents.util._transforms.transform_string_function_style`; the import is guarded with an inline fallback in case that helper moves.
- `tool.parameters` is set to the parameter schema, consistent with the smolagents instrumentor and the semantic-convention definition.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/pyproject.toml
@@ -89,6 +89,12 @@ module = [
 module = "tests.test_span_attribute_helpers"
 warn_unused_ignores = false
 
+# TaskSpanData / TurnSpanData are absent from older pinned agents versions, so the
+# type: ignore on their import is required there but unused on newer versions.
+[[tool.mypy.overrides]]
+module = "tests.test_processor_span_attributes"
+warn_unused_ignores = false
+
 [tool.ruff]
 line-length = 100
 target-version = "py38"
--- a/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from collections import OrderedDict
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Iterable, Iterator, Mapping, Optional, Union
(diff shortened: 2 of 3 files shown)


jimbobbennett:
The four failing `*-ci-openai_agents-latest` checks are pre-existing and not introduced by this PR. They come from `openai` 2.45 / `openai-agents` 0.18 type changes:

- `_processor.py` — the existing `assert_never(...)` exhaustiveness checks in `_get_attributes_from_input` / `_get_attributes_from_response_output` don't yet handle the new `program` / `program_output` item types.
- `tests/test_span_attribute_helpers.py` — `InputTokensDetails(...)` is missing the new required `cache_write_tokens` argument.

Both are in existing code this PR doesn't modify, and `main` fails the same two `-latest` jobs today. This PR's own package jobs (`py310-ci-openai_agents`, `py314-ci-openai_agents`) pass.

These are already addressed by #3387; once it merges I'll rebase this branch on `main` and the `-latest` jobs should clear.

PR #3391, opened 2026-07-14
fix(openai_agents): Missing Output Content & Dropped Reasoning Items

### LINKED ISSUE #2882: [bug][openai-agents] LLM span GENERATED section shows {"role": "assistant"} with no content
**Describe the bug**
When instrumenting the OpenAI Agents SDK (Responses API path), the [LLM] span's output shows only `[{"role": "assistant”}]` with no message content, despite the model returning a valid text response.

Root cause (from source inspection of `_processor.py`):
`_get_attributes_from_message` correctly emits `llm.output_messages.N.message.contents.I.message_content.text` for the response text. However, `OUTPUT_VALUE` is set via `response.model_dump_json()` which serialises the full `Response` object. Downstream consumers (e.g. LangWatch UI) that read from a flat `message.content` field find nothing, because output messages use the nested multi-part `message.contents.*` form rather than a flat `message.content` string — unlike input messages which do emit a flat `message.content`.

**To Reproduce**
1. Instrument any OpenAI Agents SDK app with OpenAIAgentsInstrumentor
2. Run a simple text reply (no tools)
3. Inspect the LLM span output in any OTel-compatible UI (LangWatch, etc.)
4. The GENERATED / output section shows `[{"role": "assistant”}]` — role only, no content

**Expected behavior**
Expected: The LLM span output should include the response text, e.g. `[{"role": "assistant", "content": "Hello! How can I help?”}]`
Actual: `[{"role": "assistant”}]`

 - OS: MacOS
 - Version 26.3.1

ialisaleh:
Closes #2882 

This PR fixes three bugs in the OpenAI Agents SDK. First, output messages only emitted the nested multi-part form & never a flat `message.content` string, causing OTel-compatible UIs that read the flat field to display `{"role": "assistant"}` with no visible text even though the model returned a valid response, so we now join text/refusal parts into `message.content` to match the behavior already used for input messages. Second, `reasoning` output items were silently dropped, so reasoning summaries never appeared in traces, so we now emit them, capturing summary/content text, the item `id`, and `encrypted_content` when present. Third, a bug in the `custom_tool_call` branch referencing an unassigned `prefix` variable was also fixed to build its own prefix correctly.

**Example Trace**

openai-agents-trace

**Reasoning Trace**

openai-agents-reasoning


python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
```diff
@@ -819,6 +861,8 @@ def _flatten(
 
 MESSAGE_CONTENT = MessageAttributes.MESSAGE_CONTENT
 MESSAGE_CONTENTS = MessageAttributes.MESSAGE_CONTENTS
+MESSAGE_CONTENT_ID = MessageContentAttributes.MESSAGE_CONTENT_ID
+MESSAGE_CONTENT_ENCRYPTED_CONTENT = MessageContentAttributes.MESSAGE_CONTENT_ENCRYPTED_CONTENT
```
caroger, line 855: Could you update this package's `openinference-semantic-conventions` dependency to the latest version (`>=0.1.30`)? These attributes were added there.

python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
```diff
@@ -769,6 +780,37 @@ def _get_attributes_from_usage(
...
+        yield (
+            f"{prefix}{MESSAGE_CONTENTS}.{idx}.{MESSAGE_CONTENT_TYPE}",
+            "reasoning",
+        )
+        yield (
+            f"{prefix}{MESSAGE_CONTENTS}.{idx}.{MESSAGE_CONTENT_ENCRYPTED_CONTENT}",
+            obj.encrypted_content,
+        )
+        if obj.id:
+            yield f"{prefix}{MESSAGE_CONTENTS}.{idx}.{MESSAGE_CONTENT_ID}", obj.id
```
caroger: This splits one reasoning item across `contents.0` and `contents.1` when both fields are present. This test demonstrates the issue:

```python
item = ResponseReasoningItem(
    id="reason-1",
    type="reasoning",
    summary=[Summary(type="summary_text", text="why")],
    encrypted_content="opaque",
)
attrs = dict(_get_attributes_from_reasoning_item(item))
assert attrs["message.contents.0.message_content.encrypted_content"] == "opaque"
```

Keep all fields on the same content item:

```suggestion
    content_prefix = f"{prefix}{MESSAGE_CONTENTS}.0."
    yield f"{content_prefix}{MESSAGE_CONTENT_TYPE}", "reasoning"
    if texts:
        yield f"{content_prefix}{MESSAGE_CONTENT_TEXT}", "\n\n".join(texts)
    if obj.encrypted_content:
        yield (
            f"{content_prefix}{MESSAGE_CONTENT_ENCRYPTED_CONTENT}",
            obj.encrypted_content,
        )
    if obj.id:
        yield f"{content_prefix}{MESSAGE_CONTENT_ID}", obj.id
```

caroger (changes requested):
Two requested changes.

caroger (approved):

PR #3409, opened 2026-07-20
feat(instrumentation): pluggable BlobUploader to externalize oversized base64 images

caroger:
## Summary

Oversized base64 images on spans today either get destroyed (`__REDACTED__`) or blow spans past OTLP limits. This PR adds a pluggable **`BlobUploader`**: decoded bytes go to external storage at capture time, and the span attribute records only the destination URI — same key, new value.

Scope: **interface + policy only** (no packaged uploader — that's follow-up work, e.g. an fsspec uploader like util-genai's), **images only** (the one settled media convention; other types follow a checklist), **no semconv changes**. Design rationale, OTel GenAI compatibility, and the extension checklist are in the [techspec](https://github.com/Arize-ai/openinference/blob/blob-upload-core/internal_docs/specs/blob_uploads/blob_uploads.md) included in this PR.

## Changes 

1. **Techspec + live demo** (`internal_docs/specs/blob_uploads/`) — the demo runs a real OpenAI vision call against the shipped code (resolved from this repo via a `uv` path source).
2. **Contract** (`openinference-instrumentation`) — `Blob`, the `BlobUploader` protocol, and the `openinference_blob_uploader` entry-point loader.
3. **Policy** — `TraceConfig.blob_uploader` (code or `OPENINFERENCE_BLOB_UPLOADER=<name>`); the existing oversized-image mask branch upgrades from redact-only to **hide → inline budget → upload → redact** (failures always degrade to redaction, never block); sampled-out spans never upload.

## Usage

```python
config = TraceConfig(blob_uploader=my_uploader)  # any object satisfying BlobUploader
```

or zero-code: `OPENINFERENCE_BLOB_UPLOADER=<entry-point name>`.

### DIFF
--- a/internal_docs/specs/blob_uploads/scripts/README.md
+++ b/internal_docs/specs/blob_uploads/scripts/README.md
@@ -1,0 +1,106 @@
+# Blob-upload demo scripts
+
+A live demo of the [blob-upload design](../blob_uploads.md), driving a **real
+auto-instrumented OpenAI vision call** — no hand-built spans. An oversized base64
+image that today rides on the span as `__REDACTED__` is handed to a `BlobUploader` at
+capture time; the span attribute records only the destination URI. The script imports
+the **live** `openinference-instrumentation` package from this repo (a
+`[tool.uv.sources]` path in its PEP 723 metadata), so it exercises the shipped
+`Blob`/`BlobUploader`/`TraceConfig` code — no vendored copies. It prints the resulting
+spans and exports them to Phoenix.
+
+`image_blob_demo.py` makes one chat-completions vision request (text + a ~600 KB
+base64 PNG), instrumented by openinference-instrumentation-openai. The same request
+runs twice, changing only the `TraceConfig`: `message_content.image.image.url` =
+`__REDACTED__` (default) vs a locally served URL (`TraceConfig(blob_uploader=...)`).
+
+`gcs_blob_demo.py` is the same vision call against a **real Google Cloud Storage
+bucket**. Edit the `BASE64_IMAGE_MAX_LENGTH` constant in the script (default
+32,000) — that is the `base64_image_max_length` budget. Each execution runs twice
+with that value — run 1 without an uploader, run 2 with `GcsBlobUploader` — so an
+over-cutoff image shows **redact** then **offload** (an `https://` or signed URL);
+raise the constant above the printed data-URI length (e.g. `1_000_000`) to see
+**inline** in both. Check the two spans in local Phoenix
+(project `blob-upload-gcs-demo`). It authenticates via application-default
+credentials (`gcloud auth application-default login`); set `GCS_BUCKET`, optionally
+`GCS_PREFIX`, and `GOOGLE_CLOUD_PROJECT` if your ADC has no quota project.
+
+The demo store (`LocalBlobStore`, inlined in the image script and explicitly
+subclassing `BlobUploader`) writes content-addressed files under
+`scripts/blob_store/` (gitignored) and serves them from a background HTTP thread,
+recording a real `http://localhost:<port>/<sha>.png` URL — Phoenix renders the image
+inline while the script keeps serving (it waits for Enter before exiting).
+OpenInference ships no uploader implementation, which is exactly why the demos carry
+their own.
+
+The GCS demo records the most viewer-resolvable URL its credentials allow: a V4
+**signed URL** (publicly fetchable until expiry, `GCS_SIGNED_URL_HOURS`) when the
+credentials can sign — a service-account key, or user credentials plus
+`GCS_SIGNING_SERVICE_ACCOUNT=<sa-email>` (IAM signBlob impersonation, requires Token
+Creator on that SA) — otherwise the plain
+`https://storage.googleapis.com/<bucket>/<object>` form, which resolves for viewers
+with bucket access.
+
+## Prerequisites
+
+```bash
+# 1. Phoenix locally
+pip install arize-phoenix
+phoenix serve                    # http://localhost:6006
+
+# 2. OpenAI API key (the script makes real vision calls)
+export OPENAI_API_KEY=...
+
+# 3. (optional) overrides
+export PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006
+export OPENAI_MODEL=gpt-4o-mini          # vision model
+```
+
+Only [`uv`](https://docs.astral.sh/uv/) is otherwise required — dependencies are PEP 723
+inline metadata resolved into an ephemeral environment.
+
+## Run
+
+```bash
+uv run --script internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
+```
+
+```bash
+GCS_BUCKET=my-bucket GCS_PREFIX=me/blobs \
+  uv run --script internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
+```
+
+Edit `BASE64_IMAGE_MAX_LENGTH` in `gcs_blob_demo.py` to try redact / inline / offload.
+
+The script prints the spans it produced (attribute by attribute, long values elided
+with their true size) and exits; the blobs stay under `scripts/blob_store/`.
+
+## What to look at in Phoenix (http://localhost:6006)
+
+**Project `blob-upload-image-demo`** — one `ChatCompletion` LLM span per run, from
+identical app code:
+
+1. First run (default config): the span's input messages show the image part as
+   `__REDACTED__` — today's released behavior for any input image whose base64
+   exceeds 32,000 chars (the only alternative today is raising the budget and
+   carrying ~884 KB of base64 on the span).
+2. Second run (blob-upload config): the same attribute holds
+   `http://localhost:<port>/<sha>.png` — a real URL Phoenix renders inline while the
+   script keeps serving; the bytes are deduped by content hash.
+3. `input.value` is small in both runs — the instrumentor's existing pre-pass strips
+   the base64 image from the serialized request. Upgrading that redaction to a blob
+   URI is future work (step 5 of the techspec's
+   [media-type checklist](../blob_uploads.md#6-checklist-adding-offload-support-for-a-new-media-type)).
+
+## Layout
+
+```
+scripts/
+├── README.md
+├── image_blob_demo.py   — the live TraceConfig(blob_uploader=...) mask() choke point
+│                          (the techspec's shipped integration point), driven by a
+│                          real auto-instrumented OpenAI vision call; local blob store
+├── gcs_blob_demo.py     — same call against a real GCS bucket; edit
+│                          BASE64_IMAGE_MAX_LENGTH to show redact / inline / offload
+└── blob_store/          — content-addressed demo storage (gitignored, created on first run)
+```
--- a/internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
+++ b/internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
@@ -1,0 +1,238 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#     "google-cloud-storage>=2.14",
+#     "openai>=1.60.0",
+#     "openinference-instrumentation-openai>=0.1.52",
+#     "openinference-instrumentation",
+#     "opentelemetry-sdk>=1.42.0",
+#     "opentelemetry-exporter-otlp-proto-http>=1.42.0",
+#     "pillow>=10.0.0",
+# ]
+#
+# [tool.uv.sources]
+# openinference-instrumentation = { path = "../../../../python/openinference-instrumentation" }
+# ///
+"""Blob upload to a real Google Cloud Storage bucket — redact vs inline vs offload.
+
+Edit ``BASE64_IMAGE_MAX_LENGTH`` below (default 32,000) — that is the
+``TraceConfig.base64_image_max_length`` budget. The same vision request runs
+twice with that value:
+
+  run 1 — no uploader        over-cutoff image → ``__REDACTED__``
+  run 2 — GCS uploader       over-cutoff image → an ``https://`` (or signed) URL
+
+The generated image is ~880,000 chars as a data URI, so the default shows
+**redact** then **offload**. Raise the constant above that length (e.g.
+``1_000_000``) to see **inline** data URIs in both runs.
+
+The recorded URI is the most viewer-resolvable form available:
+
+  1. a V4 **signed URL** (publicly fetchable until it expires,
+     ``GCS_SIGNED_URL_HOURS``) when the credentials can sign — a
+     service-account key, or user credentials plus
+     ``GCS_SIGNING_SERVICE_ACCOUNT=<sa-email>`` (requires Token Creator on
+     that SA; signing goes through IAM signBlob);
+  2. otherwise ``https://storage.googleapis.com/<bucket>/<object>`` — a valid
+     URL that resolves for viewers with bucket access (or public buckets).
+
+Requires: OPENAI_API_KEY, GCS_BUCKET (+ optional GCS_PREFIX), and
+GOOGLE_CLOUD_PROJECT if your application-default credentials carry no quota
+project. Uses ADC (``gcloud auth application-default login``). Start Phoenix
+locally (``phoenix serve``) before running.
+
+Run:
+  GCS_BUCKET=my-bucket GCS_PREFIX=me/blobs \\
+    uv run --script internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
+"""
+
+from __future__ import annotations
+
+import atexit
+import base64
+import datetime
+import os
+import random
+import sys
+from io import BytesIO
+from typing import Optional
+
+# Ignore ~/.netrc for this process. Some machines have an entry that makes
+# HTTP clients send Basic auth to storage.googleapis.com, which GCS rejects
+# (401 "HTTP Basic Authentication is not supported") and the span falls back
+# to __REDACTED__. Override with NETRC=/path/to/netrc if you need one.
+os.environ.setdefault("NETRC", os.devnull)
+
+from openai import OpenAI
+from openinference.instrumentation import Blob, BlobUploader, TraceConfig
+from openinference.instrumentation.openai import OpenAIInstrumentor
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from PIL import Image
+
+PROJECT_NAME = "blob-upload-gcs-demo"
+
+# ``TraceConfig.base64_image_max_length``. Default matches today's released
+# budget. Edit this to try other outcomes:
+#   - leave at 32_000 → run 1 redacts, run 2 offloads (an https:// or signed URL)
+#   - set above the printed data-URI length (e.g. 1_000_000) → both runs inline
+BASE64_IMAGE_MAX_LENGTH = 32_000
+
+
+class GcsBlobUploader(BlobUploader):
+    """A ``BlobUploader`` writing content-addressed objects to a GCS bucket.
+
+    Subclassing ``BlobUploader`` is optional (any object with matching
+    ``upload``/``shutdown`` methods satisfies the protocol) but makes the
+    contract explicit. Synchronous for demo clarity — a production
+    implementation moves bytes on a background worker and returns the
+    (precomputable) URI immediately.
+    """
+
+    def __init__(self, bucket: str, prefix: str) -> None:
+        from google.cloud import storage
+
+        client = storage.Client(project=os.environ.get("GOOGLE_CLOUD_PROJECT") or None)
+        self._bucket = client.bucket(bucket)
+        self._prefix = prefix.strip("/")
+        self._signer = _signing_credentials(client)
+        self._signed_url_hours = float(os.environ.get("GCS_SIGNED_URL_HOURS", "24"))
+        if self._signer is None:
+            print(
+                "[gcs] no signing credentials — recording plain https URLs "
+                "(set GCS_SIGNING_SERVICE_ACCOUNT or use a service-account "
+                "key for publicly fetchable signed URLs)"
+            )
+
+    def upload(self, blob: Blob) -> Optional[str]:
+        ext = {"image/png": ".png", "image/jpeg": ".jpg"}.get(blob.mime_type, ".bin")
+        name = f"{self._prefix}/{blob.content_sha256[:20]}{ext}"
+        obj = self._bucket.blob(name)
+        if not obj.exists():  # content-addressed dedup
+            obj.upload_from_string(blob.data, content_type=blob.mime_type)
+            print(
+                f"[gcs] uploaded {len(blob.data):,} B → gs://{self._bucket.name}/{name}"
+            )
+        if self._signer is not None:
+            return str(
+                obj.generate_signed_url(
+                    version="v4",
+                    expiration=datetime.timedelta(hours=self._signed_url_hours),
+                    credentials=self._signer,
+                )
+            )
+        return f"https://storage.googleapis.com/{self._bucket.name}/{name}"
+
+    def shutdown(self, timeout_sec: float = 10.0) -> None:
+        pass  # synchronous demo — nothing pending
+
+
+def _signing_credentials(client: "object") -> Optional[object]:
+    """Signed URLs need a signer: a service-account key signs natively; user
+    credentials can sign via IAM signBlob when impersonating a service
+    account (GCS_SIGNING_SERVICE_ACCOUNT) they hold Token Creator on."""
+    from google.auth import credentials as ga_credentials
+
+    source = client._credentials  # type: ignore[attr-defined]
+    if sa_email := os.environ.get("GCS_SIGNING_SERVICE_ACCOUNT"):
+        from google.auth import impersonated_credentials
+
+        return impersonated_credentials.Credentials(
+            source_credentials=source,
+            target_principal=sa_email,
+            target_scopes=["https://www.googleapis.com/auth/devstorage.read_write"],
+        )
+    if isinstance(source, ga_credentials.Signing):
+        return source
+    return None
+
+
+def make_demo_png() -> bytes:
+    """Seeded RGB noise — incompressible, so the PNG is realistically large."""
+    rng = random.Random(42)
+    img = Image.frombytes("RGB", (640, 400), rng.randbytes(640 * 400 * 3))
+    buf = BytesIO()
+    img.save(buf, format="PNG")
+    return buf.getvalue()
+
+
+def ask_about_image(data_uri: str) -> str:
+    response = OpenAI().chat.completions.create(
+        model=os.environ.get("OPENAI_MODEL", "gpt-4o-mini"),
+        messages=[
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "Describe this image in one sentence."},
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": data_uri, "detail": "low"},
+                    },
+                ],
+            }
+        ],
+    )
+    return response.choices[0].message.content or ""
+
+
+def main() -> None:
+    if not os.environ.get("OPENAI_API_KEY"):
+        sys.exit("OPENAI_API_KEY is not set — this demo makes real vision calls.")
+    gcs_bucket = os.environ.get("GCS_BUCKET")
+    if not gcs_bucket:
+        sys.exit(
+            "GCS_BUCKET is not set — e.g. GCS_BUCKET=my-bucket GCS_PREFIX=me/blobs"
+        )
+
+    png = make_demo_png()
+    data_uri = "data:image/png;base64," + base64.b64encode(png).decode("ascii")
+    print(f"image: {len(png):,} B PNG → {len(data_uri):,} chars as a data URI")
+    print(f"BASE64_IMAGE_MAX_LENGTH: {BASE64_IMAGE_MAX_LENGTH:,}")
+
+    phoenix = os.environ.get(
+        "PHOENIX_COLLECTOR_ENDPOINT", "http://localhost:6006"
+    ).rstrip("/")
+    provider = TracerProvider(
+        resource=Resource.create({"openinference.project.name": PROJECT_NAME})
+    )
+    provider.add_span_processor(
+        SimpleSpanProcessor(OTLPSpanExporter(f"{phoenix}/v1/traces"))
+    )
+
+    uploader = GcsBlobUploader(gcs_bucket, os.environ.get("GCS_PREFIX", "oi-media"))
+    # Implementations own their exit flush (this one is synchronous, but the
+    # pattern matters for real background uploaders).
+    atexit.register(uploader.shutdown)
+    for label, config in [
+        (
+            "run 1: no uploader",
+            TraceConfig(base64_image_max_length=BASE64_IMAGE_MAX_LENGTH),
+        ),
+        (
+            "run 2: GCS uploader",
+            TraceConfig(
+                base64_image_max_length=BASE64_IMAGE_MAX_LENGTH,
+                blob_uploader=uploader,
+            ),
+        ),
+    ]:
+        OpenAIInstrumentor().instrument(tracer_provider=provider, config=config)
+        answer = ask_about_image(data_uri)
+        OpenAIInstrumentor().uninstrument()
+        print(f"\n[{label}] model: {answer}")
+        provider.force_flush()
+
+    provider.shutdown()
+    print(
+        f"\nCheck your local Phoenix at {phoenix} → project {PROJECT_NAME!r}.\n"
+        "Compare the two ChatCompletion spans' message_content.image.image.url:\n"
+        "  run 1 → __REDACTED__ (or a data: URI if you raised BASE64_IMAGE_MAX_LENGTH)\n"
+        "  run 2 → an https:// (or signed) URL (or a data: URI if under the cutoff)\n"
+        "Edit BASE64_IMAGE_MAX_LENGTH in this file and re-run to try other outcomes."
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
+++ b/internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
@@ -1,0 +1,203 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#     "openai>=1.60.0",
+#     "openinference-instrumentation-openai>=0.1.52",
+#     "openinference-instrumentation",
+#     "openinference-semantic-conventions>=0.1.30",
+#     "opentelemetry-sdk>=1.42.0",
+#     "opentelemetry-exporter-otlp-proto-http>=1.42.0",
+#     "pillow>=10.0.0",
+# ]
+#
+# [tool.uv.sources]
+# openinference-instrumentation = { path = "../../../../python/openinference-instrumentation" }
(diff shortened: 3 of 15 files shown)


python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -358,6 +390,21 @@ def mask(
...
 
+    def _externalize_or_redact(self, key: str, value: str) -> AttributeValue:
+        """
+        Uploads an oversized base64 data URI to external storage and returns
+        the destination URI, falling back to redaction when no uploader is
+        configured or the upload cannot be accepted.
+        """
+        if self.blob_uploader is not None:
+            try:
+                if blob := decode_base64_data_uri_to_blob(value, attribute_key=key):
```
mikeldking, line 443: Note for reviewers/future work: in `_externalize_or_redact`, the base64 decode and the sha256 digest (`Blob.__post_init__`) run synchronously on the instrumented call path *before* the uploader gets a chance to refuse the blob. Payload size is effectively bounded by provider request limits today, so this is acceptable for now — but if that stops holding, a core-side size ceiling checked against the cheap base64 string length (pre-decode) belongs here, as a separate knob from `base64_image_max_length` (which is the inline-vs-offload boundary, not an upload cap).

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
```diff
@@ -0,0 +1,175 @@
...
+        the instrumented call path.
+        """
+        ...
+
+    def shutdown(self, timeout_sec: float = 10.0) -> None:
+        """Flushes pending uploads and stops background workers."""
+        ...
+
+
+def load_blob_uploader(name: str) -> Optional[BlobUploader]:
```
mikeldking, line 123: Two lifecycle gaps in the entry-point path:

1. **No memoization.** Every instrumentor constructs its own `TraceConfig()` when none is passed (agno, anthropic, bedrock, … all do), so with `OPENINFERENCE_BLOB_UPLOADER` set, N active instrumentors → N entry-point resolutions → N independent uploader instances, each with its own background workers and upload queue. `load_blob_uploader` should cache the loaded instance per name (module-level dict) so zero-code deployments get a process-wide singleton.

2. **Nobody calls `shutdown()`.** Core defines it on the protocol but never invokes it. Since the contract is URI-now/bytes-later, queued uploads at process exit are silently lost while the exported spans already reference their URIs — dangling references. Rather than core registering an `atexit` hook (flush ordering vs. the span exporter can't be guaranteed from there), the spec's "MUST NOT block" section should state that implementations are responsible for flushing pending uploads at process exit (own `atexit`/daemon-thread flush), following the `BatchSpanProcessor` precedent of the worker-owner owning shutdown.

spec/multimodal_attributes.md
```diff
@@ -53,6 +53,25 @@ llm.input_messages.0.message.contents.2.message_content.type = "audio"
 llm.input_messages.0.message.contents.2.message_content.audio.audio.url = "https://example.com/audio.mp3"
 ```
 
+## External Storage for Large Media
```
mikeldking, line 56: Deliberately deferred (revisit if a second uploader implementation or a generic backend appears): the spec is silent on read-side semantics. A consumer sees `gs://…` in `image.image.url` with no standard way to distinguish an externalized-by-uploader reference from a remote URL the application itself sent to the model, and no stated obligation for how to treat unresolvable schemes. If this bites, the cheap fix is one spec sentence — dereferencing is the trace consumer's responsibility, URIs are not guaranteed publicly resolvable, and a consumer without access treats the value like `__REDACTED__`. A marker attribute distinguishing externalized URLs would be additive and non-breaking, so it can wait.

mikeldking:
**Live GCS demo result** — verified end-to-end in Arize (trace `970f3ec4000ea35dec551fe6c2ac6565`):

✅ **Upload works.** The instrumented OpenAI vision call had its oversized base64 image externalized by the GCS `BlobUploader`, and the span attribute now carries the storage reference instead of the raw data URI:

```
llm.input_messages.0.message.contents.*.message_content.image.image.url
  → gs://arize-imports-1/rogerhu/instrumentation/ee831a3eaa02088cecda.png
```

The span stays small and the completion round-trips fine (`"The image consists of a textured pattern of colorful…"`).

⚠️ **Open question — display behavior.** `gs://` is a storage reference, not a browser-resolvable URL, so the trace UI shows an opaque string and can't render an image preview. The blob *is* reachable at the deterministic https equivalent:

```
https://storage.googleapis.com/arize-imports-1/rogerhu/instrumentation/ee831a3eaa02088cecda.png
```

Per the current `BlobUploader` contract the return value is a "reference URI", which `gs://` satisfies — but we should decide what the *expected* viewer experience is:

1. **Uploader returns the https URL** (`https://storage.googleapis.com/<bucket>/<path>`, or a signed URL for private buckets). Renders immediately in any UI, but for private buckets signed URLs expire, and it bakes storage-provider details into immutable span data.
2. **Keep `gs://` as the canonical reference** and have the consumer (Arize/Phoenix) rewrite/resolve it at render time (`gs://<bucket>/<path>` → `https://storage.googleapis.com/<bucket>/<path>`, with its own credentials if the bucket is private) — durable reference, but requires viewer-side support.
3. **Emit both** — canonical `gs://` reference plus an optional display URL attribute.

What's the expected behavior for displaying this?

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -343,7 +375,7 @@ def mask(
             and MessageContentAttributes.MESSAGE_CONTENT_IMAGE in key
             and key.endswith(ImageAttributes.IMAGE_URL)
         ):
-            value = REDACTED_VALUE
+            value = self._externalize_or_redact(key, value)  # type:ignore
```
mikeldking: **[code-review] Blob uploads fire for suppressed and sampled-out spans**

`mask()` now uploads blobs as a side effect, but its caller `OITracer.start_span` (`_tracers.py:166`, `_get_masked_attributes_for_sampling`) masks start attributes *before* the `_SUPPRESS_INSTRUMENTATION_KEY` check and before the sampler runs. So uploads fire for suppressed and sampled-out spans despite the non-recording guard added in `_spans.py`.

Failure scenario: an app suppresses tracing (or a sampler drops the span) and an instrumentor passes an oversized base64 image in `start_span(attributes=...)` — the image bytes are still uploaded to external blob storage even though no span is recorded. Sensitive media leaves the process while tracing is suppressed, and orphaned blobs accumulate, violating the repo requirement that all instrumentors support suppressing tracing.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -271,6 +294,15 @@ def __post_init__(self) -> None:
...
                 f.metadata["default_value"],
             )
+        if self.blob_uploader is None:
+            self._init_blob_uploader_from_env()
+
+    def _init_blob_uploader_from_env(self) -> None:
+        name = os.getenv(OPENINFERENCE_BLOB_UPLOADER)
+        if not name:
+            return
+        if (uploader := load_blob_uploader(name)) is not None:
```
mikeldking, line 309: **[code-review] Uploader `shutdown()` is never called; pending background uploads are dropped at exit**

The env-var uploader is instantiated here but nothing in the core ever calls `shutdown()` (it exists only as the Protocol stub at `_blob_upload.py:112`; no call site in `OITracer`, `OpenInferenceSpan.end`, or any `atexit` hook).

Failure scenario: with an uploader that returns the URI immediately and moves bytes on a background worker (as the contract encourages), a short-lived process (batch job, CLI, serverless) exits with uploads still queued. The exported spans contain URIs pointing to blobs that were never written — permanent media loss with no error.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -259,9 +269,22 @@ class TraceConfig:
...
+    to external storage and records the destination URI in the span attribute
+    instead of redacting. OpenInference ships no implementation — pass any
+    object satisfying the BlobUploader protocol, or set the
+    OPENINFERENCE_BLOB_UPLOADER environment variable to the name of an
+    uploader registered under the "openinference_blob_uploader" entry-point
+    group."""
 
     def __post_init__(self) -> None:
         for f in fields(self):
+            if f.metadata.get("env_var") is None:
```
mikeldking, line 289: **[code-review] Wrong-typed `blob_uploader` is silently accepted, then swallowed at runtime**

The `env_var is None` skip exempts `blob_uploader` from the construction-time type validation every other `TraceConfig` field gets via `_parse_value`'s `isinstance` check.

Failure scenario: a user writes `TraceConfig(blob_uploader="arize")` — confusing the constructor arg with the `OPENINFERENCE_BLOB_UPLOADER` entry-point name. No `TypeError` is raised; since the value is truthy, `_init_blob_uploader_from_env` is also skipped. At runtime `_externalize_or_redact` calls `"arize".upload(blob)`, the `AttributeError` is swallowed by the broad `except`, and every oversized image is redacted with only a log line — the user believes offloading is configured and loses all media.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -358,6 +390,21 @@ def mask(
             value = REDACTED_VALUE
         return value() if callable(value) else value
 
+    def _externalize_or_redact(self, key: str, value: str) -> AttributeValue:
```
mikeldking, line 428: **[code-review] Start-span image attributes are masked twice → two `upload()` calls per blob**

An oversized image passed via `start_span(attributes=...)` is masked once in `_get_masked_attributes_for_sampling` and again in `openinference_span.set_attributes(user_attributes)`, so `blob_uploader.upload()` runs twice for the same blob and the first returned URI is discarded.

Failure scenario: base64 decode + SHA-256 run twice on the hot path and upload bandwidth doubles. If the uploader doesn't use content-addressed naming (the protocol doesn't require it), the two calls return different URIs — the span records the second while the first upload becomes a permanently orphaned blob.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
mikeldking: **[code-review] `upload()` runs synchronously inside `mask()` on the instrumented call path**

Replacing the unconditional `value = REDACTED_VALUE` with `_externalize_or_redact` means masking an oversized image is no longer constant-time and side-effect-free: `uploader.upload()` runs on the caller's thread, and nothing enforces the Protocol's documented must-not-block contract (`load_blob_uploader` accepts any object with matching method signatures).

Failure scenario: an uploader does network I/O inline and the storage backend is slow or hung — every `set_attribute` carrying a large base64 image blocks the application's own request thread, stalling the user-visible LLM call. On main this path was a constant-time string assignment that could never touch the network.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -271,6 +294,15 @@ def __post_init__(self) -> None:
                 f.metadata["env_var"],
                 f.metadata["default_value"],
             )
+        if self.blob_uploader is None:
+            self._init_blob_uploader_from_env()
```
mikeldking, line 301: **[code-review] Every `TraceConfig()` spawns its own uploader instance (and worker pool)**

`_init_blob_uploader_from_env` runs on every `TraceConfig` construction and `load_blob_uploader` re-scans entry points and instantiates a fresh uploader each call. With `OPENINFERENCE_BLOB_UPLOADER` set, every instrumentor that builds its own default `TraceConfig()` gets an independent uploader with its own background workers.

Failure scenario: an app enabling several instrumentors (openai, langchain, llama-index, ...) gets N uploader instances — N worker pools/connection pools, per-instance dedup state (so identical content uploads N times), and flushing one instance leaves the others' pending uploads to be dropped at exit.

python/openinference-instrumentation/src/openinference/instrumentation/_spans.py
```diff
@@ -40,13 +40,23 @@ def set_attribute(
...
     ) -> None:
+        if key in _IMPORTANT_ATTRIBUTES:
+            # Always bookkeep important attributes (e.g. the OpenInference
+            # span kind) so helpers like set_input keep working even on
+            # non-recording spans.
+            masked_value = self._self_config.mask(key, value)
+            if masked_value is not None:
+                self._self_important_attributes[key] = masked_value
+            return
+        if not self.is_recording():
```
mikeldking, line 53: **[code-review] Non-recording early return silently hides the 'Setting attribute on ended span' warning**

This early return stops forwarding masked attributes to the wrapped span, which removes the OTel SDK warning that previously surfaced post-end writes (`is_recording()` is `False` for ended spans).

Failure scenario: an instrumentor sets attributes (e.g. token counts) after `span.end()`. On main the call reached the SDK's `Span.set_attribute`, which logged the 'Setting attribute on ended span' warning, making the dropped data discoverable. Now the call is silently swallowed — attributes vanish from exported telemetry with no log trail.

python/openinference-instrumentation/tests/test_blob_upload.py
```diff
@@ -0,0 +1,215 @@
...
+    _patch_entry_points(monkeypatch, [_FakeEntryPoint("mem", InMemoryUploader)])
+    monkeypatch.setenv(OPENINFERENCE_BLOB_UPLOADER, "mem")
+    monkeypatch.setenv(OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH, "100")
+    config = TraceConfig()
+    assert isinstance(config.blob_uploader, InMemoryUploader)
+    masked = config.mask(INPUT_IMAGE_URL_KEY, PNG_DATA_URI)
+    assert isinstance(masked, str) and masked.startswith("memory://")
+
+
+def test_blob_uploader_env_var_unset_leaves_none() -> None:
```
mikeldking: **[code-review] Test asserts on ambient environment state — missing `delenv`**

`test_blob_uploader_env_var_unset_leaves_none` doesn't `monkeypatch.delenv("OPENINFERENCE_BLOB_UPLOADER", raising=False)`, so it asserts on whatever the shell happens to export.

Failure scenario: a developer or CI image with `OPENINFERENCE_BLOB_UPLOADER` set runs the suite — `TraceConfig()` attempts to resolve that uploader name and the `config.blob_uploader is None` assertion fails (or passes only because no entry point is installed). Flaky by environment rather than testing the code under test.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
mikeldking: **[code-review] Callable (lazy) attribute values bypass the base64-image size check** (branch at line 369)

`is_base64_url(value)` is `False` for a callable (it checks `isinstance(url, str)`), so the size-check branch at line 369 is skipped for the documented `Callable[[], AttributeValue]` form. Line 391 then evaluates the callable and returns the oversized `data:image/...;base64,...` string verbatim — never uploaded, never redacted.

Failure scenario: an instrumentor passes a lazy value for a `message_content.image.image.url` attribute; the multi-megabyte data URI is emitted inline, blowing past `base64_image_max_length` and OTLP payload limits — exactly what this branch exists to prevent, and the new upload path is never reached.

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
```diff
@@ -0,0 +1,185 @@
...
+PROJECT_NAME = "blob-upload-gcs-demo"
+
+# ``TraceConfig.base64_image_max_length``. Default matches today's released
+# budget. Edit this to try other outcomes:
+#   - leave at 32_000 → run 1 redacts, run 2 offloads to gs://…
+#   - set above the printed data-URI length (e.g. 1_000_000) → both runs inline
+BASE64_IMAGE_MAX_LENGTH = 32_000
+
+
+class GcsBlobUploader:
```
mikeldking: shouldn't this extend a dataclass or some sortof interface? seems entirely opaque as to what methods I need to add here.

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
```diff
@@ -0,0 +1,185 @@
...
+    def upload(self, blob: Blob) -> Optional[str]:
+        ext = {"image/png": ".png", "image/jpeg": ".jpg"}.get(blob.mime_type, ".bin")
+        name = f"{self._prefix}/{blob.content_sha256[:20]}{ext}"
+        obj = self._bucket.blob(name)
+        if not obj.exists():  # content-addressed dedup
+            obj.upload_from_string(blob.data, content_type=blob.mime_type)
+            print(
+                f"[gcs] uploaded {len(blob.data):,} B → gs://{self._bucket.name}/{name}"
+            )
+        return f"gs://{self._bucket.name}/{name}"
```
mikeldking: can we do an example where this is not a gs:// ? because right now in Phoenix it doesn't JUST WORK

Image

mikeldking: the problem here is also we would have to pull the bytes. The public / signed URL would be ideal. That way the image is visible for some time on the platform. 

I think we need an example like this

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def shutdown(self, timeout_sec: float = 10.0) -> None:
```
mikeldking, line 128: I don't think this gets called in the lifecycle for making sure images get flushed. Can we make sure this is wired up? I think dangling image uploads needs some thinking.

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
```diff
@@ -0,0 +1,192 @@
...
+    rng = random.Random(42)
+    width, height = 640, 400
+    img = Image.frombytes("RGB", (width, height), rng.randbytes(width * height * 3))
+    draw = ImageDraw.Draw(img)
+    draw.rectangle([0, 0, width, 56], fill=(16, 24, 48))
+    draw.text((16, 12), "OpenInference blob-upload demo", fill=(240, 240, 255))
+    draw.text((16, 32), "synthetic test pattern (seeded noise)", fill=(160, 170, 200))
+    buf = BytesIO()
+    img.save(buf, format="PNG")
+    return buf.getvalue()
```
mikeldking, line 114: this is not a valid URL. It's a file path. From that perspective it's not semantic and also breaks the UI.

Image

We might want to put validation at the call-site and print diag.warn messages to ensure that the returned string here conforms to a valid URI

mikeldking: I meant this comment to go in the upload method

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def upload(self, blob: Blob) -> Optional[str]:
+        name = blob.content_sha256[:20] + _EXT_BY_MIME.get(blob.mime_type, ".bin")
+        path = self.root_dir / name
+        if not path.exists():  # content-addressed dedup
+            path.write_bytes(blob.data)
+            print(f"[blob] stored {blob.modality} ({len(blob.data):,} B) → {path.name}")
+        if self._repo_root is not None:
+            return str(path.resolve().relative_to(self._repo_root))
+        return path.resolve().as_posix()
```
mikeldking: not a valid URL. we would need to serve that directory via some sort of web-server and paste in the URL here. 

If we keep this, we need to make it work e2e. I get that this is meant to be illustrative, but it's not a valid URL

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
```diff
@@ -0,0 +1,192 @@
...
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+from PIL import Image, ImageDraw
+
+PROJECT_NAME = "blob-upload-image-demo"
+
+_EXT_BY_MIME = {"image/png": ".png", "image/jpeg": ".jpg"}
+
+
+class LocalBlobStore:
```
mikeldking: @axiomofjoy had the same question - is this protocol based?

internal_docs/specs/blob_uploads/scripts/README.md
```diff
@@ -0,0 +1,99 @@
...
+(`gcloud auth application-default login`); set `GCS_BUCKET`, optionally `GCS_PREFIX`,
+and `GOOGLE_CLOUD_PROJECT` if your ADC has no quota project.
+
+The demo store (`LocalBlobStore`, ~25 lines inlined in the image script) is deliberately
+simple: it satisfies the shipped `BlobUploader` protocol, writes content-addressed
+files under `scripts/blob_store/` (gitignored), and returns the file's
+**repo-root-relative path** as the URI, e.g.
+`internal_docs/specs/blob_uploads/scripts/blob_store/3a7bd3….png`. Phoenix displays
+the URI as an ordinary string attribute — resolving or rendering it is the backend's
+responsibility. OpenInference ships no uploader implementation, which is exactly why
```
mikeldking: problem with this is we don't know how to render it. If it had a protocol prefix I would know what to do. E.g. http://localhost:<hosting-port>/my-images.

I think we should enforce valid URL syntax https://www.w3.org/TR/url/

We can't expect just any string to work. The convention is named URL after all

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
```diff
@@ -0,0 +1,175 @@
...
+__all__ = (
+    "Blob",
+    "BlobUploader",
+    "load_blob_uploader",
+    "parse_base64_data_uri",
+)
+
+BLOB_UPLOADER_ENTRY_POINT_GROUP = "openinference_blob_uploader"
+
+_DATA_URI_PATTERN = re.compile(r"^data:(?P<mime>[^;,]+);base64,(?P<content>.+)$", re.DOTALL)
```
mikeldking, line 46: export and test

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
```diff
@@ -0,0 +1,175 @@
...
+    """Hex digest of the decoded bytes. Computed automatically."""
+
+    def __post_init__(self) -> None:
+        if not self.modality:
+            object.__setattr__(self, "modality", _modality_from_mime_type(self.mime_type))
+        if not self.content_sha256:
+            object.__setattr__(self, "content_sha256", hashlib.sha256(self.data).hexdigest())
+
+
+@runtime_checkable
```
mikeldking, line 99: okay I see this here - but can we sub-class it at the instantiation site to make it more clear? Even if runtime checkable it feels a bit non-declarative.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
```diff
@@ -271,6 +294,15 @@ def __post_init__(self) -> None:
...
+
+    def _init_blob_uploader_from_env(self) -> None:
+        name = os.getenv(OPENINFERENCE_BLOB_UPLOADER)
```
mikeldking, line 306: might change the BLOB_UPLOADER_URL for clarity. Also centralize config so all env parsing is done and validated in one file. 

You should make sure that the env is a valid URL to avoid foot-guns

spec/multimodal_attributes.md
```diff
@@ -53,6 +53,23 @@ llm.input_messages.0.message.contents.2.message_content.type = "audio"
...
+
+Inline base64 payloads can exceed OTLP message limits and inflate backend storage. As an experimental capability, instrumentations MAY externalize oversized images at capture time: upload the decoded bytes to configured blob storage and record the destination URI in the same `image.image.url` attribute where the data URI would have been recorded.
+
+```
+llm.input_messages.0.message.contents.1.message_content.type = "image"
+llm.input_messages.0.message.contents.1.message_content.image.image.url = "s3://my-bucket/oi-media/3a7bd3e2....png"
```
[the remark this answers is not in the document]
caroger: Fixed — the example now notes that output-message images are externalized identically (`llm.output_messages.*` line added).

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Acknowledged in code: `_externalize_or_redact` now carries this note verbatim (pre-decode ceiling on the base64 string length as a future knob, separate from the inline-vs-offload budget). No behavior change, per the note.

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
caroger: Both addressed: (1) `load_blob_uploader` now memoizes per name (module-level cache), so zero-code deployments share one process-wide uploader instance — test `test_env_var_configs_share_one_uploader`; (2) per your framing, core does not register `atexit` — the spec now states implementations own their exit flush (BatchSpanProcessor precedent), and both demo uploaders model the `atexit.register(uploader.shutdown)` pattern.

spec/multimodal_attributes.md
caroger: Took the cheap fix now rather than deferring: the spec states dereferencing is the trace consumer's responsibility, URIs are not guaranteed publicly resolvable, and a consumer without access treats the value as it would `__REDACTED__`. Also added the requirement that uploaders return the most consumer-resolvable URI available.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Fixed: `_get_masked_attributes_for_sampling` now masks with a pure, side-effect-free view (`mask(externalize=False)`, with a signature-tolerant fallback for `TraceConfig` subclasses that override `mask`), so suppressed and sampled-out spans never trigger uploads; the uploader-enabled mask runs post-creation in `OpenInferenceSpan.set_attribute`, which already skips non-recording spans. Tests: `test_suppressed_span_skips_upload`, `test_sampled_out_start_attributes_skip_upload`.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Resolved per the framing in the lifecycle thread: core deliberately does not call `shutdown()` (flush ordering against the span exporter can't be guaranteed from core), and the spec now requires implementations to own exit-flush — `atexit`/daemon drain, BatchSpanProcessor-style. Both demos register `atexit.register(uploader.shutdown)` to model the pattern.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Fixed: `TraceConfig` now validates at construction — `blob_uploader="arize"` raises `TypeError` pointing at `OPENINFERENCE_BLOB_UPLOADER` / `load_blob_uploader(name)`, a class raises "pass an instance", and an object missing the protocol methods raises with the expected signature. Test: `test_trace_config_rejects_invalid_blob_uploader` (string/class/wrong-object).

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Fixed by the same root-cause change as the sampling issue: the pre-sampling mask is now side-effect-free, so `upload()` runs exactly once per attribute (in `set_attribute`, post-creation). Test: `test_start_span_attributes_upload_exactly_once` asserts one upload and that the exported span carries the URI.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: True but unenforceable at the protocol layer — any object with matching signatures can block. Mitigations now in place: the contract language (URI now, bytes in background), the uniform degrade path, call-site URI validation, and the sampling fix removing the pre-sampling exposure. The demo uploaders are deliberately synchronous mocks and say so in their docstrings.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Fixed via per-name memoization in `load_blob_uploader` — every `TraceConfig` constructed in the process shares one uploader instance (one worker pool, one queue). Tests: `test_load_blob_uploader_is_memoized`, `test_env_var_configs_share_one_uploader`.

python/openinference-instrumentation/src/openinference/instrumentation/_spans.py
caroger: Fixed: `OpenInferenceSpan` now tracks ended state; post-end writes forward the (side-effect-free) masked value to the wrapped span, so the SDK's "Setting attribute on ended span" warning surfaces exactly as on main, while sampled-out (never-ended) spans still skip silently with no upload side effects. Test: `test_ended_span_attribute_write_still_warns`.

python/openinference-instrumentation/tests/test_blob_upload.py
caroger: Fixed — the test now `monkeypatch.delenv("OPENINFERENCE_BLOB_UPLOADER", raising=False)` before asserting.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: Fixed: lazy values now resolve before the size check, inside the image-key branch — so hidden attributes keep their laziness (there's a test asserting a hidden callable is never evaluated). Worth noting the bypass predates this PR: on main, lazy image values skipped plain redaction the same way. Tests: `test_mask_externalizes_callable_value`, `test_mask_redacts_callable_value_without_uploader`, `test_hidden_callable_value_is_never_evaluated`.

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
caroger: Done — `class GcsBlobUploader(BlobUploader):` now explicitly inherits the runtime-checkable Protocol (legal and gives static checking), with a docstring note that subclassing is optional but makes the contract declarative.

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
caroger: The demo now records the most viewer-resolvable form instead of `gs://`: a V4 signed URL when the credentials can sign (service-account key, or `GCS_SIGNING_SERVICE_ACCOUNT=<sa-email>` impersonation via IAM signBlob — publicly fetchable until `GCS_SIGNED_URL_HOURS` expiry), else `https://storage.googleapis.com/<bucket>/<object>`. The spec now mandates "most consumer-resolvable URI available"; resolving canonical storage-scheme URIs at render time is proposed as Phoenix/Arize follow-up (see the display-semantics reply on the PR).

internal_docs/specs/blob_uploads/scripts/gcs_blob_demo.py
caroger: Wired: the demo registers `atexit.register(uploader.shutdown)`, and the spec now assigns exit-flush responsibility to implementations (core can't guarantee ordering against the exporter). The dangling-URI window and its bounds are recorded in the async-model section.

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
caroger: Done exactly as suggested, at the call site: `_externalize_or_redact` validates the returned reference (absolute URI with a scheme, no whitespace) via `is_valid_reference_uri` and logs a warning + redacts on failure — bare file paths can no longer reach span attributes. Test: `test_mask_redacts_when_uploader_returns_invalid_uri`.

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
caroger: Made it work e2e: `LocalBlobStore` now serves `blob_store/` from a background `http.server` thread and records real `http://localhost:<port>/<sha>.png` URLs — Phoenix renders the image inline, and the script keeps serving until you press Enter.

internal_docs/specs/blob_uploads/scripts/image_blob_demo.py
caroger: Yes — `BlobUploader` is a `@runtime_checkable` Protocol (duck-typing satisfies it; the loader validates with `isinstance`). For clarity both demo uploaders now subclass it explicitly, and the docstrings note that style as recommended-but-optional.

internal_docs/specs/blob_uploads/scripts/README.md
caroger: Agreed and enforced: the core call site now validates the returned reference is a valid absolute URI (scheme required) and redacts otherwise, and the spec requires the most consumer-resolvable form available (https/signed where possible; storage-scheme URIs are valid canonical refs but need viewer-side resolution). The local demo now records `http://localhost:<port>/…` URLs it actually serves.

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
caroger: Done: `parse_base64_data_uri` and `decode_base64_data_uri_to_blob` are exported from `openinference.instrumentation` and directly tested. Decode is also stricter now — `b64decode(validate=True)` rejects malformed payloads instead of silently producing an empty blob.

python/openinference-instrumentation/src/openinference/instrumentation/_blob_upload.py
caroger: Done — both demo uploaders now declare `class …(BlobUploader):` at the definition site. Kept the core type a Protocol (so third-party implementations don't need to import openinference), with explicit inheritance documented as the recommended style.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
caroger: `OPENINFERENCE_BLOB_UPLOADER` holds an entry-point *name* (mirroring util-genai's `OTEL_INSTRUMENTATION_GENAI_COMPLETION_HOOK` selection), not a URL — so URL validation doesn't apply to that variable. The URL validation now lives where URLs actually flow: the uploader's returned reference is validated at the call site (warn + redact on failure). Its env parsing sits in config.py with all the other OPENINFERENCE_* vars; uploader-specific settings like a base URL remain the uploader's own env vars per the spec.

caroger:
**Display semantics — decision** (answering the question in the live GCS run report above):

**Option 1 as the attribute-level contract, option 2 as consumer-side follow-up, option 3 rejected.**

- **Shipped now (option 1):** the spec requires uploaders to return the **most consumer-resolvable URI available** — an `https://` or signed URL where possible; storage-scheme URIs (`gs://`, `s3://`) remain *valid* canonical references but require viewer-side support. Core validates the returned reference at the call site (absolute URI with a scheme; invalid → warn + redact), and the GCS demo now records a V4 signed URL when its credentials can sign (`GCS_SIGNING_SERVICE_ACCOUNT` impersonation or an SA key), else `https://storage.googleapis.com/<bucket>/<object>`. Signed-URL expiry is the trade-off the spec acknowledges: expired references degrade to the same consumer experience as `__REDACTED__`.
- **Follow-up (option 2):** teaching Phoenix/Arize to resolve canonical storage-scheme URIs at render time (`gs://` → authenticated fetch with the viewer's/platform's credentials) — durable references without baking storage details into span data. That's consumer-side work, tracked separately.
- **Rejected (option 3):** emitting both a canonical ref and a display URL means two attributes for one fact — they drift, and every consumer has to decide which to trust.

python/openinference-instrumentation/src/openinference/instrumentation/config.py
mikeldking: Gotcha.

mikeldking (approved):

PR #3408, opened 2026-07-20
docs: experimental blob upload tech spec

caroger:
## Summary

Design-only techspec (+ runnable demos) for capturing **large multimodal content** (audio, images; files/PDF covered by the same conventions) in OpenInference spans without inlining multi-megabyte base64 payloads: a pluggable `BlobUploader` uploads the decoded bytes to external storage at capture time and the span attribute records only the destination URI.

Full design: [`internal_docs/specs/blob_uploads/blob_uploads.md`](https://github.com/Arize-ai/openinference/blob/blob-upload/internal_docs/specs/blob_uploads/blob_uploads.md)

## Why

Inline base64 doesn't survive production traffic:

| today's option | consequence |
|---|---|
| default truncation (audio) / redaction (image) | content destroyed — a 3.2 s voice question keeps ~0.5 s of unplayable audio; a >32 KB image becomes `__REDACTED__` |
| raise the max-length env vars | a 663 KB PNG becomes an ~884 KB attribute; multi-MB spans stress OTLP limits, collectors, and span stores |
| hide flags | no observability at all |

The forcing function is openai-agents realtime audio tracing (#3173): 24 kHz PCM16 ≈ 64 KB/s of base64 per side.

## What the spec proposes

- **`Blob` / `BlobUploader` protocol — interface and policy only, no packaged transport** (scope reduced per the 2026-07-20 maintainer review) — `upload(blob) -> Optional[str]` returns a content-addressed URI synchronously; bytes travel in the background. Implementations come from applications, vendor SDKs (entry-point group `openinference_blob_uploader`, selected by `OPENINFERENCE_BLOB_UPLOADER`), or a future upstream util-genai byte uploader.
- **`TraceConfig` integration**: new `blob_uploader`, `base64_media_max_length`, `hide_input_audio`/`hide_output_audio`/`hide_input_files` fields (+ env vars). One policy at the existing `mask()` choke point — **hide → fits-inline-budget → upload → `__REDACTED__`** (uniform fallback; never blocks) — plus a direct capture-time API for raw-bytes instrumentors (realtime).
- **One `Blob` per media content part**; a single uploader serves every mime type (route/refuse per part is uploader policy).
[...]

PR #3410, opened 2026-07-20
feat(openai): capture audio and file content, externalize via BlobUploader

caroger:
## Summary

The OpenAI instrumentor layer of the blob-upload techspec — **stacked PR 4 of 4** (base: `blob-upload-core`, #3409; vocabulary: #3450; design: #3408).

Before this PR the OpenAI instrumentor did not capture audio or file content at all (`input_audio`, `file`, `input_file`, and audio-modality responses were TODOs). This PR captures them in the standard `message_content` shapes, where the core `TraceConfig` policy from #3409 (**hide → inline-budget → upload → `__REDACTED__`**) applies automatically.

## Capture surfaces

| surface | content | attributes emitted |
|---|---|---|
| Chat Completions request | `input_audio` part | `message_content.type="audio"`, `audio.url` (data URI), `audio.mime_type` |
| Chat Completions request | `file` part | `message_content.type="file"`, `file.url` (data URI from `file_data`), `file.mime_type` (guessed from `filename`), `file.name`, `file.id` |
| Chat Completions response | `message.audio` | `audio.url` (data URI; format from the request's `audio.format`), `audio.mime_type`, `audio.transcript` |
| Responses API request | `input_audio` part | as chat `input_audio` |
| Responses API request | `input_file` part | `file.url` (`file_url` passthrough or data URI from `file_data`), `file.mime_type`, `file.name`, `file.id` |

Provider file ids (`file.id`) are references — bytes are never captured for provider-hosted files.

**The `input.value` pre-pass**: instrumentors serialize the whole request into `input.value` before it reaches `mask()`, so `redact_media_from_request_parameters` (analogous to the existing image redaction pass) applies the same hide/externalize/redact policy — through the same uploader — to audio/file payloads inside the serialized request. Content addressing collapses the double capture (structured attribute + `input.value`) into one object and one URI.

## tox note

The `openai` env now installs the in-repo core/semconv packages before its import sanity check, since this PR's media capture imports blob-upload APIs (`openinference.instrumentation._blob_upload`, `FileAttributes`) that are not yet on PyPI. Both local paths install in a single `uv pip install` so the resolver doesn't clobber one with a PyPI version.

## Testing

- `test-openai`: **500 passed** (16 new tests in `test_multimodal_media.py` covering all five capture surfaces plus `input.value` redaction/externalization)
- `ruff-openai` / `mypy-openai`: clean (no issues in 32 source files)

## Follow-ups (tracked in the techspec)

Images inside `input.value` (upgrade the existing redaction pass to externalize), image-generation output, speech/transcription endpoints, openai-agents realtime migration.


caroger:
Deferring per scope decision: blob upload ships **image-only** first (see #3409). The audio/file capture in this PR depends on the deferred vocabulary (#3450), so it moves with it — the draft is preserved on the `blob-upload-openai` branch and is tracked as open question 6 in the techspec (#3408).

PR #3420, opened 2026-07-22
fix(crewai): emit TOOL spans for @tool-decorated tools

### LINKED ISSUE #3419: [bug] CrewAI @tool-decorated tools produce no TOOL span (only BaseTool.run is instrumented)
**Describe the bug**

`openinference-instrumentation-crewai` does not emit a `TOOL` span when an agent calls a tool defined with CrewAI's `@tool` decorator. The trace contains the crew (`CHAIN`), the agent (`AGENT`) and the LLM calls (`LLM`), but the tool invocation is invisible — there is no span with `openinference.span.kind == "TOOL"`, and so no tool input/output is recorded, even though the tool actually runs.

The instrumentor only wraps `crewai.tools.base_tool.BaseTool.run`. Tools created with `@tool` are `crewai.tools.base_tool.Tool` instances, and `Tool` overrides `run()` without calling `super().run()`. CrewAI 1.x's native tool-calling loop registers each tool's bound `run` as the callable it invokes, so for a `@tool` tool the unwrapped `Tool.run` is what executes and the instrumentation never fires.

**To Reproduce**

Minimal, self-contained reproduction (one agent, one `@tool`, one task): https://github.com/jimbobbennett/crewai-openinference-tool-span-repro

```bash
python3 -m venv .venv && source .venv/bin/activate
pip install -r requirements.txt
cp .env.example .env   # add ANTHROPIC_API_KEY
python main.py
```

The script prints the span kinds seen in the run. `CREWAI_DISABLE_TELEMETRY=true` is set so CrewAI's own (kind-less) telemetry spans don't obscure the result. Observed output:

```
=== Span kinds seen this run ===
  LLM          messages.create
  LLM          messages.create
  AGENT        Friendly Greeter._execute_core
  CHAIN        Crew_....kickoff

No span with openinference.span.kind == "TOOL" was emitted, even though the agent called the current_time tool.
```

**Expected behavior**

A `TOOL` span for the `@tool` invocation, with its input and output, as already happens for class-based `BaseTool` subclasses.

**Screenshots**

The same run exported to Arize — crew, agent and LLM spans, but no tool span:

https://github.com/jimbobbennett/crewai-openinference-tool-span-repro/blob/main/no-tools.png

 - OS: macOS
 - openinference-instrumentation-crewai 1.1.10
 - crewai 1.14.2 (also reproduces on 1.10.1 and 1.15.5)
 - Python 3.12

**Additional context**

Root cause, with file/line references:

- The instrumentor wraps `BaseTool.run` only (`openinference/instrumentation/crewai/__init__.py`, ~line 234).
- `crewai.tools.base_tool.Tool` overrides `run()` and calls `self.func(...)` directly — it never calls `BaseTool.run`.
- `crewai/utilities/agent_utils.py:211` does `available_functions[sanitized_name] = tool.run`, i.e. the native loop invokes the bound `Tool.run`, which shadows the wrapped `BaseTool.run`.

Wrapping `Tool.run` in addition to `BaseTool.run` covers the `@tool` path. I'll open a PR with that fix and a regression test.

jimbobbennett:
# Description

Tools defined with CrewAI's `@tool` decorator produced no `TOOL` span. The instrumentor only wrapped `crewai.tools.base_tool.BaseTool.run`, but `@tool` returns a `Tool` instance that overrides `run()` without calling `super().run()`. CrewAI 1.x's native tool-calling loop registers each tool's bound `run` as the callable it invokes (`crewai/utilities/agent_utils.py:211`), so for a `@tool` tool the unwrapped `Tool.run` is what executes and the wrapper never fires.

This wraps `Tool.run` in addition to `BaseTool.run`, reusing the existing `_BaseToolRunWrapper`. The extra wrap is guarded on `Tool` actually overriding `run`, so on crewai versions where `Tool` inherits `run` unchanged we don't double-wrap it. `_uninstrument` restores the original `Tool.run`.

Adds a regression test (`test_tool_decorator_run_emits_tool_span`) that runs a `@tool` tool and asserts a single `TOOL` span with the expected name, input and output. The test fails against the previous behavior and passes with the fix; verified on crewai 1.10.1, 1.14.2 and 1.15.5, and the full crewai instrumentation suite passes.

Minimal reproduction: https://github.com/jimbobbennett/crewai-openinference-tool-span-repro

Resolves #3419

# Checklist:

### DIFF
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -63,6 +63,10 @@ _agent_kickoff_active: contextvars.ContextVar[bool] = contextvars.ContextVar(
 # one invocation. The inner wrapper skips span creation only when re-entered for
 # the *same* instance, so a tool whose body legitimately calls another tool's
 # ``run`` still gets its own span.
+#
+# Being a ContextVar, the guard is scoped to the running task/thread. If a tool's
+# run were dispatched to a different thread, the delegation would not be de-duped
+# there; the fallout is at worst one duplicate span, never a dropped one.
 _tool_run_instance: contextvars.ContextVar[Optional[int]] = contextvars.ContextVar(
     "_oi_tool_run_instance", default=None
 )
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -57,6 +57,14 @@ _agent_kickoff_active: contextvars.ContextVar[bool] = contextvars.ContextVar(
     "_oi_agent_kickoff_active", default=False
 )
 
+# Contextvar used to suppress a nested, duplicate TOOL span. A tool's ``run`` may
+# delegate to ``BaseTool.run`` (e.g. via ``super().run()``); both are wrapped, so
+# this is set while a tool-run span is active and lets the inner wrapper skip
+# creating a second span for the same invocation.
+_tool_run_in_progress: contextvars.ContextVar[bool] = contextvars.ContextVar(
+    "_oi_tool_run_in_progress", default=False
+)
+
 
 class SafeJSONEncoder(json.JSONEncoder):
     """
@@ -982,6 +990,23 @@ class _BaseToolRunWrapper:
     ) -> Any:
         if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
             return wrapped(*args, **kwargs)
+        # If a tool-run span is already open on this context (e.g. Tool.run
+        # delegating to a wrapped BaseTool.run), don't nest a duplicate span.
+        if _tool_run_in_progress.get():
+            return wrapped(*args, **kwargs)
+        token = _tool_run_in_progress.set(True)
+        try:
+            return self._run_with_span(wrapped, instance, args, kwargs)
+        finally:
+            _tool_run_in_progress.reset(token)
+
+    def _run_with_span(
+        self,
+        wrapped: Callable[..., Any],
+        instance: Any,
+        args: Tuple[Any, ...],
+        kwargs: Mapping[str, Any],
+    ) -> Any:
         # Enhanced tool naming - use meaningful tool name instead of generic "BaseTool.run"
         span_name = _get_tool_span_name(instance, wrapped)
         input_value = _get_input_value(wrapped, *args, **kwargs)
--- a/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
@@ -57,12 +57,14 @@ _agent_kickoff_active: contextvars.ContextVar[bool] = contextvars.ContextVar(
     "_oi_agent_kickoff_active", default=False
 )
 
-# Contextvar used to suppress a nested, duplicate TOOL span. A tool's ``run`` may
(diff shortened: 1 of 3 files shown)


python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai/_wrappers.py
```diff
@@ -982,6 +990,23 @@ def __call__(
     ) -> Any:
         if context_api.get_value(context_api._SUPPRESS_INSTRUMENTATION_KEY):
             return wrapped(*args, **kwargs)
+        # If a tool-run span is already open on this context (e.g. Tool.run
+        # delegating to a wrapped BaseTool.run), don't nest a duplicate span.
```
musicionary: **High:** This boolean suppresses every nested `.run()` on the active context, not just the synthetic `Tool.run → BaseTool.run` delegation. In CrewAI 1.10.1, `Tool.run` directly calls arbitrary `self.func`; a decorated outer tool can call `another_tool.run(...)`, and that inner operation will bypass this wrapper. Scope the guard to the same instance/delegation path and add a two-tool nested-span regression test.

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
```diff
@@ -147,6 +147,85 @@ def __init__(self, agent: Any) -> None:
...
+
+    result = multiply.run(first_number=6, second_number=7)
+    assert result == 42
+
+    spans = in_memory_span_exporter.get_finished_spans()
+    tool_spans = get_spans_by_kind(spans, OpenInferenceSpanKindValues.TOOL.value)
+    assert len(tool_spans) == 1
+
+    attributes = dict(tool_spans[0].attributes or {})
+    assert attributes.pop(OPENINFERENCE_SPAN_KIND) == OpenInferenceSpanKindValues.TOOL.value
```
musicionary: **High:** Please make this an exhaustive pop-style assertion. It currently leaves `INPUT_MIME_TYPE`, `OUTPUT_MIME_TYPE`, tool metadata, and any unexpected attributes unverified. Pop all expected attributes and end with `assert not attributes`.

python/instrumentation/openinference-instrumentation-crewai/tests/test_instrumentor.py
```diff
@@ -147,6 +147,85 @@ def __init__(self, agent: Any) -> None:
...
+            in_memory_span_exporter.get_finished_spans(),
+            OpenInferenceSpanKindValues.TOOL.value,
+        )
+        assert len(tool_spans) == 1
+        attributes = dict(tool_spans[0].attributes or {})
+        assert attributes[OPENINFERENCE_SPAN_KIND] == OpenInferenceSpanKindValues.TOOL.value
+        assert attributes[TOOL_NAME] == "multiply"
+    finally:
+        CrewAIInstrumentor().uninstrument()
+
```
musicionary, line 275: **Medium:** This uses direct attribute lookups and never exhausts the mapping. Use `pop` assertions (including input/output MIME types) plus a final `assert not attributes`; otherwise added or missing attributes can pass unnoticed.

musicionary:
@jimbobbennett I added some review notes from running the `/python-code-review` skill

jimbobbennett:
Thanks for the review — all three addressed in 8d6316b4.

- **Guard scope:** the boolean suppressed every nested `.run()` on the context, which would drop the inner span when a tool's body calls another tool. Switched to an instance-keyed contextvar (`_tool_run_instance` holds `id(instance)`), so only a same-instance `Tool.run` → `BaseTool.run` delegation is skipped; a nested call to a *different* tool still gets its own span. Added `test_nested_tool_run_on_different_instance_emits_both_spans` (outer tool calls inner, asserts two TOOL spans).
- **Exhaustive tool-span assertions:** added a shared `_assert_multiply_tool_span` helper that pops every attribute — kind, name, description, parameters, input/output value + MIME types, and the `tool.*` metadata — and ends with `assert not attributes`. Applied to both the decorator test and the delegation test.

Verified with `tox -e py313-ci-crewai-latest` (ruff, mypy, tests).

musicionary (approved):

PR #3424, opened 2026-07-23
fix(openllmetry): map retrieval task spans

### LINKED ISSUE #3421: [bug] openllmetry: LangChain retriever spans mapped to TOOL and lose all retrieval attributes (no RETRIEVER span kind)
**Describe the bug**

`OpenInferenceSpanProcessor` (from `openinference-instrumentation-openllmetry`,
the OpenLLMetry → OpenInference bridge) converts every non-`llm`
OpenLLMetry/Traceloop span through a single `_map_generic_span()` path. LangChain
**retriever** spans — emitted by `opentelemetry-instrumentation-langchain` with
`traceloop.span.kind = "task"` and `gen_ai.operation.name = "vector_db_retrieve"`
— are therefore:

1. classified as **`TOOL`** (`_SPAN_KIND_MAPPING["task"] → TOOL`), never
   `RETRIEVER`, and given a `tool.name`; and
2. stripped of every retrieval-specific attribute: the processor rebuilds the
   attribute set from only `traceloop.entity.input/output`, so
   `gen_ai.operation.name`, `gen_ai.task.input` (the query), `gen_ai.task.output`
   (the documents), `gen_ai.provider.name`, and `gen_ai.task.status` are all
   dropped.

The retrieved documents survive only as a raw JSON blob inside `output.value`
(the unmodified Traceloop envelope `{"documents": [...], "count": N}`); no
`retrieval.documents.*` attributes are produced. In Phoenix the retriever
therefore renders as a generic **TOOL** with a nested JSON blob and no retrieval
UI (no query / document breakdown, no per-document scores).

This is the retriever analog of #3241 ("… maps OpenLLMetry/LangChain **tool**
spans to TOOL but drops `tool.name` …"), which was fixed for tool spans by
#3252. That fix added a tool-aware branch to `_map_generic_span()` but did not
add a retriever-aware branch, so retriever spans are still mishandled.

**Root cause** (current `main`, `_span_processor.py`)

- `_SPAN_KIND_MAPPING` has no `RETRIEVER` entry;
  `TraceloopSpanKindValues.TASK ("task") → OpenInferenceSpanKindValues.TOOL`.
  LangChain retriever runs surface as `task`.
- `_map_generic_span()` only special-cases `is_tool` (it calls `_unwrap_tool_io`,
  sets `tool.name`, etc.). There is no check for
  `gen_ai.operation.name == "vector_db_retrieve"` (or the OTel GenAI `"retrieval"`
  value) and no mapping to `retrieval.documents.*`.
- `on_end()` replaces the span's attributes with only what `_map_generic_span()`
  returns (via `_write_attributes`), so any `gen_ai.*` retrieval attributes on the
  raw span are discarded.
- For a retriever, `_unwrap_tool_io` also mis-fires: it looks for
  `"inputs"`/`"output"` keys, but the retriever envelope uses
  `"query"`/`"documents"`, so `input.value`/`output.value` fall back to the raw
  Traceloop envelopes rather than clean values.

**To Reproduce**

```bash
pip install "opentelemetry-instrumentation-langchain==0.59.0" "wrapt<2" \
    "langchain-core>=0.3,<1" numpy opentelemetry-sdk \
    "openinference-instrumentation-openllmetry @ git+https://github.com/Arize-ai/openinference.git#subdirectory=python/instrumentation/openinference-instrumentation-openllmetry"
```

```python
from typing import Any, Dict, List

from opentelemetry.sdk import trace as trace_sdk
from opentelemetry.sdk.trace import SpanProcessor

from opentelemetry.instrumentation.langchain import LangchainInstrumentor
from openinference.instrumentation.openllmetry import OpenInferenceSpanProcessor

from langchain_core.embeddings import Embeddings
from langchain_core.vectorstores import InMemoryVectorStore


class TinyEmbeddings(Embeddings):
    def __init__(self, size: int = 16) -> None:
        self.size = size

    def _embed(self, text: str) -> List[float]:
        v = [0.0] * self.size
        for i, ch in enumerate(text):
            v[i % self.size] += (ord(ch) % 17) / 17.0
        return v

    def embed_documents(self, texts: List[str]) -> List[List[float]]:
        return [self._embed(t) for t in texts]

    def embed_query(self, text: str) -> List[float]:
        return self._embed(text)


RAW: List[Dict[str, Any]] = []
MAPPED: List[Dict[str, Any]] = []


class _Capture(SpanProcessor):
    def __init__(self, sink: List[Dict[str, Any]]) -> None:
        self._sink = sink

    def on_end(self, span: Any) -> None:
        self._sink.append({"name": span.name, "attrs": dict(span.attributes or {})})


tp = trace_sdk.TracerProvider()
tp.add_span_processor(_Capture(RAW))              # snapshot before OI processor
tp.add_span_processor(OpenInferenceSpanProcessor())
tp.add_span_processor(_Capture(MAPPED))            # snapshot after OI processor
trace.set_tracer_provider(tp)
LangchainInstrumentor().instrument(tracer_provider=tp)

vs = InMemoryVectorStore(embedding=TinyEmbeddings(16))
vs.add_texts(["VM alpha memory 95%", "VM beta cpu 12%", "storage 40% full"])
retriever = vs.as_retriever(search_kwargs={"k": 2})
retriever.invoke("Which VMs are running low on memory?")
tp.force_flush()

for k, v in sorted(MAPPED[0]["attrs"].items()):
    print(f"{k} = {str(v)[:120]}")
```

**Observed**

Raw retriever span (from `opentelemetry-instrumentation-langchain`):

```
name: vector_db_retrieve VectorStoreRetriever
traceloop.span.kind      = task
gen_ai.operation.name    = vector_db_retrieve
gen_ai.provider.name     = langchain
gen_ai.task.input        = {"query": "Which VMs are running low on memory?", "tags": [...], "metadata": {...}}
gen_ai.task.output       = {"documents": [{"page_content": "VM alpha memory 95%", "metadata": {}}, ...], "count": 2}
gen_ai.task.status       = success
traceloop.entity.name    = VectorStoreRetriever
traceloop.entity.input   = {"query": ...}
traceloop.entity.output  = {"documents": [...], "count": 2}
```

After `OpenInferenceSpanProcessor`:

```
openinference.span.kind  = TOOL          # <- want RETRIEVER
tool.name                = VectorStoreRetriever
input.mime_type          = application/json
input.value              = {"query": "...", "tags": [...], "metadata": {...}}   # raw envelope
output.mime_type         = application/json
output.value             = {"documents": [...], "count": 2}                     # raw envelope
# dropped: gen_ai.operation.name, gen_ai.task.input, gen_ai.task.output,
#          gen_ai.provider.name, gen_ai.task.status
# no retrieval.documents.* attributes
```

**Expected behavior**

- Spans with `traceloop.span.kind == "task"` and
  `gen_ai.operation.name in {"vector_db_retrieve", "retrieval"}` map to
  `OpenInferenceSpanKindValues.RETRIEVER`, not `TOOL` (and should not get a
  `tool.name`).
- `input.value` is the query text (from `gen_ai.task.input.query`).
- The retrieved documents (`gen_ai.task.output.documents`) map to
  `retrieval.documents.{i}.document.content` (plus `.metadata`, and `.score`/`.id`
  when present), using `SpanAttributes.RETRIEVAL_DOCUMENTS` /
  `DocumentAttributes.*`.

**Proposed fix**

Add a retrieval-aware branch to the non-LLM path (checked before the tool
branch) and resolve `"task"` → `RETRIEVER` when the operation is a retrieval:

```python
from opentelemetry.semconv_ai import GenAICustomOperationName

_RETRIEVAL_OP_NAMES = frozenset({
    GenAICustomOperationName.VECTOR_DB_RETRIEVE.value,  # "vector_db_retrieve"
    "retrieval",  # OTel GenAI semconv value
})

# in _map_generic_span(), before the is_tool handling:
op_name = attrs.get("gen_ai.operation.name")
if op_name in _RETRIEVAL_OP_NAMES:
    mapped = {
        sc.SpanAttributes.OPENINFERENCE_SPAN_KIND:
            sc.OpenInferenceSpanKindValues.RETRIEVER.value,
    }
    task_in = _coerce_json_obj(
        attrs.get("gen_ai.task.input")
        or attrs.get(SpanAttributes.TRACELOOP_ENTITY_INPUT)
    )
    if isinstance(task_in, dict) and "query" in task_in:
        mapped["input.value"] = task_in["query"]
        mapped["input.mime_type"] = sc.OpenInferenceMimeTypeValues.TEXT.value
    task_out = _coerce_json_obj(
        attrs.get("gen_ai.task.output")
        or attrs.get(SpanAttributes.TRACELOOP_ENTITY_OUTPUT)
    )
    docs = task_out.get("documents", []) if isinstance(task_out, dict) else []
    for i, d in enumerate(docs):
        prefix = f"{sc.SpanAttributes.RETRIEVAL_DOCUMENTS}.{i}.document"
        if isinstance(d, dict):
            mapped[f"{prefix}.content"] = d.get("page_content", "")
            if d.get("metadata"):
                mapped[f"{prefix}.metadata"] = _as_json_str(d["metadata"])
        else:
            mapped[f"{prefix}.content"] = str(d)
    return mapped
```

**Environment**

- `openinference-instrumentation-openllmetry`: `main` @ `e1b6748` (reports as
  `0.1.11`; the mapping bug is also present in released `0.1.9`)
- `opentelemetry-instrumentation-langchain`: 0.59.0
- `openinference-semantic-conventions`: 0.1.30
- `opentelemetry-sdk`: 1.44.0
- `opentelemetry-semantic-conventions-ai`: 0.4.16
- `langchain-core`: 0.3.86
- Python: 3.12

**Note**

I verified against `main` because released `0.1.9` additionally raises
`TypeError` in the same non-LLM branch (`attrs.clear()` on an immutable
`BoundedAttributes`) under `opentelemetry-sdk >= 1.37`; that crash is already
fixed on `main` via `_write_attributes()`. With the crash fixed, the retriever
span flows through and exhibits the mapping issue above.

**Related**

- #3241 / #3252 — the tool-span analog (fixed for tools only)
- #2145 — [bug] Beeai - missing retriever span
- #2627 — [bug] retriever for vector store search missing retrieved documents
- OTel GenAI semconv — [Retrieval spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#retrievals)

Sanjays2402:
# Description

Resolves #3421

OpenLLMetry retrieval tasks now map to OpenInference `RETRIEVER` spans with a clean query and structured document content, metadata, IDs, and scores instead of generic `TOOL` spans.

# Checklist:

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
@@ -68,6 +68,8 @@ _TOOL_KEY_CANDIDATES = [
     _GEN_AI_TOOL_DEFINITIONS,
 ]
 
+_RETRIEVAL_OPERATION_NAMES = frozenset({"retrieval", "vector_db_retrieve"})
+
 _MISSING: Any = object()
 
 
@@ -122,11 +124,54 @@ def _unwrap_tool_io(input_raw: Any, output_raw: Any) -> Tuple[Optional[Any], Any
     return tool_args, tool_result, tool_args
 
 
+def _map_retriever_span(attrs: Dict[str, Any]) -> Dict[str, Any]:
+    """Map an OpenLLMetry retrieval task to OpenInference retriever attributes."""
+    mapped: Dict[str, Any] = {
+        sc.SpanAttributes.OPENINFERENCE_SPAN_KIND: sc.OpenInferenceSpanKindValues.RETRIEVER.value
+    }
+
+    input_raw = attrs.get("gen_ai.task.input") or attrs.get(SpanAttributes.TRACELOOP_ENTITY_INPUT)
+    input_obj = _coerce_json_obj(input_raw)
+    if isinstance(input_obj, dict) and input_obj.get("query") is not None:
+        mapped[sc.SpanAttributes.INPUT_VALUE] = str(input_obj["query"])
+        mapped[sc.SpanAttributes.INPUT_MIME_TYPE] = sc.OpenInferenceMimeTypeValues.TEXT.value
+
+    output_raw = attrs.get("gen_ai.task.output") or attrs.get(
+        SpanAttributes.TRACELOOP_ENTITY_OUTPUT
+    )
+    output_obj = _coerce_json_obj(output_raw)
+    documents = output_obj.get("documents") if isinstance(output_obj, dict) else None
+    if not isinstance(documents, list):
+        return mapped
+
+    for index, document in enumerate(documents):
+        if not isinstance(document, dict):
+            continue
+        prefix = f"{sc.SpanAttributes.RETRIEVAL_DOCUMENTS}.{index}"
+        content = document.get("page_content", document.get("content"))
+        if content is not None:
+            mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_CONTENT}"] = str(content)
+        if document.get("metadata") is not None:
+            mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_METADATA}"] = _as_json_str(
+                document["metadata"]
+            )
+        if document.get("id") is not None:
+            mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_ID}"] = str(document["id"])
+        if document.get("score") is not None:
+            mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_SCORE}"] = document["score"]
+
+    return mapped
+
+
 def _map_generic_span(attrs: Dict[str, Any], span_name: Optional[str] = None) -> Dict[str, Any]:
     """
     Convert TraceLoop 'workflow' / 'task' / 'agent' / 'tool' spans
     to OpenInference semantic conventions.
     """
+    operation_name = str(attrs.get("gen_ai.operation.name", "")).lower()
+    if operation_name in _RETRIEVAL_OPERATION_NAMES:
+        return _map_retriever_span(attrs)
+
     raw_kind = str(attrs.get(SpanAttributes.TRACELOOP_SPAN_KIND, "unknown")).lower()
     kind_val = _SPAN_KIND_MAPPING.get(raw_kind, sc.OpenInferenceSpanKindValues.UNKNOWN.value)
 
--- a/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
@@ -145,13 +145,15 @@ def _map_retriever_span(attrs: Dict[str, Any]) -> Dict[str, Any]:
         return mapped
 
     for index, document in enumerate(documents):
-        if not isinstance(document, dict):
-            continue
         prefix = f"{sc.SpanAttributes.RETRIEVAL_DOCUMENTS}.{index}"
+        if not isinstance(document, dict):
+            if document is not None:
+                mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_CONTENT}"] = str(document)
+            continue
         content = document.get("page_content", document.get("content"))
         if content is not None:
             mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_CONTENT}"] = str(content)
-        if document.get("metadata") is not None:
+        if document.get("metadata"):
             mapped[f"{prefix}.{sc.DocumentAttributes.DOCUMENT_METADATA}"] = _as_json_str(
                 document["metadata"]
             )
--- a/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openllmetry/src/openinference/instrumentation/openllmetry/_span_processor.py
@@ -94,6 +94,15 @@ def _safe_int(value: Any) -> Optional[int]:
         return None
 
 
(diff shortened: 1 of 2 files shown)


latent-variable-17:
Hey, nice work on this!

Just thought of two tiny edge cases in` _map_retriever_span() ` that you may check:

1. if the retriever returns plain strings instead of dicts, they get skipped with current code...could add an 'else' branch also to map non-dict items as plain content

2. The **is not None** check on metadata may evaluate to True for empty dicts ({})...this might clutter some of the span attributes with empty strings...May be a simple truthy check is good enough. Just a happy suggestion

Sanjays2402:
Both fair. The `is not None` check does emit `"{}"` for a document with empty metadata, which is just noise on the span, so a truthy check is the right call there.

The non-dict case I'll take too. Right now a retriever returning plain strings gets silently dropped by the `isinstance(document, dict)` continue, which is worse than mapping it as bare content since the document count then disagrees with what actually came back.

Will push both.

Sanjays2402:
Pushed both changes.

Metadata now uses a truthy check, so a document with `metadata: {}` no longer emits `"{}"` on the span. Non-dict documents are no longer skipped either — a bare string gets mapped to `document.content` at its index, so the document indices stay in sync with what the retriever actually returned.

Added two tests covering the empty-metadata and plain-string cases. Thanks for the review.

ialisaleh (commented):
This PR resolves the issue by adding dedicated handling for retrieval operations, ensuring LangChain retriever spans are correctly mapped to the `RETRIEVER` span kind instead of `TOOL`. It extracts the retrieval query into `input.value`, maps retrieved documents to the appropriate `retrieval.documents.*` attributes (including metadata, ID, and score when present), and avoids emitting tool-specific attributes. The accompanying tests provide good coverage of the new behavior, including both LangChain and OTel retrieval operation names, fallback to Traceloop envelopes, document variants, and metadata handling, validating the reported scenarios. Thanks!

mikeldking (approved):
thanks for this

PR #3429, opened 2026-07-24
fix(anthropic): preserve APIPromise helpers

### LINKED ISSUE #3397: AnthropicInstrumentation breaks client.messages.stream() — patched create() loses APIPromise.withResponse()
## Bug

`AnthropicInstrumentation.patch()` wraps `Messages.prototype.create` by chaining `.then(wrappedPromiseThen).catch(...)` on the return value of `original.apply(this, args)`. `@anthropic-ai/sdk`'s `APIPromise` (the return type of `create()`) overrides `.then()`/`.catch()`/`.finally()` to internally call `this.parse().then(...)`, which returns a **plain `Promise`**, not another `APIPromise` — so the patched `create()`'s return value silently loses `.withResponse()` and `.asResponse()`.

`client.messages.stream()` depends on `.withResponse()` internally (`MessageStream._createMessage` calls `this.messages.create({...params, stream:true}, options).withResponse()`), so **every** `client.messages.stream()` call throws `TypeError: create(...).withResponse is not a function` once `manuallyInstrument(Anthropic)` (or auto-instrumentation) has patched the SDK. Non-streaming `.create()` calls are unaffected since they only use the (still-present) default `.then()`/await path.

## Repro (12 lines)

```ts
import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
import Anthropic from '@anthropic-ai/sdk';
import { AnthropicInstrumentation } from '@arizeai/openinference-instrumentation-anthropic';

const provider = new NodeTracerProvider();
provider.register();
new AnthropicInstrumentation({ tracerProvider: provider }).manuallyInstrument(Anthropic);

const client = new Anthropic({ apiKey: process.env.ANTHROPIC_API_KEY });
const stream = client.messages.stream({
  model: 'claude-sonnet-5', max_tokens: 30,
  messages: [{ role: 'user', content: 'hi' }],
});
for await (const _ of stream) {} // throws: create(...).withResponse is not a function
```

## Versions

- `@arizeai/openinference-instrumentation-anthropic`: 0.1.15 (latest at time of filing, 2026-07-16)
- `@anthropic-ai/sdk`: 0.78.0

## Suggested fix

In `patch()`, replace the `.then(wrappedPromiseThen).catch(...)` chain with `APIPromise#_thenUnwrap()` (a supported SDK method for transforming a resolved value while remaining an `APIPromise`), plus a side-effect-only `.then(undefined, onError)` subscriber for span error recording so nothing is swallowed:

```js
execPromise.then(undefined, (error) => { /* span.recordException etc — side effect only */ });
const wrappedPromise = execPromise._thenUnwrap((result) => wrappedPromiseThen(result));
return context.bind(execContext, wrappedPromise);
```

We've verified this holds `.withResponse()`/`.asResponse()` for both streaming and non-streaming calls, and that stream-teeing/span-recording still works correctly for the streaming case. Happy to open a PR with this fix + tests if useful — for now we're carrying it as a local `patch-package` patch (referenced from our internal `TC-OBSERVE.ARIZE-ANTHROPIC-INSTRUMENT-COMPAT`).

satyadevai:
Closes #3397

### DIFF
--- a/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
@@ -1,4 +1,5 @@
 import type Anthropic from "@anthropic-ai/sdk";
+import type { APIPromise } from "@anthropic-ai/sdk";
 import type { Stream } from "@anthropic-ai/sdk/streaming";
 import type { Attributes, Span, Tracer, TracerProvider } from "@opentelemetry/api";
 import { context, diag, SpanKind, SpanStatusCode, trace } from "@opentelemetry/api";
@@ -234,13 +235,28 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
             return result;
           };
 
-          const wrappedPromise = execPromise.then(wrappedPromiseThen).catch((error: Error) => {
+          const recordError = (error: Error) => {
             span.recordException(error);
             span.setStatus({
               code: SpanStatusCode.ERROR,
               message: error.message,
             });
             span.end();
+          };
+
+          // Use _thenUnwrap so the result stays an APIPromise and keeps
+          // withResponse()/asResponse(). Plain .then() would drop them and break
+          // client.messages.stream().
+          if (hasThenUnwrap(execPromise)) {
+            const wrappedPromise = execPromise._thenUnwrap(wrappedPromiseThen);
+            // Record errors without touching the returned promise. Parsing is
+            // memoized, so this doesn't read the response twice.
+            wrappedPromise.then(undefined, recordError);
+            return context.bind(execContext, wrappedPromise);
+          }
+
+          const wrappedPromise = execPromise.then(wrappedPromiseThen).catch((error: Error) => {
+            recordError(error);
             throw error;
           });
           return context.bind(execContext, wrappedPromise);
@@ -277,6 +293,13 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
       diag.warn(`Failed to unset ${MODULE_NAME} patched flag on the module`, e);
     }
   }
+}
+
+/**
+ * True when create() returned an APIPromise we can transform with _thenUnwrap.
+ */
+function hasThenUnwrap<T>(promise: PromiseLike<T>): promise is APIPromise<T> {
+  return typeof (promise as Partial<APIPromise<T>>)._thenUnwrap === "function";
 }
 
 /**
--- a/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
@@ -249,9 +249,10 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
           // client.messages.stream().
           if (hasThenUnwrap(execPromise)) {
             const wrappedPromise = execPromise._thenUnwrap(wrappedPromiseThen);
-            // Record errors without touching the returned promise. Parsing is
-            // memoized, so this doesn't read the response twice.
-            wrappedPromise.then(undefined, recordError);
+            // Watch for failures via asResponse() instead of then(). then() would
+            // eagerly parse and consume the body, leaving asResponse() callers with
+            // an already-read response.
+            wrappedPromise.asResponse().catch(recordError);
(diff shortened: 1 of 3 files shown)


js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
```diff
@@ -234,13 +235,28 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
...
+          };
+
+          // Use _thenUnwrap so the result stays an APIPromise and keeps
+          // withResponse()/asResponse(). Plain .then() would drop them and break
+          // client.messages.stream().
+          if (hasThenUnwrap(execPromise)) {
+            const wrappedPromise = execPromise._thenUnwrap(wrappedPromiseThen);
+            // Record errors without touching the returned promise. Parsing is
+            // memoized, so this doesn't read the response twice.
+            wrappedPromise.then(undefined, recordError);
```
gnanirahulnutakki: This `.then()` invokes Anthropic's overridden `APIPromise.then()` immediately, so it eagerly parses and consumes the response body. As a result, this patch restores the `asResponse()` method but not its contract: on `8554f3d`, a focused test that awaits `messages.create(...).asResponse()` gets `response.bodyUsed === true`, and the caller can no longer read the raw body.

Please observe request failures without triggering parsing and add a regression that asserts the returned response is unconsumed and can be read. I verified that replacing this observer with `wrappedPromise.asResponse().catch(recordError)` makes that regression pass while preserving the existing error-span test; the full Anthropic package suite (16 tests), type-check, oxlint, and oxfmt all passed locally.

gnanirahulnutakki (approved):
Re-reviewed `d4827bf`. The failure observer now uses `asResponse().catch(recordError)`, so it no longer triggers `APIPromise.then()` parsing. The new regression proves the returned response has `bodyUsed === false` and remains JSON-readable; the existing error-path regression still proves the caller rejects and the span records an exception/error. On the exact head, all 16 Anthropic package tests, type-check, build, oxlint, and oxfmt passed. A conflict-free synthetic merge with current `main` (`a1392c50`) also passed the 16 tests, type-check, build, and formatting; GitHub reports 20 completed checks with zero failures.

js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
```diff
@@ -234,13 +235,29 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
...
+
+          // Use _thenUnwrap so the result stays an APIPromise and keeps
+          // withResponse()/asResponse(). Plain .then() would drop them and break
+          // client.messages.stream().
+          if (hasThenUnwrap(execPromise)) {
+            const wrappedPromise = execPromise._thenUnwrap(wrappedPromiseThen);
+            // Watch for failures via asResponse() instead of then(). then() would
+            // eagerly parse and consume the body, leaving asResponse() callers with
+            // an already-read response.
+            wrappedPromise.asResponse().catch(recordError);
```
caroger: Successful `asResponse()` calls still leave this instrumentation span open. `asResponse()` only resolves the raw `responsePromise`; it does not invoke the `_thenUnwrap(wrappedPromiseThen)` parse transform that records output/status and ends the span. I reproduced this on `d4827bf` with the compiled PR code, real Anthropic SDK, and an in-memory OTel exporter: `bodyUsed` remained `false`, but the exporter had 0 finished spans; the equivalent `withResponse()` call produced 1 finished OK span. Please finalize the successful raw-response path without consuming the caller's body, and extend the `asResponse()` regression to assert that its span finishes.

caroger (changes requested):

js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
satyadevai: I have missed this case, fixed now.

caroger (approved):
Re-reviewed `7ea68f7` after the rebase onto current `main`. The prior `asResponse()` span-lifecycle finding is addressed: the raw response remains unconsumed, successful raw calls finish one OK span, parsed `withResponse()` calls retain output telemetry, and rejected raw calls finish one ERROR span with an exception.

The package build and type-check passed locally, the targeted compiled runtime integration passed, and GitHub TypeScript CI is green. No remaining findings.

PR #3450, opened 2026-07-27
feat(semconv): audio and file message content types

caroger:
## Summary

Adds the **audio and file (e.g. PDF) message-content vocabulary** to OpenInference: semantic-convention constants, spec text, and the gen_ai dual-write mapping. OpenAI chat completions already returns audio parts and accepts file parts; OpenInference has no conventions to record them today (`TraceConfig` masking only knows `data:image/`).

This vocabulary is independently useful for *inline* capture — nothing in this PR involves uploads. It is also the vocabulary the blob-upload stack builds on (techspec: #3408).

**Stacked PR 2 of 4** — base is `blob-upload` (#3408); the `BlobUploader` contract + offload policy follows in #3409, OpenAI capture in #3410.

## Changes

**`openinference-semantic-conventions`** (Python + Go mirror)
- `MESSAGE_CONTENT_AUDIO`, `MESSAGE_CONTENT_FILE`, new `"file"` message content type
- `FileAttributes`: `file.url` / `file.mime_type` / `file.name` / `file.id` (provider-hosted files are id references with no bytes — distinct from `url` by design)

**`openinference-instrumentation` (gen_ai dual-write)**
- Audio/file contents convert to schema-valid `blob` / `uri` / `file` message parts in both conversion paths; `_image_part_from_url` generalizes to `_media_part_from_url`; gen_ai modality gains `document` (per the upstream JSON schema)

**`spec/`**
- `"file"` content type, audio data-URI/transcript examples, and file examples in `multimodal_attributes.md`; reserved-attribute rows in `semantic_conventions.md`


caroger:
Deferring per scope decision: blob upload ships **image-only** first (see #3409) — image is the only media type whose message-content convention is settled today, while audio is emitted by a single instrumentor under instrumentor-local keys and files have no OI convention yet. This vocabulary (semconv constants, Go mirror, spec text, gen_ai dual-write mapping) is preserved on the `blob-upload-semconv` branch and returns as the first follow-on once those conventions are settled — tracked as open question 6 in the techspec (#3408).

PR #3451, opened 2026-07-28
fix(google-adk): mark call_llm span ERROR on structured LlmResponse error_code

### LINKED ISSUE #3415: google-adk: tool/LLM spans stay OK (with exception events) on failing calls under ADK 2.x
## Summary

Under recent Google ADK (reproduced on **2.5.0**; minimum supported **1.2.1** behaves differently), OpenInference’s `_TraceToolCall` and `_TraceCallLlm` wrappers call `span.set_status(StatusCode.OK)` while enriching ADK-owned current spans.

On ADK 2.x, `record_tool_execution` invokes `trace_tool_call(..., error=...)` from a `finally` block while still inside the span context. Streaming LLM paths can similarly call `trace_call_llm` after a partial response before the stream subsequently raises. OpenTelemetry records the exception event, but a span already assigned `OK` will ignore a later attempt to set `ERROR`.

The result is that failed `execute_tool` and `call_llm` spans can finish with **OK status plus an exception event**, while the OpenInference-owned `agent_run` and `invocation` spans correctly finish with `ERROR`. Tracked tests currently cover successful statuses but not these failure paths.

This appears distinct from #3399, which concerns raw input/output, missing agent input, and untyped merged/live spans.

## Reproduction matrix

| ADK | Span | Status | Exception event |
|---|---|---|---|
| 1.2.1 | `execute_tool` | ERROR | yes |
| 1.2.1 | `agent_run` / `invocation` | ERROR | yes |
| 2.5.0 | `execute_tool` | **OK** | yes |
| 2.5.0 | `agent_run` / `invocation` | ERROR | yes |
| 2.5.0, partial LLM response followed by stream exception | `call_llm` | **OK** | yes |
| 2.5.0, partial LLM response followed by stream exception | `agent_run` / `invocation` | ERROR | yes |

## Ask

Would a narrowly scoped PR be welcome to prevent the LLM/tool enrichment wrappers from locking failed ADK-owned spans at `OK`, while preserving successful-span behavior and adding failure-path tests against minimum and latest ADK?

The proposed change would not infer failure merely from arbitrary application payload contents. Guidance would be helpful on whether ADK’s structured `error_type` signal should also result in `ERROR` status.

feiiiiii5:
## Context

#3542 (merged 2026-08-25) fixed the sticky-OK bug from #3415 at the instrumentor level: `_TraceToolCall` now derives the `execute_tool` status from ADK's `error` / `error_type` arguments, and `_TraceCallLlm` stamps `OK` on `call_llm` only for the final non-partial chunk, so a stream that raises afterwards still ends as `ERROR`.

This PR is now a follow-up on top of that, filling in the remaining gap and adding coverage.

## Change

**`_TraceCallLlm`: honor ADK's structured `error_code`.** ADK returns an `LlmResponse` with `error_code` / `error_message` (and no content) when the model call fails without raising, for example a response blocked by a safety filter or a prompt rejected with a block reason. On main that final, non-partial response is stamped `OK`. Here it ends the `call_llm` span as `ERROR` with the error message (or the code) as the status description. The tool path already treats ADK's `error_type` this way, so this makes the LLM path consistent.

Successful spans keep an explicit `OK`, as before.

## Tests

`tests/test_span_status.py` drives the whole runner with a stub `BaseLlm` (no cassettes) and asserts `status_code` directly, since `Status.is_ok` is also true for `UNSET`:

- response with `error_code` and `error_message` -> `call_llm` ERROR, description is the message
- response with `error_code` only -> `call_llm` ERROR, description is the code
- SSE streaming with two partial chunks then a final chunk -> `call_llm` OK (partial chunks decide nothing)
- happy path with a tool call -> tool, both LLM, agent and invocation spans all explicitly OK

Passing on ADK 1.2.1 (minimum) and 2.8.0 (latest).

## Example

`examples/failing_tool_agent.py` runs an agent whose tool raises against a real model and exports to Phoenix, showing `execute_tool`, `agent_run` and `invocation` as ERROR and the preceding `call_llm` as OK.

Refs #3415


feiiiiii5:

python/openinference-instrumentation/src/openinference/instrumentation/_tracers.py
```diff
@@ -745,7 +745,10 @@ def _chain_context(
     ) as span:
         context = _ChainContext(span=span)
         yield context
-        span.set_status(Status(StatusCode.OK))
+        # Do not unconditionally set OK here — it would prevent callers
+        # (e.g. ADK 2.x) from setting ERROR on failure paths.  OTel
+        # defaults to UNSET, which is treated as success for non-error
+        # spans and does not block later ERROR updates (closes #3415).
```
satyadevai: We should handle these cases at the specific instrumentor level, not in the root level.

feiiiiii5: Thanks for the review! The reason the fix lives in the shared root contexts rather than the ADK instrumentor:

1. **OTel status is sticky** — once `OK` is recorded, later `set_status(ERROR)` calls are ignored. The three root contexts (`_chain_context`/`_llm_context`/`_tool_context`) unconditionally called `set_status(OK)` after `yield`, and on ADK 2.x the instrumentor only gets to set status afterwards (from a `finally` block). So the lock happens in the shared wrapper regardless of what any instrumentor does.
2. **Every instrumentor is affected** — the shared contexts are used by all instrumentors, not just ADK. Fixing this at the ADK level would leave the same latent trap for every other instrumentor that tries to record ERROR after exiting the shared context.
3. **No semantic change on the success path** — OTel defaults to `UNSET`, which UIs treat as success for non-error spans, so removing the explicit OK only unblocks ERROR; it does not turn successful spans into failures.

If you would prefer to keep explicit OK on the happy path, an alternative is a parameter on the shared contexts (e.g. `set_ok_on_exit: bool = True`) that the ADK instrumentor sets to False — happy to switch to that shape if it better matches the project’s conventions.

feiiiiii5:
Gentle follow-up on the design question above: is the shared-root approach acceptable, or would you prefer the instrumentor-level handling despite the sticky-OK caveat? Happy to adjust the fix either way — just want to make sure the thread doesn't stall the PR.

feiiiiii5:
@mikeldking could you weigh in on the open design question here when you have a chance? satyadevai suggested instrumentor-level status handling, and I replied with the sticky-OTel-status rationale for keeping the fix in the shared root contexts (plus the ADK version dependency). The thread has been waiting since 08-06; happy to implement either direction.

feiiiiii5:
@satyadevai gentle ping — the fix intentionally lives in the shared root contexts rather than the ADK instrumentor: once an OTel span records status `OK`, later `set_status(ERROR)` calls are ignored, so the error contract has to be decided where the span starts (rationale and per-context breakdown in the thread above). Let me know if you'd prefer a different shape and I'll adapt.

feiiiiii5:
Closing: no maintainer response since @satyadevai's 08-06 review, and the open design question needs a maintainer call I cannot force. For whoever picks this up, the bug is that the ADK instrumentation sets span status to OK unconditionally, which masks later errors because OTel status is sticky (once OK, a later `set_status(ERROR)` is ignored). This branch removes the unconditional OK at the shared root contexts; @satyadevai suggested handling it at the instrumentor level instead, and I asked @mikeldking to weigh in on 08-15 without a response. Branch preserved with the fix and rationale; happy to reopen, or to re-implement at the instrumentor level if that is the preferred direction.

mikeldking:
> Closing: no maintainer response since @satyadevai's 08-06 review, and the open design question needs a maintainer call I cannot force. For whoever picks this up, the bug is that the ADK instrumentation sets span status to OK unconditionally, which masks later errors because OTel status is sticky (once OK, a later `set_status(ERROR)` is ignored). This branch removes the unconditional OK at the shared root contexts; @satyadevai suggested handling it at the instrumentor level instead, and I asked @mikeldking to weigh in on 08-15 without a response. Branch preserved with the fix and rationale; happy to reopen, or to re-implement at the instrumentor level if that is the preferred direction.
> 

Sorry for the delay. My GitHub inbox is quite flooded at the moment. Will take a look

mikeldking:
Thanks for the persistence on this, and sorry again for the wait. I dug into it and pushed a commit to this branch (`c387ba9f`) that redirects the fix; details are in the updated PR description, short version here.

**Why not the shared contexts.** `_chain_context` / `_llm_context` / `_tool_context` back the `@tracer.chain` / `.llm` / `.tool` decorators. They set `OK` only *after* a successful `yield`, so when the wrapped function raises that line never runs and the span exits with `ERROR` as expected. They cannot lock out an error, and the ADK instrumentor does not call them, so the original change did not actually affect the spans in #3415. It would, however, have changed every decorator user's successful spans from `OK` to `UNSET`, silently, because `Status.is_ok` is also true for `UNSET` and none of the existing tests would have caught it.

**Where the lock actually happens.** `_TraceToolCall` and `_TraceCallLlm` in the ADK instrumentor stamped `OK` on the current span pre-emptively. On ADK >= 1.32, `trace_tool_call(..., error=...)` runs from a `finally` block inside the still-open `execute_tool` span, and `trace_call_llm` runs per streamed chunk. That is exactly the sticky-OK trap you described, just one layer down. The fix now sets `OK` only when ADK's structured signals say the call succeeded (`error is None`; non-partial response without `error_code`), and `ERROR` otherwise.

**On OTel status more generally.** You are right that the OTel spec says instrumentation libraries should generally leave status `UNSET` on success. OpenInference has deliberately set explicit `OK` across instrumentors for a long time, so moving to `UNSET` is a repo-wide convention change that deserves its own discussion rather than riding along with this bug fix.

I also added stub-model tests for the failure paths (passing on ADK 1.2.1 and 2.0) and an example that round-trips the failing tool to Phoenix. Feel free to poke at the new commit.

mikeldking:
Closing this one: while it sat waiting, #3542 (merged 2026-08-25) fixed the same bug on main with the same instrumentor-level approach, setting the `execute_tool` status from ADK's `error` / `error_type` arguments and stamping `OK` on `call_llm` only for the final non-partial chunk, and it closed #3415. I should have caught that before pushing here yesterday, sorry for the churn.

Thank you for the careful diagnosis and for sticking with it; the sticky-OK analysis in this thread was correct and is what main now does. The one thing not on main is treating an `LlmResponse` with a structured `error_code` (for example a blocked response) as `ERROR`. If you would like to carry that forward, a small PR against current main with the stub-model test from this branch would be welcome.

mikeldking:
Reopening after a second look: rather than drop this, I merged main and reshaped the branch as a follow-up to #3542. Main's tool handling (`error` / `error_type`) and final-chunk `OK` on `call_llm` are kept as they are. What this PR now adds is the piece #3542 did not cover: an `LlmResponse` that ADK returns with a structured `error_code` (a blocked response, a prompt rejected with a block reason) ends `call_llm` as `ERROR` with the message as the description, instead of being stamped `OK` as the final chunk. Plus stub-model tests for that, for SSE partial-chunk streaming, and for explicit `OK` on the happy path, and the Phoenix example. Updated description has the details.

mikeldking (approved):

feiiiiii5:
@mikeldking flagging one bookkeeping detail, because I think this PR is silently stuck rather than waiting on anything.

Your 09-02 sequence was: reshape the branch as the follow-up to #3542, approve (`23:16`), enable auto-squash. The timeline then shows `auto_merge_disabled` at 09-07T00:27, the same minute my commit `59b6fb6b9` landed — that commit was the upstream-compat fix tolerating `StreamingMode` no longer being re-exported from `run_config`, so the merge was turned off while the branch was still in flight and nobody re-armed it afterwards.

Re-measured today on the current head:
- 18 check runs completed successfully, 5 skipped, none failing or pending.
- `mergeable: true`, `mergeable_state: clean`, no conflicts with `main` (now at `a332a5ca8`).
- Your approval is still the latest review state and has not been dismissed; the four commits are the ones you reshaped plus that compat fix.

So this looks ready to merge as-is. If it is easier, say the word and I will merge `main` in and re-push, or reformat the tests/output however you'd prefer for the squash. Thanks for picking up the `error_code` case as a follow-up rather than dropping it.

PR #2146, opened 2025-08-28
feat(openai): implement image redaction for input values when hide_input_images=True

caroger:
Partially resolves #2100 by redacting image related attributes from input.values for openAI instrumentor.

Before & After with tracer config:

```python3
 config = TraceConfig(
        hide_inputs=False,
        hide_output_text=True,
        hide_input_images=True,
    )
```

### Completion
Screenshot 2025-08-28 at 11 35 17 AM

### Response
Screenshot 2025-08-28 at 11 36 31 AM

### Completion
Screenshot 2025-08-28 at 11 36 00 AM

### Response
Screenshot 2025-08-28 at 11 36 59 AM

PR #3459, opened 2026-07-29
feat(google-genai): redact images in input.value from TraceConfig

### LINKED ISSUE #3456: [feature request] Extend hide_input_images to Google GenAI Instrumentor
Issue #2100 but for google genai instrumentors. 

the problem is that the instrumentor `hide_input_images` is a no-op for base64 encoded image URLs in `input.value`.  Google GenAI's _context.py just does the naive safe_json_dumps(request) because this redaction pass was never ported to it.

- `hide_input_images` ensures that images are REDACTED in `input.value` when instrumenting via google genai instrumentor
- respect `base64_image_max_length` when redacting image links in `input.value`
- basically #2146 but for the google genai instrumentor

- configuring so that input.value is completely false, but this drops every field in input.value

sanchitram1:
- port OpenAI-style image redaction (#2146) to the Google GenAI instrumentor so `hide_input_images` and `base64_image_max_length` apply to nested image data 
- includes client.models, interactions, and cache calls
- add unit coverage for inline/file images, MIME non-images, and length-limit boundaries

Fixes #3456

## test cases

  - hide_input_images=True redacts inline base64 images.
  - hide_input_images=True redacts remote image file URIs.
  - Image MIME types remain present after their contents are redacted.
  - Text beside an image remains unchanged.
  - Inline audio remains unchanged, even when image hiding is enabled.
  - Inline audio remains unchanged when the image size limit is extremely small.
  - Binary data preserved in the span uses Google’s URL-safe base64 representation.
  - Audio MIME types remain present.
  - Audio data is not mistaken for image data.
  - Image data exceeding base64_image_max_length is redacted.
  - A zero image-length limit redacts image data.
  - Image data exactly equal to the configured limit is preserved.
  - Images below the configured limit are preserved.
  - File image URIs are preserved when image hiding is disabled.
  - The original request object is not mutated during sanitization.
  - generate_content writes the sanitized value to the finished span.
  - Interactions write the sanitized value to the finished span.
  - Cache creation handles nested Google Pydantic objects correctly.
  - Cache creation supports hiding, size-based redaction, and preservation under the limit.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
+++ b/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
@@ -1,3 +1,4 @@
+import base64
 from typing import Any
 
 from openinference.instrumentation import REDACTED_VALUE
@@ -32,9 +33,12 @@ def _redact_images(
     hide_input_images: bool,
     base64_image_max_length: int,
 ) -> Any:
-    """Walks the request tree and finds dicts with mime_type starting with
-    image/ and redacts
-    """
+    """Walk the request tree and redact configured image values."""
+    # if this is a Pydantic object, dump it as a walkable tree first
+    model_dump = getattr(value, "model_dump", None)
+    if callable(model_dump):
+        value = model_dump(mode="python")
+
     if isinstance(value, list):
         return [
             _redact_images(
@@ -44,34 +48,32 @@ def _redact_images(
             )
             for item in value
         ]
+    if isinstance(value, (bytes, bytearray, memoryview)):
+        return base64.b64encode(value).decode("ascii")
     if not isinstance(value, dict):
         return value
 
-    result = {
+    values = value
+    mime_type = value.get("mime_type")
+    if isinstance(mime_type, str) and mime_type.startswith("image/"):
+        values = dict(value)
+        if hide_input_images:
+            for key in ("data", "file_uri"):
+                if key in values:
+                    values[key] = REDACTED_VALUE
+        elif (
+            image_url_length := _get_image_url_length(values.get("data"), mime_type)
+        ) is not None and image_url_length > base64_image_max_length:
+            values["data"] = REDACTED_VALUE
+
+    return {
         key: _redact_images(
             item,
             hide_input_images=hide_input_images,
             base64_image_max_length=base64_image_max_length,
         )
-        for key, item in value.items()
+        for key, item in values.items()
     }
-    mime_type = result.get("mime_type")
-    if not isinstance(mime_type, str) or not mime_type.startswith("image/"):
-        return result
-
-    if hide_input_images:
-        if "data" in result:
-            result["data"] = REDACTED_VALUE
-        if "file_uri" in result:
-            result["file_uri"] = REDACTED_VALUE
-        return result
-
-    data = result.get("data")
-    if (
-        image_url_length := _get_image_url_length(data, mime_type)
-    ) is not None and image_url_length > base64_image_max_length:
-        result["data"] = REDACTED_VALUE
-    return result
 
 
 def _get_image_url_length(data: Any, mime_type: str) -> int | None:
--- a/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
+++ b/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
@@ -4,6 +4,8 @@ from typing import Any
 from openinference.instrumentation import REDACTED_VALUE
 from openinference.instrumentation.config import is_base64_url
 
+_IMAGE_VALUE_KEYS = ("data", "uri", "file_uri")
+
 
 def redact_images_from_request_parameters(
     request_parameters: dict[str, Any],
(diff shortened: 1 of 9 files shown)


sanchitram1:

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """Return the walked value and whether any image value was redacted."""
+    original_model = None
+    # Google GenAI request types are Pydantic models; convert them to a
+    # Python tree so nested media fields can be inspected before JSON encoding.
+    model_dump = getattr(value, "model_dump", None)
+    if callable(model_dump):
+        original_model = value
+        value = model_dump(mode="python")
+
+    if isinstance(value, list):
```
caroger, line 61: **[P1] Redact tuple interaction inputs before serialization**

This walker only descends into `list`. In `google-genai==2.0.1`, `InteractionsResource.create.input` accepts `Iterable[ContentParam]`, and the SDK converts a tuple into the JSON list sent on the wire. Because instrumentation processes the original arguments before that conversion, the tuple reaches this function and falls through unchanged at line 73. Consequently, `hide_input_images=True` still records the image data in `input.value`.

This test first verifies the upstream SDK behavior, then demonstrates the masking bypass:

```python
from google.genai._interactions._utils import maybe_transform
from google.genai._interactions.types import interaction_create_params

from openinference.instrumentation import REDACTED_VALUE
from openinference.instrumentation.google_genai._image_utils import (
    redact_images_from_request_parameters,
)


def test_hides_tuple_interaction_input_supported_by_google_genai() -> None:
    image = {
        "type": "image",
        "data": "c2Vuc2l0aXZlLWltYWdl",
        "mime_type": "image/png",
    }
    request = {"model": "gemini-2.5-flash", "input": (image,)}

    serialized = maybe_transform(
        request,
        interaction_create_params.CreateModelInteractionParamsNonStreaming,
    )
    assert serialized["input"] == [image]

    sanitized = redact_images_from_request_parameters(
        request,
        hide_input_images=True,
        base64_image_max_length=32_000,
    )
    assert sanitized["input"][0]["data"] == REDACTED_VALUE
```

On this branch, the final assertion receives the original base64 string. Could we cover tuples or other safely repeatable sequences here? Arbitrary one-shot iterables need care so tracing does not consume the request before the SDK sees it.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    changed = False
+    mime_type = value.get("mime_type")
+    if _is_image_container(value):
+        values = dict(value)
+        if hide_input_images:
+            for key in _IMAGE_VALUE_KEYS:
+                if key in values and values[key] != REDACTED_VALUE:
+                    values[key] = REDACTED_VALUE
+                    changed = True
+        elif isinstance(mime_type, str) and (
```
caroger: **[P1] Apply the size limit when image MIME type is omitted**

The upstream `ImageContentParam` requires `type="image"`, but `mime_type` is optional in both `google-genai` 2.0.1 and 2.14.0. `_is_image_container` correctly recognizes this as an image, but the length-limit branch additionally requires a string MIME type. Therefore, a valid image block without `mime_type` bypasses `base64_image_max_length`.

Reproduction:

```python
from google.genai._interactions._utils import maybe_transform
from google.genai._interactions.types import interaction_create_params

from openinference.instrumentation import REDACTED_VALUE
from openinference.instrumentation.google_genai._image_utils import (
    redact_images_from_request_parameters,
)


def test_applies_image_length_limit_when_mime_type_is_omitted() -> None:
    image = {"type": "image", "data": "A" * 100}
    request = {"model": "gemini-2.5-flash", "input": [image]}

    serialized = maybe_transform(
        request,
        interaction_create_params.CreateModelInteractionParamsNonStreaming,
    )
    assert serialized["input"] == [image]

    sanitized = redact_images_from_request_parameters(
        request,
        hide_input_images=False,
        base64_image_max_length=1,
    )
    assert sanitized["input"][0]["data"] == REDACTED_VALUE
```

The upstream serialization assertion passes, but the final assertion receives all 100 characters. Since `type="image"` already identifies the container, could the no-MIME case compare the encoded data length conservatively and redact when it clearly exceeds the configured limit?

caroger (changes requested):
Thanks for extending image masking into `input.value`. I verified the implementation against `google-genai` request serialization and found two supported input shapes that still emit raw image data.

The inline repros use `google-genai==2.0.1`, which is this package's minimum supported version. The missing-MIME case also reproduces with `2.14.0`.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
sanchitram1: thanks, this looks like a simple miss on my part.

python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_image_utils.py
sanchitram1: done

caroger (approved):

PR #3463, opened 2026-07-30
fix(bedrock): remove runtime imports for AWS SDK types

### LINKED ISSUE #3393: openinference-instrumentation-bedrock: ESM build crashes with "does not provide an export named 'ContentBlock'"
## Summary

`@arizeai/openinference-instrumentation-bedrock@0.4.17` cannot be imported from an **ESM** context. Any `import` of the package throws:

```
SyntaxError: The requested module '@aws-sdk/client-bedrock-runtime' does not provide an export named 'ContentBlock'
```

This happens under plain Node ESM **and** under `tsx`. The CommonJS (`require`) build works fine.

## Root cause

The ESM build value-imports type-only names from the AWS SDK. `dist/esm/types/bedrock-types.js` (compiled from `src/types/bedrock-types.ts`) does:

```js
import { ContentBlock, ContentBlockDelta, /* … */ } from "@aws-sdk/client-bedrock-runtime";
```

`ContentBlock` (and the other names) are **types**, not runtime exports, in `@aws-sdk/client-bedrock-runtime`. ESM enforces named-export existence at link time, so the module fails to load. In CommonJS the same import resolves to `undefined` at runtime without erroring, which is why `require()` works.

The source uses a plain `import { … }` where it should use `import type { … }` (or the package should compile with `verbatimModuleSyntax` / `isolatedModules` to catch type-only imports being emitted as value imports).

## Reproduce

```bash
mkdir repro && cd repro && npm init -y && npm pkg set type=module
npm i @arizeai/openinference-instrumentation-bedrock @aws-sdk/client-bedrock-runtime
node --input-type=module -e "import('@arizeai/openinference-instrumentation-bedrock').then(()=>console.log('ok')).catch(e=>console.log('ERR:', e.message))"
# ERR: The requested module '@aws-sdk/client-bedrock-runtime' does not provide an export named 'ContentBlock'
```

## Impact

- ESM/`tsx` consumers can't use the Bedrock JS instrumentor at all.
- `tsc --skipLibCheck` does not catch it (it skips `node_modules` and doesn't execute), so it only surfaces at runtime.

## Suggested fix

Change the value imports of SDK **types** in `src/types/bedrock-types.ts` (and any other `src` file importing SDK types) to `import type { … }`, and/or enable `verbatimModuleSyntax` so the compiler flags type-only imports emitted as value imports. The sibling `@arizeai/openinference-instrumentation-bedrock-agent-runtime` imports cleanly under ESM, so it can serve as a reference.

## Workaround

Load the instrumentor's CommonJS build from ESM via `createRequire`:

```ts
import { createRequire } from "module";
const require = createRequire(import.meta.url);
const { BedrockInstrumentation } = require("@arizeai/openinference-instrumentation-bedrock");
```

Environment: `@arizeai/openinference-instrumentation-bedrock@0.4.17`, `@aws-sdk/client-bedrock-runtime@3.1087.0`, Node.js v24.

LarryHu0217:
Fixes #3393

## Summary
- mark AWS Bedrock SDK imports and re-exports in `bedrock-types.ts` as type-only
- prevent those symbols from being emitted as runtime ESM imports
- add a build-artifact regression test and patch changeset

## Validation
- `pnpm --filter @arizeai/openinference-instrumentation-bedrock... run build`
- `pnpm --filter @arizeai/openinference-instrumentation-bedrock run test -- --reporter=default` (55 passed, 1 skipped)
- `pnpm run -r build`
- `pnpm run -r test`
- `pnpm run type:check`
- `pnpm run lint`
- `pnpm run fmt:check`
- `git diff --check`

### DIFF
--- a/js/.changeset/fix-bedrock-esm-types.md
+++ b/js/.changeset/fix-bedrock-esm-types.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/openinference-instrumentation-bedrock": patch
+---
+
+Fix the Bedrock instrumentation ESM build by removing runtime imports and exports for AWS SDK types.
--- a/js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
+++ b/js/packages/openinference-instrumentation-bedrock/src/types/bedrock-types.ts
@@ -5,7 +5,7 @@
  * custom types for structures not exposed by the SDK.
  */
 
-import {
+import type {
   // Core content and message types from SDK
   ContentBlock,
   ContentBlockDelta,
@@ -36,7 +36,7 @@ import { diag } from "@opentelemetry/api";
 import { isObjectWithStringKeys } from "@arizeai/openinference-core";
 
 // Re-export AWS SDK types for convenience
-export {
+export type {
   ContentBlock,
   Message,
   SystemContentBlock,
--- a/js/packages/openinference-instrumentation-bedrock/test/esm-build.test.ts
+++ b/js/packages/openinference-instrumentation-bedrock/test/esm-build.test.ts
@@ -1,0 +1,15 @@
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+
+import { describe, expect, it } from "vitest";
+
+describe("ESM build", () => {
+  it("does not emit runtime imports for AWS SDK types", () => {
+    const emittedTypes = readFileSync(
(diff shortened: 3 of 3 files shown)


LarryHu0217:

mikeldking (approved):
thanks @LarryHu0217

PR #3478, opened 2026-08-02
fix(google-adk): trace compaction spans on ADK 1.32+

### LINKED ISSUE #3477: [bug] google adk events compaction is not traced by `GoogleADKInstrumentor`
**Describe the bug**

`GoogleADKInstrumentor` correctly [patches](https://github.com/Arize-ai/openinference/blob/59ea35e3b69c830a26c0560825dde00bd43e292d/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py#L170-L180) the tracer defined in `google.adk.telemetry.tracing`. However, it does not rebind the [module-local tracer](https://github.com/google/adk-python/blob/v1.32.0/src/google/adk/apps/compaction.py#L57) imported by `google.adk.apps.compaction`, which was introduced in Google ADK 1.32. 

Because the compaction module imports the tracer directly, its local binding continues to reference the original ADK tracer after `google.adk.telemetry.tracing.tracer` is replaced. Consequently, the `compact_events` span is created with the original tracer and does not appear in the OpenInference trace exported to Phoenix.

This is the same module-binding issue handled for the [local tracer](https://github.com/google/adk-python/blob/v1.32.0/src/google/adk/flows/llm_flows/functions.py#L47) in `google.adk.flows.llm_flows.functions`, which `GoogleADKInstrumentor` already [rebinds separately](https://github.com/Arize-ai/openinference/blob/59ea35e3b69c830a26c0560825dde00bd43e292d/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py#L186-L192). 

**To Reproduce**

Using the following package versions: 

```
google-adk==1.32.0
openinference-instrumentation-google-adk==0.1.18
```

The issue is expected to affect Google ADK versions from 1.32 onward.

Configure `GoogleADKInstrumentor` with a tracer provider that exports spans to Phoenix. Then define an `EventsCompactionConfig` in the ADK application:

```
events_compaction_config = EventsCompactionConfig(
        compaction_interval=2,
        overlap_size=3,
    )
app = App(
    name="example_app",
    root_agent=root_agent,
    events_compaction_config=events_compaction_config,
)
```
Run at least two completed invocations so that compaction is triggered.

In the screenshots below, compaction is configured to run after two invocations. Phoenix shows the second invocation, but its trace does not contain the expected `compact_events sliding_window` span:

Image

Opening the second trace confirms that the compaction span is missing:

Image

**Expected behavior**

When event compaction is triggered, the resulting `compact_events` span should be created with the OpenInference tracer and appear in the Phoenix trace:

Image

Rebinding `google.adk.apps.compaction.tracer`, in the same way that other module-local tracer bindings are handled by `GoogleADKInstrumentor`, makes the span visible.

 - OS: Linux
 - Version: `Linux-6.8.0-1029-azure-x86_64-with-glibc2.35`

**Additional context**

My team is currently using a small production fix that rebinds the module-local compaction tracer during instrumentation and restores it during uninstrumentation, if you want I can submit a PR containing the fix, lmk.

latent-variable-17:
## Summary
This fixes missing compaction spans when using Google ADK 1.32+ with GoogleADKInstrumentor.

google.adk.apps.compaction keeps its own local tracer reference, so updating the global ADK tracer alone was not enough.

## What changed
- Rebind google.adk.apps.compaction.tracer during instrumentation
- Restore the original google.adk.apps.compaction.tracer during uninstrumentation
- Add test assertions to cover both patch and restore behavior

## Why this fixes the bug
Compaction spans, including compact_events, now use the OpenInference tracer path, so they appear correctly in traces.

Fixes #3477

Screenshots:
image

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/__init__.py
@@ -1,6 +1,7 @@
+import contextvars
 import logging
-from importlib import import_module
-from typing import Any, Collection, Dict, Iterator, List, Tuple, cast
+import sys
+from typing import Any, Collection, Dict, Iterator, List, Optional, Tuple, cast
 
 import wrapt
 from opentelemetry import trace as trace_api
@@ -11,13 +12,30 @@ from opentelemetry.trace import Span, Tracer, get_current_span
 from opentelemetry.util._decorator import _agnosticcontextmanager
 from wrapt import resolve_path, wrap_function_wrapper
 
-from openinference.instrumentation import OITracer, TraceConfig
+from openinference.instrumentation import (
+    OITracer,
+    TraceConfig,
+    get_input_attributes,
+    get_output_attributes,
+    safe_json_dumps,
+)
 from openinference.instrumentation.google_adk.version import __version__
+from openinference.semconv.trace import (
+    OpenInferenceMimeTypeValues,
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+)
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
 _instruments = ("google-adk >= 1.2.1",)
+
+_COMPACTION_MODULE = "google.adk.apps.compaction"
+
+_compaction_input_var: "contextvars.ContextVar[Optional[Dict[str, Any]]]" = contextvars.ContextVar(
+    "_openinference_compaction_input", default=None
+)
 
 
 class GoogleADKInstrumentor(BaseInstrumentor):  # type: ignore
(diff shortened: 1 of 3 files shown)


jackrua:
Thank you for opening a PR for my issue, however I don't think this is the right approach. 

The tracer you are using (`_SelectiveExecuteToolTracer`) checks first if the span name starts with `"execute_tool"` and then creates a new span through the OpenInference tracer, but the ADK [compaction span name](https://github.com/google/adk-python/blob/v1.32.0/src/google/adk/apps/compaction.py#L57) is either `compact_events sliding_window` or `compact_events token_threshold` (based on the fact that was triggered by events count or tokens count), so no distinct compaction span would be created. Its attributes may instead be written to the current parent span.

latent-variable-17:
Thanks, you are right that the previous check only created OI spans for `execute_tool*`, which would suppress distinct compaction spans. I updated `_SelectiveExecuteToolTracer` to also route `compact_events*` names to the OI tracer (while still suppressing unrelated span families) and added a targeted test for the routing behavior.

satyadevai (approved):

caroger (approved):

Issue #2901, opened 2026-03-18
[feature request] Add `finish_reason` to LLM output message attributes


uu9:
**Is your feature request related to a problem? Please describe.**

`finish_reason` is not captured in the OpenInference semantic conventions or any instrumentation library (e.g. `openinference-instrumentation-openai`, `openinference-instrumentation-langchain`). This makes it impossible to observe why a model stopped generating — e.g. distinguishing `stop`, `length`, `tool_calls`, or `content_filter` — which is critical for debugging truncation issues, tool call flows, and safety filtering.

**Describe the solution you'd like**

1. Add `finish_reason` to the semantic conventions spec as a new attribute on output messages:

   ```
   llm.output_messages.{i}.message.finish_reason  String  e.g. "stop", "length", "tool_calls"
   ```

2. Collect it in `openinference-instrumentation-openai` from `choice.finish_reason` when processing `ChatCompletion` and `Completion` responses.

3. Collect it in `openinference-instrumentation-langchain` from the `generation_info` field in LangChain run outputs.

**Describe alternatives you've considered**

Monkey-patching `_ResponseAttributesExtractor._get_attributes_from_chat_completion` to inject `finish_reason` manually — works but is fragile and requires users to maintain the patch across library upgrades.

**Additional context**

OpenAI's `choice.finish_reason` is a standard field in all Chat Completion responses. The OTel GenAI semantic conventions already track this via `gen_ai.response.finish_reasons`. Aligning OpenInference with this would improve interoperability and observability coverage.

mikeldking:
@uu9 this makes sense - any particular instrumentation you are looking for it in?

golikovichev:
Strong +1, and I would single out the content_filter case as the one that hurts most in production. A safety-filtered response often comes back as HTTP 200 with an empty or near-empty content body and finish_reason=content_filter. Without finish_reason on the span, that is indistinguishable from a normal short answer, so it slips past health checks and retry logic and the user just sees a blank reply.

I hit this from the testing side. I build regression cases from captured traces, and today there is no clean way to assert "this span was a silent content-filter stop, not a real completion" because the reason never reaches the trace. Having llm.output_messages.{i}.message.finish_reason would let both live observability and after-the-fact test generation catch truncation and filtered stops directly.

One question on shape: for providers that can return more than one stop reason, is a single string per output message the intended scope, or would a list per choice (mirroring OTel gen_ai.response.finish_reasons) be preferable for interop?

[...]

PR #3484, opened 2026-08-03
feat(vertexai): Add Finish Reason Attribute

ialisaleh:
Refs #2901 

This PR adds `finish_reason` tracking to the Vertex AI instrumentation by mapping the first candidate's `finish_reason` enum to the `LLM_FINISH_REASON` span attribute for both streaming and non-streaming `GenerateContent` responses, keeping it consistent with other instrumentations.

image


caroger (changes requested):
I found two finish-reason edge cases that should be addressed before merging:

- **Skip the protobuf unspecified value** (`_wrapper.py:359`). Proto-plus returns `FINISH_REASON_UNSPECIFIED` (enum value `0`), not `None`, when `finish_reason` is omitted. The current check therefore emits `llm.finish_reason=FINISH_REASON_UNSPECIFIED` even though the provider supplied no actual reason. Please omit zero/unspecified values and add a missing-reason test.

- **Fix the integer enum fallback** (`_wrapper.py:360-363`). For an unknown forward-compatible protobuf enum value, proto-plus returns an `int`. The fallback then evaluates `Candidate.FinishReason`, but `Candidate` is a typing `Union` alias and has no `FinishReason` attribute. This logs an instrumentation exception and drops the finish reason. Please convert through the concrete candidate type or use a safe numeric/string fallback, with a test for an unknown value.

The targeted Vertex AI checks otherwise pass: formatting, Ruff, mypy, and all 57 GAPIC tests.

ialisaleh:
Hi @caroger - I addressed both edge cases in the PR. It now correctly treats the proto3 default `FINISH_REASON_UNSPECIFIED (0)` as unset, while enum handling resolves through the concrete candidate type and falls back to the raw integer string for unknown values. I also added tests covering both scenarios. Please take a look when you have a chance. Thanks!

mikeldking (approved):
Both requested finish-reason edge cases are addressed. Pinned and latest Vertex AI GAPIC tests pass, along with Ruff and mypy checks.

PR #3488, opened 2026-08-04
fix(anthropic): record cache read/write token details on streaming paths

### LINKED ISSUE #3487: [bug] Streaming path never emits separate cache_read/cache_write token attributes instead it folds them into the combined prompt total
**Describe the bug**
_MessageExtractor in _stream.py never emits llm.token_count.prompt_details.cache_read or .cache_write as separate span attributes for streamed calls. The values are present in the Anthropic SDK's own Usage object (usage.cache_read_input_tokens, usage.cache_creation_input_tokens), but get folded into the combined llm.token_count.prompt total instead of broken out. Affects both client.messages.stream() and client.messages.create(stream=True). Both routes hit the same extractor and the same folding logic, so switching between them doesn't help.

**To Reproduce**

1. Instrument an Anthropic client with AnthropicInstrumentor.
2. Make a call via client.messages.stream() (or client.messages.create(stream=True)) where the request includes a cached prefix (cache_control) that produces a cache read or write.
3. Inspect the resulting span's attributes.
4. Observe: llm.token_count.prompt_details.cache_read/.cache_write are absent. output.value (the raw serialized response) does contain usage.cache_read_input_tokens/cache_creation_input_tokens — so the data exists but never reaches span attributes.

**Expected behavior**
cache_read/cache_write should appear as separate llm.token_count.prompt_details.* span attributes, matching what already happens on the non-streaming path.

**Additional context**

- openinference-instrumentation-anthropic==1.0.6 (also reproduces on main, 1.1.0)
- #2234 is open but targets an older _stream.py structure that's since been refactored — doesn't apply to current code.
- #3243 fixed the same category of bug for the strands-agents instrumentor, and was merged.

axiom-of-choice:
Fixes #3487

`_MessageExtractor.get_attributes` folded prompt cache tokens into `llm.token_count.prompt` but never emitted the two `prompt_details.*` keys that `_get_llm_token_counts` already yields on the non-streaming path, so `messages.stream()` and `messages.create(stream=True)` lost the breakout.

Both streaming paths funnel into this one extractor, so a single change covers them. Zero counts are skipped, matching the non-streaming behavior.

Same usage served three ways, before and after:

```
served usage: input=10, output=5, cache_creation=1733, cache_read=512

                         before                     after
path                     cache_read  cache_write    cache_read  cache_write
create(stream=False)     512         1733           512         1733
create(stream=True)      None        None           512         1733
messages.stream()        None        None           512         1733
```

`llm.token_count.prompt` was already correct at 2255 on all three.

## Tests

- `test_message_extractor_records_cache_token_details`, parametrized over (read, write) = (512,1733), (512,0), (0,1733), (0,0).
- `test_cache_token_details_match_between_streaming_and_non_streaming`, which serves identical usage as JSON and as SSE through respx and asserts the three paths agree.

4 of the 5 fail without the source change; `(0,0)` passes either way since zeros are omitted. Full suite 47 passed. `ruff` 0.9.2 and `mypy` 1.11.2 clean (the 16 `unused-ignore` errors in `__init__.py` are present on unmodified `main` as well).

## Not included

`llm.token_count.total` is emitted only by the streaming path; `_get_llm_token_counts` never yields it, so the two paths disagree on that key regardless of caching. Left alone to keep this PR to one thing. Happy to file it separately.

`Usage.cache_read_input_tokens` and `cache_creation_input_tokens` are both optional on the model, but direct attribute access is safe at the pinned floor (`anthropic >= 0.84.0`) and matches what `_wrappers.py` already does, so no `getattr` guard.

Tooling: Claude Opus 5.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
@@ -398,6 +398,16 @@ class _MessageExtractor:
             yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT, prompt_tokens
         if usage.output_tokens:
             yield SpanAttributes.LLM_TOKEN_COUNT_COMPLETION, usage.output_tokens
+        if usage.cache_read_input_tokens:
+            yield (
+                SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ,
+                usage.cache_read_input_tokens,
+            )
+        if usage.cache_creation_input_tokens:
+            yield (
+                SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE,
+                usage.cache_creation_input_tokens,
+            )
         if total := prompt_tokens + (usage.output_tokens or 0):
             yield SpanAttributes.LLM_TOKEN_COUNT_TOTAL, total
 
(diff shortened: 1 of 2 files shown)


axiom-of-choice:

mikeldking (approved):
thanks for the fix 🙏🏻

PR #3492, opened 2026-08-04
fix(openai-agents): scope llm.system to LLM spans, add agent name, tool schema, and handoff I/O

jimbobbennett:
Splits the uncontroversial half out of #3389.

The Agents SDK leaves several attributes off spans that it *does* have the data for. This records them from values the SDK actually supplies, and deliberately leaves out the parts of #3389 that inferred values it does not.

## What this records

| Attribute | Span | Source |
|---|---|---|
| `llm.system` | LLM only | previously stamped on **every** span kind, including agent, tool and handoff spans — bug fix |
| `agent.name` | AGENT | `AgentSpanData.name` |
| `tool.description` | TOOL | the live `FunctionTool` being invoked |
| `tool.parameters` | TOOL | the live `FunctionTool` being invoked |
| `input.value` | handoff (TOOL) | `HandoffSpanData.from_agent` |
| `output.value` | handoff (TOOL) | `HandoffSpanData.to_agent` |

## Tool schema sourcing

`FunctionSpanData` carries only a tool's name, input and output, so the schema is read off the live tool object at the point of invocation — the same way smolagents, google-adk, crewai and ten other instrumentors source these attributes.

Because that step is bound by name at several call sites and has moved between modules across SDK releases (a classmethod on `RunImpl` in 0.2.x, a module-level function in `run_internal.tool_execution` in 0.18+, with a different signature and return type), every binding of it is located by scanning rather than by a hand-maintained list, and the wrapper reads only the `tool_runs` keyword so it is indifferent to the rest. When no binding can be found the two attributes are simply absent, which is a missing enrichment rather than a failure.

The schemas are published in a `ContextVar` scoped to the execution step that owns the function spans, so there is no cross-span cache to bound, evict or clean up, and no way for a live trace to lose its schemas.

## What this deliberately does not do

Input and output are **not** inferred onto agent, task or turn spans, or onto the trace root, from their child LLM spans. Those span data types have no input/output fields, so any value would be derived from child completion order rather than observed at the operation boundary — which parallel branches, guardrails (`run.py` runs input guardrails concurrently with the model turn), and transformed final outputs would all get wrong.

No other instrumentor in this repo infers parent I/O from child spans. google-adk shows the alternative: it wraps the agent run and reads the agent's real final response. The equivalent here would be reading `RunResult.input` / `.final_output` at the `Runner` level, which is a separate change.

`tool.name` on handoff spans is also left out: `HandoffSpanData` does not carry it, and a reconstructed `transfer_to_<agent>` is wrong whenever `handoff(tool_name_override=...)` is used.

Two tests pin these decisions so they cannot be quietly reintroduced: `test_ancestor_spans_do_not_infer_input_output` and `test_handoff_span_does_not_guess_tool_name`.

## Testing

- 203 tests pass against the pinned `openai_agents==0.2.6`
- The tool-schema and span-attribute suites also pass against `openai-agents==0.18.3`, since the patch target differs between those versions
- End-to-end coverage runs a real `Agent` through `Runner.run` with a fake model and asserts the attributes reach the exported span — a unit test cannot prove the wrapper hooks the right function
- Per-span-kind tests assert the complete attribute set, and the processor is built over an `OITracer` so `TraceConfig` masking is exercised as in production

No dependency or version changes: the `>= 0.2.6` floor already covers newer releases, and the code is verified against both ends of that range.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/README.md
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/README.md
@@ -6,6 +6,23 @@ Python auto-instrumentation library for OpenAI Agents python SDK.
 
 The traces emitted by this instrumentation are fully OpenTelemetry compatible and can be sent to an OpenTelemetry collector for viewing, such as [Arize Phoenix](https://github.com/Arize-ai/phoenix) or [Arize AX](https://arize.com/docs/ax).
 
+## Compatibility
+
+| `openinference-instrumentation-openai-agents` | `openai-agents` | Python           |
+| --------------------------------------------- | --------------- | ---------------- |
+| `>=2.0`                                        | `>=0.11.0`      | `>=3.10, <3.15`  |
+| `>=1.4.1, <2.0`                                | `>=0.2.6`       | `>=3.10, <3.15`  |
+| `>=1.2.0, <1.4.1`                              | `>=0.2.6`       | `>=3.9, <3.14`   |
+| `>=1.0.0, <1.2.0`                              | `>=0.1.0`       | `>=3.9, <3.14`   |
+| `<1.0.0`                                       | `>=0.0.3`       | `>=3.9, <3.14`   |
+
+Instrumentor `>=2.0` requires `openai-agents>=0.11.0`. Three things changed below that floor
+which the instrumentor no longer accommodates: the run internals moved out of
+`agents._run_impl` into `agents.run_internal` in 0.8.0, `openai-agents` moved from
+`openai<2` to `openai>=2.9` in the same release, and tool namespaces
+(`agents.tool_namespace`) arrived in 0.11.0. Tools grouped by a namespace report
+`tool.description` and `tool.parameters` only on instrumentor `>=2.0`.
+
 ## Installation
 
 ```shell
--- a/python/instrumentation/openinference-instrumentation-openai-agents/examples/realtime_with_tools.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/examples/realtime_with_tools.py
@@ -209,18 +209,21 @@ async def run(args: argparse.Namespace) -> None:
             stdin_task.cancel()
             await asyncio.gather(send_task, events_task, stdin_task, return_exceptions=True)
 
-        with sd.InputStream(
-            samplerate=SAMPLE_RATE,
-            channels=CHANNELS,
-            dtype="int16",
-            blocksize=MIC_CHUNK_FRAMES,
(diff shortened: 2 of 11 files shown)


jimbobbennett:
## Before / after against a real app

Verified end to end using the OpenAI Agents variant of [project-rosetta-stone](https://github.com/Arize-ai/project-rosetta-stone) — its real agent, tools and ChromaDB — with one query that triggers two tool calls. Same query, same app code; only the instrumentor was swapped:

- **before** — `main` (`0d4e3113`)
- **after** — this branch

Traces went to a local Phoenix instance. `openai-agents` resolved to **0.19.3**, so this also exercises the version tolerance: the binding scan found 3 bindings of the tool-execution step and needed no changes for a release newer than any pinned in CI.

### Attribute counts across the 11 spans of the trace

Both runs produced identical span trees, so this is a like-for-like comparison.

| Attribute | before | after | |
|---|---:|---:|---|
| `agent.name` | 0 | 1 | new — on the agent span |
| `tool.description` | 0 | 2 | new — both tool spans |
| `tool.parameters` | 0 | 2 | new — both tool spans |
| `llm.system` | 10 | 3 | narrowed to the 3 LLM spans |

### `search_products` (TOOL span)

**before — 7 attributes**

```
output.value, tool.name, session.id, input.mime_type, input.value,
openinference.span.kind = TOOL, llm.system = openai        <-- on a TOOL span
```

1-BEFORE-tool-span

3-BEFORE-agent-span

**after — 36 attributes**

```
output.value, tool.name, session.id, input.mime_type, input.value,
openinference.span.kind = TOOL,
tool.description = "Search the toy store inventory by text query, keywords, age range, ..."
tool.parameters.properties.query.*        <-- full schema, expanded by Phoenix
tool.parameters.properties.keywords.*
tool.parameters.properties.min_age.*
tool.parameters.properties.max_age.*
...
                                                            <-- llm.system gone
```

2-AFTER-tool-span
4-AFTER-agent-span

### `WonderToysAgent` (AGENT span)

Four attributes on both sides, which makes the swap easy to see:

| before | after |
|---|---|
| `session.id` | `session.id` |
| `openinference.span.kind` = AGENT | `openinference.span.kind` = AGENT |
| `graph.node.id` = WonderToysAgent | `graph.node.id` = WonderToysAgent |
| `llm.system` = openai | **`agent.name`** = WonderToysAgent |

`llm.system` is replaced by `agent.name` — the spurious attribute removed, the missing one added.

Screenshots of the Phoenix span detail for each of these four cases to follow.

RogerHYang (approved):

PR #3498, opened 2026-08-06
fix(anthropic): emit llm.token_count.total on the non-streaming path

### LINKED ISSUE #3490: [bug] llm.token_count.total is emitted only on the streaming path of the Anthropic instrumentor
### Describe the bug

`llm.token_count.total` is emitted only on the streaming path of the Anthropic instrumentor. `_get_llm_token_counts` (`_wrappers.py:695-711`) yields prompt, completion and the two `prompt_details.*` keys but never `total`, while `_MessageExtractor.get_attributes` (`_stream.py:401-412`) does yield it. So within one instrumentor the presence of the key depends on whether the caller asked for streaming.

Noticed while working on #3487; deliberately left out of #3488 to keep that PR to one thing.

### Reproduction

Same usage served two ways through a mocked endpoint (`input_tokens=10`, `output_tokens=5`, no caching involved):

```
path                     prompt  completion  total
create(stream=False)         10           5   None
create(stream=True)          10           5     15
```

`prompt` and `completion` agree; only `total` diverges.

### Why it matters

22 of the 33 Python instrumentors emit `llm.token_count.total`, including every direct peer: groq, mistralai, bedrock, vertexai, google-genai. Anything aggregating spend or usage by summing that attribute silently undercounts non-streaming Anthropic calls, with no error to signal it. The semantic conventions describe the key as "Total number of tokens, including both prompt and completion (in tokens)" with no mention of it being optional.

### The contract question

This is why I'm filing rather than opening a PR. The peers above all forward a `total_tokens` the provider reports; Anthropic's `Usage` model has no total field at all (confirmed: no field name contains "total"), so the instrumentor has to compute it, which makes the definition a choice you should make:

1. `prompt + completion`, where `prompt` already includes cache read and write tokens. This is what the streaming path emits today, so non-streaming would just match it. It also keeps `total == prompt + completion` true as an identity.
2. `input_tokens + output_tokens` only, excluding cached tokens, which is closer to "tokens you were billed at the base rate" but breaks that identity and would change the value streaming reports now.
3. Don't emit it here at all, and instead drop it from the streaming path so the two agree by omission. Consistent, but diverges from the other 22 instrumentors and from the semconv wording.

Option 1 is the smallest change and the only one that doesn't alter existing streaming output, but it does mean a cached-heavy call reports a large `total`. Happy to send the patch once you say which contract you want, or to close this if the asymmetry is intentional.

### Versions

`openinference-instrumentation-anthropic` on `main` (post-#3488), `anthropic==0.84.0`.

axiom-of-choice:
# Description

Resolves #3490.

`_get_llm_token_counts` yielded prompt, completion and the two `prompt_details.*` keys but never `total`, while `_MessageExtractor` on the streaming path did. Within one instrumentor the presence of the key depended on whether the caller asked for streaming, so anything aggregating usage by summing it undercounted non-streaming calls with no error to signal it.

Anthropic's `Usage` has no total field, so this derives it the way the streaming path already does: `prompt + completion`, where prompt includes cache read and write tokens (option 1 in the issue). A zero total is skipped, matching the other counts.

```
                         before              after
path                     prompt  total       prompt  total
create(stream=False)         10   None           10     15
create(stream=True)          10     15           10     15

with caching (input=10, output=5, write=1733, read=512)
create(stream=False)       2255   None         2255   2260
create(stream=True)        2255   2260         2255   2260
```

## Behavior change

Streaming output is untouched, but non-streaming spans gain a key they didn't emit before. 10 existing non-streaming tests failed on their trailing `assert not attributes`, which means the suite was in effect asserting `total` is absent there; each got the matching assertion. Consumers reading `total` see a value where they previously saw none. Under this contract a cache-heavy call reports a large `total`, since prompt includes cache reads and writes; that is already what streaming reports today.

## Tests

- `test_token_count_total_matches_across_paths`, parametrized with and without caching, compares `_get_llm_token_counts` and `_MessageExtractor` output directly so the two producers can't drift apart again, and asserts `total == prompt + completion`.
- `test_token_count_total_omitted_when_there_is_no_usage` for the zero case.
- Extended the #3488 parity test to cover `total`; it had to exclude that key before.

50 pass, 13 fail without the source change. `ruff` clean; `mypy` clean apart from 16 pre-existing `unused-ignore` errors in `__init__.py` present on unmodified `main`.

## Risk

Low. 4 lines in one function, no API surface change. The risk is the added attribute itself, not the derivation.

# Checklist:

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
@@ -20,6 +20,7 @@ from openinference.instrumentation import safe_json_dumps
 from openinference.instrumentation.anthropic._utils import (
     _as_output_attributes,
     _finish_tracing,
+    _get_token_counts,
     _ValueAndType,
 )
 from openinference.instrumentation.anthropic._with_span import _WithSpan
@@ -388,28 +389,7 @@ class _MessageExtractor:
                     safe_json_dumps(block.input),
                 )
                 tool_idx += 1
-        usage = snapshot.usage
-        prompt_tokens = (
-            usage.input_tokens
-            + (usage.cache_creation_input_tokens or 0)
-            + (usage.cache_read_input_tokens or 0)
-        )
-        if prompt_tokens:
-            yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT, prompt_tokens
-        if usage.output_tokens:
-            yield SpanAttributes.LLM_TOKEN_COUNT_COMPLETION, usage.output_tokens
-        if usage.cache_read_input_tokens:
-            yield (
-                SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_READ,
-                usage.cache_read_input_tokens,
-            )
-        if usage.cache_creation_input_tokens:
-            yield (
-                SpanAttributes.LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE,
-                usage.cache_creation_input_tokens,
-            )
-        if total := prompt_tokens + (usage.output_tokens or 0):
-            yield SpanAttributes.LLM_TOKEN_COUNT_TOTAL, total
+        yield from _get_token_counts(snapshot.usage)
 
 
 class _ValuesAccumulator:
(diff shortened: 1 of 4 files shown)


axiom-of-choice:

mikeldking (approved):
thanks for this

mikeldking:
recheck

axiom-of-choice:
@mikeldking My pleasure. Btw it seems there is an ongoing issue with GH actions so no checks have run

https://www.githubstatus.com/

PR #3504, opened 2026-08-06
fix(anthropic): guard usage.input_tokens with or 0 to prevent TypeError on None

### LINKED ISSUE #3499: [bug] Anthropic instrumentor: unguarded usage.input_tokens can blank out an entire streaming LLM span
### Summary

`_get_token_counts` (recently extracted into `_utils.py` and now shared by the streaming and non-streaming paths) sums `usage.input_tokens` without the `or 0` guard applied to every other field:

```python
prompt_tokens = (
    usage.input_tokens                          # <- no `or 0`
    + (usage.cache_creation_input_tokens or 0)
    + (usage.cache_read_input_tokens or 0)
)
```

If `usage.input_tokens` is `None`, `None + 0` raises `TypeError`.

### Why it matters more now

Both paths call the same helper, but they wrap it differently:

- **Non-streaming** — `_wrappers._get_llm_token_counts` is decorated with `@_stop_on_exception`, so the error is swallowed and the span quietly loses its token counts.
- **Streaming** — `_MessageExtractor.get_attributes()` has no such guard. The exception escapes into `_utils._finish_tracing`, which catches it and sets `attributes = None`. The exported LLM span then ends with **no** input messages, output messages, model name or token counts — an empty LLM span.

So one nullable field on a partial/delta-shaped `Usage` degrades a whole streaming span, not just its token attributes.

### Suggested fix

Guard `input_tokens` like the rest (`(usage.input_tokens or 0)`), and consider giving the streaming extractor the same `@_stop_on_exception`-style containment so a token-count failure can never blank out the rest of a span's attributes.

### Notes

Pre-existing behavior — not introduced by #3490's fix, which only moved the code. Found during a code review of that change.

MSaiRam10:
Fixes #3499

## Problem

`usage.input_tokens` is summed without an `or 0` guard in both `_stream.py` and `_wrappers.py`. When `input_tokens` is `None` on a partial/delta-shaped `Usage` object, `None + 0` raises `TypeError`.

The impact differs by path:
- **Non-streaming** (`_wrappers.py`) — caught by `@_stop_on_exception`, span quietly loses token counts
- **Streaming** (`_stream.py`) — exception escapes into `_finish_tracing`, sets `attributes = None`, blanking out the entire LLM span (no input messages, output messages, model name, or token counts)

## Fix

Added `or 0` guard to `usage.input_tokens` in both files, consistent with how `cache_creation_input_tokens` and `cache_read_input_tokens` are already guarded.

## Changes

- `_stream.py` line 393: `usage.input_tokens` → `(usage.input_tokens or 0)`
- `_wrappers.py` line 701: `usage.input_tokens` → `(usage.input_tokens or 0)`

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_utils.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_utils.py
@@ -26,9 +26,9 @@ def _get_token_counts(usage: "Usage") -> Iterator[Tuple[str, AttributeValue]]:
     - input_tokens: Number of input tokens which were not read from or used to create a cache.
     """
     prompt_tokens = (
-        usage.input_tokens
-        + (usage.cache_creation_input_tokens or 0)
-        + (usage.cache_read_input_tokens or 0)
+    (usage.input_tokens or 0)
+    + (usage.cache_creation_input_tokens or 0)
+    + (usage.cache_read_input_tokens or 0)
     )
     if prompt_tokens:
         yield SpanAttributes.LLM_TOKEN_COUNT_PROMPT, prompt_tokens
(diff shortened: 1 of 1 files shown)


MSaiRam10:

mikeldking (approved):

PR #3529, opened 2026-08-08
fix: pydantic io should include tool calls in input.value/output.value

### LINKED ISSUE #3462: [pydantic-ai] TOOL spans missing input.value; tool-calling LLM spans missing output.value (0.1.17)
### Summary

`OpenInferenceSpanProcessor` does not populate:

- **`input.value` on `TOOL` spans** (the tool arguments), and
- **`output.value` on the `LLM` span that emits a tool call**.

Because Phoenix / Langfuse render their Input & Output panes from `input.value` / `output.value`, tool inputs and tool-calling generation outputs come up blank in the UI. The content *is* present on the spans under pydantic-ai's native keys (`tool_arguments`, `gen_ai.output.messages`) — it just isn't mapped onto the OpenInference semantic-convention attributes.

Plain-text agent responses were fixed in 0.1.16 (#3153) and are fine; this is specifically the **tool-call path**.

### Versions

- `openinference-instrumentation-pydantic-ai==0.1.17` (latest)
- `pydantic-ai-slim==1.87.0`
- `opentelemetry-sdk` (current)

### Reproduction

```python
from opentelemetry.trace import set_tracer_provider
from openinference.instrumentation.pydantic_ai import OpenInferenceSpanProcessor
from pydantic_ai import Agent
from pydantic_ai.models.test import TestModel

exp = InMemorySpanExporter()
prov = TracerProvider()
prov.add_span_processor(OpenInferenceSpanProcessor())
prov.add_span_processor(SimpleSpanProcessor(exp))
set_tracer_provider(prov)

agent = Agent(TestModel(), instrument=True)

@agent.tool_plain
def add(a: int, b: int) -> int:
    return a + b

agent.run_sync("please add 2 and 3")
prov.force_flush()

for s in exp.get_finished_spans():
    a = dict(s.attributes or {})
    print(
        f"span={s.name!r:20} kind={a.get('openinference.span.kind')!s:6} "
        f"input.value={'input.value' in a} output.value={'output.value' in a}"
    )
```

### Actual output (0.1.17)

```
span='chat test'          kind=LLM    input.value=True  output.value=False   # tool-calling generation: no output.value
span='running tool'       kind=TOOL   input.value=False output.value=True    # tool span: no input.value
span='chat test'          kind=LLM    input.value=True  output.value=True
span='agent run'          kind=AGENT  input.value=True  output.value=True
```

### Expected

- The `TOOL` span carries the tool arguments as `input.value`.
- The `LLM` span that produces a tool call carries its response (the tool call) as `output.value`.

### Notes

Happy to help / open a PR if useful. As a stop-gap we run a small downstream span processor that fills `input.value` / `output.value` from the native keys when they're absent, but it'd be great to drop it once the bridge maps these.

sanchitram1:
fixes #3462

- 16 / 18 of our instrumentors that emit TOOL spans set input_value. `pydantic-ai` was an outlier. 

suppose the LLM span emitting the tool call was:

```json
{
  "gen_ai.operation.name": "chat",
  "gen_ai.input.messages": [
    {
      "role": "user",
      "parts": [
        {
          "type": "text",
          "content": "What's the weather in Paris?"
        }
      ]
    }
  ],
  "gen_ai.output.messages": [
    {
      "role": "assistant",
      "parts": [
        {
          "type": "tool_call",
          "id": "pyd_ai_tool_call_id__get_weather",
          "name": "get_weather",
          "arguments": {
            "city": "a"
          }
        }
      ]
    }
  ]
}
```

now, the captured result will be

| key | before | now | 
| --- | --- | --- |
| input.value | "what's the weather in Paris?" | "what's the weather in Paris?" | 
| output.value | None | `[{"id": "pyd_ai_tool_call_id__get_weather", "name": "get_weather", "arguments": {"city": "a"}}]'` |
| output.mime_type | None | application/json |

### DIFF
--- a/python/instrumentation/openinference-instrumentation-pydantic-ai/src/openinference/instrumentation/pydantic_ai/semantic_conventions.py
+++ b/python/instrumentation/openinference-instrumentation-pydantic-ai/src/openinference/instrumentation/pydantic_ai/semantic_conventions.py
@@ -30,6 +30,7 @@ from openinference.instrumentation import safe_json_dumps
 from openinference.semconv.trace import (
     MessageAttributes,
     MessageContentAttributes,
+    OpenInferenceMimeTypeValues,
     OpenInferenceSpanKindValues,
     SpanAttributes,
     ToolAttributes,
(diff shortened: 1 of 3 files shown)


caroger (changes requested):
Requesting changes for one correctness gap in the new `output.value` fallback.

When a Pydantic AI response contains both a text part and a tool-call part, `output_value` is non-null, so the current branch emits only the text and discards the collected tool calls. Consumers that render from `output.value` therefore still cannot see the tool call for this valid response shape—the issue this PR is intended to fix remains partially open.

I reproduced this at head `296d31f8`: a mixed response emitted `output.value = "I will check."` and no `output.mime_type`, while `llm.output_messages` still contained `get_weather`.

The proposed fix is included inline. The compatibility contract should remain:

- text-only: unchanged text output
- tool-only: unchanged JSON tool-call list
- mixed text + tool calls: JSON containing both `content` and `tool_calls`
- `final_result`: unchanged structured result

Please cover the mixed case through both the v1-events and `gen_ai.output.messages` extraction paths, asserting the parsed `output.value` payload and `output.mime_type == "application/json"`.

Local validation of the current branch passed Ruff, mypy, and the targeted 27-test suite; the missing mixed-response assertion is the gap.

python/instrumentation/openinference-instrumentation-pydantic-ai/src/openinference/instrumentation/pydantic_ai/semantic_conventions.py
```diff
@@ -873,7 +940,19 @@ def _extract_from_gen_ai_messages(gen_ai_attrs: Mapping[str, Any]) -> Iterator[T
...
+                                        normalized = _normalize_tool_call(
+                                            part.get(GenAIToolCallFields.ID),
+                                            part.get(GenAIFunctionFields.NAME),
+                                            part.get(GenAIFunctionFields.ARGUMENTS),
+                                        )
+                                        if normalized:
+                                            output_tool_calls.append(normalized)
             except json.JSONDecodeError:
                 pass
     if output_value is not None:
```
caroger: **[P1] Preserve tool calls when the response also contains text**

This `if` makes text win over `output_tool_calls`, so a response containing both still hides the tool call from `output.value` consumers. I reproduced `output.value = "I will check."` with no MIME type while `llm.output_messages` contained `get_weather`.

A minimal fix is to share the following selection logic with the v1-events path (and exclude `final_result` calls from `regular_tool_calls` so structured output remains unchanged):

```python
regular_tool_calls = [
    call
    for call in output_tool_calls
    if call.get(GenAIFunctionFields.NAME) != PydanticFinalResult.FINAL_RESULT
]
if regular_tool_calls:
    payload: Any = regular_tool_calls
    if output_value is not None:
        payload = {"content": output_value, "tool_calls": regular_tool_calls}
    yield SpanAttributes.OUTPUT_VALUE, safe_json_dumps(payload)
    yield SpanAttributes.OUTPUT_MIME_TYPE, OpenInferenceMimeTypeValues.JSON.value
elif output_value is not None:
    yield SpanAttributes.OUTPUT_VALUE, output_value
```

Please replace `test_llm_text_output_still_wins_over_tool_calls` with a regression asserting the mixed JSON envelope and add the equivalent v1-events case. That makes the fix fail on this head and verifies both extraction paths after the change.

caroger (approved):

PR #3530, opened 2026-08-08
feat(cohere): instrument ClientV2.rerank

### LINKED ISSUE #3515: feat(cohere): instrument ClientV2.rerank
## Summary

`openinference-instrumentation-cohere` (added in #3349) instruments the v2 chat endpoints only — `ClientV2.chat` and `ClientV2.chat_stream`, plus their async counterparts. `ClientV2.rerank` / `AsyncClientV2.rerank` are not wrapped, so rerank calls produce no spans.

## What's needed

Wrap `V2Client.rerank` / `AsyncV2Client.rerank` (in `cohere.v2.client`, the base classes `ClientV2` / `AsyncClientV2` inherit from) and emit a **RERANKER** span.

Relevant attributes from `openinference.semconv.trace`:

- `OpenInferenceSpanKindValues.RERANKER` for `openinference.span.kind`
- `RerankerAttributes.RERANKER_QUERY`, `RERANKER_MODEL_NAME`, `RERANKER_TOP_K`
- `RerankerAttributes.RERANKER_INPUT_DOCUMENTS` and `RERANKER_OUTPUT_DOCUMENTS`, flattened with the document attributes

Notes for whoever picks this up:

- Signature is `(model, query, documents, top_n, max_tokens_per_doc, priority, request_options)`; `top_n` maps to `RERANKER_TOP_K`.
- `documents` accepts strings or document objects — both shapes need handling.
- The response returns results carrying an index into the input documents plus a relevance score; output documents should be reconstructed from those indices.
- Follow the existing conventions in the package: name the span after the wrapped resource class (`ClientV2.rerank` / `AsyncClientV2.rerank`), reuse `_parse_args` so `request_options` stays excluded, and only read re-iterable sequences so caller generators are not drained.

## Context

Follow-up from the Cohere instrumentor PR: #3349

lxingy3:
Resolves #3515

## What changed

- instrument `ClientV2.rerank` and `AsyncClientV2.rerank` with RERANKER spans
- capture the query, model, top-k value, and flattened input documents
- reconstruct ranked output documents from Cohere result indices and attach relevance scores
- preserve one-shot document iterators and keep transport-only `request_options` out of span data
- retain core reranker and context attributes when large ranked results exceed the span attribute limit
- document the new coverage and add sync, async, error, suppression, structured-document, and edge-case tests

## Why

Cohere v2 rerank calls currently produce no OpenInference spans. This adds the missing coverage while following the package's existing wrapper and attribute-extraction patterns.

## Validation

- `uvx --from tox-uv==1.11.2 tox run-parallel -e py310-ci-cohere,py310-ci-cohere-latest,py314-ci-cohere,py314-ci-cohere-latest`
- focused Cohere 5.13.0 rerank compatibility smoke test


satyadevai:
@caroger I have reviewed this PR and looks good to me.

satyadevai (approved):

python/instrumentation/openinference-instrumentation-cohere/src/openinference/instrumentation/cohere/_request_attributes_extractor.py
```diff
@@ -151,6 +155,38 @@ def get_attributes_from_embed_request(
...
 
+    def get_attributes_from_rerank_request(
+        self,
+        request_parameters: Mapping[str, Any],
+    ) -> Iterator[Tuple[str, AttributeValue]]:
+        yield SpanAttributes.OPENINFERENCE_SPAN_KIND, OpenInferenceSpanKindValues.RERANKER.value
+        query = request_parameters.get("query")
+        model_name = request_parameters.get("model")
+        top_k = request_parameters.get("top_n")
+        yield from get_reranker_attributes(
```
caroger, line 166: **[P1] Apply `TraceConfig` redaction to reranker content**

With `TraceConfig(hide_inputs=True, hide_outputs=True)`, the new `reranker.query`, `reranker.input_documents.*`, and `reranker.output_documents.*` attributes remain verbatim even though `input.value` and `output.value` are redacted. I reproduced this with a query and document containing sentinel secrets; all three specialized attributes still contained them. This silently defeats the privacy setting for rerank calls. Please extend the masking logic (or suppress these attributes under the corresponding flags) and add a regression test covering both input and output reranker content.

caroger (changes requested):
Requesting changes for the inline privacy issue. The functional and telemetry-shape checks otherwise passed: ruff, mypy, and 36 Cohere tests are green.

python/instrumentation/openinference-instrumentation-cohere/src/openinference/instrumentation/cohere/_request_attributes_extractor.py
satyadevai: @caroger masking input/output issue has now been addressed in recent commit by @lxingy3. I verified the changes, and everything is working as expected. I also ran the examples and verified the span attributes. 

One additional note: the masking changes was made in openinference-instrumentation (the core package). This should also resolve the masking issues for other instrumentors that emit spans for rerankers.

Once the new packages are published, they will have updated version numbers. However, this will not be an issue because we have configured the core package as openinference-instrumentation>=0.1.51, which will install the latest compatible version of openinference-instrumentation.

satyadevai (approved):

axiomofjoy (approved):

lxingy3:
Updated this branch with current main, which includes #3592 for the unrelated py310-ci-dspy-latest failure from the previous run. The reranker privacy changes remain unchanged. @satyadevai, could you approve the fork workflows when convenient? @caroger, could you please take another look once they complete? Thanks.

caroger (approved):

caroger:
thank you for your contribution @lxingy3 !

PR #3531, opened 2026-08-09
feat(anthropic): instrument beta tool runner with parent chain and child tool spans

### LINKED ISSUE #3392: Add support for Anthropic tool helpers
Anthropic has a beta SDK for tool helpers: https://platform.claude.com/docs/en/cli-sdks-libraries/sdks/python#tool-helpers. This will automatically handle tool call responses and call the tools for you.

The anthropic autoinstrumentor should be extended to create the relevant tool spans when this is used. At the moment using this just gives you a set of traces each with one LLM span for each LLM call made to handle the tool call.

Aryan-Pardeshi:
Closes #3392

The instrumentor wraps `completions.create`, `messages.create`, `messages.stream`, `messages.parse` and the `beta.messages` equivalents, but not the beta tool runner. A run that lets the SDK resolve tool calls automatically therefore produces a set of disconnected single-LLM-span traces, with no span for any tool actually invoked and nothing tying the turns together.

This wraps `beta.messages.tool_runner` (sync and async) so a run emits a parent span for the runner with child `TOOL` spans for each tool invocation, using the existing OpenInference tool attribute conventions. Both wrappers follow the structure already in `_wrappers.py` and are unwound in `_uninstrument`, so instrumenting and uninstrumenting stays symmetric.

Also adds `examples/tool_runner.py`, mirroring the existing `end_to_end_tool_use.py`, and a CHANGELOG entry.

Written and verified against **anthropic 0.84.0**. Worth flagging that this is a beta SDK surface: the entry point was confirmed by inspecting `anthropic.resources.beta.messages.Messages` on that version rather than taken from the docs, but it can move.

Full package suite passes: 53 passed. `ruff check` and `ruff format --check` are clean across the package.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/CHANGELOG.md
+++ b/python/instrumentation/openinference-instrumentation-anthropic/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Features
+
+* **anthropic:** instrument the beta tool runner with a parent chain span and tool child spans ([#3392](https://github.com/Arize-ai/openinference/issues/3392))
+
 ## [1.1.2](https://github.com/Arize-ai/openinference/compare/python-openinference-instrumentation-anthropic-v1.1.1...python-openinference-instrumentation-anthropic-v1.1.2) (2026-08-07)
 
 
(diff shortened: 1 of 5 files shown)


Aryan-Pardeshi:

PR #3542, opened 2026-08-10
fix: Fix: google-adk tool/LLM spans locked at OK on ADK 2.x failures

### LINKED ISSUE #3415: google-adk: tool/LLM spans stay OK (with exception events) on failing calls under ADK 2.x

satyadevai:
Closes #3415

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
@@ -257,7 +257,6 @@ class _TraceCallLlm(_WithTracer):
         if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             return ans
         span = get_current_span()
-        span.set_status(StatusCode.OK)  # Pre-emptively set status to OK
         span.set_attribute(
             SpanAttributes.OPENINFERENCE_SPAN_KIND,
             OpenInferenceSpanKindValues.LLM.value,
@@ -336,6 +335,10 @@ class _TraceCallLlm(_WithTracer):
         if llm_response:
             for k, v in _get_attributes_from_llm_response(llm_response):
                 span.set_attribute(k, v)
+            if not llm_response.partial:
+                # This is the final chunk for this LLM turn, so no further
+                # trace_call_llm calls will land on this span.
+                span.set_status(StatusCode.OK)
         return ans
 
 
@@ -352,12 +355,19 @@ class _TraceToolCall(_WithTracer):
         if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             return ans
         span = get_current_span()
-        span.set_status(StatusCode.OK)  # Pre-emptively set status to OK
+        arguments = bind_args_kwargs(wrapped, *args, **kwargs)
+        error = arguments.get("error")
+        error_type = arguments.get("error_type")
+        if isinstance(error, BaseException):
+            span.set_status(StatusCode.ERROR, description=f"{type(error).__name__}: {error}")
+        elif error_type is not None:
+            span.set_status(StatusCode.ERROR, description=str(error_type))
+        else:
+            span.set_status(StatusCode.OK)
         span.set_attribute(
             SpanAttributes.OPENINFERENCE_SPAN_KIND,
             OpenInferenceSpanKindValues.TOOL.value,
         )
-        arguments = bind_args_kwargs(wrapped, *args, **kwargs)
         if base_tool := next(
             (arg for arg in arguments.values() if isinstance(arg, BaseTool)), None
         ):
--- a/python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/tests/test_instrumentor.py
@@ -11,6 +11,9 @@ import pytest
 from google.adk import Agent, __version__
 from google.adk.code_executors.built_in_code_executor import BuiltInCodeExecutor
 from google.adk.events import Event, EventActions
+from google.adk.models.base_llm import BaseLlm
+from google.adk.models.llm_request import LlmRequest
+from google.adk.models.llm_response import LlmResponse
 from google.adk.planners import BuiltInPlanner
 from google.adk.runners import InMemoryRunner
 from google.adk.tools.agent_tool import AgentTool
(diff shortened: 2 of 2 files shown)


python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
```diff
@@ -336,6 +335,10 @@ def __call__(
         if llm_response:
             for k, v in _get_attributes_from_llm_response(llm_response):
                 span.set_attribute(k, v)
+            if not llm_response.partial:
+                # This is the final chunk for this LLM turn, so no further
+                # trace_call_llm calls will land on this span.
+                span.set_status(StatusCode.OK)
```
caroger, line 341: **[P1] Do not finalize `OK` before ADK callbacks complete**

Reproduced on this head (`639b8d9`) with `google-adk==2.6.3`: a final response followed by `after_model_callback` raising `RuntimeError("after-model failed")` exports the `call_llm` span as `OK` with two exception events.

Suggested change:

```diff
-            if not llm_response.partial:
-                # This is the final chunk for this LLM turn, so no further
-                # trace_call_llm calls will land on this span.
-                span.set_status(StatusCode.OK)
```

Success can remain `UNSET`; this lets the owning ADK span record a later callback failure as `ERROR`. A regression test can use the existing fixtures:

```python
class _FinalResponseLlm(BaseLlm):
    async def generate_content_async(
        self, llm_request: LlmRequest, stream: bool = False
    ) -> AsyncGenerator[LlmResponse, None]:
        yield LlmResponse(
            content=types.Content(role="model", parts=[types.Part(text="done")])
        )


def _fail_after_model(_ctx: Any, _response: LlmResponse) -> None:
    raise RuntimeError("after-model failed")


async def test_call_llm_span_errors_when_after_model_callback_raises(
    instrument: Any, in_memory_span_exporter: InMemorySpanExporter
) -> None:
    agent = Agent(
        name="agent",
        model=_FinalResponseLlm(model="fake-model"),
        after_model_callback=_fail_after_model,
    )
    runner = InMemoryRunner(agent=agent, app_name="app")
    await runner.session_service.create_session(
        app_name="app", user_id="user", session_id="session"
    )

    with pytest.raises(RuntimeError, match="after-model failed"):
        async for _ in runner.run_async(
            user_id="user",
            session_id="session",
            new_message=types.Content(
                role="user", parts=[types.Part(text="hello")]
            ),
        ):
            pass

    [span] = [
        span
        for span in in_memory_span_exporter.get_finished_spans()
        if span.name == "call_llm"
    ]
    assert span.status.status_code == trace_api.StatusCode.ERROR
    assert any(event.name == "exception" for event in span.events)
```

caroger (changes requested):
Requesting changes for one remaining correctness gap in the LLM status fix.

The tool-failure and partial-stream failure paths are fixed, and the Google ADK 2.6.3 lane passes locally (Ruff, mypy, and 34 tests). However, `_TraceCallLlm` still assigns `OK` on a final response before ADK invokes `after_model_callback`. A callback exception therefore produces an `OK` `call_llm` span containing exception events—the same class of status-locking bug this PR is intended to eliminate.

The exact-head reproduction and suggested minimal code/test changes are in the inline comment.

python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
satyadevai: @caroger As described in the issue, we set the call_llm span to OK once the LLM call has completed with a full, successful response. If an after_model_callback subsequently fails, that exception belongs to the agent flow rather than to the LLM call itself, so marking the completed LLM span as OK is appropriate.

At the moment, we do not have a way to prevent ADK from attaching those later callback exception events to the call_llm span. Because of that, an exception event may appear on a span whose status is still OK, even though the exception was not caused by the LLM call.

For streaming responses, if the LLM response is only partial and the call does not complete successfully, we set the call_llm span to ERROR because the LLM call itself did not complete.

The parent/agent spans are marked as ERROR for the callback failure, which is the appropriate level to represent this failure, while keeping the successfully completed call_llm span as OK.

image

satyadevai: If we remove setting OK when llm_response.partial is false, we effectively revert this part of the code to the state before this ticket. The purpose of this change is to explicitly mark the LLM span as successful once the LLM has returned a complete, successful response.

A later after_model_callback failure is part of the agent flow, not a failure of the already-completed LLM call. So leaving the LLM span as UNSET just to allow a later callback exception to mark it as ERROR would bring back the original issue and incorrectly couple the LLM call status to downstream agent processing.

caroger (approved):

PR #3543, opened 2026-08-10
fix: make the instruments extras match instrumentation_dependencies()

Anai-Guo:
### Problem

Each instrumentation package declares its target library twice:

* `[project.optional-dependencies] instruments` in `pyproject.toml` — what `pip install <pkg>[instruments]` resolves;
* `_instruments` / `instrumentation_dependencies()` — what `BaseInstrumentor.instrument()` checks at runtime via `get_dependency_conflicts`.

When the extra admits a version the runtime check rejects, `instrument()` **logs an error and returns** — no exception, no spans, nothing that surfaces in the application. Three packages are in that state today:

| package | `instruments` extra | `instrumentation_dependencies()` | versions the extra admits but the instrumentor refuses |
|---|---|---|---|
| `groq` | `groq >= 0.6.0` | `groq >= 0.9.0` | `0.6.0 … 0.8.x` |
| `autogen-agentchat` | `autogen-agentchat >= 0.5.0` | `autogen-agentchat >= 0.5.1` | `0.5.0` |
| `guardrails` | `guardrails-ai` (unbounded) | `guardrails-ai>=0.4.5,<0.5.1` | everything `>= 0.5.1`, i.e. the current release |

### Reproduction (groq)

```
$ pip install "groq==0.6.0"
$ pip install "openinference-instrumentation-groq[instruments]"   # leaves groq at 0.6.0
$ pip show groq | head -2
Name: groq
Version: 0.6.0
```

```python
from groq.resources.chat.completions import Completions
from opentelemetry.instrumentation.dependencies import get_dependency_conflicts
from openinference.instrumentation.groq import GroqInstrumentor

print(get_dependency_conflicts(("groq >= 0.9.0",)))
before = Completions.create
GroqInstrumentor().instrument()
print("patched:", before is not Completions.create)
```

Before this change, with the version the extra is happy to leave in place:

```
LOG ERROR opentelemetry.instrumentation.instrumentor: DependencyConflict: requested: "groq >= 0.9.0" but found: "groq 0.6.0"
DependencyConflict: requested: "groq >= 0.9.0" but found: "groq 0.6.0"
patched: False
```

After (`pip install "groq>=0.9.0"`, which is what the corrected extra resolves to):

```
None
patched: True
```

`guardrails` fails the same way for a different reason — besides the `_instruments` range, `_instrument()` itself does

```python
if (version.major, version.minor, version.micro) >= (0, 5, 1):
    logger.info("Guardrails version >= 0.5.1 detected, skipping instrumentation")
    return
```

so the unbounded extra installs `guardrails-ai` 0.10.2 and instrumentation is guaranteed to be a no-op.

### Change

Align each extra with the range the instrumentor actually enforces. Three one-line edits, no source changes.

### Notes

* I checked all 36 Python instrumentation packages. The rest either match or have an extra that is *stricter* than `_instruments` (e.g. `openai`: extra `>= 2.8.0`, `_instruments` `>= 1.69.0`) — that direction is harmless, so I left it alone.
* The `guardrails` hunk is the only one with a trade-off: adding the upper bound means `pip install openinference-instrumentation-guardrails[instruments]` will now downgrade a newer `guardrails-ai` instead of silently installing something that is never instrumented. I think stating the supported range is the more honest option and it matches every other package, but happy to drop that hunk if you'd rather keep the extra open.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-autogen-agentchat/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-autogen-agentchat/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "autogen-agentchat >= 0.5.0",
+  "autogen-agentchat >= 0.5.1",
 ]
 
 [project.entry-points.opentelemetry_instrumentor]
--- a/python/instrumentation/openinference-instrumentation-groq/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-groq/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-    "groq >= 0.6.0",
+    "groq >= 0.9.0",
 ]
 
 [project.entry-points.opentelemetry_instrumentor]
--- a/python/instrumentation/openinference-instrumentation-guardrails/pyproject.toml
+++ b/python/instrumentation/openinference-instrumentation-guardrails/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "guardrails-ai",
+  "guardrails-ai>=0.4.5,<0.5.1",
 ]
 
 [project.entry-points.opentelemetry_instrumentor]


mikeldking (approved):
Verified against `main` — all three mismatches are real and still present:

| package | extra on main | `_instruments` | silently-broken range |
|---|---|---|---|
| `groq` | `>= 0.6.0` | `groq >= 0.9.0` | 0.6.0–0.8.x |
| `autogen-agentchat` | `>= 0.5.0` | `>= 0.5.1` | 0.5.0 |
| `guardrails` | unbounded | `>=0.4.5,<0.5.1` | >= 0.5.1 (every current release) |

The failure mode is right: `BaseInstrumentor.instrument()` runs `get_dependency_conflicts` and on a conflict only logs and returns, so the user gets no spans and no exception. Aligning the extra with the enforced range is the fix.

I also re-ran the audit across all 36 Python instrumentation packages independently and can confirm the claim in the description: every remaining divergence (`agno`, `bedrock`, `dspy`, `google-genai`, `instructor`, `langchain`, `llama-index`, `mistralai`, `openai`, `pydantic-ai`, `smolagents`) has the extra *stricter* than `_instruments`, which is the harmless direction. Nothing left in the looser direction after this.

On the `guardrails` trade-off you flagged — keeping the hunk. `_instrument()` hard-returns above 0.5.1, so the unbounded extra can only ever install something that is guaranteed not to be instrumented; and `python/tox.ini:163` already pins `uv pip install -U 'guardrails-ai<0.5.1'` for the `guardrails-latest` env, so `<0.5.1` is already the repo's own statement of the supported ceiling. Declaring it in the extra is consistent. Worth noting separately that the guardrails instrumentor being dead for every current release is its own issue — either update it or deprecate it — but that's out of scope here.

Also confirmed no CI path or script installs the `[instruments]` extras (tests go through `test-requirements.txt`), so these edits can't move test resolution.

Branch updated with `main`; 24 checks green, diff still exactly the three one-liners.

Issue #1331, opened 2025-03-03
[feature request] capture reasoning and cache tokens for cost analysis


mikeldking:
caching tokens and reasoning tokens can affect billing and cost analysis. We should capture these across

- OpenAI
- Claude
- Groq
- Bedrock
- Gemini

mikeldking:
heads up @caroger these tokens don't come back from openai streaming unless specifically flagged.

caroger:
OpenAI
Anthropic 
Langchain
Llamaindex
Caching and Audio

guillaumefrd:
Any update on this for Gemini cache tokens?

Shailendra005:
While auditing per-instrumentor token-detail handling for this issue, I found a correctness bug in the **LiteLLM** instrumentor's `_set_token_counts_from_usage`:

In `python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py` (~L841-845), `completion_tokens_details.text_tokens` — a raw token **count** — is written to `SpanAttributes.LLM_COST_COMPLETION_DETAILS_OUTPUT`, which the semantic conventions define as *"Total cost of output tokens in USD"* (`openinference-semantic-conventions/.../trace/__init__.py` L160-165). So a token count is emitted as a dollar amount and corrupts cost rollups.

Related, at ~L821-825, `prompt_tokens_details.text_tokens` is written to `LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_INPUT` ("cached input tokens"), but the cached count is already emitted correctly from `cached_tokens` (→ `CACHE_READ`, L810-813). `text_tokens` is not a cached count.

There is no semconv key for "text tokens" in either the completion-detail or prompt-detail token-count families (only `audio`/`reasoning` and `audio`/`cache_*` respectively), so the correct fix is to drop both mappings rather than invent a key. This matches the OpenAI instrumentor precedent (`openai/_response_attributes_extractor.py` L240-259), which maps only `cached_tokens→CACHE_READ`, `audio_tokens→AUDIO`, and `reasoning_tokens→REASONING`.

I've opened #3549 with a small fix + regression test. Happy to adjust per your preference.

PR #3549, opened 2026-08-12
fix(litellm): do not report token counts as USD cost

Shailendra005:
## What changed

Removed two mismappings in `_set_token_counts_from_usage` (LiteLLM instrumentor):

- `completion_tokens_details.text_tokens` was written to `LLM_COST_COMPLETION_DETAILS_OUTPUT` — a semconv attribute defined as *cost in USD*. A raw token count was being emitted as a dollar amount.
- `prompt_tokens_details.text_tokens` was written to `LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_INPUT` (cached input tokens); the cached count is already emitted from `cached_tokens`, and `text_tokens` is not a cached count.

## Why

`text_tokens` is a count, and there is no semconv key for output/prompt "text tokens" (completion-detail counts: `audio`, `reasoning`; prompt-detail counts: `audio`, `cache_*`). So the correct fix is to drop the mappings rather than invent a key. This aligns LiteLLM with the OpenAI instrumentor, which maps only `cached_tokens→CACHE_READ`, `audio_tokens→AUDIO`, `reasoning_tokens→REASONING`. Real cost handling (`response_cost`/`cost` → `LLM_COST_TOTAL`) is untouched.

## Span impact

Removes two attributes that were being populated incorrectly (`llm.cost.completion_details.output` and `llm.token_count.prompt_details.cache_input` when sourced from `text_tokens`). No new attributes added, so no trace screenshot applies.

## Tests

Adds `test_text_tokens_are_not_reported_as_cost_or_cache`, which fails on `main` and passes with this change; existing token-count tests still pass. `ruff` + `mypy` clean.

Relates to #1331.


Shailendra005:

python/instrumentation/openinference-instrumentation-litellm/tests/test_token_counts.py
```diff
@@ -121,6 +127,51 @@ def test_anthropic(
...
 
+def test_text_tokens_are_not_reported_as_cost_or_cache() -> None:
+    """Regression test for a token-count vs. USD-cost mismap.
+
+    ``completion_tokens_details.text_tokens`` and ``prompt_tokens_details.text_tokens``
+    are raw token *counts*. They must never be written to a cost attribute
+    (``llm.cost.*``, defined by the semantic conventions as USD amounts) nor to the
+    cache-input token-count key (which represents cached tokens, not text tokens).
+    """
+    exporter = InMemorySpanExporter()
```
mikeldking, line 138: nit: `tests/conftest.py` already provides `tracer_provider` and `in_memory_span_exporter` fixtures, and the autouse `instrument` fixture in this module wires them up. You could take `tracer_provider: TracerProvider, in_memory_span_exporter: InMemorySpanExporter` as parameters and call `tracer_provider.get_tracer(__name__)` instead of building a second SDK provider here. That would also let the `SDKTracerProvider` / `SimpleSpanProcessor` imports go away. Fine to leave as is.

mikeldking (approved):
Thanks for this. Confirmed the mismap is still present on `main`: `prompt_tokens_details.text_tokens` was landing on `llm.token_count.prompt_details.cache_input` and `completion_tokens_details.text_tokens` on `llm.cost.completion_details.output`, which is a USD attribute. Neither has a semconv home as a token count, and the OpenAI instrumentor does not emit `text_tokens` either, so dropping both mappings is the right call.

I rebased the branch onto current `main` (no conflicts) and pushed it back to your fork so CI runs against the new base. Locally `tox run -e ruff-mypy-test-litellm` is green (118 passed) and the new regression test fails without the source change.

One small nit inline on the test; not blocking.

PR #3554, opened 2026-08-13
fix(instrumentation): make suppress_tracing usable with `async with`

DivyaNarahari97:
`suppress_tracing` defines `__aenter__`/`__aexit__`, but they were plain methods rather than `async def`. Python's `async with` awaits whatever `__aenter__` returns, so every use raised:

    TypeError: 'async with' received an object from __aenter__ that does
    not implement __await__: suppress_tracing

Since those dunders exist only to support `async with`, the async form was entirely non-functional. The sibling context managers in this package (`using_session`, `using_user`, ...) already declare them `async def`; `suppress_tracing` was the only non-async `__aenter__`/`__aexit__` in the Python tree.

The synchronous `with` path is unchanged.

### DIFF
--- a/python/openinference-instrumentation/src/openinference/instrumentation/config.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/config.py
@@ -45,13 +45,17 @@ class suppress_tracing:
         with suppress_tracing():
             # No tracing will occur within this block
             ...
+
+        async with suppress_tracing():
+            # No tracing will occur within this block
+            ...
     """
 
     def __enter__(self) -> "suppress_tracing":
         self._token = attach(set_value(_SUPPRESS_INSTRUMENTATION_KEY, True))
         return self
 
-    def __aenter__(self) -> "suppress_tracing":
+    async def __aenter__(self) -> "suppress_tracing":
         self._token = attach(set_value(_SUPPRESS_INSTRUMENTATION_KEY, True))
         return self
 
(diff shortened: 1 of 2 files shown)


DivyaNarahari97:

DivyaNarahari97:
recheck

mikeldking:
recheck

mikeldking:
Verified the fix end to end against a local Phoenix, for both `with` and `async with`, using the new `examples/suppress_tracing_sync_async.py`. The example starts spans inside and outside each `suppress_tracing()` block, and also raises inside an `async with` block to check the context is restored afterwards.

**After (this branch, editable install of `python/openinference-instrumentation/src`)** — project `instrumentation-suppress-tracing-sync-async-after-20260914-2244`, exit 0, no export errors, 10 spans:

```
sync.traced [CHAIN]              -> sync_tool [TOOL]
sync.after [CHAIN]               -> sync_tool [TOOL]
async.traced [CHAIN]             -> async_tool [TOOL]
async.after [CHAIN]              -> async_tool [TOOL]
async.after_exception [CHAIN]    -> async_tool [TOOL]
```

Spans whose name contains `suppressed`: 0. Both the sync and async blocks dropped their spans, and tracing resumed after each block, including after the exception.

**Before (`origin/main`, same example)** — project `instrumentation-suppress-tracing-sync-async-before-20260914-2244`, exit 1 after the sync half (6 spans, sync behaviour identical to after):

```
  async with suppress_tracing():
TypeError: 'async with' received an object from __aenter__ that does not implement __await__: suppress_tracing
```

I also merged `main` into the branch, added an async exception-path test, and documented `suppress_tracing` in the package README. Ruff, mypy, and the package test suite pass locally.

mikeldking (approved):
Verified sync and async suppression end to end against local Phoenix (see evidence comment above). CI green on the previous head; auto-merge enabled for this one.

PR #3555, opened 2026-08-13
fix(instrumentation): raise a real error for unparsable bool env vars

DivyaNarahari97:
`TraceConfig._cast_value` used a bare `raise` to signal "this string is not a bool". A bare `raise` outside an `except` block does not create an error; it re-raises whatever exception is currently being handled.

Normally that degrades to `RuntimeError: No active exception to reraise`, which `_parse_value` catches, so the fallback to the default happens to work. But when a `TraceConfig` is constructed while an exception is being handled, the bare `raise` re-raises *that* exception instead. If it is a `BaseException` such as `KeyboardInterrupt`, it escapes the `except Exception` handler in `_parse_value` entirely, so an unparsable env var (e.g. `OPENINFERENCE_HIDE_INPUTS=yes`) propagates an unrelated exception into user code instead of falling back to the default.

Instrumentation must never raise into user code, so raise an explicit `ValueError` that `_parse_value` can always catch.

### DIFF
--- a/python/openinference-instrumentation/src/openinference/instrumentation/config.py
+++ b/python/openinference-instrumentation/src/openinference/instrumentation/config.py
@@ -498,7 +498,7 @@ class TraceConfig:
                 return True
             if isinstance(value, str) and value.lower() == "false":
                 return False
-            raise
+            raise ValueError(f"Cannot cast {value!r} to bool.")
         else:
             return cast_to(value)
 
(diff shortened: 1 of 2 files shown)


DivyaNarahari97:

DivyaNarahari97:
recheck

mikeldking (approved):

PR #3556, opened 2026-08-13
fix(anthropic): guard usage.input_tokens against None in _get_token_counts

### LINKED ISSUE #3499: [bug] Anthropic instrumentor: unguarded usage.input_tokens can blank out an entire streaming LLM span

Aftabbs:
## Summary

`_get_token_counts` (shared by the streaming and non-streaming paths) sums `usage.input_tokens` without the `or 0` guard applied to every other field. When `input_tokens` is `None` — possible in streaming snapshots before the final `message_delta` arrives — `None + 0` raises `TypeError`. Because `_MessageExtractor.get_attributes()` has no `@_stop_on_exception` containment, this exception escapes into `_finish_tracing`, which catches it and sets `attributes = None`. The streaming LLM span is then exported with **no** input messages, output messages, model name, or token counts: a completely blank span.

Fixes #3499

## Root Cause

```python
# _utils.py  (before)
prompt_tokens = (
    usage.input_tokens                          # ← no `or 0` guard
    + (usage.cache_creation_input_tokens or 0)
    + (usage.cache_read_input_tokens or 0)
)
```

All sibling fields have `or 0`; `input_tokens` was the only exception. The bug was latent before `_get_token_counts` was extracted as a shared helper (PR #3490); previously each path handled `None` differently.

## Changes

| File | What changed |
|------|-------------|
| `src/.../anthropic/_utils.py` | Add `or 0` to `usage.input_tokens`, matching every other nullable field |
| `tests/.../test_utils.py` | Three new unit tests: `input_tokens=None` (no TypeError), all-None (empty yield), normal values (correct arithmetic) |

## Testing

- [x] Existing tests pass
- [x] New test: `test_get_token_counts_with_null_input_tokens_does_not_raise` — verifies that passing `input_tokens=None` does not raise `TypeError` and still yields correct `completion`/`total` counts
- [x] New test: `test_get_token_counts_with_all_none_fields_does_not_raise` — all-None usage yields no attributes
- [x] New test: `test_get_token_counts_with_normal_usage` — integer fields produce correct `prompt`, `completion`, `total`, and `cache_read` attributes
- [x] Manually verified: the pre-fix code raises `TypeError` with `input_tokens=None`; the fixed code returns `[]` for `prompt` count and the correct `completion`/`total` values

## Impact

Users who receive streaming Anthropic responses where the usage snapshot briefly has `input_tokens=None` (e.g. during the `message_start` → `message_delta` transition) will now see fully-attributed LLM spans instead of blank spans missing all metadata.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_utils.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_utils.py
@@ -26,7 +26,7 @@ def _get_token_counts(usage: "Usage") -> Iterator[Tuple[str, AttributeValue]]:
     - input_tokens: Number of input tokens which were not read from or used to create a cache.
     """
     prompt_tokens = (
-        usage.input_tokens
+        (usage.input_tokens or 0)
         + (usage.cache_creation_input_tokens or 0)
         + (usage.cache_read_input_tokens or 0)
     )
--- a/python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_utils.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_utils.py
@@ -1,11 +1,12 @@
 import logging
+from types import SimpleNamespace
 from typing import Iterator, Tuple
 
 import pytest
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.util.types import AttributeValue
 
-from openinference.instrumentation.anthropic._utils import _finish_tracing
+from openinference.instrumentation.anthropic._utils import _finish_tracing, _get_token_counts
 from openinference.instrumentation.anthropic._with_span import _WithSpan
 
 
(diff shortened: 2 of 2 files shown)


Aftabbs:

feiiiiii5:
Ran this locally on current upstream and the guard does what it says; the arithmetic regression is real and the tests bite. One substantive question about reachability, and one about the resulting numbers.

What I measured — python 3.11.15, `uv venv` + this package's `test-requirements.txt` (`anthropic==0.84.0`), base `d21fe736` + head `03b98a64`, no network/keys:

```
head:  tests/openinference/anthropic/test_utils.py        -> 4 passed
head:  tests/ (whole package)                             -> 54 passed
base:  same files with only `(usage.input_tokens or 0)` reverted back to
       `usage.input_tokens`                               -> 2 failed, 52 passed
   E  TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
      src/.../anthropic/_utils.py:29: TypeError
   FAILED ...::test_get_token_counts_with_null_input_tokens_does_not_raise
   FAILED ...::test_get_token_counts_with_all_none_fields_does_not_raise
```

so the two new tests fail without the change and the other 52 stay green with it. I read `TypeError` reaching `_finish_tracing` from the extractor the same way the description does, and issue #3499 is still open with this as its fix.

The gap I could not close from the code: in the version this package pins, `input_tokens` cannot actually arrive as `None` through the SDK types.

```
anthropic.types.Usage.input_tokens      -> int, required=True
anthropic.types.beta.BetaUsage.input_tokens -> int, required=True
Usage(input_tokens=None, output_tokens=10) -> ValidationError (field required / not optional)
Usage(output_tokens=10)                   -> ValidationError: input_tokens Field required
```

and across this package's `tests/` (including the recorded cassettes) there are 30 `input_tokens` occurrences, none of them null. The new tests build the usage object with `SimpleNamespace`, so they pin the arithmetic but not the path that produces `None`. Could you point at one concrete reachable case — a `anthropic` version outside the pin where the field is `Optional`, or the specific stream event / beta call whose `usage` is not a typed `Usage` (a recorded cassette would settle it)? That decides whether this is a live-blank-span fix or defensive hardening, which changes how a maintainer triages it. If it turns out to be defensive only, the change still looks fine to merge, but it should probably say so.

Second point, about what the span now reports. With the guard in place, `input_tokens=None, output_tokens=10` yields:

```
llm.token_count.completion = 10
llm.token_count.total      = 10      # prompt omitted because prompt_tokens == 0
```

`total` is derived as `prompt_tokens + output_tokens`, so an unknown prompt collapses to `0` and the total then *equals* completion — which reads as a correct small number rather than a partial one, in exactly the billing/cost queries this fix is trying to rescue. `test_get_token_counts_with_null_input_tokens_does_not_raise` asserts that outcome, so it's intentional today; worth deciding whether unknown-input should instead omit `total` (keeping "absent" meaning "unknown", as it already does for `prompt`). Happy either way, but it should be a deliberate choice rather than inherited from the `if prompt_tokens:` branch.

Two mechanics notes: `GitHub` reports the branch as mergeable at current `main`, but it is still based on `d21fe736` from 2026-08-13, so rebasing onto `main` before review would let CI run against the suite as it stands today. And for anyone picking this up: fork PRs here sit behind the "approved workflows" gate, so the numbers above are from local runs — CI status is pending approval, not evidence either way.

Disclosure: automated contributor working for the account owner; the commands and outputs above are from my own run in a throwaway venv, and this is offered as measurements, not as a review approval.

Issue #1106, opened 2024-11-07
[enhancement][js] Ensure auto-instrumentation is functional in ESM-only projects


cephalization:
Packages are now advertised as ESM, but does auto-instrumentation work under ESM-only projects?

> While we build and ship ESM versions of our packages, we do not advertise their availability to module resolvers via package.json
> 
> We need to fill out the `exports` field.
> 
> See:
> - https://docs.npmjs.com/cli/v10/configuring-npm/package-json#exports
> - https://nodejs.org/api/packages.html#conditional-exports
>   - Conditional exports allow us to point to esm builds when `import`-ing and cjs builds when `require`-ing

cephalization:
Additionally, we should evaluate if esm support actually works correctly.

See https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md

for what the implications are when using these packages in esm-only applications

cephalization:
Also see #1144

cephalization:
As per https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md

To run native auto module patching instrumentation via import in the middle, you must exectute your program with node and the `--experimental-loader=@opentelemetry/instrumentation/hook.mjs` flag.

I have a few problems with this:
- It requires users to augment their execution command, which feels heavy handed
- It does not work with openai, our most popular instrumentation, due to import ordering issues
- It precludes support for runtimes other than node, that would otherwise be able to support module patching via import

All that being said, I think that for the time being we continue to provide a `manuallyPatch` method on our instrumentation that allows users to import a module, and patch it directly, having that module be patched for the rest of program execution (as the modules/side-effects generated from calling import are cached across import calls).

So, as of right now, users of ESM in their projects can continue to manually instrument so long as they run on a node-like environment.

This does not help browser based users however, as we still attempt to auto-patch with node dependencies, that break browser execution. If `InstrumentationNodeModuleDefinition` is removed, then the instrumentation works just fine in the browser with manual patching.

Next steps:
- Feature detect environment when instrumentation is initialized
  - If node-like environment detected, attempt auto patching but allow manual patching
  - If browser-like environment detected, do not attempt auto patching, do not import node module definition patcher, allow manual patching

cephalization:
Closing in favor of #1585

PR #3558, opened 2026-08-13
fix(js): scope double-patch guard per module so both builds of dual-package SDKs can be patched

### LINKED ISSUE #3557: [bug][js] Global _isOpenInferencePatched flag blocks manuallyInstrument for one of the two builds of dual-package SDKs
**Affected packages:** `@arizeai/openinference-instrumentation-openai` (4.1.8), `@arizeai/openinference-instrumentation-anthropic` (0.1.20) — likely all JS instrumentations sharing the pattern.

### Problem

Per #1106, `manuallyInstrument()` is the recommended path for ESM projects, since the require hook can't see ESM importers. But `patch()` guards on a module-scoped boolean in addition to the per-module marker:

```js
let _isOpenInferencePatched = false;
// ...
patch(module, moduleVersion) {
    if (module?.openInferencePatched || _isOpenInferencePatched) {
        return module;
    }
```

`openai` and `@anthropic-ai/sdk` are dual packages: the CJS (`require`) and ESM (`import`) builds are separate module objects with separate class prototypes, so patching one build does not instrument consumers of the other. The global flag makes patching both in one process impossible: whichever build is patched first flips it, and any later `patch()`/`manuallyInstrument()` call for the **other** build silently returns without patching.

Concrete scenario: an SDK wrapping your instrumentations calls `manuallyInstrument(await import("openai"))` so ESM apps get spans. Any dependency in the same process that later `require`s `openai` gets the untouched CJS build — the require-hook patch no-ops on the flag — and those calls silently produce no spans. The reverse order breaks the ESM side instead.

### Reproduction

```js
import { createRequire } from "node:module";
const require = createRequire(import.meta.url);
import { AnthropicInstrumentation } from "@arizeai/openinference-instrumentation-anthropic";

const instr = new AnthropicInstrumentation();
const esm = await import("@anthropic-ai/sdk");
const cjs = require("@anthropic-ai/sdk");
console.log(esm.Anthropic === cjs.Anthropic); // false — two separate builds

instr.manuallyInstrument(esm.default ?? esm);  // patches ESM build, sets the global flag
instr.manuallyInstrument(cjs);                 // silently no-ops: CJS build stays unpatched
```

### Proposed fix

I understand the flag was added as a double-patch guard (#491, #492) — the ask is not to drop the safety but to scope it to the object it protects. `patch()` already sets `module.openInferencePatched = true` on exactly the object it patched, which is the right granularity:

```js
if (module?.openInferencePatched) {
    return module;
}
```

Each build still gets patched at most once (no double-wrapping), but both builds of a dual package can be patched in the same process. `isPatched()` could keep its current global meaning unchanged.

One wrinkle: assigning the marker onto a native ESM namespace throws (`Cannot assign to read only property`, cf. #3336) — for namespaces the guard may need a `WeakSet` instead of a property write. Happy to open a PR if you'd take one.

*(The related loader-hook route is separately blocked by the `import-in-the-middle` 1.x/3.x split between these packages and OTel 0.2xx — but that looks covered by #3394.)*

PabloPardoGarcia:
Fixes #3557.

## Problem

`openai` and `@anthropic-ai/sdk` are dual packages: their CJS (`require`) and ESM (`import`) builds are separate module objects with separate class prototypes, so patching one build does not instrument consumers of the other. `patch()` in both instrumentations is guarded by a module-global boolean:

```ts
if (module?.openInferencePatched || _isOpenInferencePatched) {
  return module;
}
```

Whichever build is patched first flips `_isOpenInferencePatched`, and any later `patch()`/`manuallyInstrument()` call for the **other** build silently returns without patching. Since `manuallyInstrument()` is the recommended path for ESM consumers (#1106), an integrator who patches the ESM build cannot also have the require hook cover CJS consumers in the same process, and vice versa — with no error or warning either way.

## Fix

Replace the global boolean in the guard with a module-scoped `WeakSet` keyed on the class object the patch actually wraps (`module.OpenAI`; anthropic's `module.default || module`):

- **Object-scoped like the `openInferencePatched` marker**, so each build of a dual package is patched at most once, and both can be patched in one process.
- **No write to the module**, so the double-patch protection keeps working where the module is immutable (Deno, webpack) — the case the global flag was introduced for (per its own comment), and the same concern `canPatchInPlace()` addresses in the Claude Agent SDK instrumentation (#3344).
- Keyed on the class rather than the received module object, so callers that pass a fresh wrapper object per call are still guarded.

`_isOpenInferencePatched` is kept and still set/cleared in `patch()`/`unpatch()`, so `isPatched()` behavior is unchanged for existing consumers. `unpatch()` deletes the class from the `WeakSet`, so re-patching after an unpatch still works.

## Tests

New `test/dualBuild.test.ts` in each package:

- requires the CJS build and imports the ESM build, asserts they are distinct objects, patches both via `manuallyInstrument()`, and asserts **both** are wrapped — the second call was a silent no-op before this change;
- asserts a repeated patch of the same build does not double-wrap. The CJS build re-guards via the property marker; the ESM namespace rejects the marker write (frozen namespace), so its re-guard is the `WeakSet` alone — this is the regression test for the immutable-module case the global flag used to cover.

Both packages' full suites pass (`19` and `67` tests), plus `lint`, `fmt:check`, and `type:check`.

## Scope

Deliberately limited to the two dual-package provider instrumentations, per the discussion in #3557. The same global-flag pattern exists in `langchain`, `langchain-v0`, `openai-agents`, `beeai`, and `claude-agent-sdk`; happy to follow up with the same treatment there if you'd like consistency.

### DIFF
--- a/js/.changeset/dual-build-patch-guard.md
+++ b/js/.changeset/dual-build-patch-guard.md
@@ -1,0 +1,6 @@
+---
+"@arizeai/openinference-instrumentation-openai": patch
+"@arizeai/openinference-instrumentation-anthropic": patch
+---
+
+Scope the double-patch guard to the module object so both the CJS and ESM builds of a dual-package SDK can be patched in the same process. Previously the module-global `_isOpenInferencePatched` flag made whichever build was patched first silently block `patch()`/`manuallyInstrument()` for the other build (#3557). The guard is now a `WeakSet` keyed on the patched class, which needs no write to the module and therefore also keeps protecting immutable modules (Deno, webpack) — the case the global flag existed for. `isPatched()` behavior is unchanged.
--- a/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
+++ b/js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts
@@ -74,12 +74,14 @@ let _isOpenInferencePatched = false;
  * The Anthropic classes that have already been patched, tracked by identity.
  * The SDK ships separate CJS and ESM builds with separate class objects, so a
  * module-global boolean cannot guard them independently: whichever build was
- * patched first would block the other one forever (#3557). A WeakSet is
+ * patched first would block the other one forever (#3557). A Set is
  * scoped to the object, and needs no write to the module, so it also keeps
  * the double-patch guard working when the module is immutable (e.g. Deno,
- * webpack) and the `openInferencePatched` property cannot be set.
+ * webpack) and the `openInferencePatched` property cannot be set. Entries
+ * are removed when their wrappers are removed, so the Set does not retain
+ * unpatched SDK builds.
  */
-const _patchedModules = new WeakSet<object>();
+const _patchedModules = new Set<object>();
 
 /**
  * function to check if instrumentation is enabled / disabled
@@ -113,6 +115,10 @@ export class AnthropicInstrumentation extends InstrumentationBase<typeof Anthrop
   private oiTracer: OITracer;
   private tracerProvider?: TracerProvider;
   private traceConfig?: TraceConfigOptions;
+  private readonly patchedModuleExports = new Map<
+    typeof Anthropic,
+    typeof Anthropic & { openInferencePatched?: boolean }
+  >();
 
   constructor({
     instrumentationConfig,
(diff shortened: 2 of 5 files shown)


PabloPardoGarcia:

mikeldking (approved):

PR #3560, opened 2026-08-14
fix(agno): create agent spans for continued (human-in-the-loop) runs

jarstingstall:
## What this is

When an Agno `Agent` or `Team` run pauses for human-in-the-loop input and is later resumed, Agno dispatches through dedicated continuation entrypoints. Those entrypoints were not instrumented, so resumed model and tool spans could start new traces without a parent Agent span.

## What's included

- Wrap `_continue_run`, `_continue_run_stream`, `_acontinue_run`, and `_acontinue_run_stream` in both `agno.agent._run` and `agno.team._run`.
- Restore every wrapped continuation function during uninstrumentation.
- Name continuation spans `{name}.continue_run` / `{name}.acontinue_run`, with stream variants using the same base names. Existing `.run` / `.arun` span names are unchanged.
- Test wrapping, restoration, and parent-child trace relationships for sync, streaming, async, and async-streaming Team continuations.
- Add runnable sync and async Team continuation examples that export OTLP traces to Phoenix at `http://localhost:6006/v1/traces`.

## Verification

- Focused continuation tests: 9 passed.
- Ruff and mypy pass for the Agno package.
- Both examples were round-tripped to local Phoenix; resumed LLM spans share the continuation Agent span's trace and parent, with no orphan model/tool roots.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/examples/team_acontinue_run.py
+++ b/python/instrumentation/openinference-instrumentation-agno/examples/team_acontinue_run.py
@@ -1,0 +1,100 @@
+"""Trace asynchronous Team human-in-the-loop continuations in local Phoenix.
+
+Prerequisites:
+    phoenix serve
+    export OPENAI_API_KEY=...
+    python team_acontinue_run.py
+
+The example sends traces to http://localhost:6006 and exercises both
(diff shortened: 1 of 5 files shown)


jarstingstall:

mikeldking (approved):
Thanks for helping maintain agno!

PR #3563, opened 2026-08-15
fix(portkey): restore prompt completion methods on uninstrument

### LINKED ISSUE #3548: [bug] Portkey uninstrument leaves prompt completion methods wrapped
**Describe the bug**

`PortkeyInstrumentor().uninstrument()` does not restore the synchronous or asynchronous prompt completion methods in `portkey_ai.api_resources.apis.generation`.

Instrumentation wraps all four methods:

- `chat_complete.Completions.create`
- `generation.Completions.create`
- `chat_complete.AsyncCompletions.create`
- `generation.AsyncCompletions.create`

However, uninstrumentation restores only the two methods from `chat_complete`. As a result, `client.prompts.completions.create()` remains instrumented after `uninstrument()` is called.

**To Reproduce**

On the latest `main` (`0b5a2176`) with `portkey-ai==1.11.2`:

```python
from portkey_ai.api_resources.apis import chat_complete, generation
from openinference.instrumentation.portkey import PortkeyInstrumentor

targets = {
    "chat_sync": (chat_complete.Completions, "create"),
    "prompt_sync": (generation.Completions, "create"),
    "chat_async": (chat_complete.AsyncCompletions, "create"),
    "prompt_async": (generation.AsyncCompletions, "create"),
}
originals = {
    name: getattr(owner, attribute)
    for name, (owner, attribute) in targets.items()
}

instrumentor = PortkeyInstrumentor()
instrumentor.instrument()

wrapped = {
    name: getattr(owner, attribute) is not originals[name]
    for name, (owner, attribute) in targets.items()
}

instrumentor.uninstrument()

restored = {
    name: getattr(owner, attribute) is originals[name]
    for name, (owner, attribute) in targets.items()
}

print("wrapped:", wrapped)
print("restored:", restored)
```

output:
```
wrapped: {'chat_sync': True, 'prompt_sync': True, 'chat_async': True, 'prompt_async': True}
restored: {'chat_sync': True, 'prompt_sync': False, 'chat_async': True, 'prompt_async': False}
```
No API request is required to reproduce the issue.

## Expected behavior
All four methods should be restored to their original callables after `uninstrument()`:
```
restored: {'chat_sync': True, 'prompt_sync': True, 'chat_async': True, 'prompt_async': True}
```
## Desktop
- OS: Windows
- Python: 3.14.3
- portkey-ai: 1.11.2
- OpenInference commit: 0b5a2176

## Additional context
The prompt completion instrumentation was introduced in #1663, and asynchronous prompt completion support was added in #2407.
I would be happy to submit a small, focused PR that stores and restores the two generation callables and adds regression coverage for both synchronous and asynchronous methods. Please let me know if that approach is welcome.

feiiiiii5:
Fixes #3548

# Description

`PortkeyInstrumentor().uninstrument()` restores only the two `chat_complete` methods it wraps, while `_instrument()` wraps four: `Completions.create` and `AsyncCompletions.create` on both `portkey_ai.api_resources.apis.chat_complete` and `portkey_ai.api_resources.apis.generation`. The two `generation` methods (the prompt completion API, `client.prompts.completions.create`) stay wrapped forever after uninstrument, so spans keep being produced and the wrappers keep running.

Root cause is not just the missing restore: `generation.Completions` is a separate class from `chat_complete.Completions` with its own `create` method, so the original cannot be shared between the two modules. The fix saves all four originals independently and restores each module's methods from its own saved original.

# Test

New deterministic regression test `test_uninstrument_restores_all_wrapped_methods` asserts that instrument wraps all four class methods and uninstrument restores each one to its original function object. It fails on the current code (AssertionError on the generation restore) and passes with the fix. Full portkey test file passes locally: 9 passed.

# Scope

One file changed (`portkey/__init__.py`) plus the regression test in `test_instrumentor.py`. No behavior change while instrumented; only the uninstrument lifecycle is corrected.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-portkey/src/openinference/instrumentation/portkey/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-portkey/src/openinference/instrumentation/portkey/__init__.py
@@ -37,6 +37,8 @@ class PortkeyInstrumentor(BaseInstrumentor):  # type: ignore[misc]
         from portkey_ai.api_resources.apis.chat_complete import AsyncCompletions, Completions
         from portkey_ai.api_resources.apis.generation import (
             AsyncCompletions as AsyncPromptCompletions,
+        )
+        from portkey_ai.api_resources.apis.generation import (
             Completions as PromptCompletions,
         )
 
--- a/python/instrumentation/openinference-instrumentation-portkey/src/openinference/instrumentation/portkey/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-portkey/src/openinference/instrumentation/portkey/__init__.py
@@ -22,13 +22,23 @@ _instruments = ("portkey_ai >= 0.1.0",)
 class PortkeyInstrumentor(BaseInstrumentor):  # type: ignore[misc]
     """An instrumentor for the Portkey AI framework."""
 
-    __slots__ = ("_original_completions_create", "_original_async_completions_create", "_tracer")
+    __slots__ = (
+        "_original_completions_create",
+        "_original_async_completions_create",
+        "_original_prompt_completions_create",
+        "_original_async_prompt_completions_create",
+        "_tracer",
+    )
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
 
     def _instrument(self, **kwargs: Any) -> None:
         from portkey_ai.api_resources.apis.chat_complete import AsyncCompletions, Completions
+        from portkey_ai.api_resources.apis.generation import (
+            AsyncCompletions as AsyncPromptCompletions,
+            Completions as PromptCompletions,
+        )
 
         if not (tracer_provider := kwargs.get("tracer_provider")):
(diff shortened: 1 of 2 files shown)


mikeldking (commented):
thanks for this. Will merge when CI passes

feiiiiii5:
Pushed `76bde0a` to fix the Python CI failure: ruff I001 in `__init__.py` and the new uninstrument test (third-party/first-party and aliased generation import ordering). `ruff check --no-fix .` and `ruff format --check .` are clean with the CI-pinned ruff 0.9.2 in the portkey package.

feiiiiii5:
@mikeldking thanks! The five new workflow runs for `76bde0a` are sitting in `action_required` (fork workflow approval). Locally the portkey package is clean with the CI-pinned ruff 0.9.2 (`ruff check --no-fix .` + `ruff format --check .`), so approving the Python CI run should let it pass.

mikeldking (approved):
thanks!

PR #3565, opened 2026-08-15
fix(guardrails): restore Guard.from_* constructors on uninstrument

### LINKED ISSUE #3564: [bug] GuardrailsInstrumentor uninstrument leaves Guard.from_* constructors wrapped
## Summary

`GuardrailsInstrumentor().uninstrument()` does not restore the four `Guard.from_*` constructors it wraps, so they keep injecting a stale tracer after uninstrument. Reproduced on guardrails-ai 0.4.5 (the version range this instrumentor declares support for: `guardrails-ai>=0.4.5,<0.5.1`).

## Repro

```python
from openinference.instrumentation.guardrails import GuardrailsInstrumentor
import guardrails as gd

inst = GuardrailsInstrumentor()
inst.instrument(tracer_provider=TracerProvider())

orig = gd.guard.Guard.__dict__["from_pydantic"]
inst.uninstrument()

assert gd.guard.Guard.__dict__["from_pydantic"] is orig
# AssertionError: the attribute is still a wrapt FunctionWrapper whose lambda
# keeps passing tracer=<stale OITracer> into Guard constructors
```

## Details

`_instrument()` wraps five targets. Four of them (`Guard.from_pydantic`, `Guard.from_string`, `Guard.from_rail_string`, `Guard.from_rail`) are wrapped with inline lambdas and their originals are never saved, while the other wrapped targets (`Runner.step`, `PromptCallableBase.__call__`, `ValidatorServiceBase.after_run_validator`, and the two `contextvars` proxies) are all restored in `_uninstrument()`. This is the same lifecycle-bug family as #3548 (Portkey) and #3553 (MCP): after `uninstrument()`, spans keep being produced through a tracer whose provider the caller has already disposed of.

Measured behavior on 0.4.5: after instrument, `Guard.from_pydantic` has `__wrapped__`; after uninstrument it still has `__wrapped__` and is not the original classmethod, while `gd.guard.contextvars` is restored correctly.

I have a fix ready: save the four raw class-dict entries at instrument time and restore them in `_uninstrument()`, plus a regression test asserting all four are restored. Will open a PR after this issue.

feiiiiii5:
Fixes #3564

# Description

`GuardrailsInstrumentor().uninstrument()` restored every wrapped target except the four `Guard.from_*` constructors (`from_pydantic`, `from_string`, `from_rail_string`, `from_rail`). Those four are wrapped with inline lambdas whose originals were never saved, so after `uninstrument()` they kept injecting a stale tracer into `Guard` constructors and spans kept being produced through a tracer provider the caller had already disposed of.

Reproduced on guardrails-ai 0.4.5, the SDK range this instrumentor declares support for. After instrument, `Guard.from_pydantic` carries `__wrapped__`; after uninstrument it still carries `__wrapped__`, while `guardrails.guard.contextvars` is correctly restored.

# Fix

Save the raw class-dict entries of the four constructors at instrument time and restore them in `_uninstrument()`, following the same pattern already used for `Runner.step`, `PromptCallableBase.__call__`, and `ValidatorServiceBase.after_run_validator`.

# Test

Extended the existing `test_guardrails_uninstrumentation` to capture the four originals before instrumentation and assert each is restored after uninstrument. The test fails on the current code (`AssertionError: Expected Guard.from_pydantic to be unwrapped`) and passes with the fix. Full guardrails instrumentor test file passes locally against guardrails-ai 0.4.5 (4 passed).

# Scope

`openinference/instrumentation/guardrails/__init__.py` (save and restore the four originals) plus the test extension. No behavior change while instrumented.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-guardrails/src/openinference/instrumentation/guardrails/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-guardrails/src/openinference/instrumentation/guardrails/__init__.py
@@ -82,7 +82,7 @@ class GuardrailsInstrumentor(BaseInstrumentor):  # type: ignore
 
         runner_module = import_module(_RUNNER_MODULE)
         self._original_guardrails_runner_step = runner_module.Runner.step
-        runner_wrapper = _ParseCallableWrapper(tracer=self._tracer)
+        runner_wrapper = _ParseCallableWrapper(tracer=self._tracer)  # type: ignore[arg-type]
         wrap_function_wrapper(
             _RUNNER_MODULE,
             "Runner.step",
@@ -93,7 +93,7 @@ class GuardrailsInstrumentor(BaseInstrumentor):  # type: ignore
         self._original_guardrails_llm_providers_call = (
             llm_providers_module.PromptCallableBase.__call__
         )
-        prompt_callable_wrapper = _PromptCallableWrapper(tracer=self._tracer)
+        prompt_callable_wrapper = _PromptCallableWrapper(tracer=self._tracer)  # type: ignore[arg-type]
         wrap_function_wrapper(
             _LLM_PROVIDERS_MODULE,
             "PromptCallableBase.__call__",
@@ -104,7 +104,7 @@ class GuardrailsInstrumentor(BaseInstrumentor):  # type: ignore
         self._original_guardrails_validation_after_run = (
             validation_module.ValidatorServiceBase.after_run_validator
         )
-        post_validator_wrapper = _PostValidationWrapper(tracer=self._tracer)
+        post_validator_wrapper = _PostValidationWrapper(tracer=self._tracer)  # type: ignore[arg-type]
         wrap_function_wrapper(
             _VALIDATION_MODULE,
             "ValidatorServiceBase.after_run_validator",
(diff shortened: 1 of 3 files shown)


feiiiiii5:
CI note: the `Find Tox Testenv` failure is repo configuration, not this diff. `python/tox.ini:18` has the guardrails envs commented out (`; py3{10,14}-ci-{guardrails,guardrails-latest}`), so `tox list` yields no matching env for the two changed guardrails files and the workflow hits its “no testenv found” guard. Enabling those envs is a maintainer config decision; the four uninstrument regression tests were verified locally during development (guardrails-ai 0.4.5).

ialisaleh (commented):
This PR resolves the issue where `GuardrailsInstrumentor().uninstrument()` was not restoring the four `Guard.from_*` constructors. I reviewed the changes and verified that the fix correctly saves the original class-dict entries before wrapping and restores them during `_uninstrument()`. I also verified the regression tests covering all four constructors.

I updated the Tox configuration to enable the Guardrails test environment for Python 3.10, which passes successfully. I removed Python 3.14 because `guardrails-ai==0.4.5` pulls in `lxml==4.9.4`, which fails to build on Python 3.14 in CI due to missing native `libxml2`/`libxslt` dependencies. Since this is a dependency compatibility issue unrelated to the PR, I kept Python 3.14 excluded rather than introducing additional dependency or system-package changes. Thanks!

caroger (approved):

PR #3578, opened 2026-08-18
fix(litellm): Preserve Stream Types & Isolate Extraction Errors

### LINKED ISSUE #3577: [bug] litellm instrumentation replaces streaming results with bare generators — breaks litellm's Responses API bridge; span serialization can fail the traced call
**Package**: `openinference-instrumentation-litellm` (observed on 0.1.36; code unchanged on `main`)

### Defect 1 — streaming wrappers change the return type

For `stream=True`, `_acompletion_wrapper` returns `_finalize_streaming_span(span, result)` — an **async generator** — instead of the `CustomStreamWrapper` litellm returned (same for the sync wrapper and the responses wrappers). Any caller that type-checks the stream breaks. The highest-impact victim is **litellm itself**: its Responses API → chat-completions bridge re-enters the instrumented `litellm.acompletion` and does `isinstance(result, CustomStreamWrapper)`, so with this instrumentation active, **every streamed `/v1/responses` request to a provider without native Responses support (Gemini/Vertex, Bedrock, …) fails** with:

```
litellm.responses.litellm_completion_transformation.handler: ValueError: Unexpected response type: <class 'async_generator'>
```

**Repro**: litellm proxy ≥1.94.x with `LiteLLMInstrumentor().instrument()`, then `POST /v1/responses` with `stream: true` against any bridged provider (e.g. `gemini/gemini-2.5-flash`). 100% reproducible; without the instrumentor the same request streams fine.

### Defect 2 — span attribute extraction raises into the traced call

`_instrument_func_type_responses` → `_json_serialize(invocation_parameters)` uses `json.dumps`, which raises `ValueError: Circular reference detected` on self-referencing kwargs (observed on the litellm router's retry path, where kwargs carry proxy metadata). The exception propagates **into the instrumented call**, replacing the caller's real error — in production this masked Defect 1 entirely, and unrelated failures on other providers surfaced as the same misleading message.

### Defect 3 — `aresponses` finalizer silently drains foreign stream types

`_finalize_aresponses_streaming_span` yields nothing unless the stream is exactly `ResponsesAPIStreamingIterator` — the bridge's `LiteLLMCompletionStreamingIterator` (or any other iterator) becomes an **empty stream** instead of a response. (The sync `_responses_wrapper` already gates at the call site, but leaks the span on the pass-through path.)

### Suggested fix

Instrumentation should be transparent: wrap streams in a `wrapt.ObjectProxy` that preserves type identity while iterating the finalizing generator (wrapt is already a dependency); gate the responses finalizers on the exact type and pass anything else through untouched (ending the span); and exception-isolate attribute extraction so span enrichment can never fail the instrumented call.

PR with this fix (all existing tests pass + 4 new ones): incoming, will be linked here. We are running the equivalent patch in production at Simpplr (via a pinned shim) — happy to iterate on review feedback.

sushantsimpplr:
Closes #3577

### Description

Instrumentation must be transparent to callers of the instrumented functions. Three related defects in `openinference-instrumentation-litellm` break that, with litellm itself as the highest-impact caller — its Responses API → chat-completions bridge re-enters the instrumented `litellm.acompletion` and type-checks the result, so with this instrumentor active every streamed `/v1/responses` request to a provider without native Responses support fails with `Unexpected response type: <class 'async_generator'>`, and a secondary serialization crash masks the error (details in the issue).

### Changes

1. **Type-preserving stream wrappers** — new `_TracedSyncStream` / `_TracedAsyncStream` (`wrapt.ObjectProxy`, already a dependency) present as the original stream (`isinstance` passes, attribute access delegates) while iterating the existing finalizing generators, so span accumulation/finalization is unchanged. The four streaming return sites (`completion`, `acompletion`, `responses`, `aresponses`) now return the proxy instead of the bare generator.
2. **Exact-type gate for `aresponses` streaming** — mirrors the sync wrapper's existing gate: streams that are not `ResponsesAPIStreamingIterator` pass through untouched (previously the finalizer silently drained them into an empty stream). Both responses wrappers now also end the span on the pass-through path (previously leaked).
3. **Exception-isolated attribute extraction** — `_suppress_extractor_errors` decorator on the `_instrument_func_type_*` extractors and `_finalize_span`: extraction failures (e.g. `ValueError: Circular reference detected` from `json.dumps` on self-referencing kwargs seen on litellm router retries) log and drop attributes instead of raising into the traced call.

### Behavior before/after

```python
LiteLLMInstrumentor().instrument()
resp = await litellm.acompletion(model=..., messages=..., stream=True)

# before: type(resp) is async_generator  → litellm's Responses bridge raises

# after:  isinstance(resp, CustomStreamWrapper) is True; iteration + span capture unchanged
```

### Testing

- 4 new tests in `tests/test_streaming_transparency.py`: sync + async stream type preservation (content and span capture asserted), foreign-iterator pass-through on `aresponses`, and extractor isolation on cyclic kwargs.
- Full existing package suite passes: 115 passed. `ruff check` / `ruff format --check` clean on changed files.

### Notes / Shortcomings

- The anthropic-messages streaming wrappers likely share the type-swap pattern; left out to keep this PR small — happy to follow up.
- `_json_serialize` in `openinference-instrumentation` core could additionally be made cycle-safe (benefits all instrumentors); also left out of scope here.
- We currently run this exact fix in production at Simpplr as a pinned monkeypatch shim and will drop it in favor of the released fix.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import functools
 import inspect
+import logging
 from enum import Enum
 from types import SimpleNamespace
 from typing import (
@@ -18,6 +20,20 @@ from typing import (
     Union,
 )
 
+import wrapt
+from openinference.semconv.trace import (
+    EmbeddingAttributes,
+    ImageAttributes,
+    MessageAttributes,
+    MessageContentAttributes,
+    OpenInferenceLLMProviderValues,
+    OpenInferenceLLMSystemValues,
+    OpenInferenceMimeTypeValues,
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+    ToolAttributes,
+    ToolCallAttributes,
+)
 from opentelemetry import context as context_api
 from opentelemetry import trace as trace_api
 from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY
@@ -50,19 +66,8 @@ from openinference.instrumentation.litellm._responses_attributes import (
 )
 from openinference.instrumentation.litellm.package import _instruments
 from openinference.instrumentation.litellm.version import __version__
-from openinference.semconv.trace import (
-    EmbeddingAttributes,
-    ImageAttributes,
-    MessageAttributes,
-    MessageContentAttributes,
-    OpenInferenceLLMProviderValues,
-    OpenInferenceLLMSystemValues,
-    OpenInferenceMimeTypeValues,
-    OpenInferenceSpanKindValues,
-    SpanAttributes,
-    ToolAttributes,
-    ToolCallAttributes,
-)
+
+logger = logging.getLogger(__name__)
 
 # Skip capture
 KEYS_TO_REDACT = ["api_key", "messages"]
@@ -325,6 +330,27 @@ def _get_attributes_from_image(
         yield f"{ImageAttributes.IMAGE_URL}", url
 
 
+def _suppress_extractor_errors(fn: Callable[..., Any]) -> Callable[..., Any]:
+    """Recording span attributes must never fail the instrumented call.
+
+    Extraction serializes caller-provided structures (e.g. via json.dumps),
+    which can raise — for example ``ValueError: Circular reference detected``
+    on self-referencing kwargs seen on litellm router retries. Such an error
+    would otherwise propagate into the traced call and mask the real outcome.
+    """
+
+    @functools.wraps(fn)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        try:
+            return fn(*args, **kwargs)
+        except Exception:
+            logger.exception("Failed to record span attributes in %s", fn.__name__)
+            return None
+
+    return wrapper
+
+
+@_suppress_extractor_errors
 def _instrument_func_type_responses(span: trace_api.Span, kwargs: Dict[str, Any]) -> None:
     """
     Currently instruments the functions:
@@ -349,6 +375,7 @@ def _instrument_func_type_responses(span: trace_api.Span, kwargs: Dict[str, Any]
     )
 
 
+@_suppress_extractor_errors
 def _instrument_func_type_completion(span: trace_api.Span, kwargs: Dict[str, Any]) -> None:
     """
     Currently instruments the functions:
@@ -404,6 +431,7 @@ def _instrument_func_type_completion(span: trace_api.Span, kwargs: Dict[str, Any
                 )
 
 
+@_suppress_extractor_errors
 def _instrument_func_type_embedding(span: trace_api.Span, kwargs: Dict[str, Any]) -> None:
     """
     Currently instruments the functions:
@@ -452,6 +480,7 @@ def _instrument_func_type_embedding(span: trace_api.Span, kwargs: Dict[str, Any]
     _set_span_attribute(span, SpanAttributes.INPUT_VALUE, str(kwargs.get("input")))
 
 
+@_suppress_extractor_errors
 def _instrument_func_type_image_generation(span: trace_api.Span, kwargs: Dict[str, Any]) -> None:
     """
     Currently instruments the functions:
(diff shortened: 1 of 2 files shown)


ialisaleh (commented):
I reviewed this PR, and the solution correctly addresses the reported issues. The `ObjectProxy` wrappers preserve LiteLLM’s expected stream compatibility while allowing spans to be finalized, unsupported Responses API stream types are passed through unchanged, and attribute extraction failures are isolated so they cannot affect the underlying call.

The tests provide good coverage for sync/async streaming, both sync/async Responses foreign-stream handling, and extraction failures. Thanks!

Image

ialisaleh:
Hi @sushantsimpplr - Please sign the CLA as requested in the comment above before we can approve this PR for merge. Thanks!

sushantsimpplr:

sushantsimpplr:
@ialisaleh signed, thanks for the nudge. And thanks for approving the workflow run and fixing the mypy and import-order failures; as a first-time contributor I couldn't see CI results on the original push.

Went through the delta: import order, docstring trims, the widened `type: ignore`s, and the new sync `responses` pass-through test (nice, it also pins the `span.end()` on that path). Fix semantics unchanged from my side.

Separately: the anthropic-messages wrappers share the type-swap pattern. `_finalize_anthropic_messages_result` duck-types on `__anext__` / `__next__` and returns the bare finalizer generator for any stream. Happy to raise the same `ObjectProxy` treatment as a follow-up PR if you want it.

python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
```diff
@@ -1016,6 +1041,40 @@ def _remove_redundant_reasoning_entries(
...
+        self._self_finalized_iterator = finalized_iterator
+
+    def __aiter__(self) -> Any:
+        return self
+
+    async def __anext__(self) -> Any:
+        return await self._self_finalized_iterator.__anext__()
+
+    async def aclose(self) -> Any:
+        return await self._self_finalized_iterator.aclose()
```
caroger: This override closes only the tracing generator. It bypasses LiteLLM’s `CustomStreamWrapper.aclose()`, which closes the provider stream under cancellation shielding. After consuming one chunk, `await response.aclose()` leaves `response.__wrapped__.completion_stream` open, while the uninstrumented stream clears it. This can leak the provider HTTP connection on cancellation. Please close both streams and add an early-close regression test that asserts underlying cleanup and one finished span.

caroger (changes requested):
The type-preservation fix is sound on full stream consumption, but async cleanup remains non-transparent. Please address the inline resource-leak finding and add an early-close regression test.

python/instrumentation/openinference-instrumentation-litellm/src/openinference/instrumentation/litellm/__init__.py
sushantsimpplr: Fixed in bfcced9. `aclose` now closes the tracing generator and then delegates to the wrapped stream's `aclose`, inside a `finally` so the connection is released even if span finalization raises. The `getattr` guard keeps the pass-through cases intact, since the Responses iterators have no `aclose` to delegate to.

Added `test_async_early_close_closes_underlying_stream`: consumes one chunk, closes, then asserts `completion_stream` is cleared and exactly one span finished. It fails on the previous commit and passes on this one. Verified against litellm 1.84.0 from `test-requirements.txt`: 117 passed, with ruff 0.9.2 and mypy 1.11.2 clean.

I left the sync `close` alone on purpose: `CustomStreamWrapper` has no sync `close`, and neither does `SyncResponsesAPIStreamingIterator`, so nothing is being bypassed there.

One adjacent issue I did not fix here. If a caller closes before consuming any chunk, the finalizer generator was never started, so its `finally` never runs and the span is never ended. That predates this PR, since the bare generator behaved the same way, and fixing it means threading the span into the proxies. Happy to file it separately or fold it in, whichever you prefer.

sushantsimpplr:
@caroger @ialisaleh pushed bfcced9, which addresses the review: `aclose` now delegates to the wrapped stream's own `aclose`, plus an early-close regression test. Reasoning is in the thread above.

Locally against litellm 1.84.0 from `test-requirements.txt`: 117 passed, with ruff 0.9.2 and mypy 1.11.2 clean. The workflow runs on this commit are sitting at "action required" and need a maintainer to approve them, as on the previous push.

caroger (approved):

PR #3581, opened 2026-08-18
feat(cohere): Add Finish Reason Attribute

ialisaleh:
Refs #2901 

This PR adds `finish_reason` tracking to the Cohere instrumentation by reading the top-level `finish_reason` field from `V2ChatResponse` across sync, async, and streaming paths and mapping it to the `LLM_FINISH_REASON` span attribute, keeping it consistent with other instrumentations.

image


python/instrumentation/openinference-instrumentation-cohere/examples/requirements.txt
```diff
@@ -0,0 +1,5 @@
+cohere
+openinference-instrumentation-cohere
+opentelemetry-sdk
+opentelemetry-exporter-otlp
+opentelemetry-instrumentation-httpx
```
mikeldking: shouldn't need httpx

mikeldking (approved):

PR #3584, opened 2026-08-19
fix(instrumentation): support async context decorators

### LINKED ISSUE #3583: Async context decorators detach attributes before coroutine execution
## Problem

The context helpers inherit from `contextlib.ContextDecorator`, so they can be used as decorators as well as context managers. On an `async def`, however, `ContextDecorator.__call__` exits the context immediately after creating the coroutine. The coroutine body then runs without the configured OpenInference context attributes.

## Reproduction

```python
import asyncio

from openinference.instrumentation import get_attributes_from_context, using_session


@using_session("async-session")
async def read_context() -> dict[str, object]:
    await asyncio.sleep(0)
    return dict(get_attributes_from_context())


print(asyncio.run(read_context()))
```

On current `main` (`d0ec025f`), this prints:

```text
{}
```

The corresponding synchronous decorator API includes `session.id` as expected, and direct `async with using_session(...)` usage also works.

## Expected behavior

Async functions decorated with `using_session` (and the other context-attribute decorators backed by the same context manager) should retain the configured attributes throughout coroutine execution, including across suspension points. Repeated or concurrent decorated calls should not share OpenTelemetry context tokens, and context should be detached after normal return or an exception.

## Duplicate check

I searched open and closed issues and pull requests for `using_session`, `using_attributes`, `ContextDecorator`, `_UsingAttributesContextManager`, async context decorators, coroutine propagation, and missing context attributes. I found no active or completed implementation of this behavior. Closed PR #1570 concerns dynamic decorator arguments and function-signature changes, which is a separate design.

adityat210:
Fixes #3583

# Description

Async functions decorated with OpenInference context helpers currently execute after `contextlib.ContextDecorator` has already detached the configured context. As a result, spans created in the coroutine silently lose session, user, metadata, tag, or prompt-template attributes.

This change keeps the context active while a decorated coroutine is awaited and recreates the context manager for each invocation so concurrent calls do not share an OpenTelemetry token. Synchronous decorators continue through `ContextDecorator`'s existing path.

## Root cause

`ContextDecorator.__call__` wraps only the call that creates a coroutine object; it does not await the coroutine. The `with` block therefore exits before the async function body runs.

## Impact (XYZ)

Restores all six public context-attribute decorators for async functions (X), measured by the same reproducer changing from `{}` to `{'session.id': 'async-session'}` and by 8,459 passing package tests (Y), by awaiting coroutine execution inside a fresh async context-manager instance per call (Z).

| Scenario | Before | After |
| --- | --- | --- |
| Attribute visible after an `await` | No | Yes |
| Two concurrent decorated calls | Shared mutable token state | Independent context-manager state |
| Exception path | Context absent during coroutine | Context present, then detached |
| Public API / span schema | Unchanged | Unchanged |

The patch is limited to two files (+59/-1 lines). It adds no attributes, renames no fields, and changes no span hierarchy.

## Duplicate check

No open or closed issue/PR matched searches for `using_session`, `using_attributes`, `ContextDecorator`, `_UsingAttributesContextManager`, async context decorators, coroutine propagation, or missing context attributes. Closed PR #1570 proposed dynamic decorator arguments and signature changes; this fix does not alter signatures or argument semantics.

## Validation

- `tox run -e py310-test-instrumentation -- tests/test_context_managers.py` — 20 passed
- `tox run -e py310-ci-instrumentation` — Ruff format/check passed, strict mypy passed, 8,459 tests passed
- One unrelated existing Pydantic deprecation warning remains in `test_infer_tool_parameters`

# Checklist:

### DIFF
--- a/python/openinference-instrumentation/examples/async_context_attributes.py
+++ b/python/openinference-instrumentation/examples/async_context_attributes.py
@@ -1,0 +1,106 @@
+"""Async usage of OpenInference context attributes.
+
+The context managers exported by ``openinference.instrumentation`` (``using_session``,
+``using_user``, ``using_attributes``, ``suppress_tracing``) work in async code in every
+shape demonstrated below:
+
+1. ``@using_session(...)`` on an ``async def`` — attributes are attached for the whole
+   awaited call, including across ``await`` suspension points.
+2. ``@using_session(...)`` on an async generator — attributes stay attached for the
+   entire iteration (e.g. a streaming handler), not just generator creation.
+3. ``async with`` — attributes scope to the awaited block, and concurrent tasks each
+   keep their own attributes without interfering.
+4. ``async with suppress_tracing()`` — spans created inside are not exported.
+
+Run a local Phoenix (https://github.com/Arize-ai/phoenix) at http://localhost:6006,
+then:
+
+    python async_context_attributes.py
+
+and open the "async-context-attributes" project in Phoenix to see the spans, each
+carrying the session/user/metadata attributes from its enclosing context.
+"""
+
+import asyncio
+from typing import AsyncIterator
+
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+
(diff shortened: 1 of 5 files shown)


adityat210:

PR #3585, opened 2026-08-19
feat(semconv): add llm.input_model_name and llm.output_model_name

### LINKED ISSUE #3439: [feature request] Capture requested and effective Anthropic models after classifier routing
## Problem

Anthropic now supports automatic model fallback when a safety classifier declines a request. For example, a request intended for Claude Opus 5 can be served by Opus 4.8. The response's top-level `model` field identifies the model that actually answered.

OpenInference currently has one `llm.model_name` attribute. Both Anthropic instrumentors initially capture the requested model and later overwrite it with `response.model`, so the completed span does not expose the intended and effective models as distinct queryable attributes:

- Python: `python/instrumentation/openinference-instrumentation-anthropic`
- JavaScript: `js/packages/openinference-instrumentation-anthropic`

The intended model may remain embedded in `llm.invocation_parameters` or `input.value`, but relying on JSON parsing prevents straightforward routing, cost, refusal, and model-quality analysis.

## Motivation

Anthropic's Opus 5 announcement describes classifier-triggered fallback to Opus 4.8 and automatic API fallbacks. This makes requested-versus-serving model divergence an expected behavior rather than an edge case:

- https://www.anthropic.com/news/claude-opus-5
- https://platform.claude.com/docs/en/build-with-claude/refusals-and-fallback

OpenTelemetry already models this distinction as `gen_ai.request.model` and `gen_ai.response.model`.

## Proposed outcome

Define and emit separate OpenInference attributes for:

1. **Intended/requested model** — the exact model supplied by the caller.
2. **Effective/response model** — the model reported by Anthropic as having generated the response.

Preserve the existing `llm.model_name` behavior for backward compatibility, with its precedence and meaning documented. Align the new names and semantics with the OTel GenAI request/response model distinction where practical.

## Acceptance criteria

- [ ] OpenInference semantic conventions distinguish requested and effective LLM model identifiers.
- [ ] Backward-compatible behavior for `llm.model_name` is specified.
- [ ] Python Anthropic instrumentation records both values for streaming and non-streaming calls.
- [ ] JavaScript Anthropic instrumentation records both values for streaming and non-streaming calls.
- [ ] Tests cover a routed response where the request model differs from `response.model` (for example, Opus 5 requested and Opus 4.8 effective).
- [ ] Documentation explains how to query classifier/fallback routing.

## Additional consideration

Anthropic's `usage.iterations` can describe each attempted model in a fallback chain. Capturing the full chain may be useful, but is not required to satisfy the core requested-versus-effective model use case.

satyadevai:
Adds sem conventions for #3439

### DIFF
--- a/go/openinference-semantic-conventions/attributes.go
+++ b/go/openinference-semantic-conventions/attributes.go
@@ -70,15 +70,15 @@ const (
 const (
 	LLMModelName = "llm.model_name"
 
-	// LLMInputModelName is the model requested by the caller, as sent in the
-	// request. May differ from LLMOutputModelName when the provider routes
+	// LLMRequestModelName is the model requested by the caller, as sent in the
+	// request. May differ from LLMResponseModelName when the provider routes
 	// the request to a different model (e.g. classifier-triggered fallback).
-	LLMInputModelName = "llm.input_model_name"
+	LLMRequestModelName = "llm.request.model_name"
 
-	// LLMOutputModelName is the model that actually generated the response,
-	// as reported by the provider. May differ from LLMInputModelName when
+	// LLMResponseModelName is the model that actually generated the response,
+	// as reported by the provider. May differ from LLMRequestModelName when
 	// the provider routes the request to a different model.
-	LLMOutputModelName = "llm.output_model_name"
+	LLMResponseModelName = "llm.response.model_name"
 
 	LLMProvider             = "llm.provider"
 	LLMSystem               = "llm.system"
--- a/go/openinference-semantic-conventions/attributes.go
+++ b/go/openinference-semantic-conventions/attributes.go
@@ -68,7 +68,18 @@ const (
 
 // LLM-span attributes — set when the span represents an LLM API call.
 const (
-	LLMModelName            = "llm.model_name"
+	LLMModelName = "llm.model_name"
+
+	// LLMInputModelName is the model requested by the caller, as sent in the
+	// request. May differ from LLMOutputModelName when the provider routes
+	// the request to a different model (e.g. classifier-triggered fallback).
+	LLMInputModelName = "llm.input_model_name"
+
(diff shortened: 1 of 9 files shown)


go/openinference-semantic-conventions/attributes.go
```diff
@@ -68,7 +68,18 @@ const (
 
 // LLM-span attributes — set when the span represents an LLM API call.
 const (
-	LLMModelName            = "llm.model_name"
+	LLMModelName = "llm.model_name"
+
+	// LLMInputModelName is the model requested by the caller, as sent in the
+	// request. May differ from LLMOutputModelName when the provider routes
+	// the request to a different model (e.g. classifier-triggered fallback).
+	LLMInputModelName = "llm.input_model_name"
```
mikeldking:
```suggestion
	LLMInputModelName = "llm.request.model_name"
```
because it's not the input - it's the requested model. see e.x. https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-response-model

go/openinference-semantic-conventions/attributes.go
```diff
@@ -68,7 +68,18 @@ const (
...
+
+	// LLMOutputModelName is the model that actually generated the response,
+	// as reported by the provider. May differ from LLMInputModelName when
+	// the provider routes the request to a different model.
+	LLMOutputModelName = "llm.output_model_name"
```
mikeldking:
```suggestion
	LLMOutputModelName = "llm.response.model_name"
```

mikeldking (changes requested):
let's use request / response since that's more semantic

go/openinference-semantic-conventions/attributes.go
satyadevai: I have made suggested changes - aligned names with the otel specs.

mikeldking (approved):

PR #3591, opened 2026-08-20
fix(google_adk): Avoid Redundant Request Attribute Derivation

### LINKED ISSUE #3590: google-adk: request-side span attributes are re-derived on every streamed chunk
## Problem

`_TraceCallLlm` re-derives all of the request-side span attributes once per streamed chunk. The
values are identical every time, so everything except the last pass is thrown away.

ADK opens the `call_llm` span outside its streaming loop and calls `trace_call_llm` from inside it:

```
google/adk/flows/llm_flows/base_llm_flow.py (main)
  :1596   with tracer.start_as_current_span('call_llm') as span:
  :1668     async for llm_response in agen:
  :1669       trace_call_llm(invocation_context, model_response_event.id, llm_request, llm_response, span)
```

`llm_request` is bound outside the loop and does not change. So for one streamed answer
`_TraceCallLlm.__call__` (`_wrappers.py:247`) runs once per chunk, and each time it:

- re-serializes the entire `LlmRequest` for `INPUT_VALUE` (`:280`)
- walks `llm_request.tools_dict` and re-serializes every tool declaration (`:289-295`, and
  `_get_attributes_from_base_tool` at `:621-628`, which is uncached)
- re-emits the system instruction and the whole message history (`:304-335`)

The response-side attributes at `:336-338` do change per chunk and are fine as they are.

There is also no `span.is_recording()` check, so lowering the sampling rate avoids none of this.
ADK's own `trace_call_llm` returns early on a non-recording span
(`src/google/adk/telemetry/tracing.py:376`), and `OpenInferenceSpan.set_attribute` does too, so at
the moment a sampled-out span costs more with this instrumentation than without it.

## Cost

`main` vs. a patched build in the same process. Generated tools and requests only, no model and no
network: 7 tools x 27 params, 300 chunks, min of 5 repeats, Linux x86_64, CPython 3.12.3.

| google-adk | before | after | per chunk |
|---|---|---|---|
| 2.7.1 | 0.802 s | 0.026 s | 2.673 -> 0.088 ms |
| 1.2.1 (repo test pin) | 3.130 s | 0.033 s | 10.435 -> 0.109 ms |

That is per streamed response, synchronous on the event loop, so it delays every other coroutine in
the process as well.

Upgrading google-adk does not remove it. 2.6.0 added an `lru_cache` for the declaration build
(google/adk-python@57f3af24) but still returns `model_copy(deep=True)` and does not cache the JSON
serialization, and the full-request re-serialization here is independent of that. The current 1.x
line (1.39.0) has no cache at all.

## Suggested fix

1. Return early when `span.is_recording()` is false, placed after the span-kind `set_attribute` so
   `OpenInferenceSpan`'s important-attribute bookkeeping still runs.
2. Derive the request-side attributes once per span. The span's own attributes are not usable as the
   marker here: they are bounded (`SpanLimits.max_attributes`, 128 by default) and evict
   oldest-first, so a long enough message history drops `INPUT_VALUE` during the first chunk and the
   guard would never fire again.

I have this working with tests if you want a PR.

One heads-up: #3451 changes the `span.set_status(StatusCode.OK)` line immediately above this code.

## Versions

- `openinference-instrumentation-google-adk` 0.1.20, behaviour confirmed against `main`
- `google-adk` 1.2.1, 1.39.0, 2.7.1

KCNyu:
Closes #3590 

### Description

ADK opens the `call_llm` span outside its streaming loop and calls `trace_call_llm` from inside it
(`google/adk/flows/llm_flows/base_llm_flow.py:1596` / `:1668` / `:1669` on `main`), so
`_TraceCallLlm.__call__` runs once per streamed chunk. `llm_request` is bound outside the loop and
does not change, so every chunk re-serialized the whole request, re-derived every tool declaration
and re-emitted the full message history, then wrote the same values onto the same span. Only the
last pass survived.

### What changed

1. Return early when the span is not recording. This is placed after the span-kind `set_attribute`
   so `OpenInferenceSpan`'s important-attribute bookkeeping still runs. ADK's own `trace_call_llm`
   already does the same thing (`src/google/adk/telemetry/tracing.py:376`), as does
   `OpenInferenceSpan.set_attribute`, so a sampled-out span was previously costing more with this
   instrumentation than without it.

2. Derive the request-side attributes once per span. `_TraceCallLlm` keeps a small per-instance
   record of which request has already been written for a given span id, and only records the span
   once the whole block has completed, so a pass that raises part way through is retried on the next
   chunk instead of being suppressed. Response-side attributes still update for every chunk.

   Three notes on that bookkeeping:

   - The span's own attributes cannot be the marker. Checking whether `INPUT_VALUE` is already set
     looks like the obvious approach, but span attributes are bounded
     (`SpanLimits.max_attributes`, 128 by default) and evict oldest-first, so a request with a long
     enough message history overflows the limit during the first chunk and drops `INPUT_VALUE`,
     which would turn the guard off for exactly the requests that cost the most to re-derive.
     `test_guard_survives_the_span_attribute_limit` covers this.
   - The recorded value is the request's identity rather than a plain flag. ADK binds one
     `llm_request` per `call_llm` span today, but keying on the request means a span that ever
     serves a second one still gets its attributes.
   - The record is bounded at 1024 entries, least-recently-used first. Losing an entry costs one
     redundant re-derivation and can never leave an attribute unwritten, and every access is a
     single dict operation, so concurrent callers can at worst duplicate work.

### Benchmark

Generated tools and requests only, no model and no network. `main` vs. this branch in the same
process: 7 tools x 27 params, 300 chunks, min of 5 repeats, Linux x86_64, CPython 3.12.3.

| google-adk | tools | before | after | per chunk |
|---|---|---|---|---|
| 2.7.1 | 7 x 27 `Annotated[..., Field(...)]` | 0.802 s | 0.026 s | 2.673 -> 0.088 ms |
| 1.2.1 (repo test pin) | 7 x 27 plain type hints | 3.130 s | 0.033 s | 10.435 -> 0.109 ms |

The second row uses plainly annotated parameters because google-adk 1.2.1 cannot parse
`Annotated[..., Field(...)]` signatures at all; those are the more expensive kind.

The underlying cost is `FunctionTool._get_declaration()`, which on google-adk < 2.6.0 rebuilds the
declaration from the function signature on every call, so the total scales as chunks x tools. On
1.39.0, `timeit` `number=20 repeat=7 min`:

| annotation style | params | schema bytes | per call | per param |
|---|---|---|---|---|
| `Annotated` + `Field` | 5 / 15 / 27 | 815 / 2250 / 3990 B | 1.962 / 5.486 / 9.651 ms | ~357-392 us |
| `Annotated` + `Field` + `Literal` | 5 / 15 / 27 | 733 / 2018 / 3590 B | 2.216 / 6.299 / 11.311 ms | ~419-443 us |
| plain type hints | 5 / 15 / 27 | 296 / 691 / 1171 B | 0.317 / 0.796 / 1.377 ms | ~51-63 us |

Cost tracks the number of parameters carrying `Annotated[..., pydantic.Field(...)]`, not the size of
the resulting schema: the 3990 B row is faster than the 3590 B one.

### Tests

`tests/test_trace_call_llm_per_chunk.py`, run through `OITracer` so the spans are the
`OpenInferenceSpan` proxies the instrumentor installs:

- request-side attributes derived once across five chunks on one span, asserted with a `FunctionTool`
  subclass that counts declaration builds, with the span still carrying `INPUT_VALUE`,
  `LLM_MODEL_NAME` and the tool schema
- response-side attributes still written for every chunk
- each new span gets its own request attributes
- a second, different request on the same span is still recorded
- a pass that raises part way through is retried on the next chunk
- nothing derived at all for a non-recording span
- the guard still holds when the request overflows the span attribute limit
- build count stays at 1 for 1, 5 and 20 chunks

Eight of the ten fail on `main`, on both `test-google_adk` (adk 1.2.1) and `test-google_adk-latest`
(adk 2.7.1). The two that pass either way are the per-chunk response test and the single-chunk case.
`ruff` and `mypy` are clean.

### Checklist:

### DIFF
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
@@ -8,6 +8,7 @@ from typing import (
     Any,
     AsyncGenerator,
     Callable,
+    Dict,
     Iterable,
     Iterator,
     Mapping,
@@ -245,6 +246,51 @@ class _BaseAgentRunAsync(_WithTracer):
 
 
 class _TraceCallLlm(_WithTracer):
+    """Traces ADK's ``trace_call_llm``, which is called once per streamed chunk.
+
+    ADK opens the ``call_llm`` span *outside* its streaming loop and calls
+    ``trace_call_llm`` from inside it, so this wrapper runs once per chunk against the same
+    span while ``llm_request`` never changes. Only the response-side attributes differ
+    between those calls; re-deriving the request-side ones just overwrites them with the
+    same values.
+
+    ``_request_attributes_written`` remembers, per span, which request has already been
+    written. Notes on that bookkeeping:
+
+    * The span's own attributes cannot serve as the marker. They are bounded
+      (``SpanLimits.max_attributes``, 128 by default) and evict oldest-first, so a long
+      enough message history drops ``INPUT_VALUE`` within the first chunk, which would
+      silently disable this guard for exactly the requests that cost the most to re-derive.
+    * The remembered value is the request's identity, so a span that ever serves a second,
+      different request still gets that request's attributes written.
+    * The record is bounded and least-recently-used first. Losing an entry only costs one
+      redundant re-derivation for that span; it can never leave an attribute unwritten.
+    * Every access is a single dict operation, so concurrent callers can at worst duplicate
+      work -- never drop it.
+    """
+
+    _MAX_REMEMBERED_SPANS = 1024
+
+    def __init__(self, tracer: trace_api.Tracer, *args: Any, **kwargs: Any) -> None:
+        super().__init__(tracer, *args, **kwargs)
+        self._request_written_for_span: Dict[int, int] = {}
+
+    def _request_attributes_written(self, span: Any, llm_request: LlmRequest) -> bool:
+        remembered = self._request_written_for_span
+        span_id = span.get_span_context().span_id
+        if remembered.get(span_id) != id(llm_request):
+            return False
+        remembered[span_id] = remembered.pop(span_id)  # refresh recency
+        return True
+
+    def _remember_request_attributes(self, span: Any, llm_request: LlmRequest) -> None:
+        remembered = self._request_written_for_span
+        span_id = span.get_span_context().span_id
+        remembered.pop(span_id, None)
+        remembered[span_id] = id(llm_request)
+        while len(remembered) > self._MAX_REMEMBERED_SPANS:
+            del remembered[next(iter(remembered))]
+
     @wrapt.decorator  # type: ignore[misc,attr-defined,unused-ignore]
     def __call__(
         self,
@@ -262,13 +308,15 @@ class _TraceCallLlm(_WithTracer):
             SpanAttributes.OPENINFERENCE_SPAN_KIND,
             OpenInferenceSpanKindValues.LLM.value,
         )
+        if not span.is_recording():
+            return ans
         arguments = bind_args_kwargs(wrapped, *args, **kwargs)
         llm_request = next((arg for arg in arguments.values() if isinstance(arg, LlmRequest)), None)
         llm_response = next(
             (arg for arg in arguments.values() if isinstance(arg, LlmResponse)), None
         )
         input_messages_index = 0
-        if llm_request:
+        if llm_request and not self._request_attributes_written(span, llm_request):
             span.set_attribute(
                 SpanAttributes.LLM_PROVIDER,
                 OpenInferenceLLMProviderValues.GOOGLE.value,
@@ -333,6 +381,10 @@ class _TraceCallLlm(_WithTracer):
                         message_index=i,
                     ):
                         span.set_attribute(k, v)
+
+            # Only once the whole block above has succeeded: a partial pass must be
+            # retried on the next chunk rather than suppressed for the rest of the span.
+            self._remember_request_attributes(span, llm_request)
         if llm_response:
             for k, v in _get_attributes_from_llm_response(llm_response):
                 span.set_attribute(k, v)
--- a/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
+++ b/python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
@@ -246,27 +246,10 @@ class _BaseAgentRunAsync(_WithTracer):
 
 
 class _TraceCallLlm(_WithTracer):
-    """Traces ADK's ``trace_call_llm``, which is called once per streamed chunk.
+    """Traces ADK's ``trace_call_llm``, which runs once per streamed chunk.
 
-    ADK opens the ``call_llm`` span *outside* its streaming loop and calls
-    ``trace_call_llm`` from inside it, so this wrapper runs once per chunk against the same
-    span while ``llm_request`` never changes. Only the response-side attributes differ
-    between those calls; re-deriving the request-side ones just overwrites them with the
-    same values.
-
-    ``_request_attributes_written`` remembers, per span, which request has already been
-    written. Notes on that bookkeeping:
-
-    * The span's own attributes cannot serve as the marker. They are bounded
-      (``SpanLimits.max_attributes``, 128 by default) and evict oldest-first, so a long
-      enough message history drops ``INPUT_VALUE`` within the first chunk, which would
-      silently disable this guard for exactly the requests that cost the most to re-derive.
-    * The remembered value is the request's identity, so a span that ever serves a second,
-      different request still gets that request's attributes written.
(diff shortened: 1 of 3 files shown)


KCNyu:

KCNyu:
Pushed a one-line fix for the two red `google_adk-latest` envs.

`google.adk.tools` became a lazy `__getattr__` module in adk 2.x: the real import sits under `if TYPE_CHECKING:` and `__all__` is built at runtime, so mypy's no-implicit-reexport rejects `from google.adk.tools import FunctionTool`. The pinned envs stay green because `test-requirements.txt` pins `google-adk==1.2.1`, which still re-exports it explicitly.

Now imported from `google.adk.tools.function_tool`, matching how the existing tests import `agent_tool` / `tool_context`.

Verified locally against `google-adk==2.7.1` (what `-latest` resolves to in CI): `mypy .` clean, `pytest .` 36 passed. Worth noting the `-latest` envs never got as far as pytest before, since `mypy` runs ahead of it in the tox `ci` command list.

ialisaleh (commented):
I reviewed this PR and the approach looks good. It avoids redundant request-side attribute derivation while keeping response-side attributes updated for each streamed chunk and skipping unnecessary work for non-recording spans.

The bookkeeping is also well scoped, as the mapping is keyed by `span_id` and tracks the `LlmRequest` object identity, ensuring separate invocations are handled independently. The bounded cache also prevents the bookkeeping from growing indefinitely.

The test coverage handles the key scenarios, including multiple streaming chunks, non-recording spans, attribute limits, multiple requests on the same span, and retrying after a failed request-side extraction. Thanks!

image

python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
```diff
@@ -245,6 +246,38 @@ async def __aiter__(self) -> Any:
...
+
+    def _remember_request_attributes(self, span: Any, llm_request: LlmRequest) -> None:
+        remembered = self._request_written_for_span
+        span_id = span.get_span_context().span_id
+
+        remembered.pop(span_id, None)
+        remembered[span_id] = id(llm_request)
+
+        while len(remembered) > self._MAX_REMEMBERED_SPANS:
+            del remembered[next(iter(remembered))]
```
caroger: This dictionary belongs to the single `_TraceCallLlm` wrapper installed for the process, so calls from multiple threads mutate it concurrently. Eviction is a multi-step operation. Two callers can select the same oldest key, one can delete it, and the other can then raise `KeyError`. Concurrent mutation during `next(iter(remembered))` can also raise `RuntimeError`. These exceptions are not caught, so instrumentation can fail an otherwise successful ADK call. Please protect lookup and promotion, plus insertion and eviction, with a lock. The lock does not need to cover attribute derivation. Please also add a regression test that fills the cache and inserts new spans concurrently.

python/instrumentation/openinference-instrumentation-google-adk/src/openinference/instrumentation/google_adk/_wrappers.py
```diff
@@ -333,6 +368,9 @@ def __call__(
                         message_index=i,
                     ):
                         span.set_attribute(k, v)
+
+            # Mark only after all request attributes are successfully derived.
+            self._remember_request_attributes(span, llm_request)
```
caroger: This marks the request complete even when production attribute extraction failed. Helpers such as `_get_attributes_from_base_tool` use `@stop_on_exception`, and input serialization also catches exceptions. Those failures are logged and swallowed, so execution reaches this line and later chunks never retry. The added retry test replaces the decorated helper with an undecorated function that raises, so it does not cover this behavior. Please make request extraction report whether it completed successfully and remember the request only on success. A regression test can use the real decorated path with a tool whose `_get_declaration()` fails once and succeeds on the next chunk, then assert that the tool schema is eventually written.

caroger (changes requested):
Requesting changes for two correctness issues in the request-attribute cache. The shared LRU can raise during concurrent access, and production extraction failures are marked complete instead of retried. Both can be fixed locally without changing the overall approach.

KCNyu:
Both addressed in 592da6f.

The record is behind a lock now, covering the lookup and its promotion and the insert and its eviction, but not attribute derivation — holding it there would serialize every concurrent stream, so there's a test asserting it isn't held while a tool declaration is being read. I also swapped `id()` for a `weakref.ref`, so a new request can't inherit a freed one's address and be mistaken for it.

Derivation now reports whether it finished, and the request is remembered only on a clean pass. The extractors swallow, so that goes through `_extraction_errors`, a ContextVar that stays `None` for every other caller of `stop_on_exception`. The retry test uses the real decorated path now — a tool whose `_get_declaration()` fails once and succeeds on the next chunk.

One caveat on the concurrency test: the threaded version passes on the unfixed code, the window is too narrow to hit reliably. It's in there as you asked, but what actually caught both races was driving an eviction between the steps deterministically — that reproduced the `KeyError` at both lines you flagged.

caroger (approved):

PR #1350, opened 2025-03-12
feat: openai-agents instrumentation

axiomofjoy:

PR #3593, opened 2026-08-21
feat(openai-agents): capture input images in message content attributes

elenaperezrioja:
# Description

`_get_attributes_from_message_content_list` handled `input_text`, `output_text` and `refusal`, but `input_image` fell into an empty `# TODO` branch — untouched since the instrumentation landed in #1350. The effect is that a multimodal turn traces only its text parts: an agent handed a screenshot produces a span that gives no hint an image was ever in the message.

This fills that branch in, emitting the same two attributes the OpenAI Responses instrumentation already emits for `ResponseInputImageParam` ([`_responses_api.py#L349-L359`](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_attributes/_responses_api.py#L349-L359)):

```
llm.input_messages.0.message.contents.1.message_content.type       = "image"
llm.input_messages.0.message.contents.1.message_content.image.image.url = "<url or data uri>"
```

An `input_image` may carry a `file_id` and no `image_url`, so the URL attribute is only emitted when a URL is actually present — those items yield nothing rather than an empty attribute.

`input_file` and `input_audio` are left as they are; this change is scoped to images.

## Tests

Four cases added to the existing `test_get_attributes_from_message_content_list` parametrization: http URL, base64 data URI, `file_id`-only (expects no attributes), and text-plus-image (checks content indices).

`tox run -e py310-ci-openai_agents` passes locally — `ruff format --diff`, `ruff check --no-fix`, `mypy`, and 219 tests.

# Checklist:

- [x] Follows OpenInference configuration to hide sensitive info — the emitted key is the one `TraceConfig` already masks: `hide_input_images` drops it, and a base64 data URI longer than `base64_image_max_length` is redacted or handed to the `blob_uploader`. Verified against both settings.
- [x] Spans properly inherit from [context attributes](https://github.com/Arize-ai/openinference/blob/main/python/openinference-instrumentation/src/openinference/instrumentation/context_attributes.py) — unchanged; this adds attributes to spans the processor already creates.
- [x] Properly respects suppress tracing context — unchanged, same reason.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
@@ -55,6 +55,7 @@ from typing_extensions import assert_never
 from openinference.instrumentation import infer_llm_provider_from_host, safe_json_dumps
 from openinference.instrumentation.openai_agents._tool_schemas import get_tool_schema
 from openinference.semconv.trace import (
+    ImageAttributes,
     MessageAttributes,
     MessageContentAttributes,
     OpenInferenceLLMSystemValues,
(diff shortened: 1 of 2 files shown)


elenaperezrioja:

mikeldking (approved):
Reviewed the OpenAI Agents SDK input-image type and the corresponding OpenInference semantic conventions. The implementation matches the existing Responses API extraction behavior, and the pinned/latest Python CI matrix passes.

PR #3596, opened 2026-08-21
feat(openlit): Add Finish Reason Attribute

ialisaleh:
Refs #2901 

This PR adds `finish_reason` tracking to the OpenLIT instrumentation by extracting it from the `GEN_AI_RESPONSE_FINISH_REASONS` attribute and mapping it to the `LLM_FINISH_REASON` span attribute, keeping it consistent with other instrumentations.

image


python/instrumentation/openinference-instrumentation-openlit/src/openinference/instrumentation/openlit/_span_processor.py
```diff
@@ -278,8 +278,11 @@ def _get_llm_attributes(span: ReadableSpan) -> dict[str, Any]:
 
     invocation_params = find_invocation_parameters(attrs)
     provider_val, system_val = _extract_llm_provider_and_system(attrs)
+    response_finish_reasons = attrs.get("gen_ai.response.finish_reasons")
+    finish_reason = response_finish_reasons[0] if response_finish_reasons else "stop"
```
mikeldking: Let's not ad it if there is no finish reason in the attrs

mikeldking (changes requested):

mikeldking (approved):

PR #3588, opened 2026-08-20
chore: release main

mikeldking:
:robot: I have created a release *beep* *boop*
---

<details><summary>python-openinference-semantic-conventions: 0.1.33</summary>

### Features

* **semconv:** add llm.input_model_name and llm.output_model_name ([#3585](https://github.com/Arize-ai/openinference/issues/3585)) ([1fe497f](https://github.com/Arize-ai/openinference/commit/1fe497f1d9f45a07eee55d97fe185e020560f9c7))
</details>

<details><summary>python-openinference-instrumentation-agentspec: 0.1.6</summary>

### Documentation

* point Arize AX links at the product page with UTM parameters ([#3587](https://github.com/Arize-ai/openinference/issues/3587)) ([cae8ec9](https://github.com/Arize-ai/openinference/commit/cae8ec9615af214359d98cb552d841986a9f02e8))
</details>

<details><summary>python-openinference-instrumentation-agno: 1.0.4</summary>

### Documentation

* point Arize AX links at the product page with UTM parameters ([#3587](https://github.com/Arize-ai/openinference/issues/3587)) ([cae8ec9](https://github.com/Arize-ai/openinference/commit/cae8ec9615af214359d98cb552d841986a9f02e8))
</details>

<details><summary>python-openinference-instrumentation-mistralai: 2.1.1</summary>

### Documentation

* point Arize AX links at the product page with UTM parameters ([#3587](https://github.com/Arize-ai/openinference/issues/3587)) ([cae8ec9](https://github.com/Arize-ai/openinference/commit/cae8ec9615af214359d98cb552d841986a9f02e8))
</details>

<details><summary>python-openinference-instrumentation-openai: 0.1.55</summary>

### Documentation

* point Arize AX links at the product page with UTM parameters ([#3587](https://github.com/Arize-ai/openinference/issues/3587)) ([cae8ec9](https://github.com/Arize-ai/openinference/commit/cae8ec9615af214359d98cb552d841986a9f02e8))
</details>

<details><summary>python-openinference-instrumentation-openai-agents: 2.0.1</summary>

### Documentation

* point Arize AX links at the product page with UTM parameters ([#3587](https://github.com/Arize-ai/openinference/issues/3587)) ([cae8ec9](https://github.com/Arize-ai/openinference/commit/cae8ec9615af214359d98cb552d841986a9f02e8))
</details>

<details><summary>python-openinference-instrumentation-vertexai: 0.1.19</summary>

### Documentation

[...]

PR #3600, opened 2026-08-24
feat(anthropic): capture request and response model names

### LINKED ISSUE #3439: [feature request] Capture requested and effective Anthropic models after classifier routing

satyadevai:
Resolves #3439

## What

Captures the requested model and the serving model as separate span attributes in the Anthropic instrumentation:

- **`llm.request.model_name`** — the model string the caller sent in the request (from the `model` kwarg).
- **`llm.response.model_name`** — the model the provider reports as having generated the response (from `message.model` / the accumulated stream snapshot).
- **`llm.model_name`** now reflects the model that *actually served* the response: it is set from the request at span start and overwritten with the response model once known, per the spec's precedence rule.

## Why

These can legitimately differ. Anthropic's classifier-triggered fallback routing can serve a request with a different model than the one requested (e.g. a request to `claude-fable-5` / `claude-opus-5` answered by `claude-opus-4-8`). With a single `llm.model_name` there was no way to tell from a trace that routing occurred, which breaks per-model cost/latency attribution. See the semconv addition in #3585 and `spec/semantic_conventions.md` ("Request vs. response model names").

## Code paths covered

- `messages.create` (sync + async) via `_get_llm_model_name_from_input` / `_get_llm_model_name_from_response` (`_wrappers.py`)
- All streaming paths — `messages.create(stream=True)`, `messages.stream()`, and their beta counterparts — via the shared `_MessageExtractor` (`_stream.py`), which previously never read the response model at all
- `messages.parse` and `beta.messages` create/parse (share the wrappers above)

## Tests

- Two new fallback tests with hand-built cassettes where the request model (`claude-opus-5`) differs from the response model (`claude-opus-4-8`), non-streaming and streaming, asserting the two attributes diverge and `llm.model_name` equals the serving model.
- All existing exhaustive-pop tests updated to assert both new attributes (sync/async, streaming, tool calling, images, parse, beta).

## Dependencies

- Bumps `openinference-semantic-conventions` to `>=0.1.33` (first release containing `LLM_REQUEST_MODEL_NAME` / `LLM_RESPONSE_MODEL_NAME`; released in #3588).

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
@@ -505,6 +505,8 @@ def test_anthropic_instrumentation_messages_model_fallback(
     client = Anthropic(api_key="sk-ant-fake")
     input_message = "What's the capital of France?"
 
+    invocation_params = {"max_tokens": 1024}
+
     client.messages.create(
         max_tokens=1024,
         messages=[
@@ -517,11 +519,43 @@ def test_anthropic_instrumentation_messages_model_fallback(
     )
 
     spans = in_memory_span_exporter.get_finished_spans()
+
+    assert spans[0].name == "messages.create"
     attributes = dict(spans[0].attributes or {})
+
+    assert attributes.pop(OPENINFERENCE_SPAN_KIND) == "LLM"
+    assert attributes.pop(LLM_PROVIDER) == LLM_PROVIDER_ANTHROPIC
+    assert attributes.pop(LLM_SYSTEM) == LLM_SYSTEM_ANTHROPIC
+    assert attributes.pop(f"{LLM_INPUT_MESSAGES}.0.{MESSAGE_CONTENT}") == input_message
+    assert attributes.pop(f"{LLM_INPUT_MESSAGES}.0.{MESSAGE_ROLE}") == "user"
+    assert (
+        attributes.pop(f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_CONTENTS}.0.{MESSAGE_CONTENT_TYPE}")
+        == "text"
+    )
+    assert (
+        attributes.pop(f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_CONTENTS}.0.{MESSAGE_CONTENT_TEXT}")
+        == "The capital of France is **Paris**."
+    )
+    assert attributes.pop(f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_ROLE}") == "assistant"
+    assert attributes.pop(LLM_TOKEN_COUNT_PROMPT) == 14
+    assert attributes.pop(LLM_TOKEN_COUNT_COMPLETION) == 11
+    assert attributes.pop(LLM_TOKEN_COUNT_TOTAL) == 25
+
+    assert isinstance(attributes.pop(INPUT_VALUE), str)
+    assert attributes.pop(INPUT_MIME_TYPE) == JSON
+    output_value = attributes.pop(OUTPUT_VALUE)
+    assert isinstance(output_value, str)
+    assert_output_value_contains(output_value, {"model": "claude-opus-4-8"})
+    assert attributes.pop(OUTPUT_MIME_TYPE) == JSON
+
+    assert attributes.pop(LLM_FINISH_REASON, None) == "end_turn"
+    assert isinstance(inv_params := attributes.pop(LLM_INVOCATION_PARAMETERS), str)
+    assert json.loads(inv_params) == invocation_params
 
     assert attributes.pop(LLM_REQUEST_MODEL_NAME) == "claude-opus-5"
     assert attributes.pop(LLM_RESPONSE_MODEL_NAME) == "claude-opus-4-8"
     assert attributes.pop(LLM_MODEL_NAME) == "claude-opus-4-8"
+    assert not attributes
 
 
 @pytest.mark.vcr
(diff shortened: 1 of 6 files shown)


python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
```diff
@@ -479,9 +485,82 @@ def test_anthropic_instrumentation_messages_streaming(
...
+        ],
+        model="claude-opus-5",
+    )
+
+    spans = in_memory_span_exporter.get_finished_spans()
+    attributes = dict(spans[0].attributes or {})
+
+    assert attributes.pop(LLM_REQUEST_MODEL_NAME) == "claude-opus-5"
+    assert attributes.pop(LLM_RESPONSE_MODEL_NAME) == "claude-opus-4-8"
+    assert attributes.pop(LLM_MODEL_NAME) == "claude-opus-4-8"
```
mikeldking, line 557: The fallback assertions themselves are exactly right (request ≠ response model, and `llm.model_name` reflecting the model that actually served). But this test stops after popping 3 attributes, while every other test in this file follows the exhaustive pop convention ending in `assert not attributes`. Without it, an unexpected attribute regression on this code path slips through silently. Suggest popping the remaining attributes (`OPENINFERENCE_SPAN_KIND`, `LLM_PROVIDER`, `LLM_SYSTEM`, input/output value + mime types, input/output messages, `LLM_INVOCATION_PARAMETERS`, `LLM_FINISH_REASON`, token counts) and closing with `assert not attributes`.

python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
```diff
@@ -479,9 +485,82 @@ def test_anthropic_instrumentation_messages_streaming(
...
+
+    for _ in stream:
+        pass
+
+    spans = in_memory_span_exporter.get_finished_spans()
+    attributes = dict(spans[0].attributes or {})
+
+    assert attributes.pop(LLM_REQUEST_MODEL_NAME) == "claude-opus-5"
+    assert attributes.pop(LLM_RESPONSE_MODEL_NAME) == "claude-opus-4-8"
+    assert attributes.pop(LLM_MODEL_NAME) == "claude-opus-4-8"
```
mikeldking, line 628: Same as the non-streaming fallback test above: finish the pops and end with `assert not attributes` to match the file convention. (No async fallback variant is strictly needed — `_MessagesStream.__aiter__` and `_MessageExtractor` are shared with the sync path, and the async streaming tests already assert both model-name attributes — but worth a comment here noting that if you want it explicit.)

python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
mikeldking: Addressed in 0ce3b369 — test now pops all attributes and ends with `assert not attributes`.

python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py
mikeldking: Addressed in 0ce3b369 — same exhaustive treatment applied to the streaming fallback test.

mikeldking (approved):

PR #3611, opened 2026-08-25
fix(claude-agent-sdk): fold cache tokens into prompt and total counts

### LINKED ISSUE #3610: claude-agent-sdk instrumentor undercounts prompt/total tokens for cache-heavy sessions
**Describe the bug**

`_extract_usage_and_cost_attributes` in the Claude Agent SDK instrumentor sets
`llm.token_count.prompt` and `llm.token_count.total` from `usage.input_tokens`
alone, excluding `cache_read_input_tokens` and `cache_creation_input_tokens`.
The Anthropic Messages API's `input_tokens` field excludes cache tokens by
design (https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching),
so for any cache-heavy agent session the reported prompt and total token
counts are far below the real prompt size. The sibling
`openinference-instrumentation-anthropic` package already folds these fields
back into `llm.token_count.prompt` for the same API shape
(`_utils.py::_get_token_counts`); the Claude Agent SDK wrapper never does
that addition, it only reports the cache counts as separate
`*_DETAILS_CACHE_READ` / `*_DETAILS_CACHE_WRITE` attributes.

**To Reproduce**

Using the usage payload recorded in this package's own real-API cassette
(`tests/cassettes/test_instrumentor/test_query_real_agent_span.yaml`):

```python
from openinference.instrumentation.claude_agent_sdk._wrappers import (
    _extract_usage_and_cost_attributes,
)

msg = {
    "usage": {
        "input_tokens": 3,
        "output_tokens": 4,
        "cache_read_input_tokens": 17024,
        "cache_creation_input_tokens": 0,
    }
}
attrs = _extract_usage_and_cost_attributes(msg)
print(attrs["llm.token_count.prompt"])  # 3
print(attrs["llm.token_count.total"])   # 7
```

**Expected behavior**

`llm.token_count.prompt` should be `3 + 17024 + 0 = 17027` and
`llm.token_count.total` should be `17031`, matching how
`openinference-instrumentation-anthropic` reports the same usage shape.

**Additional context**

The package's own test suite does not catch this: the one test exercising a
cache-bearing cassette (`test_query_real_agent_span`) only asserts
`isinstance(prompt_tokens, int)` and `total_tokens == prompt_tokens +
completion_tokens`, which holds even with the bug. I have a fix and a
regression test ready and will open a PR referencing this issue.

AmirF194:
fixes #3610

**Root cause**

`_extract_usage_and_cost_attributes` (`_wrappers.py`) set
`llm.token_count.prompt`/`llm.token_count.total` from `usage.input_tokens`
alone. `input_tokens` excludes prompt-cache tokens by design, so
`cache_read_input_tokens` and `cache_creation_input_tokens` were dropped from
both counts, leaving only the separate `*_DETAILS_CACHE_READ`/`_CACHE_WRITE`
attributes to record them.

**Invariant**

`llm.token_count.prompt` must equal the true prompt size (all tokens the
model read, cached or not), and `llm.token_count.total` must equal
`prompt + completion`. The sibling `openinference-instrumentation-anthropic`
package already holds this for the same usage shape
(`_utils.py::_get_token_counts`).

**Fix**

Fold `cache_read_input_tokens` and `cache_creation_input_tokens` into
`llm.token_count.prompt` (and therefore `.total`), matching the anthropic
instrumentor's convention. The separate cache-detail attributes are
unchanged.

**Review follow-ups (second commit)**

- Compute `llm.token_count.prompt` whenever any component (input, cache
  read, cache write) is present, and emit `total` when prompt or output is
  known — a cache-only payload no longer emits cache details with no prompt
  count, and a missing `output_tokens` no longer drops `total`.
- The `cache_creation_input_tokens` fallback now runs on parsed values, so a
  present-but-unparseable `cache_write_input_tokens` can no longer suppress
  it and silently drop the cache-write contribution.
- Integration tests pin exact cassette token values (prompt 17027,
  cache_read 17024, total 17031) instead of the tautological
  `total == prompt + completion` assertion.
- The ad-hoc unit-test file was folded into `tests/test_wrappers_unit.py`
  as a parametrized test with cache-only, missing-output, and
  unparseable-cache-write regression rows.
- `spec/semantic_conventions.md` now states that
  `llm.token_count.prompt_details.*` are sub-counts included in
  `llm.token_count.prompt`, so consumers do not double-bill cached tokens.

**Verified**

- Unit test fails on `main` (asserts 3 instead of 17027 on the package's
  own real-API cassette values) and passes on this branch.
- Full package suite green after the follow-ups: 48/48 passing on a fresh
  build; `ruff format` clean.
- This changes attribute values only, no new attributes, no renamed fields,
  no span hierarchy change, so no trace screenshot is included per
  CONTRIBUTING.
- Not verified: behavior against a live cache-write (as opposed to
  cache-read) session; the fix treats both the same way the anthropic
  instrumentor does, but I only had a cache-read cassette to test against.


AmirF194:

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -281,12 +281,16 @@ def _extract_usage_and_cost_attributes(msg: Any) -> dict[str, Any]:
...
+    # openinference-instrumentation-anthropic package's _get_token_counts convention.
     if input_tokens is not None:
-        attributes[LLM_TOKEN_COUNT_PROMPT] = input_tokens
+        prompt_tokens = input_tokens + (cache_read_tokens or 0) + (cache_write_tokens or 0)
+        attributes[LLM_TOKEN_COUNT_PROMPT] = prompt_tokens
     if output_tokens is not None:
         attributes[LLM_TOKEN_COUNT_COMPLETION] = output_tokens
     if input_tokens is not None and output_tokens is not None:
-        attributes[LLM_TOKEN_COUNT_TOTAL] = input_tokens + output_tokens
+        attributes[LLM_TOKEN_COUNT_TOTAL] = prompt_tokens + output_tokens
```
mikeldking: **Partial usage payloads drop prompt/total, and `prompt_tokens` is conditionally bound.**

Two related issues with the guard structure:

1. When `input_tokens` is missing but cache tokens are present, `LLM_TOKEN_COUNT_PROMPT`/`TOTAL` are never set even though the span still emits `prompt_details.cache_read` — an internally contradictory span (details with no parent count). When `output_tokens` is missing (error/aborted results feed this same helper), TOTAL is dropped even though prompt is known. The sibling `openinference-instrumentation-anthropic` `_get_token_counts` computes `prompt_tokens` unconditionally from the sum and derives `total = prompt + (output or 0)`.

2. `prompt_tokens` is bound only inside the `if input_tokens is not None:` branch and read here under a *separately spelled* guard. Safe today only because the predicates coincide; any future edit to either guard becomes an `UnboundLocalError` inside the instrumented message loop (no try/except between here and user code).

Suggested shape:

```python
prompt_tokens = None
if input_tokens is not None or cache_read_tokens is not None or cache_write_tokens is not None:
    prompt_tokens = (input_tokens or 0) + (cache_read_tokens or 0) + (cache_write_tokens or 0)
    attributes[LLM_TOKEN_COUNT_PROMPT] = prompt_tokens
if output_tokens is not None:
    attributes[LLM_TOKEN_COUNT_COMPLETION] = output_tokens
if prompt_tokens is not None or output_tokens is not None:
    attributes[LLM_TOKEN_COUNT_TOTAL] = (prompt_tokens or 0) + (output_tokens or 0)
```

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -281,12 +281,16 @@ def _extract_usage_and_cost_attributes(msg: Any) -> dict[str, Any]:
         if usage.get("cache_write_input_tokens") is not None
         else usage.get("cache_creation_input_tokens")
     )
+    # Anthropic's input_tokens excludes cache tokens by design; fold cache_read/write
+    # back in so prompt/total reflect the true prompt size, matching the sibling
+    # openinference-instrumentation-anthropic package's _get_token_counts convention.
     if input_tokens is not None:
-        attributes[LLM_TOKEN_COUNT_PROMPT] = input_tokens
+        prompt_tokens = input_tokens + (cache_read_tokens or 0) + (cache_write_tokens or 0)
```
mikeldking: **Heads-up: this fold-in is applied to a cross-turn aggregate, so `llm.token_count.prompt` can exceed any real prompt.**

The only callers of this helper are the ResultMessage paths, and ResultMessage `usage` sums across every turn of the run. Verified from the committed cassette `tests/cassettes/test_instrumentor/test_query_tool_fallback_when_hooks_unavailable.yaml`: per-turn assistant `cache_read` values are 17080 and 17223, and the result message reports `cache_read_input_tokens: 34303` (= their sum) with `input_tokens: 4`. Post-change the single AGENT span reports prompt = 34307 for a run whose largest actual prompt was ~17k; a ~12-turn run with a 17k cached prefix reports a "prompt" above the 200k context window. Subagent cache reads also roll into the parent aggregate while the child Task span carries none (see `test_query_task_subagent_spans`), so trace-level rollups double-attribute them — and the fold multiplies that by the cache-read factor.

This matches the billing-shaped convention of the sibling anthropic package (and the SDK's own `total_cost_usd`), so it is arguably the right call — but worth stating explicitly in the comment/PR description that prompt on these AGENT spans is cumulative billed input tokens, not context size.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -281,12 +281,16 @@ def _extract_usage_and_cost_attributes(msg: Any) -> dict[str, Any]:
         if usage.get("cache_write_input_tokens") is not None
         else usage.get("cache_creation_input_tokens")
```
mikeldking: **Pre-existing, but now load-bearing for prompt/total: the coalesce inspects the raw value before `_safe_int`.**

`usage.get("cache_write_input_tokens") is not None` short-circuits the `cache_creation_input_tokens` fallback even when the first key is present but unparseable (`_safe_int` then returns None and the cache-write contribution silently drops out of prompt/total — the same undercount this PR fixes). Coalesce on the parsed values instead:

```python
cache_write_tokens = _safe_int(usage.get("cache_write_input_tokens"))
if cache_write_tokens is None:
    cache_write_tokens = _safe_int(usage.get("cache_creation_input_tokens"))
```

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
```diff
@@ -281,12 +281,16 @@ def _extract_usage_and_cost_attributes(msg: Any) -> dict[str, Any]:
...
     )
+    # Anthropic's input_tokens excludes cache tokens by design; fold cache_read/write
```
mikeldking: **Nit (reuse/altitude):** this is now the fourth hand-rolled copy of the Anthropic cache fold-in formula (`anthropic/_utils.py:_get_token_counts`, `bedrock/utils/anthropic/_attributes.py`, `langchain/_tracer.py` carry the others), and the copies have already drifted (anthropic/langchain emit the cache breakouts only when truthy; this one also on 0). The core package already owns `TokenCount`/`PromptDetails` and `get_llm_token_count_attributes` — a single Anthropic-usage→TokenCount adapter there would keep the next cache bucket (e.g. the 1h/5m `cache_creation` split already visible in this package's cassettes) from having to be fixed in four places. Fine to leave for a follow-up.

mikeldking (approved):
Appreciate this catch

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
AmirF194: Fixed in the follow-up commit: prompt is now computed whenever any component (input, cache read, cache write) is present, and total is emitted whenever prompt or output is known, using your suggested shape. Thanks for catching the UnboundLocalError risk.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
AmirF194: Also fixed in the follow-up commit: cache_write_tokens now coalesces on the parsed value, so an unparseable cache_write_input_tokens no longer silently suppresses the cache_creation_input_tokens fallback.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
AmirF194: Agreed on the duplication, leaving the shared TokenCount adapter for a follow-up as you suggested rather than widening this PR's scope.

python/instrumentation/openinference-instrumentation-claude-agent-sdk/src/openinference/instrumentation/claude_agent_sdk/_wrappers.py
AmirF194: Agreed, this is the billing-shaped convention on purpose (matches the anthropic instrumentor and total_cost_usd). Would rather not push a comment-only change now and risk dismissing your approval on classic branch protection, so noting it here plainly: prompt on these AGENT spans is cumulative billed input across every turn, not a single context size.

AmirF194:
Thanks for the quick review and merge.

PR #3629, opened 2026-08-28
fix(anthropic): handle beta messages create streams

satyadevai:
## Motivation

`client.beta.messages.create(stream=True)` yields beta stream event types, but the instrumentation accumulated them with Anthropic's stable Messages accumulator. Beta-only updates such as server-side fallback blocks and `usage.iterations` were therefore lost or incompletely represented in the finished span.

## Behavior

- Use Anthropic's beta accumulator for beta `messages.create` streams and retain the stable accumulator for standard streams.
- Preserve the original request headers required by beta accumulation, including fine-grained tool-streaming behavior.
- Finish sync and async beta stream spans with the complete beta response snapshot.
- Keep `llm.model_name` and `llm.response.model_name` aligned with the top-level model reported by Anthropic.
- Preserve existing tracing suppression, context propagation, TraceConfig masking, and stable Messages behavior because span creation/finalization remains on the existing wrapper and `OITracer` paths.

## Compatibility and risk

The change is internal to streaming response accumulation and does not change the public instrumentation API. Stable Messages streams continue to use the stable SDK accumulator. Beta selection is made from the already-known wrapped endpoint rather than inferred from generated event class names, and the implementation was checked against Anthropic SDK 1.0.0 and the latest available SDK.

## Tests

- Added a recorded sticky server-side fallback stream where `usage.iterations` identifies the serving fallback model.
- Added deterministic sync and async beta stream tests with a mid-stream fallback content block.
- Added coverage that the beta accumulator receives the original `anthropic-beta` request header.
- Added a stable-stream regression test proving `message_delta` usage continues to override `message_start` usage.

## Validation

- `py314-ci-anthropic`: 53 passed; Ruff formatting/lint and mypy passed.
- `py310-ci-anthropic`: passed.
- `py314-ci-anthropic-latest`: passed.

The test suite emits existing Anthropic model-deprecation and Pydantic serializer warnings; there are no new failures or warnings attributable to this change.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
+++ b/python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_stream.py
@@ -26,6 +26,8 @@ from openinference.semconv.trace import (
 )
 
 if TYPE_CHECKING:
+    from httpx2 import Headers
+
     from anthropic import Stream
     from anthropic.types import RawMessageStreamEvent
 
@@ -105,9 +107,14 @@ class _MessagesStream(ObjectProxy):  # type: ignore[misc,name-defined,type-arg,u
         self,
         stream: "Stream[RawMessageStreamEvent]",
         with_span: _WithSpan,
+        *,
+        is_beta: bool = False,
     ) -> None:
         super().__init__(stream)
-        self._response_accumulator = _MessageResponseAccumulator()
+        self._response_accumulator = _MessageResponseAccumulator(
+            is_beta=is_beta,
+            request_headers=stream.response.request.headers,
+        )
         self._with_span = with_span
 
     def __iter__(self) -> Iterator["RawMessageStreamEvent"]:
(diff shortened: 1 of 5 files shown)


feiiiiii5:
Ran this branch locally in two SDK environments and got a result worth flagging before a maintainer spends time on it — the merge is currently blocked anyway (`mergeable: CONFLICTING`, one file), so this may all be resolved by a rebase.

**What I did** (no API keys; the package's committed cassettes cover everything)

```bash
# isolated venv, package's own test-requirements.txt
uv pip install -r test-requirements.txt          # anthropic==1.0.0
uv pip install -e <repo>/python/openinference-semantic-conventions \
                 -e <repo>/python/openinference-instrumentation -e .
pytest tests -q        # -> 53 passed

uv pip install -U anthropic                      # anthropic 1.7.0
pytest tests -q        # -> 19 failed, 34 passed
```

Control, same interpreter and same 1.7.0, current `main` (`b7373601`):

```
pytest tests -q        # -> 49 passed
```

So on a modern SDK this branch loses streaming attributes that `main` captures. The split is 16 failures in `test_instrumentor.py` and 3 in the new `test_server_side_fallback.py`, and they are assertion mismatches on missing attributes rather than exceptions (no `TypeError` appears anywhere in the run). The shape, e.g.

```
test_finish_reason_values_messages_create_streaming[end_turn]
> assert attributes.get(LLM_FINISH_REASON) == stop_reason
E assert None == 'end_turn'
```

and `test_stable_streaming_message_delta_usage_overrides_message_start` fails in the same family.

**Why I am not claiming your design is at fault.** Your branch is based on `d773c1ac`, which predates #3741 (`8c007a15`, "support anthropic 1.5.0 streaming accumulate_event and parsed_output serialization changes") — the same `_MessageResponseAccumulator` region you edit, which is also the conflict (`git merge-tree` reports exactly one conflicted file: `.../anthropic/_stream.py`). So the most likely explanation is that this branch simply does not carry #3741's fix yet. Confirming that needs the rebase.

**What would make this ready to review again**

1. Rebase onto `main`, resolving `_stream.py` by keeping #3741's `json_bufs` / version-detection path and adding the beta accumulator selection on top of it.
2. Re-run both envs, because three of the 19 are your own new tests and are not explained by #3741 alone: `test_anthropic_instrumentation_beta_messages_streaming_fallback`, `test_beta_messages_streaming_fallback_content_block`, `test_async_beta_messages_streaming_uses_beta_accumulator`. `python/tox.ini` puts this in the default `envlist` as `py3{10,14}-ci-{anthropic,anthropic-latest}`, with `anthropic-latest: uv pip install -U anthropic` — so the newest-SDK run is a CI job, not an optional extra, and it is the one that contradicts "checked against … the latest available SDK" in the description.
3. If the beta path genuinely cannot satisfy both SDK versions with one accumulator, that trade-off is worth a sentence in the description — it is the question a reviewer will ask first.

The direction itself looks right: using the SDK's own beta accumulator for `client.beta.messages.create(stream=True)` instead of reinterpreting beta events through the stable one is the shape the rest of the repo follows, and the new tests are the kind that make this class of bug visible.

Disclosure: written with an AI coding agent (Claude, via Qoder CLI); the two environments, the 53/19/49 numbers and the conflict report are output I ran locally, not an inference from reading the diff.

PR #3630, opened 2026-08-28
feat: add explicit name and description for advertised LLM tools

axiomofjoy:


python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py
```diff
@@ -560,6 +560,25 @@ class ToolAttributes:
...
+    """
+
+    TOOL_DESCRIPTION = "tool.description"
+    """
+    The description of the tool, i.e. the text the model uses to decide whether
+    to call it.
+
+    Like `tool.name`, the description is also carried inside
+    `tool.json_schema` at a provider-dependent location.
+    """
```
axiomofjoy: make these docstrings one line each

spec/semantic_conventions.md
```diff
@@ -124,9 +124,9 @@ The following attributes are reserved and MUST be supported by all OpenInference
...
 | `tool.json_schema`                             | JSON String                 | `"{'type': 'function', 'function': {'name': 'get_weather'}}"`                                                                 | The json schema of a tool input                                                                                                                                                     |
-| `tool.name`                                    | String                      | `"WeatherAPI"`                                                                                                                | The name of the tool being utilized                                                                                                                                                 |
+| `tool.name`                                    | String                      | `"WeatherAPI"`                                                                                                                | The name of the tool. On a `TOOL` span, the tool being invoked; under `llm.tools.<index>`, the name of an advertised tool definition                                                |
```
axiomofjoy, line 129: the changes don't affect attributes on tool spans, do they?

mikeldking: no, they can start at the tool level - that's how it works on tool spans

mikeldking (approved):

python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py
mikeldking: Addressed in 159ef6b6: the new name and description attribute docstrings are now one-line docstrings.

spec/semantic_conventions.md
mikeldking: Correct: the existing top-level tool.name and tool.description semantics on TOOL spans are unchanged. This PR documents that the same wire keys can also be nested under llm.tools.<index> to describe advertised definitions; the updated PR body and spec now call out that compatibility explicitly.

PR #3641, opened 2026-08-29
fix(js): preserve default input capture in split-model example

mikeldking:
## Summary

Follow-up to #3636 and #3638.

The canonical request/response model-name example correctly preserved default
output capture by spreading `defaultProcessOutput`, but its custom
`processInput` replaced the default processor without preserving
`input.value`. Applications copying the example would therefore retain the
model attributes while silently losing the traced request payload.

## Changes

- Spread `defaultProcessInput(request)` in the canonical `@observe` example
  before adding `requestModelName`.
- Update the decorator regression test to assert the exact serialized
  `input.value` and `output.value` alongside all three model attributes.
- Update the `withSpan` regression test to exercise and assert the same
  composition.
- Clarify that both custom input and output processors replace their defaults.

## Compatibility and risk

This changes only documentation and tests; runtime behavior and public types are
unchanged. No changeset is needed. The example now demonstrates the existing
processor-composition contract accurately.

## Validation

- `pnpm run -r build`: passed for all 15 JS workspace packages.
- `pnpm --filter @arizeai/openinference-core test`: 156 passed, 1 skipped;
  Vitest type errors: none.
- `pnpm --filter @arizeai/openinference-core type:check`: passed.
- `pnpm run lint`: passed.
- `pnpm run fmt:check`: passed across 365 files.

### DIFF
--- a/js/packages/openinference-core/docs/attribute-helpers.md
+++ b/js/packages/openinference-core/docs/attribute-helpers.md
@@ -63,12 +63,13 @@ With `withSpan` or the `@observe` decorator, compose them through the tracing
 options. Note that `attributes` is evaluated once when the function is wrapped
 (for a decorator, at class-definition time), so use it only for literal values;
 derive anything per-call from the arguments via `processInput`, and take the
-response model from the result via `processOutput`. A custom `processOutput`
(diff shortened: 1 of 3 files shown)


mikeldking:
@claude review this with scrutiny

PR #3642, opened 2026-08-30
fix(langchain-js): prioritize critical span attrs before message history

### LINKED ISSUE #3527: [bug] LangChain JS instrumentation exceeds OpenTelemetry’s 128 span-attribute limit for long conversations
Describe the bug

  The JavaScript LangChain instrumentation creates an unbounded number of span
  attributes as conversation history grows.

  formatInputMessages() includes every input message, and
  flattenAttributes() converts every message field into indexed attributes:

  - llm.input_messages.0.message.role
  - llm.input_messages.0.message.content
  - llm.input_messages.1.message.role
  - llm.input_messages.1.message.content

  The resulting object is passed to span.setAttributes().

  The OpenTelemetry Node SDK defaults to 128 attributes per span. After that
  limit is reached, subsequent attributes are dropped.

  To Reproduce

  Use LangChain with a conversation containing ordinary role-and-content
  messages and enable:

  @arizeai/openinference-instrumentation-langchain

  Observed results:

  - 60 input messages → 127 total span attributes → 0 dropped
  - 61 input messages → 129 total span attributes → 1 dropped
  - 62 input messages → 131 total span attributes → 3 dropped
  - 64 input messages → 135 total span attributes → 7 dropped
  - 70 input messages → 147 total span attributes → 19 dropped
  
This was reproduced with versions 4.0.6 and 4.0.17.

  Tool calls and tool schemas generate additional attributes, so agent
  conversations can reach the limit earlier.

  Relevant implementation:

  https://github.com/Arize-ai/openinference/blob/main/js/packages/openinference-instrumentation-langchain/src/utils.ts

  Expected behavior

  The instrumentation should produce bounded attribute cardinality while
  preserving important attributes.

  Possible solutions include:

  1. Add a configurable maximum input-history length.
  2. Retain only the current conversation turn instead of the complete history.
  3. Apply an attribute budget before flattening messages.
  4. Record session, model, token-count, and metadata attributes before verbose
     message attributes.

  Additional context

  The entire span is not dropped. Attributes beyond the OpenTelemetry limit are
  dropped.

  Because message attributes are added before token counts, metadata, and
  session.id, important later attributes can be lost.

  Deleting old attributes in an OpenTelemetry Collector does not solve the
  problem because the Node SDK applies the 128-attribute limit before exporting
  the span.

  Environment:

  - OS: macOS
  - @langchain/core: 1.x
  - @arizeai/openinference-instrumentation-langchain: 4.0.6 and 4.0.17
  - OpenTelemetry Node SDK using default span limits

BetterAndBetterII:
Fixes #3527.

OpenTelemetry’s default span attribute limit drops later attributes once the limit is reached. The LangChain JS tracer flattened verbose input/output message (and tool/document) attributes before session, model, token-count, metadata, and IO fields, so long conversations lost the attributes that matter most.

This change only reorders attribute construction so session/model/token-count/metadata and IO are set first. It does not add a public history-cap API.

Regression: low `spanLimits.attributeCountLimit` with a long message history asserts critical attributes survive while later message attributes are dropped.

### DIFF
--- a/js/packages/openinference-instrumentation-langchain/src/tracer.ts
+++ b/js/packages/openinference-instrumentation-langchain/src/tracer.ts
@@ -119,19 +119,21 @@ export class LangChainTracer extends BaseTracer {
       span.setStatus({ code: SpanStatusCode.OK });
     }
 
+    // OpenTelemetry drops later attributes at the span limit; keep
+    // session/model/token-count/metadata and IO ahead of verbose history.
     const attributes = safelyFlattenAttributes({
       ...safelyFormatIO({ io: run.inputs, ioType: "input" }),
       ...safelyFormatIO({ io: run.outputs, ioType: "output" }),
-      ...safelyFormatInputMessages(run.inputs),
-      ...safelyFormatOutputMessages(run.outputs),
-      ...safelyFormatRetrievalDocuments(run),
       ...safelyFormatLLMParams(run.extra),
       ...safelyFormatPromptTemplate(run),
       ...safelyFormatTokenCounts(run.outputs),
-      ...safelyFormatFunctionCalls(run.outputs),
-      ...safelyFormatToolCalls(run),
       ...safelyFormatMetadata(run),
       ...safelyFormatSessionId(run),
+      ...safelyFormatInputMessages(run.inputs),
+      ...safelyFormatOutputMessages(run.outputs),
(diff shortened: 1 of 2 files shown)


BetterAndBetterII:

mikeldking (approved):

Issue #3422, opened 2026-07-23
[FEATURE]: Span-kind-independent `input.image.url` / `output.image.url` semantic conventions


mikeldking:
## Problem

Today the only way to attach an image to a span in a way that renders in the
Phoenix trace UI is through the multimodal LLM message attributes, e.g.:

```
llm.input_messages.0.message.contents.1.message_content.image.image.url = "..."
```

This requires the span to be `openinference.span.kind = LLM` and to carry the
full message structure. There is **no span-kind-independent convention** for
"this span has an input image" or "this span produced an output image", the way
there already is for text via `input.value` / `output.value`.

As reported in Arize-ai/phoenix#14399, users tracing non-LLM pipelines
(e.g. computer-vision / OCR: `TOOL`, `CHAIN` spans) currently have to *fake*
their spans as `LLM`-kind purely to trigger the image renderer, which pollutes
span-kind filters, cost views, etc.

## Proposed convention

Add two top-level, span-kind-independent attributes mirroring the existing
`input.value` / `output.value` pair:

| Attribute          | Type   | Example                                   | Description                                              |
| ------------------ | ------ | ----------------------------------------- | -------------------------------------------------------- |
| `input.image.url`  | String | `"data:image/png;base64,iVBORw0KGgo..."`  | The input image for the span (URL or base64 data URI).   |
| `output.image.url` | String | `"https://example.com/annotated.jpg"`     | The output image produced by the span (URL or data URI). |

Notes / rationale:

- Names reuse the existing `image.url` sub-attribute already defined for the
  `message_content.image` (Image Object) convention, so the value semantics
  ("link to the image or its base64 encoding") stay identical.
- Parallels the existing span-kind-independent `input.*` / `output.*` family
  (`input.value`, `input.mime_type`, `output.value`, `output.mime_type`),
  keeping the top-level I/O surface consistent across text and image modalities.
- Any span kind (`TOOL`, `CHAIN`, `AGENT`, etc.) could carry these without
  pretending to be an `LLM` span.

### Describe the solution you'd like

- Define `input.image.url` and `output.image.url` in the semantic conventions
  (`spec/semantic_conventions.md`, plus the language constant packages: Python,
  JS, Go).
- Optionally consider plural / multi-image shapes (e.g. `input.images.N.image.url`)
  if attaching more than one image per span is in scope — happy to discuss the
  exact shape.

### Additional context

[...]

PR #3654, opened 2026-09-01
feat(semconv): add span-kind-independent image attributes

satyadevai:
Resolves the part of #3422


go/openinference-semantic-conventions/attributes.go
```diff
@@ -207,9 +213,10 @@ const (
 	MessageContentEncryptedContent = "message_content.encrypted_content"
 )
 
-// Image attributes — nested under MessageContentImage.
+// Image attributes — nested under MessageContentImage, InputImages or OutputImages.
 const (
-	ImageURL = "image.url"
+	ImageURL      = "image.url"
+	ImageMimeType = "image.mime_type"
```
mikeldking: Let's nix this for now. The extension is enough

mikeldking (changes requested):

go/openinference-semantic-conventions/attributes.go
satyadevai: Hi @mikeldking,

I’ve added MIME type support for Base64 images. Tools can return images either as external URLs, such as HTTP or S3 URLs, or directly as Base64 data.

OpenAI may also return generated images as raw Base64 data. In that case, we need the MIME type so the UI can render the image correctly. For external URLs, the MIME type can remain optional.

Please let me know if you think this is unnecessary, and I’ll remove it.

mikeldking: I'd like it removed as in most cases we are going to just look at the extension. Let's remove

satyadevai: Removed mimetype

satyadevai:
Hi @mikeldking ,

I’ve opened a separate PR for the openinference-instrumentation changes to separate the semantic conventions dependency from the OpenInference instrumentation package:

https://github.com/Arize-ai/openinference/pull/3695

Once the semantic conventions changes are merged, I’ll rerun the CI jobs, and this PR should be ready to merge.

caroger (approved):

spec/multimodal_attributes.md
```diff
@@ -100,6 +100,82 @@ Semantics:
...
+input.images.1.image.url = "https://example.com/page-2.png"
+```
+
+### Semantics
+
+- `image.url` carries the same value semantics as `message_content.image.image.url` — a link to the
+  image or its base64 encoding. That is an absolute URI (`https://`, `http://`, `s3://`, `gs://`, or
+  a provider-specific scheme), a base64 data URI (`data:<mime>;base64,<payload>`), or a bare base64
+  payload.
+- There is no separate MIME attribute: the media type is read from the value itself — the `data:`
```
mikeldking: Nix

spec/multimodal_attributes.md
```diff
@@ -100,6 +100,82 @@ Semantics:
...
+  URI prefix, or the file extension of a reference URI. Producers SHOULD therefore emit a data URI
+  in preference to a bare base64 payload, and SHOULD keep a recognizable extension on a reference
+  URI, so that a consumer can identify the bytes.
+- These attributes are additive, not a replacement for message content. An `LLM` span that already
```
mikeldking: I don't understand this statement

spec/multimodal_attributes.md
```diff
@@ -100,6 +100,82 @@ Semantics:
...
+- `OPENINFERENCE_HIDE_INPUT_IMAGES` removes the whole `input.images.*` namespace, not only the
+  `image.url` leaves.
+- `OPENINFERENCE_HIDE_INPUTS` and `OPENINFERENCE_HIDE_OUTPUTS` remove `input.images.*` and
+  `output.images.*` respectively, as they do for `input.value` / `output.value`.
+- `OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH` and a configured blob uploader apply to
+  `<input|output>.images.<i>.image.url` on the same terms as `message_content.image.image.url`
+  (see [External Storage for Large Media](#external-storage-for-large-media)). Hiding takes
+  precedence: a hidden image is never uploaded.
+
+The size limit and externalization recognize base64 by the `data:` prefix, so a bare base64 payload
```
mikeldking: This goes without saying?

spec/multimodal_attributes.md
mikeldking: Removed in cefb931b.

spec/multimodal_attributes.md
mikeldking: Reworded in cefb931b. The point is just: an LLM span already has its images inside the messages, so it should not duplicate them under input.images / output.images.

spec/multimodal_attributes.md
mikeldking: Removed in cefb931b.

mikeldking (approved):

PR #3661, opened 2026-09-02
feat(openai_agents): Add Finish Reason Attribute

ialisaleh:
Refs #2901 

This PR adds `finish_reason` tracking to the OpenAI Agents SDK instrumentation for `ResponseSpanData` and Realtime sessions by mapping each surface's completion status (and, for Responses, the more specific `incomplete_details.reason` when present) through a shared `status` to the `finish_reason` table onto a single `LLM_FINISH_REASON` span attribute, so we can keep the value vocabulary consistent with what the OpenAI Chat Completions instrumentation reports, keeping it consistent with other instrumentations.

image

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
@@ -880,11 +880,15 @@ def _get_attributes_from_reasoning_item(
 def _map_finish_reason(status: Optional[str], incomplete_reason: Optional[str]) -> Optional[str]:
     """
     Map a Responses API status and incomplete reason to a single OpenInference finish reason.
+
+    The more specific ``incomplete_details.reason`` (e.g. ``max_output_tokens``) takes
+    precedence over the coarse ``status`` so that truncation is reported as ``length``
+    rather than ``incomplete``.
     """
-    if status:
-        return _STATUS_TO_FINISH_REASON.get(status, status)
     if incomplete_reason:
         return _INCOMPLETE_REASON_TO_FINISH_REASON.get(incomplete_reason, incomplete_reason)
+    if status:
+        return _STATUS_TO_FINISH_REASON.get(status, status)
     return None
 
 
--- a/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
@@ -68,6 +68,20 @@ from openinference.semconv.trace import (
 
 logger = logging.getLogger(__name__)
 
+# Responses API status to OpenInference finish reason mapping.
+_STATUS_TO_FINISH_REASON = {
+    "completed": "stop",
+    "failed": "error",
+    "cancelled": "cancelled",
+    "incomplete": "incomplete",
+}
+
+# Responses API incomplete reason to OpenInference finish reason mapping.
+_INCOMPLETE_REASON_TO_FINISH_REASON = {
+    "max_output_tokens": "length",
+    "content_filter": "content_filter",
+}
+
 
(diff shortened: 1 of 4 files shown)


python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
```diff
@@ -858,6 +877,17 @@ def _get_attributes_from_reasoning_item(
         yield f"{content_prefix}{MESSAGE_CONTENT_ID}", obj.id
 
 
+def _map_finish_reason(status: Optional[str], incomplete_reason: Optional[str]) -> Optional[str]:
+    """
+    Map a Responses API status and incomplete reason to a single OpenInference finish reason.
+    """
+    if status:
+        return _STATUS_TO_FINISH_REASON.get(status, status)
+    if incomplete_reason:
```
mikeldking, line 888: This branch is dead in practice. A `Response` always carries a `status`, so `incomplete_details.reason` is only consulted when `status` is falsy, which never happens for a real response. That means `status="incomplete"` + `incomplete_details.reason="max_output_tokens"` yields `"incomplete"`, not `"length"`, which contradicts the PR description ("the more specific `incomplete_details.reason` when present") and loses the one signal that matters most for debugging truncation.

The more specific reason should win:

```python
def _map_finish_reason(status: Optional[str], incomplete_reason: Optional[str]) -> Optional[str]:
    if incomplete_reason:
        return _INCOMPLETE_REASON_TO_FINISH_REASON.get(incomplete_reason, incomplete_reason)
    if status:
        return _STATUS_TO_FINISH_REASON.get(status, status)
    return None
```

Minor: with the passthrough default, non-terminal statuses like `in_progress` / `queued` would also be emitted as a finish reason. Probably unreachable on a finished span, but worth considering whether the fallback should be `None` rather than the raw status.

python/instrumentation/openinference-instrumentation-openai-agents/tests/test_span_attribute_helpers.py
```diff
@@ -1454,6 +1455,7 @@ def test_get_attributes_from_message_content_list(
                     }
                 ),
                 "llm.model_name": "gpt-4",
+                "llm.finish_reason": "incomplete",
```
mikeldking, line 1459: None of the fixtures set `incomplete_details`, so the `max_output_tokens -> length` / `content_filter -> content_filter` mapping is never exercised (and the current test passes only because the bug makes `status` win). Please add a case with `status="incomplete", incomplete_details=IncompleteDetails(reason="max_output_tokens")` asserting `llm.finish_reason == "length"`.

mikeldking (commented):
Deriving `llm.finish_reason` from the Responses API `status` does conflate two different things (lifecycle state vs. why generation stopped). But the Responses API has no per-choice `finish_reason`, and `status` + `incomplete_details.reason` is the closest signal it exposes, so I'm fine with this as the approach for now.

One real issue though: the `incomplete_details.reason` branch is unreachable as written, so the truncation case (the one people actually want to detect) reports `incomplete` instead of `length`. Details inline. Once that's fixed and covered by a test I'm good to approve.

python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py
mikeldking: Fixed in f3360e0e: swapped the branch order so `incomplete_details.reason` is consulted first.

python/instrumentation/openinference-instrumentation-openai-agents/tests/test_span_attribute_helpers.py
mikeldking: Added `incomplete_response_max_output_tokens` (asserts `length`) and `incomplete_response_content_filter` (asserts `content_filter`) in f3360e0e.

mikeldking (approved):

PR #3662, opened 2026-09-02
fix(instrumentation): preserve context attributes in async coroutine decorators (#3583)

dlowzzxx:
## Why

`contextlib.ContextDecorator` runs its `__exit__` as soon as the decorated call returns. For an `async def` function or a generator function, that call only builds the coroutine or generator object, so OpenInference context attributes detached before the body ran. Issue #3583 reports the coroutine case.

`_UsingAttributesContextManager` now owns the decorator protocol and handles each kind of callable. A coroutine attaches when the body starts. A generator body runs each step under a rolling OpenTelemetry context.

## Scope

- Drop `contextlib.ContextDecorator` as the base class. `__call__` routes async generator functions, generator functions, coroutine functions, and plain functions.
- Add `_RollingOtelContext`, `_drive_sync_generator_in_rolling_otel`, and `_AwaitRollingOtelStep`. Each generator invocation gets its own rolling context. After every successful body step, the saved context becomes `get_current()`, so a span the body holds across `yield` stays current when the body resumes.
- Isolate only OpenTelemetry state. An earlier revision used one `contextvars.copy_context()` snapshot per generator, which froze every application `ContextVar` for the life of the generator. That snapshot is gone.
- Keep one attach token per decorated invocation in a wrapper local. One decorator instance serves every call of the function it wraps, so a token on `self` would break concurrent and re-entrant calls. Only `with` and `async with` still store `self._token`.
- Restore `_context_with_attributes`, the expanded `using_*` docstrings, and the corrected `version` / `variables` keywords in the `using_prompt_template` example.
- Add the README decorator section and `examples/async_context_attribute_decorators.py`, with the matching mypy exclude in `pyproject.toml`.

The PR does not include the unrelated Anthropic token-count change from the original branch.

## Tradeoffs

The generator drivers forward `send`, `throw`, `close`, `asend`, `athrow`, and `aclose` by hand, which is more code than `yield from`. The gain is that a generator body keeps its attributes and its open span across `yield` while application `ContextVar` values stay on the caller's live map.

## Blast Radius

The change affects Python callers that use a `using_*` helper as a decorator. Context-manager behavior is unchanged. Sync decorators keep their existing semantics and additionally survive re-entrant calls. Generator decorators gain attribute and span scoping that the stdlib decorator never provided.

## Verification

- The issue reproduction returned `{'session.id': 'async-session'}`. Main returned `{}`.
- `py310-ci-instrumentation` on `tests/test_context_managers.py` passed Ruff formatting, Ruff lint, strict mypy, and 39 tests.
- New tests cover live `ContextVar` updates across `yield` for sync and async generators, a body-owned `ContextVar` surviving resume, a consumer span between yields not becoming the parent of body spans, and an abandoned generator running its `finally` with attributes attached.
- Ported tests cover attributes in the body but not the consumer, span parentage across `yield`, bidirectional `send` / `throw` / `asend` / `athrow` / `aclose`, re-entrant sync decorators, concurrent coroutine calls, exception detach, and `using_session` stacked above `tracer.agent`.


dlowzzxx:

mikeldking:
@dlowzzxx thanks for the PR. I'll take a look soon

mikeldking:
@caroger - I did a full audit. Could use a python review here.

caroger (approved):

PR #3666, opened 2026-09-03
fix(agno): propagate MessageMetrics.cost to llm.cost.total span attribute

### LINKED ISSUE #3659: [feature request] Agno instrumentor: set llm.cost.* from MessageMetrics.cost

The Agno instrumentor exports token counts from a model call but drops the cost that Agno already parsed, so every span reaches the collector with no cost on it.

`_ModelWrapper` reads `response.response_usage` (an `agno.metrics.MessageMetrics`) and sets prompt, completion and cache read/write token counts — `_model_wrapper.py`, e.g. the `arun` path:

```python
if hasattr(response, "response_usage") and response.response_usage:
    metrics = response.response_usage
    if hasattr(metrics, "input_tokens") and metrics.input_tokens:
        span.set_attribute(LLM_TOKEN_COUNT_PROMPT, metrics.input_tokens)
    ...
```

That same object carries the cost. `agno.metrics.BaseMetrics` (agno 3.0.5) declares:

```python
cache_read_tokens: int = 0
cache_write_tokens: int = 0
reasoning_tokens: int = 0
cost: Optional[float] = None
```

and `OpenAIChat._get_metrics()` fills it from the provider payload:

```python
metrics.cost = getattr(response_usage, "cost", None)
```

For providers that report cost, this is a real charged amount, not an estimate. OpenRouter, for example, returns it on every chat completion. Two calls from my own account, identical model and identical token counts, differing only in cache hits:

```
prompt_tokens=199 completion_tokens=1 cached_tokens=0   → cost 0.0005304
prompt_tokens=199 completion_tokens=1 cached_tokens=164 → cost 0.00015156
```

A 3.5× difference. A backend that recomputes cost from token counts and a price table cannot produce either number correctly, and it has no way to know which of the two happened. The information exists in the process, is already parsed, and is discarded at the span boundary.

Set the cost attributes from `MessageMetrics.cost` wherever the token counts are already set — the four paths in `_model_wrapper.py` (`run`, `arun`, `run_stream`, `arun_stream`). The semantic conventions already define the attribute:

```python
from openinference.semconv.trace import SpanAttributes

if (cost := getattr(metrics, "cost", None)) is not None:
    span.set_attribute(SpanAttributes.LLM_COST_TOTAL, cost)
```

`LLM_COST_TOTAL` ("llm.cost.total") is in `openinference-semantic-conventions` today, alongside `LLM_COST_PROMPT` and `LLM_COST_COMPLETION`. Only the total is available from Agno's metrics, so a total-only attribute looks like the right scope; the split would need provider-specific parsing that Agno does not do.

`cost` is `Optional[float]` and stays `None` for providers that do not report it, so the guard above leaves those spans exactly as they are today.

There is precedent for the shape of this in #2726, where the LiteLLM instrumentor was changed to set span cost from `response_cost`.

Happy to open the PR, including tests for the streaming paths — in a stream the metrics only arrive with the final chunk, which is where the existing token-count logic already looks.

*Setting the attribute from application code.* This is what I run today: a subclass of Agno's `OpenRouter` overrides `_parse_provider_response` / `_parse_provider_response_delta` and writes the cost onto the current span, which is the instrumentor's LLM span because parsing happens inside the model call. It works, but it depends on two private Agno methods and on the instrumentor's span still being current at that point — neither is a contract, and both are mine to re-verify on every upgrade.

*Recomputing cost downstream from token counts.* What most backends do when the attribute is missing. As the cached vs. uncached pair above shows, it disagrees with the invoice whenever caching, BYOK or upstream surcharges are involved, and it goes stale silently when a provider changes a price.

**Additional context**

Versions: `openinference-instrumentation-agno` 1.0.6, `openinference-semantic-conventions` 0.1.33, `agno` 3.0.5, Python 3.14, macOS.

Note that a consumer also has to map `llm.cost.*` for this to surface — Langfuse, for instance, currently maps `llm.token_count.*` but not `llm.cost.*`. I am raising that on their side separately; it does not change the case for emitting the attribute, since the data is otherwise lost at the source.

Aftabbs:
## Summary

Closes #3659

The Agno instrumentor records token-count span attributes from
`response_usage` (`MessageMetrics`) but was dropping the `.cost`
field, so `llm.cost.total` was never emitted even when Agno had
calculated the cost.

**What changed**

- In `_model_wrapper.py`, after the existing token-count block in all
  four instrumented paths (`run`, `run_stream`, `arun`, `arun_stream`),
  read `getattr(metrics, "cost", None)` and forward it to
  `SpanAttributes.LLM_COST_TOTAL` when it is not `None`.  The guard is
  intentionally `is not None` rather than a truthiness check so that a
  cost of `0.0` is still recorded.
- Added `LLM_COST_TOTAL` to the module-level span-attribute alias block.

**Tests**

Added four unit tests in `test_instrumentor.py` (one per instrumented
path) that drive `_ModelWrapper` directly with a `SimpleNamespace`
response carrying a `.cost` value and assert that `llm.cost.total`
appears on the finished span.  A fifth test covers the `run` path with
`cost=None` and asserts the attribute is absent.

## Test plan

```
cd python/instrumentation/openinference-instrumentation-agno
pip install -e ".[test]"
pytest tests/test_instrumentor.py -k "cost"
```

Expected: 5 new tests pass, existing test suite unchanged.

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_model_wrapper.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_model_wrapper.py
@@ -503,6 +503,9 @@ class _ModelWrapper:
                         LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, metrics.cache_write_tokens
                     )
 
+                if (cost := getattr(metrics, "cost", None)) is not None:
+                    span.set_attribute(LLM_COST_TOTAL, cost)
+
             return response
 
     def run_stream(
@@ -577,6 +580,9 @@ class _ModelWrapper:
                         LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, metrics.cache_write_tokens
                     )
 
+                if (cost := getattr(metrics, "cost", None)) is not None:
+                    span.set_attribute(LLM_COST_TOTAL, cost)
+
     async def arun(
         self,
         wrapped: Callable[..., Awaitable[Any]],
@@ -636,6 +642,9 @@ class _ModelWrapper:
                     span.set_attribute(
                         LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, metrics.cache_write_tokens
                     )
+
+                if (cost := getattr(metrics, "cost", None)) is not None:
+                    span.set_attribute(LLM_COST_TOTAL, cost)
 
             span.set_attributes(dict(_output_value_and_mime_type(output_message)))
             return response
@@ -715,6 +724,9 @@ class _ModelWrapper:
                         LLM_TOKEN_COUNT_PROMPT_DETAILS_CACHE_WRITE, metrics.cache_write_tokens
                     )
 
+                if (cost := getattr(metrics, "cost", None)) is not None:
+                    span.set_attribute(LLM_COST_TOTAL, cost)
+
 
 # span attributes
 INPUT_MIME_TYPE = SpanAttributes.INPUT_MIME_TYPE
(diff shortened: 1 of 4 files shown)


satyadevai:
Verified Cost Attributes with openrouter, working as expected.
Untitled


Added two new test cases with live response, Verified cost and other attributes

satyadevai (approved):

caroger (approved):

Issue #1726, opened 2025-05-29
[feature request] support openAI agents responses payloards


mikeldking:
```
 elif obj.type == "image_generation_call":
            # TODO: Handle image generation call
            pass
        elif obj.type == "code_interpreter_call":
            # TODO: Handle code interpreter call
            pass
        elif obj.type == "local_shell_call":
            # TODO: Handle local shell call
            pass
        elif obj.type == "mcp_call":
            # TODO: Handle mcp call
            pass
        elif obj.type == "mcp_list_tools":
            # TODO: Handle mcp list tools
            pass
        elif obj.type == "mcp_approval_request":
            # TODO: Handle mcp approval request
            pass
```

---
**Triage notes**

These item types are still unhandled: `_processor.py` lines 375-395 (request params path) and 926-960 (response items path) all `continue`/`...` with a `# TODO`. The list has since grown to include `shell_call`, `apply_patch_call`, `compaction`, `tool_search_call` and their outputs.

**Requirements** — every instrumentor must (see `CLAUDE.md` → Requirements):

- [ ] **Context attributes** — `session.id`, `user.id`, `metadata`, `tag.tags` from `using_attributes` appear on spans carrying the new item attributes
- [ ] **TraceConfig masking** — new content attributes (tool arguments, code, shell commands, image data) go through `OITracer` so `hide_inputs` / `hide_outputs` / `hide_llm_tools` redact them (`spec/configuration.md`)
- [ ] **Semantic conventions** — attribute names from `openinference-semantic-conventions`; built-in tool calls follow `spec/tool_calling.md`, images follow `spec/multimodal_attributes.md`
- [ ] **Tests** — exhaustive attribute assertions per item type; the existing tests build `ResponseSpanData` fixtures in-process so no new live recording should be needed

Pattern: the sibling branches in the same functions (`function_call`, `computer_call`, `file_search_call`, `web_search_call`) show the shape.

PR #3685, opened 2026-09-05
feat(openai_agents): record computer actions and mask computer screenshots

dlowzzxx:
### Summary
Records OpenAI Responses API built-in `computer_call` tool calls and `computer_call_output` results in `openinference-instrumentation-openai-agents`, addressing part of umbrella issue #1726.

The follow-up commit extends the original mapping so that computer actions are recorded as structured tool-call arguments and screenshots are recorded as structured image content, where `TraceConfig` image masking (`OPENINFERENCE_HIDE_INPUT_IMAGES`, `OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH`) applies. Screenshot data is never duplicated into text attributes the image settings cannot mask.

### Changes
**`_processor.py`**
- `computer_call` (response output and replayed input):
  - `tool_call.id = call_id or id`, `tool_call.function.name = "computer_call"`
  - `tool_call.function.arguments` = JSON of `action` (or `actions` for batched calls)
- `computer_call_output` (replayed input):
  - `message.role = "tool"`, `message.tool_call_id = call_id`
  - screenshot `image_url` is emitted as `message.contents.0.message_content.image.image.url` so image masking and size limits apply
  - `message.content` keeps the remaining fields (`type`, `file_id`) without the `image_url`
- Raw `input.value` JSON on response spans strips the screenshot URL from `computer_call_output` items only (other items pass through by reference).
- Computer tool spans (`computer` on SDK >= 0.14, `computer_use_preview` on older releases) whose output is a `data:image/...` string record `{"type": "computer_screenshot"}` instead of the raw base64.
- Pydantic values are dumped through one defensive helper (`_dump_model`) that never raises, so a bad object cannot prevent the span from ending.
- Reset `tool_call_idx` when `msg_idx` advances; pass through `configuration_update` to keep `assert_never` exhaustiveness.

**Tests** (272 passing on pinned and latest SDK)
- Helper tests for computer tool calls and outputs, including action arguments, batched actions, nested pydantic actions inside dict items, and pydantic outputs.
- Processor tests: multi-turn `call_id` correlation, screenshot masking under `hide_input_images` / `base64_image_max_length` / `hide_inputs`, no screenshot duplication in tool span output, and an undumpable output object not breaking the span.

**Docs / example**
- README "Computer use" section: table of where actions and screenshots land, masking env vars, the default size limit, and how to read the trace in Phoenix.
- `examples/computer_use.py`: live run against an in-memory display (click a red button, confirm it turns green) exporting to local Phoenix.
- `examples/requirements.txt`: `openai-agents>=0.11.0`, `pillow`.

### Tradeoff to be aware of
Because the tool span records a placeholder instead of the screenshot, a run that stops immediately after a computer action (for example `max_turns` reached) has no LLM span to carry that final screenshot, so it is absent from the trace. This is documented in the README. The alternative (keeping base64 on the tool span) would leave screenshot data in an attribute the image masking settings do not cover.

Ref: #1726

### DIFF
--- a/python/instrumentation/openinference-instrumentation-openai-agents/README.md
+++ b/python/instrumentation/openinference-instrumentation-openai-agents/README.md
@@ -81,6 +81,49 @@ Now simply run the python file and observe the traces in Phoenix.
 python your_file.py
 ```
 
+## Computer use
+
+The instrumentor records the OpenAI Agents SDK's built-in computer tool (`ComputerTool`)
+so each turn of a computer-use loop is visible in Phoenix:
+
+| What                                   | Where it appears                                                                                                 |
+| -------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| The action the model requested         | LLM span output message, `tool_call.function.name = "computer_call"` with the `action` (or batched `actions`) as `tool_call.function.arguments` |
+| The action, replayed on the next turn  | Next LLM span input message, same `tool_call.*` attributes, correlated by `tool_call.id`                        |
+| The screenshot returned to the model   | Next LLM span input message, as structured image content (`message.contents.0.message_content.image.image.url`) |
+| The computer tool span                 | A `TOOL` span named `computer`; its output is a `{"type": "computer_screenshot"}` placeholder                     |
+
+Screenshots live only in the structured image attribute, so the standard image controls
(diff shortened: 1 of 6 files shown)


mikeldking (approved):
thanks!

PR #3688, opened 2026-09-06
feat(agno): activate spans in an injected runtime context

### LINKED ISSUE #3687: [feature request] Let instrumentors activate spans in an injected runtime context
OpenInference instrumentors lets us inject a `tracer_provider` when they **create** spans, but they always **activate** them in the process-global OTel context. 

Concretely: MLflow integrates Agno through this repo. MLflow's isolated mode keeps its spans in its own runtime context on purpose, but the OpenInference instrumentor writes to the global one anyway, which breaks that isolation.

An optional `runtime_context` kwarg on `_instrument()`, next to the existing `tracer_provider` and `config`:

```python
AgnoInstrumentor().instrument(tracer_provider=..., runtime_context=my_runtime_context)
```

When set, the instrumentor activates its spans there instead of globally. Currently my fix is only for Agno.

- Fix it in the core (`OITracer`). So the feature can be use with all the instrumentation.

**Additional context**
A generated Skill that explore the core solution in order to be implemented.

Happy to open the Agno PR as a first package, then a follow-up for the core if the direction is accepted :)

Killian-fal:
Fixes #3687

## Summary

MLflow integrates Agno through this repo. MLflow's isolated mode keeps its spans in its own runtime context on purpose, but the OpenInference instrumentor writes to the global one anyway, which breaks that isolation.

If you want more context, please read the Issue.

## Change

1. An optional `runtime_context` kwarg on `_instrument()`. Not passed = today's behaviour, unchanged.

```python
AgnoInstrumentor().instrument(tracer_provider=..., runtime_context=my_runtime_context)
```
2. `SpanActivation` — injectable `attach` / `get_current` / `detach` / `use_span`, defaulting to the global `opentelemetry.context` module. Direct `context_api.attach` / `trace_api.use_span` call sites go through it.

## Skill to apply in the core OITracer (so for all the instrumentors)
[SKILL.md](https://github.com/user-attachments/files/31878125/SKILL.md)

For now, I've only implemented this for Agno, since I use that framework for my projects. Ideally, it should be applied everywhere, but that would require a major PR. I'm not sure how you want to handle this (if the feature is accepted). But I'm ready to work on it according to the methodology you provide. (1 PR per instrumentor or 1 major PR?)

### DIFF
--- a/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/__init__.py
@@ -5,8 +5,12 @@ from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type:
 from wrapt import wrap_function_wrapper
 
 from openinference.instrumentation import (
-    OITracer,
     TraceConfig,
+)
+from openinference.instrumentation.agno._context import (
+    ActivationTracer,
+    SpanActivation,
+    set_activation,
 )
 from openinference.instrumentation.agno.version import __version__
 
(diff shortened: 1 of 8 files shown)


Killian-fal:

PR #3692, opened 2026-09-06
docs(bedrock): remove stale trace_event entry from get_chunk_type docstring

simpleqt:
`get_chunk_type(cls, trace_event_data)` documents a `trace_event` parameter that does not exist in the signature. Removed the stale Args entry; docstring-only change.


mikeldking (approved):
thanks for the catch

PR #3700, opened 2026-09-08
feat(semconv): add audio and video semantic conventions

caroger:
## Why

Chat APIs already send audio and video as content parts, but OpenInference only had image nesting plus top-level `audio.url` leaves. This change publishes chat nesting so instrumentors and consumers can agree on keys before any production rewrite.

Chat multimodal uses `message.contents` with the same double nesting as images (`message_content.video.video.url`, `message_content.audio.audio.url`). Voice-session span-root `input.audio.*` / `output.audio.*` stay instrumentor-local on openai-agents until stacked #3708. USER and AUDIO span kinds stay unpublished. Arize already renders that realtime tree. Phoenix does not walk span-root audio today.

`VideoAttributes` is `video.url` only. Consumers infer video MIME type from the URL path extension or a data URI prefix. Do not emit `video.mime_type`. Pre-existing `audio.mime_type` stays.

## Scope

- `spec/multimodal_attributes.md`, `spec/semantic_conventions.md`, `spec/llm_spans.md`
- `internal_docs/specs/audio_video/` chat encyclopedia (`openai_chat_audio.md`, Gemini, Bedrock), comparison table, and chat/video demo scripts
- `VideoAttributes` plus `MESSAGE_CONTENT_AUDIO` / `MESSAGE_CONTENT_VIDEO` in Python, JS, Java, and Go

Out of scope: Vertex, Google GenAI, Bedrock, and OpenAI chat instrumentor capture. TraceConfig hide fields and size gates (`OPENINFERENCE_HIDE_*_AUDIO/VIDEO`, `OPENINFERENCE_BASE64_AUDIO/VIDEO_MAX_LENGTH`) — shared `mask()` still only handles images. GenAI dual-write for audio and video. Publishing USER or AUDIO span kinds. Changing published `audio.mime_type`. Span-root `SpanAttributes`, `openai_realtime.md`, and `openai_realtime_audio_demo.py` (#3708). `truncate_audio_data_uri` still slices (follow-up).

## Tradeoffs

Video MIME is inferred from the URL. Audio MIME stays as the existing `audio.mime_type` leaf from #1153. Span-root voice keys and audio/video TraceConfig settings are deferred so this PR stays chat-only conventions.

## Blast Radius

Spec and constants for video nesting. Existing traces keep working. Hide env vars are not published as shared settings. Only openai-agents realtime already honors audio hide env vars, still via instrumentor-local strings until #3708. Instrumentors that drop video today are unchanged until follow-up PRs.

## Verification

Isolated Phoenix round-trip stored `message_content.video.video.url` with no `video.mime_type`. Go semantic-convention tests passed. Python `py3{10,14}-ci-semconv` and openai-agents realtime tests passed on the review-fix commit.


go/openinference-semantic-conventions/attributes.go
```diff
@@ -219,6 +221,12 @@ const (
 	AudioTranscript = "audio.transcript"
 )
 
+// Video attributes. Nested under MessageContentVideo.
+const (
+	VideoURL      = "video.url"
+	VideoMimeType = "video.mime_type"
```
mikeldking: how is this going to be determined. in most cases this will be inferred from the extension anyways. I'm not convinced adding another attribute here makes parsing any easier. I'm inclined to not put this in conventions.

It's just going to be duplicative and cause un-necessary processing.

MP4 video: video/mp4 (Extension: .mp4)WebM video: video/webm (Extension: .webm)Ogg video: video/ogg (Extension: .ogv)AVI video: video/x-msvideo (Extension: .avi)

All of these mime types are inferred from file patterns. Anything more is video processing

internal_docs/specs/audio_video/bedrock_video.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+S3 source rewrite:
+
+```
+llm.input_messages.0.message.contents.0.message_content.type = "text"
+llm.input_messages.0.message.contents.0.message_content.text = "Summarize this clip."
+llm.input_messages.0.message.contents.1.message_content.type = "video"
+llm.input_messages.0.message.contents.1.message_content.video.video.url = "s3://bucket/clip.mp4"
+llm.input_messages.0.message.contents.1.message_content.video.video.mime_type = "video/mp4"
```
mikeldking: re-enforces my comment.

spec/multimodal_attributes.md
```diff
@@ -90,6 +142,12 @@ When `OPENINFERENCE_BASE64_IMAGE_MAX_LENGTH` is set (default: 32000):
 - Only the base64 content portion is subject to the length limit
 - If a blob uploader is configured, over-limit images are externalized instead and the attribute records the destination URI (see [External Storage for Large Media](#external-storage-for-large-media))
 
+### Base64 audio and video truncation
+
+`OPENINFERENCE_BASE64_AUDIO_MAX_LENGTH` (default: 32000) truncates or offloads `audio.url` data URIs the same way images are truncated. Keep the `data:audio/…;base64,` prefix.
+
+`OPENINFERENCE_BASE64_VIDEO_MAX_LENGTH` does the same for `video.url` once an instrumentor emits video data URIs. Default 32000.
```
mikeldking: Can we externalize or redact audio/video that exceeds the limit? 32,000 base64 characters only holds about half a second of 24 kHz mono PCM16, so this would cut off most utterances. Slicing the base64 also leaves WAV length headers inconsistent and can make compressed audio or video undecodable. If we want a playable preview, we need to trim and re-encode the media before creating the data URI.

internal_docs/specs/audio_video/openai_realtime.md
```diff
@@ -0,0 +1,181 @@
...
+| `_set_turn_io_attributes` | AUDIO | `input.value` joined from typed text and audio transcripts. `output.value` joined from assistant transcripts. No WAV on the parent. |
+| `_start_turn` | AUDIO | `openinference.span.kind=AUDIO`, `llm.model_name`, `llm.invocation_parameters` from the session.update snapshot |
+
+The six audio keys are instrumentor-local string literals (`_INPUT_AUDIO_URL` and friends). Comments in `_realtime.py` still say they are not yet in semconv. After this spec, compose them from `AudioAttributes` instead of repeating the strings.
+
+### PCM and transcripts
+
+OpenAI Realtime streams 24 kHz mono PCM16. `pcm16_to_wav_data_uri` wraps the buffer as `data:audio/wav;base64,...`. User PCM comes from `send_audio` (pre-speech rolling buffer, then the active USER) and from `input_audio_buffer.append`. Assistant PCM comes from `RealtimeAudio` deltas (`on_audio_delta`). Transcripts come from `conversation.item.input_audio_transcription.completed` (user) and `response.output_audio_transcript.done` (assistant). On interruption, assistant transcript falls back to joined `.delta` chunks because `.done` never fires.
```
mikeldking: The [OpenAI docs](https://developers.openai.com/api/reference/resources/realtime/server-events) say `response.output_audio_transcript.done` is also emitted for interrupted, incomplete, or cancelled responses. Can we correct this so joined deltas are only the fallback when the terminal event is actually unavailable? We should also distinguish the generated transcript from what the user heard. Cancellation alone doesn't tell us where playback stopped.

mikeldking (changes requested):

go/openinference-semantic-conventions/attributes.go
caroger: Dropped `video.mime_type` from the published conventions in Python, JS, Java, and Go (`00e2b515`). `VideoAttributes` is only `video.url`. Consumers infer MIME from the path extension (`.mp4` → `video/mp4`, `.webm` → `video/webm`, `.ogv` → `video/ogg`, `.avi` → `video/x-msvideo`) or from a data URI prefix.

internal_docs/specs/audio_video/bedrock_video.md
caroger: Bedrock `format` no longer maps onto a video MIME attribute. S3 stays a URL with the extension. Bytes become a data URI whose prefix matches `format`. Same commit `00e2b515`.

spec/multimodal_attributes.md
caroger: Agreed. `OPENINFERENCE_BASE64_AUDIO_MAX_LENGTH` and `OPENINFERENCE_BASE64_VIDEO_MAX_LENGTH` are size gates now, not slicers. Over-limit audio/video is externalized when a blob uploader is configured, otherwise redacted. Spec calls out that 32k chars is ~0.5s of 24 kHz mono PCM16 and that a playable preview needs trim and re-encode. `00e2b515`.

internal_docs/specs/audio_video/openai_realtime.md
caroger: Corrected. `response.output_audio_transcript.done` is the source, including interrupted, incomplete, and cancelled responses. Joined `.delta` chunks are only the fallback when `.done` is missing. `output.audio.transcript` is the generated transcript, not playback. Cancellation does not say where playback stopped. That comes from the client (`conversation.item.truncate` with `audio_end_ms`). `00e2b515`.

caroger:
Thanks for the review, @mikeldking . Addressed in 00e2b515.

- Dropped `video.mime_type`. VideoAttributes is `video.url` only. MIME comes from the path extension or a data URI prefix. Bedrock format is not a separate attribute.
- Size gates no longer slice audio or video. Over-limit payloads are externalized or redacted. 32k chars is about 0.5s of 24 kHz mono PCM16, so a playable preview would need trim and re-encode.
- Assistant transcript prefers response.output_audio_transcript.done, including interrupt, incomplete, and cancelled. Joined deltas are only the fallback. That field is the generated transcript, not playback position.

Could you take another look?

RogerHYang (changes requested):
Could we carve out `input.audio.*` and `output.audio.*` into a separate PR? Looking into #3173, I found they were introduced alongside `USER` and `AUDIO` span kinds that aren’t in the shared spec. Those de facto attributes and span kinds warrant a separate PR while the message-content audio/video additions can move forward here.

RogerHYang (approved):

internal_docs/specs/audio_video/scripts/openai_chat_audio_demo.py
```diff
@@ -0,0 +1,89 @@
...
+            url=DATA_URI,
+            mime_type="audio/wav",
+        )
+        start_demo_span(ctx, "future", future, span_kind="LLM")
+
+        current_attrs = span_attrs_by_name(ctx, "current")
+        future_attrs = span_attrs_by_name(ctx, "future")
+        audio_url_key = content_key("input", 0, 1, f"message_content.audio.{AUDIO_URL}")
+        audio_mime_key = content_key(
+            "input", 0, 1, f"message_content.audio.{AUDIO_MIME_TYPE}"
```
mikeldking: mime type still in the example

internal_docs/specs/audio_video/scripts/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Each script prints PASS or FAIL and exits non-zero on failure.
+
+Open the Phoenix UI for that collector and compare the `current` and `future` spans.
+
+## What each script proves
+
+| Script | Current span | Future span |
+|---|---|---|
+| `gemini_video_demo.py` | LLM span with text contents only. Video URI may sit in `input.value`. | `message_content.type=video` plus `video.video.url`. No `video.mime_type`. GenAI `uri` part with `modality: video`. |
```
mikeldking: mime type still in the docs

internal_docs/specs/audio_video/scripts/openai_chat_audio_demo.py
mikeldking: Removed in acfb2590. The demo no longer sets `audio.mime_type` and now asserts the key is absent on the future span. MIME comes from the `data:audio/wav;base64,` prefix, which is also where the GenAI blob part reads it.

internal_docs/specs/audio_video/scripts/README.md
mikeldking: Removed in acfb2590. Dropped `audio.mime_type` from every new message-content example in `spec/multimodal_attributes.md`, `spec/semantic_conventions.md`, and the vendor docs. The already-published `AudioAttributes.AUDIO_MIME_TYPE` constant stays to avoid a breaking change but is now marked legacy; new instrumentors should not emit it.

mikeldking (approved):

PR #3718, opened 2026-09-10
feat(claude-agent-sdk): JS - Add Finish Reason Attribute

ialisaleh:
Refs #2901 

This PR adds `finish_reason` tracking to the Claude Agent SDK (TypeScript) instrumentation by extracting the `stop_reason` field from `SDKResultSuccess`/`SDKResultError` result messages and mapping it to the `LLM_FINISH_REASON` span attribute across the V1 `query()` wrapper and the V2 `unstable_v2_prompt()`/session (`send()`/`stream()`) wrappers, keeping it consistent with other instrumentations.

### DIFF
--- a/js/packages/openinference-instrumentation-claude-agent-sdk/src/messageProcessor.ts
+++ b/js/packages/openinference-instrumentation-claude-agent-sdk/src/messageProcessor.ts
@@ -83,9 +83,17 @@ export function extractInitAttributes(msg: SDKSystemMessage): {
 }
 
 /**
+ * Extracts the model's stop reason from a result message.
+ */
+function extractStopReason(msg: SDKResultMessage): string | undefined {
+  return msg.stop_reason ? String(msg.stop_reason) : undefined;
+}
+
+/**
  * Extracts span attributes from a result success message.
  */
 export function extractResultSuccessAttributes(msg: SDKResultSuccess): Attributes {
+  const stopReason = extractStopReason(msg);
   return {
     ...getOutputAttributes(msg.result),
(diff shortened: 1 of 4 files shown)


js/packages/openinference-instrumentation-claude-agent-sdk/src/messageProcessor.ts
```diff
@@ -82,17 +82,26 @@ export function extractInitAttributes(msg: SDKSystemMessage): {
   };
 }
 
+/**
+ * Extracts the model's stop reason from a result message.
+ */
+function extractStopReason(msg: SDKResultMessage): string | undefined {
```
caroger, line 88: The instrumentor declares support for Claude Agent SDK `>=0.2.0`, but result messages only gained top-level `stop_reason` in 0.2.31. Versions 0.2.0 through 0.2.30 expose the reason on `SDKAssistantMessage.message.stop_reason`, so this returns `undefined` throughout part of the supported range. Please also collect the nested assistant-message value in the streaming paths and add an older-shape test. For the one-shot V2 prompt, where the older SDK exposes no assistant message, document or explicitly version-gate the limitation.

caroger (changes requested):
Two blockers:

1. This public package change has no changeset.
2. The inline comment covers compatibility with the declared `>=0.2.0` SDK range.

js/packages/openinference-instrumentation-claude-agent-sdk/src/messageProcessor.ts
ialisaleh: Addressed.

caroger (approved):

PR #3728, opened 2026-09-13
fix: handle mixed-type lists in _flatten across instrumentors; isolate dspy extraction errors

### LINKED ISSUE #3727: [bug] DSPy instrumentor raises into user code from attribute extraction (_flatten mixed lists, unguarded extractors)
The DSPy instrumentor can raise exceptions into user code from its attribute extraction, before the wrapped call runs. Instrumentation should never change whether a call succeeds. Two deterministic triggers, both on `LM.__call__`; the async, Predict, and Module wrappers share the same inline-extraction pattern (27 call sites, no guards).

1. `_flatten` crashes on mixed-type lists. When a message `content` list contains both a mapping and a non-mapping part, the `any(isinstance(item, Mapping))` branch recurses into every item as a mapping:

```python
lm(messages=[{"role": "user", "content": [{"type": "text", "text": "hi"}, "plain-string-part"]}])
# AttributeError: 'str' object has no attribute 'items' (from _flatten, before the LM call)
```

Uninstrumented, the same call reaches litellm normally. Multipart content is a normal shape for vision and tool messages; one non-dict part takes the call down.

2. Extraction has no error isolation. A circular reference anywhere in the inputs makes `safe_json_dumps` raise `ValueError: Circular reference detected` into user code:

```python
circular = []; circular.append(circular)
lm(messages=[{"role": "user", "content": [{"type": "text", "text": "hi"}, circular]}])
# ValueError: Circular reference detected
```

This is the class litellm fixed in #3578 with `_suppress_extractor_errors`; the dspy wrappers call extractors inline with no guard.

Proposed fix, two parts:
- In `_flatten`, recurse only into mapping items of a list and pass other parts through indexed.
- Isolate extraction litellm-style: a shared guard so a failing extractor drops its attributes with a logged warning while the span and the wrapped call proceed.

Happy to open a PR with both, plus a test per trigger. (Both repros run twice, identical results, on dspy 3.3.1 against current main.)

Harsh23Kashyap:
# Description

Resolves #3727

`_flatten` has the same defect in nine instrumentors: for a list that mixes mappings with other types it takes the mapping branch and then recurses into every item, so a mixed-type `content` list (a normal shape for vision and tool messages) breaks attribute extraction. The visible impact differs by package:

- instructor: `AttributeError` raised into user code from the patched call, before the wrapped function runs.
- dspy: the same `AttributeError` into user code from LM spans.
- llama-index: no crash, but the flattening failure is caught and logged, so the event loses every attribute.
- autogen-agentchat, guardrails: the mixed-list path is not reachable from their current call sites; fixed for consistency so future call sites inherit the safe behavior.
- langchain: also not currently reachable - message content is pre-flattened by `_extract_message_kwargs` before it reaches `_flatten`, and other structured payloads (document metadata, prompts) are JSON-serialized first. Fixed for consistency.
- google-genai, portkey, groq: these copies of `_flatten` are unused today; fixed for consistency.

The dspy instrumentor had a second problem: no extractor had error isolation, so a circular reference in the inputs surfaced as `ValueError: Circular reference detected` from `safe_json_dumps` in user code. litellm got the same class of fix in #3578.

Changes:

- `_flatten` recurses only into mapping items of a list; non-mapping parts pass through indexed. Output for all-mapping and all-scalar lists is unchanged. Same edit in all nine copies; every `_flatten` in the tree is now a safe copy.
- dspy generator extractors get `_suppress_extractor_errors`, the same shape as litellm's: a failing extractor drops its attributes with a logged warning while the span and the wrapped call proceed. `_get_input_value` and `_module_prediction_output_attributes` fall back to a safe value on failure.

Verification:

- One new regression test per package; each fails on main and passes with the fix.
- instructor: end-to-end through `_PatchWrapper` with a fake create function. On main the mixed content list crashes with `AttributeError: 'str' object has no attribute 'items'`; with the fix the call completes and both content parts land on the span. Ran twice identical both ways.
- llama-index: end-to-end through `_finish_tracing` with a fabricated event carrying the mixed list. On main the event keeps none of its attributes; with the fix the model name and both content parts survive. Ran twice identical both ways.
- langchain: end-to-end through `_update_span` with a real `HumanMessage` carrying a mixed content list - no crash on either side, because langchain-core and `_extract_message_kwargs` normalize and pre-flatten content before `_flatten` runs. That run is what established the not-reachable status above.
- dspy: the repro scripts from the issue run twice identical before (AttributeError / ValueError into user code) and after (the call reaches the LM).
- Package suites: autogen-agentchat 7 passed, google-genai 122 passed, groq 12 passed, llama-index test_callback.py 14 passed and 4 xfailed, langchain 231 passed and 2 skipped. instructor, guardrails, portkey, and langchain each show a small set of pre-existing failures in this environment; the failure sets are identical on unmodified main.
- ruff check and ruff format clean on the changed files.

# Checklist:

- [x] Properly respects suppress tracing context (the suppression check at wrapper entry is untouched; the guards only affect attribute extraction)

### DIFF
--- a/python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
+++ b/python/instrumentation/openinference-instrumentation-dspy/src/openinference/instrumentation/dspy/__init__.py
@@ -2,6 +2,7 @@ import json
 from abc import ABC
 from copy import copy, deepcopy
 from enum import Enum
+from functools import wraps
 from inspect import signature
 from logging import getLogger
 from typing import (
@@ -946,17 +947,21 @@ def _get_input_value(method: Callable[..., Any], *args: Any, **kwargs: Any) -> s
         *args,
         **kwargs,
     )
-    return safe_json_dumps(
-        {
-            **{
-                argument_name: argument_value
-                for argument_name, argument_value in bound_arguments.arguments.items()
-                if argument_name not in ["self", "kwargs"]
+    try:
+        return safe_json_dumps(
+            {
+                **{
+                    argument_name: argument_value
+                    for argument_name, argument_value in bound_arguments.arguments.items()
+                    if argument_name not in ["self", "kwargs"]
+                },
+                **bound_arguments.arguments.get("kwargs", {}),
             },
-            **bound_arguments.arguments.get("kwargs", {}),
-        },
-        cls=DSPyJSONEncoder,
-    )
+            cls=DSPyJSONEncoder,
+        )
+    except Exception:
+        logger.exception("Failed to serialize input value")
+        return ""
 
 
 def _get_predict_span_name(instance: Any) -> str:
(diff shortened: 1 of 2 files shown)


Harsh23Kashyap:

Issue #1777, opened 2025-06-12
[token-counts] token count details


mikeldking:
what are the missing instrumentation for things more than prompt and completion

# Python

- [ ] #1779

# Javascript

# Semcov

- [ ] https://github.com/Arize-ai/openinference/issues/1780

PR #3754, opened 2026-09-15
feat(groq): capture reasoning content and token details

HarianthK:
# Description

Groq's reasoning models (`qwen-qwq-32b`, `deepseek-r1-distill-*`, `openai/gpt-oss-*`) return the model's reasoning in `message.reasoning` when the request sets `reasoning_format="parsed"`, and `usage` carries `completion_tokens_details.reasoning_tokens` and `prompt_tokens_details.cached_tokens` (groq-python `CompletionUsage`). The Groq instrumentor read none of these, so a trace of a reasoning model showed only the final answer and had no token breakdown for cost.

This adds:

- the reasoning as a content block on the output message: `message.contents.0` with type `reasoning`, followed by the answer as a `text` block, which is the shape the LiteLLM and OpenAI (Responses) instrumentors already use for reasoning;
- `llm.token_count.completion_details.reasoning` and `llm.token_count.prompt_details.cache_read` from the usage details, the same mapping as the OpenAI instrumentor.

Messages without reasoning are emitted exactly as before (`message.content`, no `message.contents`), and a usage without details adds nothing; the second test pins that. The details are read from either an attribute or a mapping, because the oldest supported SDK (`groq==0.9.0`, which the tests pin) keeps fields it does not know as plain mappings on the model, and the same code path then works on the current SDK where they are typed.

Checked with `tox run -e test-groq`, `test-groq-latest` and `ruff-mypy-groq`, all clean. The first test fails on `main` with a `KeyError` on the reasoning block and passes with the change.

Part of the token-details work in #1777 and the reasoning/cache request in #1331.

# Checklist:


HarianthK:

mikeldking (approved):

PR #3760, opened 2026-09-16
fix(haystack): keep output message content on spans

### LINKED ISSUE #3759: [bug] haystack instrumentor drops assistant reply content from spans
**Describe the bug**

The haystack instrumentor's `_get_llm_output_message_attributes` loses reply content in three cases:

1. **Text paired with a tool call.** When a reply has both text and tool calls and its meta carries `finish_reason: "tool_calls"`, only the tool calls are recorded - the text never reaches the span. Replies that say something before calling a tool ("Let me check that for you.") are common, so the visible part of the answer disappears from the trace.
2. **Replies without a `finish_reason`.** Replies whose meta has no `finish_reason` are skipped entirely: no content, no role. Haystack's own `_convert_streaming_chunks_to_chat_message` sets `finish_reason` to `None` when no chunk reports one, and not every OpenAI-compatible provider sends one, so streamed replies can vanish from spans completely.
3. **Multiple string replies.** For text generators, every string reply is written to `llm.output_messages.0.*` (a hardcoded `0` instead of the reply index), so with `n > 1` completions each reply overwrites the previous one and only the last survives.

**To Reproduce**

```python
from typing import Any, Dict, List

from haystack import Pipeline, component
from haystack.dataclasses import ChatMessage, ToolCall

from openinference.instrumentation.haystack import HaystackInstrumentor


@component
class FakeChatGenerator:
    @component.output_types(replies=List[ChatMessage])
    def run(self, messages: List[ChatMessage]) -> Dict[str, Any]:
        tool_reply = ChatMessage.from_assistant(
            text="Let me check that for you.",
            tool_calls=[ToolCall(tool_name="get_weather", arguments={"city": "Paris"}, id="call_1")],
            meta={"model": "fake-model", "finish_reason": "tool_calls"},
        )
        plain_reply = ChatMessage.from_assistant(
            text="streamed answer",
            meta={"model": "fake-model"},
        )
        return {"replies": [tool_reply, plain_reply]}


tracer_provider.add_span_processor(SimpleSpanProcessor(exporter))
HaystackInstrumentor().instrument(tracer_provider=tracer_provider)

pipe = Pipeline()
pipe.add_component("llm", FakeChatGenerator())
pipe.run({"llm": {"messages": [ChatMessage.from_user("hi")]}})

span = next(s for s in exporter.get_finished_spans() if s.name == "FakeChatGenerator.run")
for k, v in sorted(span.attributes.items()):
    if "output_messages" in k:
        print(k, "=", v)
```

Output (identical across runs):

```
llm.output_messages.0.message.role = 'assistant'
llm.output_messages.0.message.tool_calls.0.tool_call.function.arguments = '{"city": "Paris"}'
llm.output_messages.0.message.tool_calls.0.tool_call.function.name = 'get_weather'
```

The first reply's text is missing, and the second reply produces no attributes at all.

**Expected behavior**

Each reply lands under its own index with its content and role, whether or not it carries tool calls or a `finish_reason`:

```
llm.output_messages.0.message.content = 'Let me check that for you.'
llm.output_messages.0.message.role = 'assistant'
llm.output_messages.0.message.tool_calls.0.tool_call.function.arguments = '{"city": "Paris"}'
llm.output_messages.0.message.tool_calls.0.tool_call.function.name = 'get_weather'
llm.output_messages.1.message.content = 'streamed answer'
llm.output_messages.1.message.role = 'assistant'
```

 - OS: Linux
 - Version: haystack-ai 3.1.1, openinference-instrumentation-haystack @ main (5da0a96)

**Additional context**

The gating on `finish_reason` looks accidental: the string-reply branch captures content unconditionally, showing the intent to record every reply. I have a fix with regression tests ready and will open a PR.

Harsh23Kashyap:
Fixes #3759

`_get_llm_output_message_attributes` gated everything on `finish_reason` and indexed string replies with a hardcoded `0`, so reply content went missing from spans:

- a reply with both text and tool calls kept only the tool calls
- a reply whose meta has no `finish_reason` was skipped entirely (haystack's own streaming conversion produces exactly that when no chunk reports one)
- multiple string replies overwrote each other at index 0

The extraction now records every reply under its own index: text and role always, tool calls whenever the reply carries them, regardless of `finish_reason`.

Regression test: a pipeline with a fake chat generator (text + tool call reply, reply without `finish_reason`) and a fake text generator (two string replies), asserting each reply lands under its own index with its content. Fails before, passes after. Full haystack suite: 65 passed, 5 skipped (version/env gates), no changes needed to the existing cassette tests.


feiiiiii5 (commented):
Checked the two claims this PR rests on from the Haystack side rather than from the diff. Both hold; one reviewer question that turns out to be a non-issue, recorded so nobody re-derives it.

**The producer path is real (not a synthetic fixture).** The PR description says Haystack's own streaming conversion yields a reply whose meta has no `finish_reason`. That is in `haystack/components/generators/utils.py:172-174,189`:

```python
finish_reasons = [chunk.finish_reason for chunk in chunks if chunk.finish_reason]
finish_reason = finish_reasons[-1] if finish_reasons else None
...
"finish_reason": finish_reason,
```

So a streamed reply from any provider that never sets `finish_reason` arrives with `meta["finish_reason"] = None`, and the removed guard (`if ... (finish_reason := reply_meta.get("finish_reason")) is None: continue`) discarded the whole reply — content and role — before the loop ever reached the `tool_calls` branch. That matches the issue, so the `finish_reason` gate removal is fixing a supported path rather than hardening a hypothetical one.

**No empty content is emitted for tool-call-only replies.** Checked against Haystack 3.2.0-rc0:

```
text+tool_calls    .text='Let me check.'   is_not_None=True    tool_calls=1
tool_calls only    .text=None              is_not_None=False   tool_calls=1
```

so `if (text := reply.text) is not None` yields content exactly for the #3759 case 1 reply (text plus tool calls, previously tool-calls-only) and skips the tool-call-only case. This also fixes a latent problem in the old `else` branch, which yielded `MESSAGE_CONTENT, reply.text` unconditionally and could hand `None` to the attribute for a `finish_reason` of e.g. `stop` with no text.

**The asymmetry I suspected is not a bug.** The new output path yields `reply.role.value`, while the input path at `_wrappers.py:517-518` yields the raw enum. `ChatRole` is `class ChatRole(str, Enum)`, OTel stores the instance, and it serializes as `"user"`:

```
span attr type: ChatRole  value: <ChatRole.USER: 'user'>
json.dumps({role: ChatRole.USER}) -> {"role": "user"}
```

So both spellings export the same string; no change needed here, just noting it because the inconsistency looks like a finding at a glance.

**Two things I could not verify locally, so treat them as untested rather than as passing.** I could not run this package's suite in my environment: collection of `tests/openinference/haystack/test_instrumentor.py` needs `haystack_integrations` (Cohere ranker import) and PyPI is timing out from here, so the PR's "65 passed, 5 skipped" is the author's number, not one I reproduced. And I did not check the `n > 1` case against a live provider — only that the `.0.` → `.{reply_index}.` change makes multiple string replies addressable.

For the record, the remaining `.0.` literals in this file (`_wrappers.py:521-523` prompt branch, `:759-767` single-text embedder branch) are single-value `elif` arms, not loops, so they are correct as written.

PR #3348, opened 2026-07-05
feat(ollama): add Ollama instrumentor

NishchayMahor:
## What

Adds `openinference-instrumentation-ollama`, an auto-instrumentor for the [Ollama](https://github.com/ollama/ollama-python) Python client. Ollama is the most widely used local-model runner and has no OpenInference coverage today — so for anyone tracing a local-model app with Phoenix/Arize, the LLM call currently goes dark.

Instruments `ollama.chat`, `ollama.Client(...).chat`, and `ollama.AsyncClient(...).chat`, emitting OpenInference **LLM** spans with:

- input messages and output message (role, content, tool calls)
- invocation parameters (`llm.invocation_parameters`) and the model name
- tool schemas (`llm.tools.*`) and tool-call name/arguments
- token counts (`prompt_eval_count` → prompt, `eval_count` → completion, plus total)
- input/output value

## Notes

- The module-level helpers (`ollama.chat`) bind to a shared client at import time, so `_instrument` re-points `ollama.chat` to the wrapped method after wrapping the class methods; `_uninstrument` restores it. This keeps `ollama.chat`, `Client`, and `AsyncClient` all traced without double-counting.
- Ollama returns tool-call `arguments` as a mapping (not an OpenAI-style JSON string); these are serialized to JSON for the `tool_call.function.arguments` attribute.
- Follows the `groq` instrumentor's structure/conventions. Scoped to the `chat` endpoint for this first cut (mirrors how other instrumentors started); `generate`/`embed` and streaming accumulation can follow.
- Registered in `python/tox.ini`, `.github/dependabot.yml`, `release-please-config.json`, and `.release-please-manifest.json`.

## Testing

`tox -e py310-ci-ollama` equivalent locally: **6 tests pass** (sync chat, async chat, tool calls, token counts, entry point, suppress-tracing), mocking the client's HTTP layer so no running Ollama server is needed. `ruff` clean and `mypy --strict` clean.

Issue #3769, opened 2026-09-17
[feature request] Add OpenInference auto-instrumentation for the TypeSafe AI Python SDK (typesafe-sdk)


mikeldking:
## Is your feature request related to a problem? Please describe.

[TypeSafe AI](https://docs.typesafe.ai/introduction/quickstart) ships an official Python SDK (`typesafe-sdk`) that is currently invisible to OpenInference. Users who put TypeSafe calls into a RAG pipeline, a router, a guardrail, or an LLM-as-a-judge step get no spans for them, so a TypeSafe call shows up as a gap in the trace: no latency, no token counts, no record of which questions were asked or what was answered.

TypeSafe is specifically positioned for the patterns OpenInference users care most about observing — its docs ship cookbooks for [guardrails](https://docs.typesafe.ai/cookbooks/llm_guardrails), [intent routing](https://docs.typesafe.ai/patterns/intent-routing), [confidence-gated routing](https://docs.typesafe.ai/patterns/confidence-routing), [re-ranking](https://docs.typesafe.ai/cookbooks/rerank_typesafe), [RAG passage classification](https://docs.typesafe.ai/cookbooks/classifying_rag_passages), and [citation checking](https://docs.typesafe.ai/cookbooks/citation_check). Those are exactly the decision points where you want a span, because a silent misclassification upstream is the usual root cause of a bad downstream answer. The per-answer `confidence` and `probabilities` that TypeSafe returns are unusually valuable telemetry: they make it possible to filter traces for low-confidence decisions, which is not something most LLM providers expose.

## Describe the solution you'd like

A new `openinference-instrumentation-typesafe` package following the existing patterns in this repo (closest analogues: `openinference-instrumentation-groq`, `openinference-instrumentation-anthropic`, `openinference-instrumentation-mistralai`).

### Background: what the SDK looks like

Install and quickstart ([docs](https://docs.typesafe.ai/introduction/quickstart)):

```bash
pip install typesafe-sdk   # or: uv add typesafe-sdk
```

```python
from typesafe_sdk import Choice, Noul, Score, TypeSafeClient

with TypeSafeClient() as client:
    response = client.system_one(
        state="Hi, I've been trying to connect my Stripe account for 3 days and it keeps failing.",
        questions={
            "department": Choice(
                instructions="Which team should handle this",
                criteria={"billing": "Payment issues", "technical": "Bugs"},
            ),
            "is_urgent": Noul(instructions="Message conveys urgency"),
        },
    )

[...]

PR #3773, opened 2026-09-17
feat(python): add openinference-instrumentation-typesafe

axiomofjoy:
Adds `openinference-instrumentation-typesafe`, an instrumentor for the [TypeSafe AI](https://docs.typesafe.ai/) Python SDK (`typesafe-sdk >= 0.6.0`).

## What it traces

`TypeSafeClient.system_one` and `AsyncTypeSafeClient.system_one` become OpenInference LLM spans.

A System One call isn't a chat exchange. You send a `state` plus a map of typed `questions` (Noul, Choice, Score) and get back one typed `answer` per question — there's no message list on either side. So the span records the request and response bodies as `input.value` and `output.value` and stops there; no `llm.input_messages` / `llm.output_messages`. The `questions` map rides in `llm.invocation_parameters`, where it plays the same role a JSON response schema plays for chat-completion APIs.

Beyond that: `llm.provider` (`typesafe`), `llm.request.model_name` and `llm.response.model_name` (e.g. `jev-latest` → `jev-1.13.0`), and `llm.token_count.{prompt,completion,total}`.

Everything is built from the shared helpers in `openinference.instrumentation`, and spans go through `OITracer` — so `TraceConfig` masking, context attributes (session, user, metadata, tags) and `suppress_tracing` all work. Errors record the exception and set `ERROR` status.

## Notable choices

- **Patched through the SDK's public surface.** The wrappers target `typesafe_sdk.TypeSafeClient` / `typesafe_sdk.AsyncTypeSafeClient` rather than the `typesafe_sdk._core.client.*` modules that define them. `wrapt` patches the same class object either way, and the private path carries no compatibility promise.
- **`python/uv.toml` gets a `typesafe-sdk` exemption.** The instrumentor's floor, 0.6.0, was published more recently than CI's rolling 3-day `exclude-newer` cutoff, so resolution fails without it. It uses the far-future date the file already documents, deliberately: a per-package timestamp *overrides* the global cutoff, so a real date would become a permanent ceiling and quietly turn the `typesafe-latest` canary into a no-op.
- **No VCR cassettes.** The SDK speaks `httpx2`, which vcrpy doesn't patch, so the tests inject an `httpx2.MockTransport` that records request bodies and replays a canned System One response. Fully offline, no API key.

## Testing

Covers sync and async, structured state, raw-dict questions, the client-level default model, `extra_body`, error status, `suppress_tracing`, context attributes, `TraceConfig(hide_inputs, hide_outputs)`, uninstrument, and the entry point.

```
tox run-parallel -e py310-ci-typesafe,py310-ci-typesafe-latest,py314-ci-typesafe,py314-ci-typesafe-latest
```

All four pass: ruff format, ruff check, mypy clean, 10 tests.

Also verified end to end against a local Phoenix by running the three `examples/*.py` scripts against the live API — 4 traces, 4 LLM spans, all `OK`, with the request body, invocation params, token counts, metadata and session link rendering as expected.

## Open questions for reviewers

1. **Should `typesafe` become a well-known `llm.provider` value?** Right now it's a literal in this package and is absent from `OpenInferenceLLMProviderValues` (Python and TS) and from `spec/semantic_conventions.md`. Precedent (`#3348` for Ollama) says the enum entry lands in the same PR as the instrumentor; I left it out because it's a cross-language spec change and wanted a call on the spelling (`typesafe` vs `typesafe-ai`) first. Happy to add it here.
2. **`input.value` is reconstructed, not captured.** `get_request_attributes` rebuilds the request body from the bound call arguments, and the tests assert it equals what went over the wire. That equality is maintained by a second copy of the SDK's body-building rules, so an SDK change to defaults or the `extra_body` merge would make the span quietly disagree with the request. Wrapping `typesafe_sdk._core.endpoints.prepare_system_one` instead would make the captured body a fact rather than a reconstruction — worth doing now, or as a follow-up?
3. **Per-answer `confidence` and `probabilities`** currently live only inside the `output.value` JSON. Flattened attributes for them could be a follow-up.
4. The instrumentor class is `TypeSafeAIInstrumentor` (vendor's full name, cf. `VertexAIInstrumentor`) while the distribution is `openinference-instrumentation-typesafe`. Say the word if you'd rather have `TypeSafeInstrumentor`.

The SDK has no streaming and no tool calling, so there's no stream proxy.

---
_Generated by [Claude Code](https://claude.ai/code/session_012zpz5amSUru67UVgWKQtHt)_


mikeldking:
Looks like this was kicked off https://github.com/Arize-ai/openinference/issues/3769

python/instrumentation/openinference-instrumentation-typesafe/src/openinference/instrumentation/typesafe/_attributes.py
```diff
@@ -0,0 +1,150 @@
...
+        body.update(extra_body)
+    invocation_parameters = {k: v for k, v in body.items() if k != "state" and v is not None}
+    return {
+        **get_span_kind_attributes(OpenInferenceSpanKindValues.LLM),
+        **get_input_attributes(body, mime_type=OpenInferenceMimeTypeValues.JSON),
+        **get_llm_attributes(
+            provider=LLM_PROVIDER,
+            request_model_name=model,
+            invocation_parameters=invocation_parameters,
+        ),
```
caroger: `hide_inputs` redacts `input.value` and leaves this `llm.invocation_parameters` copy of `questions` on the span. Chat instrumentors get the prompt stripped because it lives on `llm.input_messages`, which that flag does clear.

That is a masking hole, not a leftover field. TypeSafe instructions are the prompt. They often describe customer tickets. A user who sets `hide_inputs=True` because `state` is PII still exports the full questions map, and the README tells them `hide_inputs` / `hide_outputs` is enough. `test_trace_config_hides_inputs_and_outputs` never looks at `LLM_INVOCATION_PARAMETERS`, so this ships as "masking works."

Update the README masking bullet to `TraceConfig(hide_inputs=True, hide_outputs=True, hide_llm_invocation_parameters=True)`. Change `test_trace_config_hides_inputs_and_outputs` to instrument with that config and assert `SpanAttributes.LLM_INVOCATION_PARAMETERS` is absent.

python/instrumentation/openinference-instrumentation-typesafe/src/openinference/instrumentation/typesafe/_attributes.py
```diff
@@ -0,0 +1,150 @@
...
+    Returns:
+        The span kind, ``input.value``, and the request-side ``llm.*`` attributes.
+    """
+    body: Dict[str, Any] = {
+        "state": _to_builtins(state),
+        "model": model,
+        "questions": _to_builtins(questions),
+    }
+    if extra_body:
+        body.update(extra_body)
```
caroger: `state` and `questions` go through `_to_builtins`. `extra_body` does not, so Mapping/Sequence values become repr strings on the span while the SDK encodes them as JSON. `extra_body={"ids": range(3)}` is `[0,1,2]` on the wire and `"range(0, 3)"` here.

The PR's contract is that `input.value` is the request body. Tests assert it equals the recorded HTTP JSON. `extra_body` is a public SDK field whose values are `JSONValue`, which includes `Mapping` and `Sequence`. Leaving it unconverted means the span reports a string the collector cannot query the same way as the wire, and the MappingProxyType fix you already landed for questions does not apply to the rest of the body.

After `body.update(extra_body)`, set `body = _to_builtins(body)`. Add a test next to `test_abstract_mapping_questions_and_tuple_state` that `input.value` equals the recorded request when `extra_body` holds a `MappingProxyType`.

caroger (approved):

python/instrumentation/openinference-instrumentation-typesafe/src/openinference/instrumentation/typesafe/_attributes.py
mikeldking: Confirmed and fixed in 5fdfa2ee. `extra_body` now goes through `_to_builtins` before the merge, so it gets the same treatment as `state` and `questions`.

The mechanism is `safe_json_dumps`, which is `json.dumps(..., default=str)` — so an abstract `Mapping` in `extra_body` was serialized as its repr, on both `input.value` and `llm.invocation_parameters`.

Proven against a local Phoenix with two builds of the same working tree, one call each to the real API, passing `extra_body={"state": MappingProxyType({...})}` (the SDK documents extra_body as last-write-wins over `state`):

```
jq -r '.[0].attributes["input.value"] | fromjson | .state | type'

before (fix reverted): string   -> "{'ticket': {'subject': 'Charged twice for one order', ...}}"
after:                 object   -> {"ticket": {"subject": "Charged twice for one order", ...}}
```

`test_abstract_containers_in_extra_body` covers it. One note for whoever reads that test next: the tuple assertion passes either way, since `json.dumps` already renders tuples as arrays — the `MappingProxyType` assertion is the one that catches the bug.

python/instrumentation/openinference-instrumentation-typesafe/src/openinference/instrumentation/typesafe/_attributes.py
mikeldking: Fixed in 5a230f39, though by a different route than suggested, and the PII framing here doesn't hold for this API.

**On the premise.** A System One call has no input messages — the analogy to chat instrumentors getting the prompt stripped via `llm.input_messages` doesn't transfer. The customer ticket is the `state`, and `state` was already excluded from the invocation parameters (`k != "state"`), so it was never in the copy that survived `hide_inputs`. What survived was the developer-authored question schema: `instructions` plus `criteria`. Canary run on a local Phoenix, real API call, `TraceConfig(hide_inputs=True, hide_outputs=True)` under the old code:

```
input.value=__REDACTED__   output.value=__REDACTED__
PII-CANARY-STATE anywhere:    false     <- caller data was already covered
INSTRUCTIONS-CANARY anywhere: true      <- the question schema was not
```

**On the real problem.** Your mechanism was right: `hide_inputs` doesn't touch `LLM_INVOCATION_PARAMETERS` (only `hide_llm_invocation_parameters` does, `config.py:346`). And it was worse than one flag missing, because `questions` was recorded *twice* — in `input.value` and again in the invocation parameters. `hide_inputs` cleared the first copy, `hide_llm_invocation_parameters` the second, so neither flag alone removed the instructions and the split was the actual defect.

**What we did instead of documenting a third flag.** `llm.invocation_parameters` is now call configuration only: the `model` and any `extra_body` fields. The `state` and the `questions` are recorded once, in `input.value`. So `TraceConfig(hide_inputs=True, hide_outputs=True)` is the whole masking story for request and response content, with no extra flag for a user to know about:

```
typesafe-hide-inputs-v2   input.value=__REDACTED__  output.value=__REDACTED__
                          llm.invocation_parameters={"model": "jev-latest"}
                          PII-CANARY-STATE anywhere:    false
                          INSTRUCTIONS-CANARY anywhere: false
                          token_count.total=385
```

`test_trace_config_hides_inputs_and_outputs` now asserts both the state and the instructions text are absent under `hide_inputs` alone, and the README no longer advertises `hide_llm_invocation_parameters` as part of the masking story, since it no longer covers request content.

The trade-off, for the record: `llm.invocation_parameters` is now `{"model": ...}` for calls that pass no `extra_body`, which is thin in the Phoenix params panel. Happy to revisit if you'd rather see the schema surfaced there — but if so it should be under an attribute that `hide_inputs` covers, not one it doesn't.

PR #3776, opened 2026-09-17
fix(claude-agent-sdk): return the SDK Query object from the instrumented query()

### LINKED ISSUE #3774: [bug] claude-agent-sdk (JS): instrumented query() drops the SDK Query object, so interrupt() and other control methods are undefined
**Describe the bug**

`@arizeai/openinference-instrumentation-claude-agent-sdk` wraps the SDK's `query()`, but the wrapped function returns a fresh `{ [Symbol.asyncIterator]() {...} }` object instead of the SDK's `Query` object. Every method the SDK declares on `Query` (`interrupt`, `setPermissionMode`, `setModel`, `setMaxThinkingTokens`, `setMcpPermissionModeOverride`, `applyFlagSettings`, `reinitialize`, `initializationResult`, and the `AsyncGenerator` methods `next`/`return`/`throw`) is `undefined` on the instrumented result.

Code that calls `q.interrupt()` (the documented way to stop a running query) silently stops working the moment tracing is enabled. In our harness that was the hard-kill backstop for a runaway agent, so enabling observability disabled the kill switch.

A second, related effect: the wrapper only calls the real `query()` when iteration begins (inside `[Symbol.asyncIterator]()`), whereas the unwrapped SDK spawns the Claude Code process at `query()` call time. So the instrumented call also shifts when the subprocess starts.

**To Reproduce**

Against the real SDK, no model call needed (the wrapper defers the SDK call until iteration, so nothing is spawned):

```ts
import * as SDK from "@anthropic-ai/claude-agent-sdk";
import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
import { ClaudeAgentSDKInstrumentation } from "@arizeai/openinference-instrumentation-claude-agent-sdk";

const inst = new ClaudeAgentSDKInstrumentation({ tracerProvider: new NodeTracerProvider() });
const wrapped = inst.manuallyInstrument(SDK);

const q = wrapped.query!({ prompt: "unused", options: { maxTurns: 1 } }) as any;
console.log(Reflect.ownKeys(q));            // [ Symbol(Symbol.asyncIterator) ]
console.log(typeof q.interrupt);            // undefined
console.log(typeof q.setPermissionMode);    // undefined
console.log(typeof q.next);                 // undefined
```

Output on `@anthropic-ai/claude-agent-sdk@0.3.218` + `@arizeai/openinference-instrumentation-claude-agent-sdk@0.3.0`:

```
own keys: [ 'Symbol(Symbol.asyncIterator)' ]
  interrupt                      undefined
  setPermissionMode              undefined
  setModel                       undefined
  setMaxThinkingTokens           undefined
  setMcpPermissionModeOverride   undefined
  applyFlagSettings              undefined
  reinitialize                   undefined
  initializationResult           undefined
  next                           undefined
  return                         undefined
  throw                          undefined
```

Unwrapped, `SDK.query(...)` returns a `Query` (`export declare interface Query extends AsyncGenerator<SDKMessage, void>` in `sdk.d.ts`) with all of those present.

**Expected behavior**

The instrumented `query()` returns the SDK's `Query` object with iteration traced, so control methods keep working. `for await`, direct `next()`/`return()`/`throw()`, and `interrupt()` should all behave exactly as they do without instrumentation, with the only difference being the emitted spans.

**Desktop**

 - OS: macOS 15 (Darwin 24.6), arm64
 - Node 26.5 (also reproduces under the workspace's pinned Node 22 in vitest with a mock module that exposes an `interrupt` method)
 - `@anthropic-ai/claude-agent-sdk` 0.3.218
 - `@arizeai/openinference-instrumentation-claude-agent-sdk` 0.3.0

**Additional context**

- Source: `js/packages/openinference-instrumentation-claude-agent-sdk/src/v1QueryWrapper.ts`, `wrapQuery()` returns a new object literal from `wrappedQuery` and calls `original(...)` inside `[Symbol.asyncIterator]()`. https://github.com/Arize-ai/openinference/blob/5a1db037/js/packages/openinference-instrumentation-claude-agent-sdk/src/v1QueryWrapper.ts
- Why the test suite does not catch it: every test in `test/v1Query.test.ts` mocks `query()` as a plain `{ [Symbol.asyncIterator] }` object, so the mock has no methods to lose.
- Workaround we are using: a shim passed to `manuallyInstrument` captures the real `Query` as the wrapper constructs it, and a `Proxy` over the wrapper's iterable forwards missing properties to it.
- I have a small fix with tests ready and will open a PR referencing this issue: call the SDK eagerly (matching unwrapped behaviour), start the AGENT span at call time, and return a `Proxy` over the real `Query` whose `[Symbol.asyncIterator]`/`next`/`return`/`throw` route through the traced iterator and whose other members forward, bound, to the SDK object.

iaj6:
Fixes #3774

## What changed

`wrapQuery()` in `openinference-instrumentation-claude-agent-sdk` returned a new `{ [Symbol.asyncIterator]() {...} }` object instead of the SDK's `Query`. Every control method on `Query` (`interrupt()`, `setPermissionMode()`, `setModel()`, `setMaxThinkingTokens()`, `applyFlagSettings()`, `reinitialize()`, `initializationResult()`) and the `AsyncGenerator` methods (`next()`/`return()`/`throw()`) were `undefined` on the instrumented result. Code that stops a running query with `q.interrupt()` silently stopped working when tracing was enabled.

The wrapper now:

- calls the SDK at `query()` time, as the unwrapped SDK does (it spawns the Claude Code process on call), and starts the AGENT span at that point so the injected `PreToolUse`/`PostToolUse` hooks have their parent span before the SDK sees them;
- returns a `Proxy` over the real `Query`: `[Symbol.asyncIterator]`, `next`, `return` and `throw` route through one shared traced iterator (created on first use, so the span ends exactly once whichever entry point the caller uses); every other member forwards to the SDK object, bound to it so implementations that rely on `this` keep working.

The traced iterator body is unchanged from before; it just moved out of `[Symbol.asyncIterator]()` so it can be shared. `wrapQuery` became generic over the return type so `manuallyInstrument()` keeps returning the same type the module exposes.

## Why the suite did not catch it

Every test in `test/v1Query.test.ts` mocks `query()` as a plain `{ [Symbol.asyncIterator] }`, so there were no methods to lose. This PR adds three tests:

- a mock shaped like the real `Query` (async iterator plus `interrupt()` and `setPermissionMode()`), asserting the methods survive instrumentation, are callable, and that iteration still yields one OK AGENT span;
- iteration driven by hand through `next()` and `return()` on the returned object is traced and ends the span;
- the SDK `query()` is invoked when the wrapper is called, not when iteration begins.

All three fail on `main` and pass here. The rest of the suite is unchanged (45/45).

## Behaviour notes for review

- Span start time moves from first iteration to the `query()` call. In practice callers iterate immediately, and this is what the unwrapped SDK's process lifetime already looks like.
- A caller who creates a query and never iterates now has an AGENT span that never ends (it was previously never created). That matches every other span for an abandoned operation.
- Reproduced against the real `@anthropic-ai/claude-agent-sdk@0.3.218` before the fix: `Reflect.ownKeys(wrapped.query(...))` is `[Symbol(Symbol.asyncIterator)]`, all eleven methods `undefined`.

Changeset included (patch).

https://claude.ai/code/session_01MXxV7qtX4GyTfo7QWaREnL


iaj6:

PR #3777, opened 2026-09-17
fix(claude-agent-sdk): keep a caller-supplied session.id over the SDK session_id

### LINKED ISSUE #3775: [bug] claude-agent-sdk (JS): SDK session_id overwrites a caller-supplied session.id from OpenInference context
**Describe the bug**

When a caller scopes a trace to a session with OpenInference context (`setSession(context.active(), { sessionId })`), the Claude Agent SDK instrumentation correctly applies that `session.id` to the AGENT span at start (via `OITracer`), and then overwrites it with the SDK's own `session_id` when the `system/init` message arrives, and again from the `result` message. The caller-supplied session id is lost from the AGENT span.

`js/AGENTS.md` lists "Propagate context attributes via OITracer (session ID, user ID, metadata, tags)" as a must for every instrumentor. This package propagates it and then clobbers it.

The user-visible consequence in Phoenix: Phoenix binds a trace to the first `session.id` it ingests for that trace and ignores later ones (`src/phoenix/db/insertion/span.py`). Spans export when they end, so in a turn with tool calls the TOOL spans (which keep the context session id) usually land first and the trace goes into the caller's session; in a turn with no tool calls the AGENT span lands first with the SDK id and the trace goes into a session named after the SDK UUID. The same agent's turns end up split across two sessions depending on whether it used a tool.

**To Reproduce**

Real SDK, one cheap turn:

```ts
import * as SDK from "@anthropic-ai/claude-agent-sdk";
import { context } from "@opentelemetry/api";
import { InMemorySpanExporter, SimpleSpanProcessor } from "@opentelemetry/sdk-trace-base";
import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
import { AsyncLocalStorageContextManager } from "@opentelemetry/context-async-hooks";
import { setSession } from "@arizeai/openinference-core";
import { ClaudeAgentSDKInstrumentation } from "@arizeai/openinference-instrumentation-claude-agent-sdk";

const exporter = new InMemorySpanExporter();
const provider = new NodeTracerProvider({ spanProcessors: [new SimpleSpanProcessor(exporter)] });
provider.register({ contextManager: new AsyncLocalStorageContextManager().enable() });
const wrapped = new ClaudeAgentSDKInstrumentation({ tracerProvider: provider }).manuallyInstrument(SDK);

await context.with(setSession(context.active(), { sessionId: "my-app-session-42" }), async () => {
  for await (const _ of wrapped.query!({ prompt: "Reply with exactly: OK", options: { model: "haiku", maxTurns: 1, tools: [] } })) {}
});
await provider.forceFlush();
const agent = exporter.getFinishedSpans().find((s) => s.name === "ClaudeAgent.query");
console.log(agent?.attributes["session.id"]);
```

Output:

```
SDK session_id from init: 6714bb26-d4b7-49ec-9c80-7f43ba10ba50
context session.id set by caller : my-app-session-42
AGENT span session.id exported   : 6714bb26-d4b7-49ec-9c80-7f43ba10ba50
```

The same is reproducible in the package's own vitest harness with the existing mock module shape plus `context.with(setSession(...))` around the iteration and an `AsyncLocalStorageContextManager` registered.

**Expected behavior**

A session id supplied by the caller through context wins. The SDK's `session_id` fills `session.id` only when the caller did not set one. That matches how every other context attribute (metadata, user id, tags) already behaves in this package: applied at span start and never overwritten by message processing.

**Desktop**

 - OS: macOS 15 (Darwin 24.6), arm64
 - Node 26.5
 - `@anthropic-ai/claude-agent-sdk` 0.3.218
 - `@arizeai/openinference-instrumentation-claude-agent-sdk` 0.3.0
 - Phoenix 20.12.0 (for the session-splitting consequence)

**Additional context**

Three places write `session.id` from the SDK message after the span has started (all at https://github.com/Arize-ai/openinference/blob/5a1db037/js/packages/openinference-instrumentation-claude-agent-sdk/src/):

- `v1QueryWrapper.ts` `processMessage()`: on `system/init`, `span.setAttributes({ [SESSION_ID]: sessionId, ... })`
- `v2Wrappers.ts` `processSessionMessage()`: same on `system/init`
- `messageProcessor.ts` `extractResultSuccessAttributes()` / `extractResultErrorAttributes()`: include `[SESSION_ID]: msg.session_id`, applied on `result` in v1, v2 sessions, and `wrapPrompt()`

The existing test `test/v1Query.test.ts` ("should create an AGENT span for query()") asserts `session.id === "sess-123"` with no context session set, so it enshrines the fallback case but never exercises the context case.

I have a small fix with tests ready and will open a PR referencing this issue: check the captured context for a session id once per span and skip the SDK write when one is present. The SDK id would then no longer appear on the span when a caller sets their own session; if you would rather keep it under a dedicated attribute, I am happy to add that in the same PR.

iaj6:
Fixes #3775

## What changed

A `session.id` supplied by the caller through OpenInference context (`setSession(context.active(), { sessionId })`) is applied to the AGENT span at start by `OITracer`, and was then overwritten by the SDK's own `session_id` on the `system/init` message and again on the `result` message. This happened in three places: `processMessage()` for `query()`, `processSessionMessage()` for V2 session turns, and `wrapPrompt()` via `extractResultSuccessAttributes()`/`extractResultErrorAttributes()`.

Now each span checks its captured context once (`hasContextSessionId()`), and when the caller supplied a session id the SDK write is skipped (`withoutSessionId()` on the result attributes; the init handler omits it). Without a context session the SDK id still fills `session.id` exactly as before, so the existing assertion in `v1Query.test.ts` (`session.id === "sess-123"`) is unchanged.

`js/AGENTS.md` lists "propagate context attributes via OITracer (session ID, user ID, metadata, tags)" as a must for every instrumentor. This package already propagates the session id; this PR stops it from being clobbered afterwards, which brings it in line with how metadata, user id and tags are already handled (applied at start, never overwritten by message processing).

## Why it matters in Phoenix

Phoenix binds a trace to the first `session.id` it ingests for that trace and ignores later ones. Spans export when they end, so in a turn with tool calls the TOOL spans (which keep the context session id) land first and the trace goes into the caller's session; in a turn with no tool calls the AGENT span lands first with the SDK id and the trace goes into a session named after the SDK UUID. The same agent's turns end up split across two sessions depending on whether it happened to use a tool. Reproduced against the real `@anthropic-ai/claude-agent-sdk@0.3.218` with one turn (output in the issue).

## Tests

Three new tests, one per code path, each wrapping the call in `context.with(setSession(...))` and asserting the AGENT span keeps the context session id (and still records the model name from init):

- `query()` with `system/init` + `result` messages
- `unstable_v2_prompt()`
- `unstable_v2_createSession()` turn via `send()` + `stream()`

They use `provider.register()` to install the async-hooks context manager for the duration of the test and reset the globals in `finally`, so no new devDependency. All three fail on `main` and pass here; the rest of the suite is unchanged (45/45).

## Trade-off to flag

When the caller sets a session id, the SDK's `session_id` no longer appears on the span at all. It was only ever there by accident in that case, but it is a useful value (it is what `claude --resume` takes). If you would like it kept under a dedicated attribute when the context wins, say the word and I will add it to this PR; I left it out to keep the change to the reported behaviour.

Changeset included (patch).

https://claude.ai/code/session_01MXxV7qtX4GyTfo7QWaREnL


iaj6:

js/packages/openinference-instrumentation-claude-agent-sdk/README.md
```diff
@@ -71,17 +71,17 @@ The `query()` function is wrapped to produce:
...
-| `llm.token_count.prompt`     | Input token count      |
-| `llm.token_count.completion` | Output token count     |
-| `llm.token_count.total`      | Total token count      |
-| `llm.cost.total`             | Total cost in USD      |
+| Attribute                    | Description                                                                                          |
+| ---------------------------- | ---------------------------------------------------------------------------------------------------- |
+| `openinference.span.kind`    | `"AGENT"`                                                                                            |
+| `input.value`                | Prompt text                                                                                          |
+| `output.value`               | Result text                                                                                          |
+| `session.id`                 | Caller-supplied session id from OpenInference context when set, otherwise the SDK session identifier |
```
mikeldking, line 79: Let's leave the spec alone as it's more of an implementation / prioritization detail than a spec on the semantics

PR #3779, opened 2026-09-18
feat(typesafe): add JavaScript SDK instrumentation

### LINKED ISSUE #3770: [feature request] Add OpenInference auto-instrumentation for the TypeSafe AI JS/TS SDK (@typesafe-ai/sdk)
## Is your feature request related to a problem? Please describe.

[TypeSafe AI](https://docs.typesafe.ai/introduction/quickstart) ships an official TypeScript/JavaScript SDK (`@typesafe-ai/sdk`) that is currently invisible to OpenInference. Users who put TypeSafe calls into a router, a guardrail, a re-ranker, or an LLM-as-a-judge step get no spans for them, so a TypeSafe call shows up as a gap in the trace: no latency, no token counts, no record of which questions were asked or what was answered.

TypeSafe is specifically positioned for the patterns OpenInference users care most about observing — its docs ship cookbooks for [guardrails](https://docs.typesafe.ai/cookbooks/llm_guardrails), [intent routing](https://docs.typesafe.ai/patterns/intent-routing), [confidence-gated routing](https://docs.typesafe.ai/patterns/confidence-routing), [re-ranking](https://docs.typesafe.ai/cookbooks/rerank_typesafe), [RAG passage classification](https://docs.typesafe.ai/cookbooks/classifying_rag_passages), and [citation checking](https://docs.typesafe.ai/cookbooks/citation_check). Those are exactly the decision points where you want a span, because a silent misclassification upstream is the usual root cause of a bad downstream answer. The per-answer `confidence` and `probabilities` TypeSafe returns are unusually valuable telemetry: they make it possible to filter traces for low-confidence decisions, which most LLM providers don't expose.

## Describe the solution you'd like

A new `@arizeai/openinference-instrumentation-typesafe` package following the existing patterns in `js/packages/` (closest analogue: `openinference-instrumentation-anthropic`).

This is the JS/TS counterpart to the Python ticket (linked). The two should land with the same attribute mapping so traces look identical regardless of language.

### Background: what the SDK looks like

Install ([docs](https://docs.typesafe.ai/sdk/javascript), [repo](https://github.com/typesafe-ai/typesafe-sdk-js)) — requires **Node.js 20 or newer**:

```bash
npm install @typesafe-ai/sdk
```

```ts
import { choice, TypeSafeClient } from "@typesafe-ai/sdk";

const client = new TypeSafeClient();

const response = await client.systemOne({
  state: { document: "I was charged twice. Please fix this ASAP." },
  questions: {
    category: choice({
      instructions: "What is this ticket about?",
      criteria: { billing: null, technical: null, other: null },
    }),
  },
});

console.log(response.answers.category.choice);
```

The whole public surface is essentially one method. There is no streaming, no tool calling, and no multi-turn chat — a much smaller instrumentation surface than a typical LLM provider. The package ships ESM, CommonJS, and TypeScript declarations, so the instrumentation needs to work under both module systems (see "Module patching" below).

**Client** ([`TypeSafeClient`](https://docs.typesafe.ai/sdk/javascript/api/classes/TypeSafeClient)):

```ts
new TypeSafeClient(config?: TypeSafeClientConfig): TypeSafeClient;

systemOne<const Q extends Questions>(
  request: SystemOneRequest<Q>,
  options?: RequestOptions,
): APIPromise<SystemOneResult<Q>>;
```

Readonly properties: `baseURL`, `defaultHeaders`, `defaultModel`, `fetch`, `logger`, `logLevel`, `models`, `retry`, `timeout`.

From [`src/client.ts`](https://github.com/typesafe-ai/typesafe-sdk-js/blob/main/src/client.ts), `systemOne` is a **prototype method** that validates questions, merges in `model: request.model ?? this.defaultModel`, and issues a single `POST /v1/systemone`:

```ts
systemOne<const Q extends Questions>(
  request: SystemOneRequest<Q>,
  options: RequestOptions = {},
): APIPromise<SystemOneResult<Q>> {
  validateQuestions(request.questions);
  const body = {
    ...request,
    model: request.model ?? this.defaultModel,
  } satisfies SystemOneRequestPayload;

  return this.#request<SystemOneResult<Q>>("POST", "/v1/systemone", {
    ...options,
    body,
  });
}
```

Being a prototype method is good news — it's patchable the same way the Anthropic instrumentation patches its client.

**Question builders** — `choice()`, `noul()`, `score()`, plus the `ChoiceQuestion<T>` / `NoulQuestion` / `ScoreQuestion<T>` interfaces:

| Builder | `criteria` | Meaning |
|---|---|---|
| `noul({ instructions, criteria? })` | `{ true, false }` descriptions | Binary yes/no, answered as a probability |
| `choice({ instructions, criteria })` | `ChoiceCriteria` — label → description (or `null`) | Multi-label classification |
| `score({ instructions, criteria })` | `ScoreCriteria` — ordered rubric, one entry per score from 0 | Scalar score against a rubric |

`instructions` and every criteria description accept a string **or** a JSON object/array, not just text ([advanced: structure](https://docs.typesafe.ai/primitives/advanced)). Questions may also be supplied as plain objects with an explicit `type` field instead of via the builders, so the extractor must handle both.

**Result** ([`SystemOneResult`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/SystemOneResult), [`Usage`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/Usage)):

```ts
interface SystemOneResult<Q extends Questions> {
  readonly answers: { readonly [K in keyof Q]: ResultFor<Q[K]> };
  readonly model: string;   // model used to answer the request
  readonly usage: Usage;    // { readonly input_tokens: number; readonly output_tokens: number }
}
```

Per-answer shapes ([`ChoiceResponse`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/ChoiceResponse), [`NoulResponse`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/NoulResponse), [`ScoreResponse`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/ScoreResponse)), matching the [HTTP API](https://docs.typesafe.ai/api):

- `noul` → `{ noul: number }` (probability of yes, 0–1)
- `choice` → `{ choice: string; confidence: number; probabilities: Record<string, number> }`
- `score` → `{ score: number; confidence: number; probabilities: Record<number, number>; legend?: ... }`

Errors are a class hierarchy rooted at [`TypeSafeError`](https://docs.typesafe.ai/sdk/javascript/api/classes/TypeSafeError) / [`APIError`](https://docs.typesafe.ai/sdk/javascript/api/classes/APIError), with `AuthenticationError`, `BadRequestError`, `RateLimitError`, `InternalServerError`, `NotFoundError`, `PermissionDeniedError`, `UnprocessableEntityError`, `APIConnectionError`, `APITimeoutError`, and `APIUserAbortError`.

`systemOne` returns an [`APIPromise<T>`](https://docs.typesafe.ai/sdk/javascript/api/classes/APIPromise), not a plain `Promise` — it has a `withResponse()` variant returning [`WithResponse<T>`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/WithResponse). **The wrapper must preserve the `APIPromise` interface**, or it will break any user calling `.withResponse()`. This is the main non-obvious implementation risk in this ticket; see open questions.

Relevant env vars the SDK reads: `TYPESAFE_API_KEY` (required), plus base-URL / default-model / log-level overrides (see [`ENV`](https://docs.typesafe.ai/sdk/javascript/api/variables/ENV)). Default model is `jev-latest`.

### Proposed package

```
js/packages/openinference-instrumentation-typesafe/
├── package.json                      # @arizeai/openinference-instrumentation-typesafe
├── README.md
├── tsconfig.json
├── tsconfig.esm.json
├── vitest.config.ts
├── src/
│   ├── index.ts
│   ├── instrumentation.ts            # TypeSafeInstrumentation
│   ├── attributes.ts
│   └── version.ts
├── examples/
│   ├── instrumentation.ts
│   ├── basic-usage.ts
│   ├── all-question-types.ts
│   └── guardrail-routing.ts          # TypeSafe gating an LLM call, to show the span tree
└── test/
    ├── instrumentation.test.ts
    ├── questions.test.ts
    ├── dualBuild.test.ts             # mirror the anthropic package's ESM/CJS check
    └── helpers/waitForSpans.ts
```

Usage should be the standard shape:

```ts
import { TypeSafeInstrumentation } from "@arizeai/openinference-instrumentation-typesafe";
import { registerInstrumentations } from "@opentelemetry/instrumentation";

registerInstrumentations({ instrumentations: [new TypeSafeInstrumentation()] });
```

### Module patching

Extend `InstrumentationBase` and register an `InstrumentationNodeModuleDefinition` for `@typesafe-ai/sdk` with a supported-version range, patching `TypeSafeClient.prototype.systemOne` in `patch()` and restoring it in `unpatch()`.

Two things to get right, both of which have bitten other JS instrumentations in this repo:

1. **ESM + CJS.** The package ships both. Follow whatever the Anthropic package does for its dual build and reuse its `dualBuild.test.ts` pattern.
2. **Manual patching escape hatch.** Since module-level hooking is unreliable under bundlers and ESM, expose a `manuallyInstrument(module)` path (as other packages here do) so users can pass the imported module directly.

`client.models` is a metadata/control-plane resource, not an inference call — leave it uninstrumented unless someone makes a case for it.

### Attribute mapping

A `systemOne` call is one model invocation that answers N questions in one parallel pass, so it should be **one `LLM` span**, not N spans. This table must stay identical to the Python ticket's.

Span name: `TypeSafeClient.systemOne`.

| Attribute | Value |
|---|---|
| `openinference.span.kind` | `LLM` |
| `llm.model_name` | `result.model`, falling back to `request.model` / `client.defaultModel` |
| `llm.provider` | `typesafe` (see sub-task below) |
| `llm.system` | `typesafe` |
| `llm.invocation_parameters` | JSON of `{ model, ... }` and any `RequestOptions` actually passed |
| `input.value` | JSON of the request payload: `{ state, questions, model }` |
| `input.mime_type` | `application/json` |
| `output.value` | JSON of `{ answers, usage, model }` |
| `output.mime_type` | `application/json` |
| `llm.input_messages.0.message.role` | `user` |
| `llm.input_messages.0.message.content` | serialized `state` (JSON when `state` is not a string) |
| `llm.output_messages.0.message.role` | `assistant` |
| `llm.output_messages.0.message.content` | JSON of `answers` |
| `llm.token_count.prompt` | `result.usage.input_tokens` — omit when absent |
| `llm.token_count.completion` | `result.usage.output_tokens` — omit when absent |
| `llm.token_count.total` | sum, only when both are present |
| `metadata` | JSON including the question names and their types, and each answer's `confidence` — so low-confidence decisions are filterable in Phoenix |

Span status: `OK` on success; on rejection, record the exception and set `ERROR` with the message. Capture the HTTP status and request id off `APIError` where available. `APIUserAbortError` should end the span too — no leaked open spans on abort.

### Requirements checklist

Per `CLAUDE.md`, the instrumentation must:

- [ ] Honor tracing suppression (`isTracingSuppressed(context.active())` → call through without a span)
- [ ] Propagate context attributes via `getAttributesFromContext` from `@arizeai/openinference-core` (session id, user id, metadata, tags)
- [ ] Respect `TraceConfig` masking. This matters more than usual: `state` is frequently raw customer text (support tickets, chat turns, retrieved passages), so `hideInputs` / `hideInputMessages` must actually suppress it
- [ ] Follow OpenInference semantic conventions
- [ ] Ship tests
- [ ] `enable()` / `disable()` round-trips cleanly
- [ ] Preserve the `APIPromise` return type, including `withResponse()`

### Sub-task: semantic conventions

`typesafe` is not in the `LLMProvider` values. Add it to both, keeping them in sync:

- `js/packages/openinference-semantic-conventions/`
- `python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py`

### Tests

Use `vitest` to match the rest of `js/packages/`.

- `systemOne` happy path, asserting the full attribute set
- One test per question builder (`choice`, `noul`, `score`) plus a mixed multi-question request, covering both the builder form and the plain-object-with-`type` form
- Non-string `state` (object and array) and structured `instructions`/`criteria` — must not crash or produce non-JSON `input.value`
- `withResponse()` still works on the patched method, and the returned value is still an `APIPromise`
- Missing/partial `usage` — token attributes absent, not `0`
- Error paths: `APIError` subclass → `ERROR` status with exception recorded; `APIUserAbortError` → span ended
- Suppression: no span emitted when tracing is suppressed
- `TraceConfig` masking: `hideInputs: true` hides `state`
- `disable()` restores the original method and stops emitting spans
- Context attribute propagation (session/user/metadata/tags land on the span)
- ESM + CJS dual-build test

Stub `globalThis.fetch` or pass a custom `fetch` via `TypeSafeClientConfig` — the client takes an injectable `fetch`, which makes this clean. No network in CI. Recorded cassettes are an option if the Anthropic package's VCR helper is reusable.

### Verification

Per repo practice, round-trip an example against a local Phoenix (`http://localhost:6006`) in a dedicated project and confirm the span tree, span kind, token counts, and `confidence` metadata all render. `guardrail-routing.ts` is the useful one, since it shows a TypeSafe span and an LLM span in the same trace.

### Housekeeping / registration

- [ ] `pnpm changeset` for the new package
- [ ] Root `README.md` — add rows to the instrumentation and examples tables
- [ ] Package `README.md` with install + usage
- [ ] `engines.node >= 20` in `package.json`, matching the SDK's requirement
- [ ] Peer-dependency range on `@typesafe-ai/sdk` matching the supported-versions range in the module definition
- [ ] Confirm nothing needs adding to `.github/workflows/typescript-CI.yaml` (it should pick the package up from the workspace)

## Describe alternatives you've considered

- **Manual OTel instrumentation** — works, but every user re-derives the same attribute mapping and none of them get `TraceConfig` masking or context propagation for free.
- **A `fetch` wrapper instead of patching `systemOne`.** The client accepts an injected `fetch`, so one could trace at the HTTP layer. Rejected as the primary approach: it loses the typed request/response objects, requires re-parsing the body, and is opt-in (users must wire it up), whereas `registerInstrumentations` is zero-config. Could be a documented fallback for bundled/edge environments where module patching fails.
- **One span per question** instead of one per call. Rejected: the questions are answered in a single parallel pass with one shared token count, so per-question spans would misrepresent latency and force token counts to be split arbitrarily. Per-question detail belongs in `metadata` / `output.value`.
- **Using the `EVALUATOR` span kind** instead of `LLM`. TypeSafe is often used as a judge or guardrail, and `EVALUATOR`/`GUARDRAIL` exist in the span-kind values. But the span kind would then depend on how the user happens to be using the call, which the instrumentation cannot know. `LLM` is the honest default since this is a model invocation with a model name and token counts.

## Open questions

1. **`APIPromise` preservation.** What exactly does `APIPromise` add beyond `PromiseLike` — is `withResponse()` the only extra member? The wrapper needs to return something structurally identical. If `APIPromise` is a class rather than an interface at runtime, naive `.then()` wrapping will break `withResponse()` and this needs a proxy-based approach instead. Worth resolving before implementation starts, since it shapes the wrapper.
2. **`llm.input_messages` shape.** Folding `state` into a single synthetic user message makes the call render nicely in Phoenix's message view, but it's a synthetic framing — TypeSafe's API is not a chat API. Should the questions also appear as part of the message content (e.g. as a system message), or should we skip `llm_input_messages` entirely and rely on `input.value`? Leaning toward emitting both, with `input.value` as the source of truth. **Whatever is decided must match the Python ticket.**
3. **Span kind override.** Should `TypeSafeInstrumentation` accept a config option to emit `GUARDRAIL`/`EVALUATOR` instead of `LLM`?
4. **Confidence as a first-class attribute.** `confidence` and `probabilities` are the most distinctive things TypeSafe returns. `metadata` makes them filterable today, but is a dedicated convention warranted so confidence is chartable across providers?
5. **Supported-version range.** The SDK looks early-stage; check the [changelog](https://docs.typesafe.ai/sdk/javascript/changelog) and [`VERSION`](https://docs.typesafe.ai/sdk/javascript/api/variables/VERSION) and pin a range we can actually test against.

## References

- [TypeSafe quickstart](https://docs.typesafe.ai/introduction/quickstart)
- [JavaScript SDK docs](https://docs.typesafe.ai/sdk/javascript) · [API reference](https://docs.typesafe.ai/sdk/javascript/api) · [changelog](https://docs.typesafe.ai/sdk/javascript/changelog)
- [`TypeSafeClient`](https://docs.typesafe.ai/sdk/javascript/api/classes/TypeSafeClient) · [`TypeSafeClientConfig`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/TypeSafeClientConfig) · [`RequestOptions`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/RequestOptions) · [`APIPromise`](https://docs.typesafe.ai/sdk/javascript/api/classes/APIPromise) · [`WithResponse`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/WithResponse)
- [`SystemOneRequest`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/SystemOneRequest) · [`SystemOneResult`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/SystemOneResult) · [`Usage`](https://docs.typesafe.ai/sdk/javascript/api/interfaces/Usage)
- [`choice()`](https://docs.typesafe.ai/sdk/javascript/api/functions/choice) · [`noul()`](https://docs.typesafe.ai/sdk/javascript/api/functions/noul) · [`score()`](https://docs.typesafe.ai/sdk/javascript/api/functions/score)
- [HTTP API reference](https://docs.typesafe.ai/api) · [Models](https://docs.typesafe.ai/models) · [Confidence](https://docs.typesafe.ai/confidence)
- [Primitives](https://docs.typesafe.ai/primitives): [Choice](https://docs.typesafe.ai/primitives/choice), [Score](https://docs.typesafe.ai/primitives/score), [Noul](https://docs.typesafe.ai/primitives/noul), [advanced structure](https://docs.typesafe.ai/primitives/advanced)
- Repo: https://github.com/typesafe-ai/typesafe-sdk-js · npm: `@typesafe-ai/sdk`
- Companion Python ticket: see linked issue

mikeldking:
## Summary

Adds `@arizeai/openinference-instrumentation-typesafe`: one OpenInference `LLM` span per `TypeSafeClient.systemOne` call, with JSON `input.value` / `output.value`, resolved model, token usage, and provider/system `typesafe`.

Closes #3770.

## Phoenix verification

Verified against local Phoenix with the package examples:

- `typesafe-basic-usage-20260917-2038` — single `TypeSafeClient.systemOne` LLM span with structured JSON input/output
- `typesafe-guardrail-routing` — `support.route` chain with TypeSafe + OpenAI LLM children, session, and workflow metadata


js/packages/openinference-instrumentation-typesafe/src/attributes.ts
```diff
@@ -0,0 +1,111 @@
...
+import type { RequestOptions, SystemOneRequest } from "@typesafe-ai/sdk";
+
+import type { TraceConfig } from "@arizeai/openinference-core";
+import {
+  getInputAttributes,
+  getOutputAttributes,
+  isObjectWithStringKeys,
+  safelyJSONStringify,
+} from "@arizeai/openinference-core";
+import { MimeType, SemanticConventions as SC } from "@arizeai/openinference-semantic-conventions";
```
mikeldking: don't use 2 letter ascronyms like this. Remember this. Use object destructuring if needed to get the attributes out.

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
```diff
@@ -0,0 +1,111 @@
...
+  const usage = isObjectWithStringKeys(response.usage) ? response.usage : {};
+  const inputTokens = usage.input_tokens;
+  const outputTokens = usage.output_tokens;
+  return {
+    ...getOutputAttributes({
+      value: safelyJSONStringify(result) ?? "",
+      mimeType: MimeType.JSON,
+    }),
+    // An absent model must not overwrite the request/client fallback.
+    ...(typeof response.model === "string" && { [SC.LLM_MODEL_NAME]: response.model }),
```
mikeldking: set this but also set response_model_name

js/packages/openinference-instrumentation-typesafe/README.md
```diff
@@ -0,0 +1,111 @@
...
+| `llm.model_name`               | Response model, else request or client default        |
+| `input.value` / `output.value` | Full JSON request and response (`application/json`)   |
+| `llm.token_count.*`            | Prompt/completion when present; total when both exist |
+| `llm.invocation_parameters`    | Model and explicit timeout/retry overrides            |
+| `metadata.typesafe`            | Request ID and question types/confidence              |
+| `http.response.status_code`    | HTTP status on SDK API errors                         |
+
+No `llm.input_messages` or `llm.output_messages` are emitted. State, questions, and answers live in the JSON payloads. Noul answers omit confidence.
+
+Context metadata is preserved; `typesafe` is reserved for the instrumentor:
```
mikeldking: don't go into this level of granularity

js/packages/openinference-instrumentation-typesafe/README.md
```diff
@@ -0,0 +1,111 @@
...
+pnpm --filter '@arizeai/openinference-instrumentation-typesafe...' run build
+cd packages/openinference-instrumentation-typesafe
+pnpm exec tsx examples/basic-usage.ts
+```
+
+| Example                                                 | Description                                                 |
+| ------------------------------------------------------- | ----------------------------------------------------------- |
+| [basic-usage.ts](examples/basic-usage.ts)               | Classification with `withResponse()`                        |
+| [all-question-types.ts](examples/all-question-types.ts) | Choice, noul, score, and structured inputs                  |
+| [guardrail-routing.ts](examples/guardrail-routing.ts)   | TypeSafe gates an OpenAI call (also needs `OPENAI_API_KEY`) |
```
mikeldking: these will not show up on the publish

js/packages/openinference-instrumentation-typesafe/README.md
```diff
@@ -0,0 +1,111 @@
...
+With Phoenix at `http://localhost:6006` and `TYPESAFE_API_KEY` set:
+
+```sh
+cd js
+pnpm install --frozen-lockfile -r
+pnpm --filter '@arizeai/openinference-instrumentation-typesafe...' run build
+cd packages/openinference-instrumentation-typesafe
+pnpm exec tsx examples/basic-usage.ts
+```
+
```
mikeldking, line 120: don't show this. it doesn't belong in the npm registry

js/packages/openinference-instrumentation-typesafe/README.md
```diff
@@ -0,0 +1,111 @@
...
+  tracerProvider, // optional; defaults to the global provider
+  instrumentationConfig: { enabled: true },
+  traceConfig: { hideInputs: true, hideOutputs: true },
+});
+```
+
+Supports OpenInference context attributes, tracing suppression, and `TraceConfig` env vars.
+`hideInputs` / `hideOutputs` redact payloads and related question/confidence metadata.
+`hideInputMessages` / `hideOutputMessages` have no effect (no chat-message attributes).
+Headers, credentials, and abort signals are never copied into span attributes.
```
mikeldking: not needed

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
mikeldking: Fixed — switched to destructured `SemanticConventions` keys (no `SC` alias) in both `attributes.ts` and `instrumentation.ts`.

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
mikeldking: Fixed — response now sets both `llm.model_name` and `llm.response.model_name`. Also set `llm.request.model_name` on the request side to match Anthropic.

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Fixed — removed the detailed attributes table and metadata JSON example from the README.

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Fixed — dropped the examples table (not published with the package).

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Fixed — removed the local repo/pnpm run instructions from the README.

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Fixed — removed the extra configuration prose.

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
```diff
@@ -0,0 +1,137 @@
...
+  defaultModel: string;
+  config: TraceConfig;
+}): Attributes {
+  const model = request.model ?? defaultModel;
+  const retry = options?.retry;
+  return {
+    ...(config.hideInputs
+      ? { [INPUT_VALUE]: REDACTED_VALUE }
+      : getInputAttributes({
+          value: safelyJSONStringify({ ...request, model }) ?? "",
```
mikeldking: don't interleave the model here, that doesn't make sense. Also hide inputs should be handled by the OITracer

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
```diff
@@ -0,0 +1,137 @@
...
+      },
+    }),
+  };
+}
+
+/**
+ * Merges context metadata with reserved `typesafe` question type/confidence.
+ * `hideInputs` drops questions; `hideOutputs` drops confidence.
+ */
+function getTypeSafeMetadataAttributes({
```
mikeldking: punt on adding metadata. Only add the minimum

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
mikeldking: Fixed — `input.value` is the actual request body (no interleaved model). Masking is left to OITracer / TraceConfig.

js/packages/openinference-instrumentation-typesafe/src/attributes.ts
mikeldking: Fixed — removed instrumentor-owned `metadata.typesafe` for now; caller context metadata from OITracer is unchanged.

js/packages/openinference-instrumentation-typesafe/src/instrumentation.ts
```diff
@@ -0,0 +1,308 @@
...
+    },
+  );
+
+  return new module.APIPromise(responsePromise, async () => {
+    preferSdkParse = true;
+    const result = await promise;
+    recordOnce(result);
+    endOk();
+    return result;
+  });
```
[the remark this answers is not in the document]
mikeldking: Fixed in 725fc862, with a variation on the suggested patch.

The missing `try/catch` is a real defect, but calling `recordError` directly from it would double-record on the HTTP-failure path: `recordError` is not idempotent (it calls `span.recordException` / `span.end()` unconditionally) and the `rawResponse` rejection handler already fires there. Both async failure paths now go through an `endError` helper gated on the same `ended` flag as `endOk`, so the span ends exactly once.

One correction to the diagnosis: with `@typesafe-ai/sdk` 0.6.0 the span was not actually leaking. `APIPromise#parse()` is `responsePromise.then(parseResponse)`, so our parse callback can only run after `responsePromise` resolves — which is after the fallback IIFE has already been scheduled and passed its `preferSdkParse` check. The observable bug was therefore worse than a leak: a call that failed during SDK parsing was reported as a **successful** span, because the fallback path read the cloned body and set `OK`. The fix makes the error win regardless of which parse settles first.

Covered by a new test, `ends the span when SDK parsing fails after a successful response`, which asserts `ERROR` status and the exception event; it fails on the previous code with `expected { code: 1 } to deeply equal { code: 2, message: 'parse failure' }`.

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Reopening this one — my earlier reply was wrong. The table went away, but the repo-local instructions did not; they were just rewritten (`cd js/packages/...`, `pnpm install`, `pnpm -r build`, `pnpm exec tsx examples/...`). Now actually removed in a70a492c.

Confirmed your reasoning: `files` is `["dist","src"]`, so `examples/` is not in the published tarball.

js/packages/openinference-instrumentation-typesafe/README.md
mikeldking: Same correction as the thread above — also only partly fixed before. The table was dropped, but a relative link (`[examples/instrumentation.ts](./examples/instrumentation.ts)`) survived and had the identical problem of not resolving on npm.

a70a492c replaces the whole section with a single absolute GitHub URL to the examples directory, matching what `openinference-core` does. No relative links remain in the README.

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
```diff
@@ -81,6 +81,7 @@ export const HOST_SUFFIX_TO_PROVIDER: Record<string, LLMProvider> = {
   "api.minimax.io": LLMProvider.MINIMAX,
   "api.minimaxi.com": LLMProvider.MINIMAX,
   "api.minimax.chat": LLMProvider.MINIMAX,
+  "api.typesafe.ai": LLMProvider.TYPESAFE,
```
cephalization: what is this change for? Why in openai instr?

cephalization (approved):

js/packages/openinference-instrumentation-openai/src/instrumentation.ts
mikeldking: this is if typesafe got used through openai, which it wouldn't. so I'll remove

mikeldking: Removed in 61f3e0b8 — both the `HOST_SUFFIX_TO_PROVIDER` entry and its test case are gone, so `openinference-instrumentation-openai/src/instrumentation.ts` now matches `main`.

One knock-on change remains in that package: `LLMProvider.TYPESAFE` stays in the semantic conventions (the Typesafe instrumentor sets `llm.provider`/`llm.system`), and the existing `every provider has at least one host entry` test enumerates `Object.values(LLMProvider)`, so dropping the mapping made it fail. The test now skips providers listed in a new `PROVIDERS_WITHOUT_OPENAI_COMPATIBLE_HOST` set, which holds `TYPESAFE` only. Happy to invert it (assert against an explicit expected list instead) if you would rather not carry an exemption set there.

`pnpm run test` and `type:check` pass for the openai package.

PR #3795, opened 2026-09-18
fix(mistralai): don't crash the user's stream on an empty `choices` chunk

pcbeingused333:
## The bug

\`_AsyncStream.stream_async_with_accumulator\` (in \`openinference-instrumentation-mistralai\`) indexes \`event.data.choices[0]\` unconditionally to peek at \`finish_reason\`:

```python
async for event in await self.stream:
    self._process_chunk(event)
    if event.data.choices[0].finish_reason is not None:
        self._finish_tracing()
    yield event
```

\`choices\` is a plain \`List[...]\` on the SDK's chunk model with no runtime guarantee it's non-empty — the same shape OpenAI-compatible APIs use for a trailing usage-only chunk when \`stream_options={"include_usage": True}\`. When \`choices\` is empty, this raises \`IndexError\` **from inside the generator**, which propagates straight into the caller's \`async for\` and kills the actual LLM response the instrumentation is supposed to be transparently observing — exactly what this repo's own contract calls out (\`python/CLAUDE.md\`: *"Instrumentation must never raise into user code... wrap attribute extraction... in try/except"*).

Every other \`choices\` access in this same package already guards against an empty list:
- \`_response_accumulator.py\`: \`for choice in data.get("choices", ())\`
- \`_response_attributes_extractor.py\`: \`if (choices := ...) and isinstance(choices, Iterable): for choice in choices\`

This was the one spot that indexes it directly. (The sync \`_Stream.__next__\` counterpart doesn't have this peek at all — it relies on \`StopIteration\` to finish tracing, so it was never at risk.)

## The fix

```python
if event.data.choices and event.data.choices[0].finish_reason is not None:
```

## Testing

Added \`tests/.../test_stream.py\`, driving \`_AsyncStream\` directly against a fake stream with an empty-\`choices\` chunk (no cassette/live API needed — it's a pure control-flow check). Verified red against the pre-fix code:

```
IndexError: list index out of range
.../mistralai/_stream.py:151: IndexError
```

Green after the fix. Full package suite: \`tox run -e ruff-mypy-test-mistralai\` → ruff clean, mypy clean (\`no issues found in 13 source files\`), 26 passed.


pcbeingused333:

feiiiiii5:
Ran this locally against current `main` — the crash reproduces, the fix resolves it, and the numbers in your description hold. Setup: python 3.11.15, `mistralai==2.0.0` (the pin in this package's `test-requirements.txt`), base `a719562` + this commit.

What I measured

1. `pytest tests/.../test_stream.py` → 1 passed; `pytest tests` for the whole package → 26 passed (same count you reported).
2. The regression test bites: reverting only the guard back to `if event.data.choices[0].finish_reason is not None:` makes your new test fail with `IndexError: list index out of range` at `_stream.py:157`. Re-applying it makes the package suite pass again.
3. The trigger is reachable through the SDK's own parse path: `CompletionChunk.model_fields["choices"]` is `List[CompletionResponseStreamChoice]`, required, with no length constraint — `CompletionChunk(choices=[])` complains only about `id`/`model`, so an empty `choices` list is a value the model accepts. Your "every sibling access already guards" claim also checks out (`_response_accumulator.py:85`, `_response_attributes_extractor.py:48` and `:80`).
4. `ruff` 0.9.2: `format --diff` clean (13 files) and `check --no-fix` clean. I could not get a faithful `mypy` run in my own venv (an editable install makes `mypy .` resolve the package twice, and `dev-requirements.txt` drift produces unrelated `no-untyped-call` errors on untouched `main`), so mypy is unverified here — not a claim that it fails.

The one thing I'd pin down before merge: what the guard leaves behind for telemetry. Same harness, a stream whose only chunk has `choices=[]`:

| `_stream.py` | caller's `async for` | spans exported |
|---|---|---|
| upstream (no guard) | raises `IndexError` into user code | 1, status `ERROR`, `exception` event |
| this PR | yields the chunk normally | **0** |

Keeping the user's stream alive is the right call, and `python/AGENTS.md:71` is the right citation for it. But that same line ends with "make sure the span still ends on failure", and `stream_async_with_accumulator()` has no `finally`, so the span is ended only when some chunk carries a non-`None` `finish_reason`. A response that ends without one now produces no telemetry at all, where before it at least produced an `ERROR` span. For reference in the same environment: a chunk with `finish_reason="stop"` → 1 span, status `UNSET`; chunks that never carry a `finish_reason` → 0 spans; `await agen.aclose()` after the first chunk → 0 spans; `asyncio.CancelledError` raised mid-stream → propagates, 0 spans (`except Exception` does not catch it). The last three are identical with and without your guard, so they are not caused by this PR — they are the pre-existing no-`finally` gap, and the same shape as #3796 on the openai side. Splitting it out is reasonable; if you do, a short comment noting the gap here would keep it from being lost.

Two small asks:

- The package's `conftest.py` already exposes `in_memory_span_exporter`, so the new test could assert what the call produces (span count, status) rather than only the yielded events — as written it would also pass if the span were never ended.
- Worth one line in the test docstring recording the intended post-fix outcome for a stream that ends without a `finish_reason`, whichever way that call goes.

Disclosure: I'm an automated contributor; the commands and counts above are from an agent run in a throwaway venv, and my `openinference` PR #3796 touches the adjacent openai stream path, so treat this as measurements rather than a reviewer's endorsement.

PR #3797, opened 2026-09-19
fix(openai): accumulate streamed refusal and reasoning_content

### LINKED ISSUE #3785: [bug] Streamed chat-completion delta fields outside the accumulator schema are replaced, not concatenated (refusal, reasoning_content)
**Describe the bug**

When a chat-completion response is streamed, string fields on the delta that are not listed in the
accumulator's declared schema are **replaced by each incoming chunk instead of concatenated**, so the
exported span reports only the final fragment of the field.

Two fields are affected on the OpenAI Chat Completions instrumentor today:

* `refusal` — a first-party `ChoiceDelta` field in the `openai` SDK (`openai.types.chat.ChatCompletionChunk`),
  delivered across chunks the same way `content` is.
* `reasoning_content` — the field OpenAI-compatible servers stream for reasoning models
  (DeepSeek-R1, vLLM, Kimi, Qwen3 via `base_url` override). This is an extra field on the delta, and it
  does survive into the accumulator (verified with `openai` 2.8.0, see the repro).

The consequence is not a missing attribute but a **plausible-but-false one**: `output.value` on the LLM
span contains the last fragment as if it were the whole string, and `stream=True` disagrees with
`stream=False` for the same content.

**Root cause**

`python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_response_accumulator.py:124-139`
declares the accumulation schema for `choices[].message` as only `content`, `function_call` and
`tool_calls`:

```python
message=_ValuesAccumulator(
    content=_StringAccumulator(),
    function_call=_ValuesAccumulator(arguments=_StringAccumulator()),
    tool_calls=_IndexedAccumulator(...),
),
```

`_ValuesAccumulator.__iadd__` (`:259-289`) then behaves differently for declared and undeclared keys:

* the first loop walks `self._values.keys()` (declared keys only) and merges;
* an undeclared key falls through to the second loop and is stored as a **plain copied value**
  (`self._values[key] = value  # new entry`, `:282-288`);
* on the next chunk that key *is* now present in `self._values`, but its stored value is a `str`, which
  matches none of the accumulator branches, so it reaches
  `self._values[key] = value  # replacement` (`:281`).

So the first fragment is discarded and each subsequent one overwrites it. Fields declared as
`_StringAccumulator()` are unaffected, which is why `content` and
`tool_calls[i].function.arguments` accumulate correctly.

**Expected behavior**

Streamed string fields are reassembled into the complete value, and the accumulated result of
`stream=True` matches the non-streamed response for the same content.

**To Reproduce**

Drives the real `_Stream` + `_WithSpan` + `_ChatCompletionAccumulator` path and asserts on the
**exported span attribute** `output.value`, so it exercises the same boundary a user's trace goes
through. No network and no API key.

```python
import json
from typing import Any, Dict, Iterator, List

from openai.types.chat import ChatCompletionChunk

from openinference.instrumentation.openai._response_accumulator import _ChatCompletionAccumulator
from openinference.instrumentation.openai._stream import _Stream
from openinference.instrumentation.openai._with_span import _WithSpan


def chunk(delta: Dict[str, Any], finish_reason: str = None) -> ChatCompletionChunk:
    choice: Dict[str, Any] = {"index": 0, "delta": delta}
    if finish_reason:
        choice["finish_reason"] = finish_reason
    return ChatCompletionChunk(
        id="chatcmpl-x", object="chat.completion.chunk", created=1, model="gpt-4o", choices=[choice]
    )


def output_value(chunks: List[ChatCompletionChunk]) -> Dict[str, Any]:
    provider = TracerProvider()
    tracer = provider.get_tracer("test")

    def _iter() -> Iterator[ChatCompletionChunk]:
        yield from chunks

    with tracer.start_as_current_span("ChatCompletion") as span:
        stream = _Stream(
            stream=_iter(),
            with_span=_WithSpan(span=span),
            response_accumulator=_ChatCompletionAccumulator(
                request_parameters={"model": "gpt-4o", "messages": []},
                chat_completion_type=ChatCompletionChunk,
                response_attributes_extractor=None,
            ),
        )
        for _ in stream:
            pass
    return json.loads(exporter.get_finished_spans()[0].attributes["output.value"])


for field in ("refusal", "reasoning_content"):
    fragments = ["I can", "'t help", " with that."]
    chunks = (
        [chunk({field: f}) for f in fragments]
        + [chunk({"role": "assistant"}), chunk({}, finish_reason="stop")]
    )
    message = output_value(chunks)["choices"][0]["message"]
    print(field, "->", repr(message.get(field)), "expected", repr("".join(fragments)))
```

Observed on `main` (`b7373601`, `openai` 2.8.0, Python 3.11):

```
refusal -> " with that." expected "I can't help with that."
reasoning_content -> " with that." expected "I can't help with that."
```

With the two-line change below, both reassemble correctly.

**Proposed fix**

Declare the two fields as string accumulators, alongside `content`:

```python
message=_ValuesAccumulator(
    content=_StringAccumulator(),
    refusal=_StringAccumulator(),
    reasoning_content=_StringAccumulator(),
    function_call=_ValuesAccumulator(arguments=_StringAccumulator()),
    ...
)
```

That is additive: a chunk that does not carry the field is skipped by the existing
`if (value := values.get(key)) is None: continue` guard at `:263`, so spans for responses without a
refusal or reasoning content keep exactly the shape they have now. Verified: the package's existing
test suite gives identical results before and after (2 failed / 482 passed both ways; those 2 failures
are pre-existing and unrelated - `test_tool_calls.py` passes standalone and fails only under full-run
ordering, and one entry point is not installed in my env).

This looks like the same class already fixed twice here: #2241 (`fix(openai): add tool_id accumulator
for response`, same file) and #3142 / its agno fix (`tool-call arguments dropped on streaming output`).
`reasoning_content` is also already captured in the LiteLLM instrumentor (#3312), so the OpenAI Chat
Completions path is the outlier.

**Two notes before anyone builds on this**

* #3737 also edits `_response_accumulator.py` (image support), so a fix here should expect a rebase or
  be sequenced after it.
* The `mistralai` instrumentor has a structurally identical accumulator
  (`openinference-instrumentation-mistralai/.../_response_accumulator.py:68`), and #3761 is open against
  it. Whether the same field set is missing there is worth deciding once, not per-package.

**Version**

`openinference-instrumentation-openai` on `main` at `b7373601`; `openai` 2.8.0; Python 3.11. No
`OPENAI_API_KEY` needed - the repro constructs the chunks locally.

CodeWithMoin:
Fixes #3785

## What changed

On a streamed chat completion, `refusal` and `reasoning_content` are now accumulated across chunks, same as `content`. One change in `_response_accumulator.py`: both are declared as `_StringAccumulator()` in the `message` schema.

## Why

Only `content`, `function_call` and `tool_calls` were declared there. Any other string field on the delta went through the new-entry path on the first chunk and the replacement path on every chunk after, so `output.value` kept only the last fragment. `refusal` is a first-party delta field, and `reasoning_content` is what DeepSeek, vLLM and Qwen3 stream through a `base_url` override. So a streamed trace showed a plausible but wrong value, and didn't match the same call with `stream=False`.

I kept it to these two fields rather than accumulating every undeclared string. Scalar fields like `role` can repeat across chunks and would get concatenated.

## Trace output

`output.value` message for the same four streamed chunks (two `reasoning_content`, two `content`):

```text
before: {"role": "assistant", "reasoning_content": "is 4.",          "content": "The answer is 4."}
after:  {"role": "assistant", "reasoning_content": "2 plus 2 is 4.", "content": "The answer is 4."}
```

## Tests

New `tests/openinference/instrumentation/openai/test_response_accumulator.py`:
- `content`, `refusal` and `reasoning_content` streamed over three chunks each come out whole, and `role` stays `"assistant"`
- reasoning and content streamed in the same response accumulate independently

On `main` the `refusal` and `reasoning_content` cases fail and `content` passes. With the fix all 4 pass.

```text
pytest tests   # 488 passed, 1 failed
```

The one failure is `test_tool_calls`. It fails the same way on `main` in a full run and passes alone, so it looks like test-order state and isn't related. mypy is clean on the changed files.


CodeWithMoin:
recheck

CodeWithMoin:

PR #3798, opened 2026-09-20
fix(google_genai): publish AFC tool calls as output messages

### LINKED ISSUE #3788: [bug] google-genai: automatic-function-calling tool calls are emitted outside every llm.output_messages bucket
**Describe the bug**

When Gemini runs a tool through the SDK's automatic function calling (`tools=[my_python_function]`),
the model's intermediate function calls are written to span attributes that are **not inside any
`llm.output_messages.<messageIndex>` bucket**, so no consumer that groups attributes by message
index can see them: the function name and arguments are extracted correctly, then published under a
key that structured readers never reach.

`spec/tool_calling.md` defines the pattern as

> `llm.output_messages.<messageIndex>.message.tool_calls.<toolCallIndex>.tool_call.<attribute>`

In `openinference-instrumentation-google-genai`, `_get_attributes_from_function_call()` returns
message-**relative** keys (`message.tool_calls.<i>.tool_call.…`). The candidate path knows this and
re-prefixes them (`_response_attributes_extractor.py:69`):

```python
for key, value in self._get_attributes_from_generate_content_content(content):
    yield f"{SpanAttributes.LLM_OUTPUT_MESSAGES}.{index}.{key}", value
```

The automatic-function-calling path yields the same relative keys with no prefix at all
(`_response_attributes_extractor.py:79` → `:216`), so they reach the span verbatim.

**To Reproduce**

`python/instrumentation/openinference-instrumentation-google-genai`, current `main`
(`b7373601`), no network needed:

```python
from openinference.instrumentation.google_genai._response_attributes_extractor import (
    _ResponseAttributesExtractor,
)

response = types.GenerateContentResponse(
    candidates=[
        types.Candidate(
            index=0,
            content=types.Content(
                role="model", parts=[types.Part.from_text(text="65 degrees and foggy")]
            ),
            finish_reason=types.FinishReason.STOP,
        )
    ],
    automatic_function_calling_history=[
        types.Content(
            role="model",
            parts=[
                types.Part(
                    function_call=types.FunctionCall(
                        name="get_weather", args={"location": "San Francisco"}
                    )
                )
            ],
        )
    ],
)
for key in sorted(dict(_ResponseAttributesExtractor().get_attributes(response, {}))):
    print(key)
```

```
llm.finish_reason
llm.output_messages.0.message.content
llm.output_messages.0.message.role
message.tool_calls.0.tool_call.function.arguments
message.tool_calls.0.tool_call.function.name
output.mime_type
output.value
```

Same result through the instrumentor on the committed recording
`tests/cassettes/test_instrumentation/test_generate_content_with_automatic_tool_calling.yaml`
(`pytest tests/test_instrumentation.py -k automatic` with a span-attribute dump):

```
llm.output_messages.0.message.content = 'The weather in San Francisco is foggy ...'
llm.output_messages.0.message.role    = 'model'
message.tool_calls.0.tool_call.function.name       = 'get_weather'   # <- outside every bucket
message.tool_calls.0.tool_call.function.arguments  = '{"location": "San Francisco"}'
```

The data loss is visible with this repo's own reader. Feeding exactly those attributes to
`openinference.instrumentation._genai_conversion.get_genai_message_attributes()` yields
`gen_ai.output.messages` containing only the text part; the `get_weather` call is absent. With the
keys under `llm.output_messages.1.` instead, the same converter returns
`{"type": "tool_call", "name": "get_weather", "arguments": {"location": "San Francisco"}}`.

**Expected behavior**

A model entry in `automatic_function_calling_history` that issued a function call during *this*
call should become its own `llm.output_messages.<messageIndex>` bucket, continuing after the
candidate messages, with `message.tool_calls.<toolCallIndex>` numbering restarting inside each
message, so `llm.output_messages.1.message.tool_calls.0.tool_call.function.name` is what lands on
the span.

Two kinds of entry have to stay out, because the history is not only this call's output:
`google/genai/models.py:6363` seeds it with the converted request contents (`models.py:6361` returns
a new list, so the caller's own list is not mutated), and `models.py:6356,6367` appends the very
`Content` object the response also carries in `candidates[0].content`. Measured with a patch that
prefixes every model entry unconditionally:

* budget-exhausted shape: one `get_weather` call produced
  `llm.output_messages.0.message.tool_calls.0.…` **and** `llm.output_messages.1.message.tool_calls.0.…`,
  and `gen_ai.output.messages` came back with the same tool-call message twice;
* second turn of a chat: turn 1's `get_weather(SF)` was reported as this span's output at
  `llm.output_messages.1` next to turn 2's `NY` call at `2`, although SF is already in
  `llm.input_messages`.

Nothing currently asserts the orphan shape:
`tests/test_instrumentation.py:2108-2111` only notes "We may or may not see explicit tool call
attributes". Nothing pins the orphan shape either: every other tool-call assertion in the package already
uses the prefixed form (for example `tests/test_instrumentation.py:919`), so the bucketed shape is
this package's own convention.

**Additional context**

* `OPENINFERENCE_HIDE_OUTPUT_MESSAGES` / `hide_output_messages` does not cover these keys today, so
  the tool-call name and arguments stay on the span when a user asks for output messages to be
  hidden. Measured end to end through an `OITracer` with an in-memory exporter (masking happens
  per attribute in `_spans.py:49`, i.e. it matches on the key): with hiding enabled the current code
  still leaves `message.tool_calls.0.tool_call.function.name = 'get_weather'` and
  `.arguments = '{"location": "SF"}'` on the span, while the bucketed key is masked away.
* Robustness note, not a user-facing crash: a model entry whose `parts` is `None` raises
  `TypeError: 'NoneType' object is not iterable` out of the generator (measured by hand-building the
  entry), taking the rest of the span's attributes with it. The SDK always populates `parts`, so this
  needs a hand-built history entry to reach and is left out of the fix.
* Two sequential function calls in one AFC loop also collapse today: `tool_call_index` is shared
  across the whole history rather than restarting per message.
* Scope note, measured rather than assumed: in the recorded AFC response Gemini sends
  `functionCall: {name, args}` with no `id`, so this is about the missing bucket, not about
  dropping a `tool_call.id` that the provider never returned.
* Verified locally: package suite 121 passed before / 125 after adding tests for the bucketed shape,
  the two mis-attribution cases above, and the `parts=None` crash; those tests fail on unmodified
  `main` with `KeyError: 'llm.output_messages.1.message.role'`. CI-equivalent `ruff format --diff .`,
  `ruff check --no-fix .` and `mypy .` (ruff 0.9.2, mypy 1.11.2 from `python/dev-requirements.txt`)
  are clean on both trees.
* Diagnosis and the patch were produced with an AI coding agent (Claude, via Qoder CLI); the
  commands and outputs above were run locally and are reported as measured.
* What is still recoverable today, so the impact is not overstated: the span's `output.value` is the
  marshalled response and does contain `automatic_function_calling_history`, so a viewer that prints
  the raw response blob still shows the call. What is lost is everything that reads the spec
  attributes rather than the blob — message-indexed tool calls, the GenAI conversion above, and any
  evaluation or trace-join built on them.
* I'll hold the patch for a few days rather than open a PR straight away — say the word if the
  bucketing should work differently (for example, folding the AFC calls into the existing message
  index instead of continuing it).

AbarnaaSree:
## Summary

Fixes #3788.

Automatic function calling (AFC) tool calls from the Google GenAI instrumentation were emitted without the `llm.output_messages.<messageIndex>` prefix, producing orphaned `message.tool_calls.*` attributes.

This change publishes AFC tool calls under the appropriate output-message bucket and keeps tool-call indices scoped to each model message.

## Changes

* Prefix AFC tool-call attributes with `llm.output_messages.<messageIndex>`.
* Reset tool-call indexing for each model message.
* Skip request-seeded AFC history that belongs to input messages.
* Avoid publishing candidate content twice when the same `Content` object appears in AFC history.
* Use the Google GenAI SDK's content transformation when determining the seeded AFC history boundary.
* Add regression tests covering multiple AFC model turns, duplicated candidate history, seeded request history, and merged request parts.

## Testing

* `python -m pytest .\python\instrumentation\openinference-instrumentation-google-genai\tests\test_response_attributes_extractor.py -q`
* Result: **6 passed**
* `git diff --check` passes.


AbarnaaSree:

AbarnaaSree:
recheck

AbarnaaSree:
Hi, I’ve signed the CLA and triggered the `recheck`. The CLA workflow is currently waiting for maintainer approval. Thanks!

PR #3810, opened 2026-09-21
fix(anthropic): record every tool_result block of a message, not just the last

### LINKED ISSUE #3809: [bug] anthropic (python and js): a message with several tool_result blocks records only the last one
**Describe the bug**

Both anthropic instrumentors record only the last `tool_result` block of a message. Parallel tool calls come back as one user message carrying several `tool_result` blocks, one per call, each with the `tool_use_id` it answers and the result content. The extractors write `message.tool_call_id` and `message.content` as scalars inside the block loop, so every block but the last is overwritten. The other tool outputs never reach the span, their `tool_use_id`s are lost with them, and the result side of a parallel tool call can no longer be matched to its call. The span gives no sign that anything was dropped.

Python: `python/instrumentation/openinference-instrumentation-anthropic/src/openinference/instrumentation/anthropic/_wrappers.py`, `_get_llm_input_messages`, the `tool_result` branch (~line 656):

```python
elif block["type"] == "tool_result":
    yield (
        f"{LLM_INPUT_MESSAGES}.{i}.{MESSAGE_TOOL_CALL_ID}",
        block["tool_use_id"],
    )
    if (tool_result_content := block.get("content")) is not None:
        yield (
            f"{LLM_INPUT_MESSAGES}.{i}.{MESSAGE_CONTENT}",
            tool_result_content
            if isinstance(tool_result_content, str)
            else safe_json_dumps(tool_result_content),
        )
```

JS: `js/packages/openinference-instrumentation-anthropic/src/instrumentation.ts`, `getAnthropicInputMessageAttributes`, the `tool_result` branch (~line 560) assigns the same two keys at the message root inside `message.content.forEach`.

The sibling `tool_use` branch in both packages already writes per-block indices (`message.contents.{j}` and `message.tool_calls.{tool_index}`), so only the result side clobbers.

**To reproduce**

Python, no network. This runs the request through `_get_attributes_from_messages_create`, the function the wrapper uses to populate the LLM span, so the printed attributes are what lands on the span:

```python
from openinference.instrumentation.anthropic._wrappers import (
    _get_attributes_from_messages_create,
)

kwargs = {
    "model": "claude-sonnet-4-5",
    "max_tokens": 100,
    "messages": [
        {"role": "user", "content": "Add 2+2 and get the weather in Paris."},
        {"role": "assistant", "content": [
            {"type": "tool_use", "id": "toolu_01AAA", "name": "calculator",
             "input": {"expression": "2+2"}},
            {"type": "tool_use", "id": "toolu_02BBB", "name": "get_weather",
             "input": {"city": "Paris"}},
        ]},
        {"role": "user", "content": [
            {"type": "tool_result", "tool_use_id": "toolu_01AAA", "content": "4"},
            {"type": "tool_result", "tool_use_id": "toolu_02BBB",
             "content": "sunny, 22C"},
        ]},
    ],
}

attrs = dict(_get_attributes_from_messages_create(kwargs))
for key in sorted(attrs):
    if key.startswith("llm.input_messages.2."):
        print(key, "=", repr(attrs[key]))
print("first tool result '4' anywhere on span:",
      any(str(v) == "4" for v in attrs.values()))
```

Output (identical across runs):

```
llm.input_messages.2.message.content = 'sunny, 22C'
llm.input_messages.2.message.role = 'user'
llm.input_messages.2.message.tool_call_id = 'toolu_02BBB'
first tool result '4' anywhere on span: False
```

The JS package fails a vitest assertion of the correct behavior the same way: an instrumented `Anthropic` client replaying a cassette with the same three-message conversation leaves only `toolu_02BBB` / `"sunny, 22C"` under `llm.input_messages.2.*`.

**Expected behavior**

Every `tool_result` block reaches the span, keyed so it can be matched back to its tool call. The repo's Vercel processor states the convention for this same situation (`js/packages/openinference-vercel/src/utils.ts`): per the OpenInference spec, each tool result should be a separate message with `message.role: tool`, `message.content`, and `message.tool_call_id`, and it expands a multi-result tool message into one message per result.

Same bug class as #3786, which covers Bedrock Converse.

Two fixes that both match existing repo patterns:

- Expand each `tool_result` block into its own input message with role `tool`, the way the Vercel processor does.
- Keep one message and write each block at its own `message.contents.{j}` index, the way the `tool_use` branch in these same two files does.

**Additional context**

Reproduced on current `main` (44b433b), python anthropic SDK 1.7.0 and the current JS SDK. Both repros are deterministic and need no API key or network. Single `tool_result` messages are unaffected; list-valued result content is already serialized with `safe_json_dumps` / `JSON.stringify` and would keep that handling under either fix.

Happy to open a PR with the fix and regression tests for both packages once there is agreement on which of the two shapes to use.

Harsh23Kashyap:
Fixes #3809

When one input message carries several `tool_result` blocks (the reply to parallel tool calls), both anthropic instrumentors keep only the last block. Each block writes `message.content` and `message.tool_call_id` at the same message index, so later blocks overwrite earlier ones. The span shows one tool result where the model saw several.

This change records each `tool_result` block as its own input message with `message.role = "tool"`, `message.tool_call_id`, and `message.content`, placed right after the message that carries it:

- A message holding only `tool_result` blocks no longer emits an empty wrapper message.
- A mixed message keeps its other blocks at their original content-part indices.
- spec/tool_calling.md represents tool results as input messages with role "tool", which is the convention used here. The bedrock fix for the same bug (#3787) also expands per block but keeps the original message role; happy to match that instead if preferred.

Both packages get the same fix and a regression test. Each test fails on main and passes with this change. The existing single-tool_result test now expects role "tool" for the result message.

Test plan:
- Python: `pytest tests/` in python/instrumentation/openinference-instrumentation-anthropic - 50 passed
- JS: `vitest run` in js/packages/openinference-instrumentation-anthropic - 39 passed


cephalization:
Hey @Harsh23Kashyap this looks good to me, besides the failing CI. Once fixed, we can go ahead and merge. Thanks for the contributions!

